PR #102, opened 2015-01-19
Added 3D texture and 3D integer math

davidpiuva:


source/draw/gpu/opengl/OpenGLES3/VolumeTexture.ooc
```diff
@@ -0,0 +1,49 @@
...
+use ooc-math
+use ooc-base
+import include/gles
+import Context
+VolumeTexture: class {
+	_backend: UInt
+	backend: UInt { get { this _backend } }
+	width: UInt
+	height: UInt
+	depth: UInt
```
davidhesselbom: `width, height, depth: UInt`

PR #142, opened 2015-03-07
Fix stride size in RasterBgr

zhaihj:
`draw/RasterBgr -> ResterBgr -> save` has incorrect stride size, which causes segmental fault.

### DIFF
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -120,7 +120,7 @@ RasterBgr: class extends RasterPacked {
 		result
 	}
 	save: func (filename: String) -> Int {
-		StbImage writePng(filename, this size width, this size height, this bytesPerPixel, this buffer pointer, this buffer size)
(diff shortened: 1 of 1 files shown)


source/draw/RasterBgr.ooc
```diff
@@ -120,7 +120,7 @@ RasterBgr: class extends RasterPacked {
 		result
 	}
 	save: func (filename: String) -> Int {
-		StbImage writePng(filename, this size width, this size height, this bytesPerPixel, this buffer pointer, this buffer size)
+		StbImage writePng(filename, this size width, this size height, this bytesPerPixel, this buffer pointer, this size width * this bytesPerPixel)
```
davidhesselbom: Thanks for the pull request! And sorry for leaving it open for so long :(

`RasterBgr` inherits from `RasterPacked` which has a `_stride` value, which is only equal to `this size width * this bytesPerPixel` if `align` is 0 ([look for yourself](https://github.com/cogneco/ooc-kean/blob/master/source/draw/RasterPacked.ooc#L37)). So, rather than `this size width * this bytesPerPixel`, it should be `this stride`.

PR #155, opened 2015-03-18
Adapted to public API

emilwestergren:


source/base/Debug.ooc
```diff
@@ -34,8 +34,14 @@ Debug: class {
 		This printFunction = f
 	}
 	print: static func (printOut: String, level: Int = 1) {
-		if (This _level == level || (This _level == DebugLevel Everything) ) {
-			This printFunction(printOut)
+		version(debugPrint) {
+			if (This _level == level || (This _level == DebugLevel Everything) ) {
```
davidhesselbom: What if `level < This _level`? Pretty sure the idea is that if the level of a message is above or equal to the current level, it prints, otherwise it doesn't. So if the level is "notification", you print everything. This requires the DebugLevel enum to be sorted in order of severity, of course...

source/draw/gpu/android/GpuPackerBin.ooc
```diff
@@ -19,15 +19,20 @@ import structs/FreeArrayList, GpuImage, GpuMonochrome, GpuBgr, GpuBgra, GpuUv, G
...
-			packer := this _packers[i]
-			packer free()
+			this _packers removeAt(0, true)
 		}
-		this _packers clear()
 	}
-	add: func (packer: GpuPacker) { this _packers add(packer) }
+	add: func (packer: GpuPacker) {
+		if (this _packers size >= this _limit) {
+			tmp := this _packers[0]
```
davidhesselbom: What happens to `tmp` here?

source/base/Debug.ooc
```diff
@@ -34,8 +34,14 @@ Debug: class {
 		This printFunction = f
 	}
 	print: static func (printOut: String, level: Int = 1) {
-		if (This _level == level || (This _level == DebugLevel Everything) ) {
-			This printFunction(printOut)
+		version(debugPrint) {
```
davidhesselbom: No version block here, version block the calls instead using module specific version flags.

emilwestergren:
Updated.

emilwestergren:
Updated.

PR #160, opened 2015-03-23
Complex numbers simple operators.

bettinaselig:


source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,35 @@
...
+	init: func@ (=real, =imaginary)
+	conjugate ::= FloatComplex new(this real, - this imaginary)
+	absoluteValue ::= (this real pow(2) + this imaginary pow(2)) sqrt()
+	operator + (other: This) -> This {This new(this real + other real, this imaginary + other imaginary)}
+	operator - (other: This) -> This {This new(this real - other real, this imaginary - other imaginary)}
+	operator - -> This {This new(-this real, -this imaginary)}
+	operator * (other: Float) -> This {This new(other * this real, other * this imaginary)}
+	operator * (other: This) -> This {This new(this real * other real - this imaginary * other imaginary, this real * other imaginary + this imaginary * other real)}
+	operator / (other: Float) -> This {This new(this real / other, this imaginary / other)}
+	operator / (other: This) -> This {(this * other conjugate) / (other absoluteValue pow(2))}
```
davidhesselbom: Leave a single space after `{` and before `}`, please (as on line 35).

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,35 @@
...
+
+
```
davidhesselbom: Why the empty lines?

test/math/FloatComplexTest.ooc
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+FloatComplexTest new() run()
+
+
+/*
+expect(this complexNumber0 == this complexNumber0, is true)
+expect(this complexNumber0 != this complexNumber1, is true)
+expect(this complexNumber1 == this complexNumber2, is true)
+*/
```
davidhesselbom: What are these comments for? Will they be used later?

test/math/FloatComplexTest.ooc
```diff
@@ -0,0 +1,67 @@
...
+			expect((this complexNumber0 + this complexNumber1) real, is equal to(this complexNumber2 real))
+			expect((this complexNumber0 + this complexNumber1) imaginary, is equal to(this complexNumber2 imaginary))
+		})
+		this add("subtraction", func() {
+			expect((this complexNumber2 - this complexNumber1) real, is equal to(this complexNumber0 real))
+			expect((this complexNumber2 - this complexNumber1) imaginary, is equal to(this complexNumber0 imaginary))
+		})
+		this add("negative", func() {
+			expect(-this complexNumber0 real == -2, is true)
+			expect(-this complexNumber0 imaginary == -1, is true)
```
davidhesselbom: I would write 
```ooc
expect((-this complexNumber0) real == -2, is true)
expect((-this complexNumber0) imaginary == -1, is true)
```
just to be safe, since we've had some problems with unary minus operators in the past.

test/math/FloatComplexTest.ooc
```diff
@@ -0,0 +1,67 @@
...
+			expect((this complexNumber0 * this complexNumber1) imaginary == 7, is true)
+		})
+		this add("scalar multiplication", func() {
+			expect(((-1.0f) * this complexNumber0) real, is equal to((-complexNumber0) real))
+			expect(((-1.0f) * this complexNumber0) imaginary, is equal to((-complexNumber0) imaginary))
+		})
+		this add("division", func() {
+			expect((complexNumber2 / complexNumber0) real == 2.6f, is true)
+			expect((complexNumber2 / complexNumber0) imaginary == 0.2f, is true)
+		})
```
davidhesselbom: I would have asked for tests of compound assignment operators (`+=`, `-=`, `*=` and so on), but since I know those operators are currently broken in the compiler itself, I will instead ask you to leave a TODO line:
```ooc
TODO: Add tests for compound assignment operators
```

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,35 @@
...
+}
+operator * (left: Float, right: FloatComplex) -> FloatComplex { FloatComplex new(left * right real, left * right imaginary) }
```
davidhesselbom: Any reason why only the `*` operator was implemented?

test/math/FloatComplexTest.ooc
```diff
@@ -0,0 +1,67 @@
...
+
+	complexNumber0 := FloatComplex new (2,1)
+	complexNumber1 := FloatComplex new (3,2)
+	complexNumber2 := FloatComplex new (5,3)
+
+	init: func() {
+		super("FloatComplex")
+		this add("equality", func() {
+			expect(this complexNumber0 real == 2, is true)
+			expect(this complexNumber0 imaginary == 1, is true)
```
davidhesselbom: How about a test for `==` and `!=`, in addition?

See e.g. `FloatPoint2D.ooc` for an example of how to add equality operators.

source/math/FloatComplex.ooc
davidhesselbom: A `toString` function might be useful.

bettinaselig:
Updated :-)

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,79 @@
...
+			imaginaryResult = array[1] trimRight('i') toFloat()
+		}
+		This new (realResult, imaginaryResult)
+	}
+	exponential: func -> This {
+		(this real) exp() * This new((this imaginary) cos(), (this imaginary) sin())
+	}
+	/*logarithm: func -> This {
+		This new(this absoluteValue log(), atan2(this imaginary, this real))
+	}*/
```
davidhesselbom: Why commit commented out code?

test/math/FloatComplexTest.ooc
```diff
@@ -0,0 +1,75 @@
...
+			expect(this complexNumber0 == this complexNumber0, is true)
+			expect(this complexNumber0 != this complexNumber1, is true)
+			expect(this complexNumber1 == this complexNumber2, is false)
+		})
+		this add("toString", func() {
+			expect(this complexNumber0 toString(), is equal to("2.00 +1.00i"))
+			//expect(this complexNumber0 toString() parse(), is equal to(this complexNumber0))
```
davidhesselbom: More commented out code...

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,79 @@
...
+	operator * (other: Float) -> This { This new(other * this real, other * this imaginary) }
+	operator * (other: This) -> This { This new(this real * other real - this imaginary * other imaginary, this real * other imaginary + this imaginary * other real) }
+	operator / (other: Float) -> This { This new(this real / other, this imaginary / other) }
+	operator / (other: This) -> This { (this * other conjugate) / (other absoluteValue pow(2)) }
+	operator == (other: This) -> Bool { this real == other real && this imaginary == other imaginary }
+	operator != (other: This) -> Bool { !(this == other) }
+	toString: func -> String {
+		if (this imaginary < 0) "#{this real toString()} #{this imaginary toString()}i"
+		else "#{this real toString()} +#{this imaginary toString()}i"
+	}
```
davidhesselbom: Unfortunately, this will leak memory.
```ooc
    toString: func -> String {
		this real toString() >> (this imaginary > 0 ? " +" : " ") & this imaginary toString() >> "i"
    }
```
will delete any temporary strings along the way. We should fix this in the other classes as well, eventually, but that's not your job.

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,79 @@
...
+		if (array size > 1) {
+			realResult = array[0] toFloat()
+			imaginaryResult = -array[1] trimRight('i') toFloat()
+		} else {
+			array := input split('+')
+			realResult = array[0] toFloat()
+			imaginaryResult = array[1] trimRight('i') toFloat()
+		}
+		This new (realResult, imaginaryResult)
+	}
```
davidhesselbom: This also leaks memory, but there's no way to avoid it if we want to use `split`.

There's also a bit of duplicate code.
```ooc
	parse: static func (input: String) -> This {
		realResult, imaginaryResult: Float
		minus := input contains?('-')
		array := input split(minus ? '-' : '+')
			realResult = array[0] toFloat()
			imaginaryString := array[1] trimRight('i')
			imaginaryResult = minus ? -imaginaryString toFloat() : imaginaryString toFloat()
			imaginaryString free()
		array free()
		This new (realResult, imaginaryResult)
	}
```
that leaks less memory, and has less duplicate code.

bettinaselig:
New try.

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,75 @@
...
+	parse: static func (input: String) -> This {
+    realResult, imaginaryResult: Float
+    minus := input contains?('-')
+    array := input split(minus ? '-' : '+')
+    realResult = array[0] toFloat()
+    imaginaryString := array[1] trimRight('i')
+    imaginaryResult = minus ? -imaginaryString toFloat() : imaginaryString toFloat()
+    imaginaryString free()
+    array free()
+    This new (realResult, imaginaryResult)
```
davidhesselbom: Something is wrong with the indentation here - make sure you're using tabs.

test/math/FloatComplexTest.ooc
```diff
@@ -0,0 +1,75 @@
...
+		this add("equality", func() {
+			expect(this complexNumber0 real == 2, is true)
+			expect(this complexNumber0 imaginary == 1, is true)
+			expect(this complexNumber0 == this complexNumber0, is true)
+			expect(this complexNumber0 != this complexNumber1, is true)
+			expect(this complexNumber1 == this complexNumber2, is false)
+		})
+		this add("toString", func() {
+			expect(this complexNumber0 toString(), is equal to("2.00 +1.00i"))
+			expect((FloatComplex parse("2.00 +1.00i")) == this complexNumber0, is true)
```
davidhesselbom: Add tests for all combinations of positive and negative real and imaginary values, e.g. 
2+1i
2-1i
-2+1i
-2-1i
I'm pretty sure negative real values will break the `parse` method in its current implementation...

test/math/FloatComplexTest.ooc
```diff
@@ -0,0 +1,75 @@
...
+		})
+		this add("exponential", func() {
+			expect(this complexNumber0 exponential() real, is equal to(this complexNumber0 real exp() * this complexNumber0 imaginary cos()))
+			expect(this complexNumber0 exponential() imaginary, is equal to(this complexNumber0 real exp() * this complexNumber0 imaginary sin()))
+		})
```
davidhesselbom: Should there be a test for the DFT as well?

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,75 @@
...
+	}
+	logarithm: func -> This {
+		This new(this absoluteValue log(), atan2(this imaginary, this real))
+	}
+	rootOfUnity: func ~one (n: Int) -> This {
+		this rootOfUnity(n, 1)
+	}
+	rootOfUnity: func ~two (n: Int, k: Int) -> This {
+		This new(0, 2 * k * PI / n)
+	}
```
davidhesselbom: You can reduce this to one function with parameter defaults:
```ooc
rootOfUnity: func (n: Int, k := 1) -> This {
```
`k` will then default to 1 if `rootOfUnity` is called with only one parameter, but you can also call `rootOfUnity` with two parameters. This means you can also remove the suffix, since there's only one function named `rootOfUnity`.

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,75 @@
...
+FloatComplex: cover {
+	real, imaginary: Float
+	init: func@ (=real, =imaginary)
+	conjugate ::= FloatComplex new(this real, - this imaginary)
+	absoluteValue ::= (this real pow(2) + this imaginary pow(2)) sqrt()
+	operator + (other: This) -> This { This new(this real + other real, this imaginary + other imaginary) }
+	operator - (other: This) -> This { This new(this real - other real, this imaginary - other imaginary) }
+	operator - -> This { This new(-this real, -this imaginary) }
+	operator * (other: Float) -> This { This new(other * this real, other * this imaginary) }
+	operator * (other: This) -> This { This new(this real * other real - this imaginary * other imaginary, this real * other imaginary + this imaginary * other real) }
```
davidhesselbom: Strictly for aesthetical reasons, this would be better as two lines:
```ooc
operator * (other: This) -> This { 
    This new(
        this real * other real - this imaginary * other imaginary, 
        this real * other imaginary + this imaginary * other real
    ) 
}
```

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,75 @@
...
+	}
+	discreteFourierTransform: func (input: ArrayList<This>) -> ArrayList<This> {
+		result := ArrayList<This> new(input size)
+		if (input size > 0) {
+			for (i in 0..(input size - 1)) {
+				for (j in 0..(input size - 1)) {
+					result[i] = result[i] + input[j] * this rootOfUnity(input size, -i * j)
+				}
+			}
+		}
```
davidhesselbom: This program
```ooc
size := 5
for (i in 0..size)
	(i toString()) print()
```
will output `01234`, so `for (i in 0..(input size - 1))` will run from `0` to `input size - 2`. When `input size` is 1, it will run 0 times. When `input size` is 2, it will run 1 time. Is this the intended behavior?

If not, you want 
```ooc
for (i in 0..(input size))
    for (j in 0..(input size))
```
and if `input size` is 0 the `for` loops will run 0 times, so there's no need to check that `input size > 0` beforehand. 

Regardless, you can remove all of these brackets without affecting the scope:
```ooc
discreteFourierTransform: func (input: ArrayList<This>) -> ArrayList<This> {
    result := ArrayList<This> new(input size)
    if (input size > 0)
        for (i in 0..(input size - 1))
             for (j in 0..(input size - 1))
                 result[i] = result[i] + input[j] * this rootOfUnity(input size, -i * j)
    result
```

source/math/FloatComplex.ooc
bettinaselig: Here, you used both >> and & to concatente strings. Is there a difference between them?

bettinaselig:
New version.

bettinaselig:
And suddenly everything became much easier.

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,90 @@
...
+	toString: func -> String {
+		this real toString() >> (this imaginary > 0 ? " +" : " ") & this imaginary toString() >> "i"
+	}
+	parse: static func (input: String) -> This {
+		realResult, imaginaryResult: Float
+		array: ArrayList <String>
+		array = input split(' ')
+		realResult = array[0] toFloat()
+		imaginaryResult = array[1] toFloat()
+		array free()
```
davidhesselbom: This will free the ArrayList, but not its content. Use a FreeArrayList instead of ArrayList, and free() will also free the contents.

source/math/FloatComplex.ooc
```diff
@@ -0,0 +1,90 @@
...
+	operator / (other: This) -> This { (this * other conjugate) / (other absoluteValue pow(2)) }
+	operator == (other: This) -> Bool { this real == other real && this imaginary == other imaginary }
+	operator != (other: This) -> Bool { !(this == other) }
+	toString: func -> String {
+		this real toString() >> (this imaginary > 0 ? " +" : " ") & this imaginary toString() >> "i"
+	}
+	parse: static func (input: String) -> This {
+		realResult, imaginaryResult: Float
+		array: ArrayList <String>
+		array = input split(' ')
```
davidhesselbom: This can be reduced to `array: FreeArrayList <String> = input split(' ')`

davidhesselbom: A better name for `array` is, perhaps, `parts`?

davidhesselbom:
ooc-kean was updated - please rebase before I merge.

bettinaselig:
Everything done.

source/collections/Vector.ooc
```diff
@@ -98,6 +99,14 @@ HeapVector: class <T> extends Vector<T> {
 	__destroy__: func {
 		gc_free(this _backend)
 	}
+
+		operator [] (index: Int) -> T {
+			this _backend[index]
+		}
+
+		operator []= (index: Int, item: T) {
```
davidhesselbom: One last thing: Before `this _backend[index] = item`, do this:
```ooc
if (T inheritsFrom?(Object)) {
	old := this[index] as Object
	old free()
}
```
or else whatever is already in `_backend[index]` will leak.

bettinaselig:
Fixed what you said in Vector.

PR #163, opened 2015-03-30
Added new FloatVector collection. Added a two functions to euclid transform.

niklasborwell:


source/collections/FloatVector.ooc
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+	variance: Float {
+		get {
+			squaredSum := 0.0f
+			for (i in 0..this count)
+				squaredSum += pow((this[i] - this mean), 2.0f)
+			squaredSum / this count
+		}
+	}
```
davidhesselbom: getters can be written like this:
`mean ::= this sum / this count`

source/math/FloatPoint3D.ooc
```diff
@@ -67,7 +67,7 @@ FloatPoint3D: cover {
 	operator <= (other: This) -> Bool { this x <= other x && this y <= other y && this z <= other z }
 	operator >= (other: This) -> Bool { this x >= other x && this y >= other y && this z >= other z }
 	operator as -> String { this toString() }
-	toString: func -> String { "#{this x toString()}, #{this y toString()}, #{this z toString()}" }
+	toString: func -> String { "#{"%.8f" format(this x)}, #{"%.8f" format(this y)}, #{"%.8f" format(this z)}" }
```
davidhesselbom: This looks painfully leaky.

source/collections/FloatVector.ooc
```diff
@@ -0,0 +1,99 @@
...
+		this _vector[index] as Float
+	}
+	operator []= (index: Int, item: Float) {
+		this _vector[index] = item
+	}
+	toString: func() -> String {
+		result := ""
+		for (i in 0..this _count)
+			result += this[i] toString() + "\n"
+		result
```
davidhesselbom: It should be possible to use one or several of the `<<` (free the right operand), `>>` (free the left operand), or `&` (free both operands) operators to make this function completely free from memory leaks.

source/math/FloatEuclidTransform.ooc
```diff
@@ -97,6 +100,26 @@ FloatEuclidTransform: cover {
...
+	toCameraCoordinates: func(zDistance: Float) -> This {
+		correction := Quaternion getTransform(-this rotation, zDistance) * FloatPoint2D new(0.0f, 0.0f)
+		This new(
+			this translationX + correction x,
+			this translationY + correction y,
+			this rotationX,
+			this rotationY,
+			this rotationZ,
+			this scaling)
+	}
```
davidhesselbom: This could also be made to leak less using the `<<` `>>` `&` operators.

source/math/FloatPoint3D.ooc
niklasborwell: That shouldn't be the main focus of toString()s, since we mainly use it for debugging. Two zeroes is not enough for everyday use for me so i have to write huge println lines all the time. And thats annoying.

source/math/FloatEuclidTransform.ooc
niklasborwell: Fixed

source/collections/FloatVector.ooc
niklasborwell: fixed

source/math/FloatPoint3D.ooc
davidhesselbom: `format()` leaks, but the rest of the function doesn't have to.
```ooc
toString: func -> String { "%.8f" format(this x) & "%.8f" format(this y) & "%.8f" format(this z) }
```
still leaks a lot, but only about half as much.

source/math/FloatEuclidTransform.ooc
```diff
@@ -97,12 +100,32 @@ FloatEuclidTransform: cover {
...
-		" Scale: " + "%8f" format(this scaling) +
-		" Rx: " + "%8f" format(this rotationX) +
-		" Ry: " + "%8f" format(this rotationY) +
-		" Rz: " + "%8f" format(this rotationZ)
+		" tx: " >> "%8f" format(this translationX) >>
+		" ty: " >> "%8f" format(this translationY) >>
+		" Scale: " >> "%8f" format(this scaling) >>
+		" Rx: " >> "%8f" format(this rotationX) >>
+		" Ry: " >> "%8f" format(this rotationY) >>
+		" Rz: " >> "%8f" format(this rotationZ)
```
davidhesselbom: This actually made " tx: " disappear from the next `toString()` call (the `>>` operation deleted the literal string. Oops). Try
```ooc
	toString: func -> String {
		" tx: " << "%8f" format(this translationX) >>
		" ty: " & "%8f" format(this translationY) >>
		" Scale: " & "%8f" format(this scaling) >>
		" Rx: " & "%8f" format(this rotationX) >>
		" Ry: " & "%8f" format(this rotationY) >>
		" Rz: " & "%8f" format(this rotationZ)
	}
```

davidhesselbom: Wow, I'd entirely forgotten about this: You can use `formatFloat` instead of `format`. Faster, no leaks, easy to use.

source/math/FloatPoint3D.ooc
davidhesselbom: Same here, use `formatFloat` instead of `format` and the leaks will be all gone.

davidhesselbom: Sorry, I forgot about the commas...
```ooc
	toString: func -> String { "%.8f" formatFloat(this x) >> ", " & "%.8f" formatFloat(this y) >> ", " & "%.8f" formatFloat(this z) }
```

source/math/FloatEuclidTransform.ooc
```diff
@@ -97,12 +100,32 @@ FloatEuclidTransform: cover {
...
 	}
+	toImageCoordinates: func(zDistance: Float) -> This {
+		correction := Quaternion getTransform(this rotation, zDistance) * FloatPoint2D new(0.0f, 0.0f)
+		This new(
+			this translationX + correction x,
+			this translationY + correction y,
+			this rotationX,
+			this rotationY,
+			this rotationZ,
+			this scaling)
```
davidhesselbom: I normally like compact code, but in this case I wonder if this wouldn't look better with the `)` on its own line. Opinions?

source/math/FloatEuclidTransform.ooc
```diff
@@ -18,10 +18,13 @@ import math
...
+import FloatPoint3D
 import FloatTransform2D
+import Quaternion
 
 FloatEuclidTransform: cover {
 	translationX, translationY, rotationX, rotationY, rotationZ, scaling: Float
+	rotation ::= FloatPoint3D new(rotationX, rotationY, rotationZ)
 	imageWidth := 1920
 	fov := 50.0f
 	k ::= tan(this fov * 0.01745329252f / 2.0f) / (this imageWidth / 2)
```
davidhesselbom: Should we document this magic number here? Where does it come from?

niklasborwell: This is on it's way out. We're gonna do something better.

source/math/FloatEuclidTransform.ooc
niklasborwell: This is gonna be deleted so it doesn't matter that much, but I think we should keep a standard and follow that so we don't need to discuss individual cases. Whether we do new line here or not i don't really care, but I'd do a new line here like if we were using { }

source/math/FloatEuclidTransform.ooc
niklasborwell: great

niklasborwell:
Rebased.

PR #166, opened 2015-03-31
Improvements to plot library and Queue added in collections

alexandersundin:


source/collections/Queue.ooc
```diff
@@ -0,0 +1,51 @@
...
+            if (this _head != this _tail || this isFull())
+                this _head = (this _head + 1) % this _capacity
+            this _count -= 1
+        }
+				return this _queue[tempHead]
+    }
+
+    peek: func -> T {
+        return this _queue[this _head]
+    }
```
davidhesselbom: Lots of strange indentation in this file...

source/plot/Axis.ooc
```diff
@@ -42,56 +43,63 @@ Axis: class {
...
-			while (tickValue <= this max) {
-				result = result & this getTickSVG(tickValue, radix, position, numberOffset, rightTickMarkerStartOffset, tickMarkerEndOffset, "end")
-				tickValue += this tick
-				position y += - scaling y * tick
+				tickValue := this getFirstTickValue()
+				position y += translationToOrigo y - plotAreaSize y - scaling y * tickValue
+				while (tickValue <= this max) {
+					result = result & this getTickSVG(tickValue, radix, position, numberOffset, rightTickMarkerStartOffset, tickMarkerEndOffset, "end")
+					tickValue += this tick
+					position y += - scaling y * tick
```
davidhesselbom: Why not `position y -= scaling y * tick`?

source/plot/LinePlotData2D.ooc
```diff
@@ -2,39 +2,75 @@ use ooc-collections
...
-			for (j in 1..this dataSeries count)
-				result = result & (scaling x * this dataSeries[j] x) toString() & " " clone() & (- scaling y * this dataSeries[j] y) toString() & " " clone()
-			result = result & "'/>\n" clone()
+				result = result & "<path stroke='" clone() & this color clone() & "' stroke-opacity='" clone() & this opacity toString() & "' fill='none' stroke-width='" clone() & this lineWidth toString() & "' d='M " clone() & (scaling x * this dataSeries[0] x) toString() & " " clone() & (- scaling y * this dataSeries[0] y) toString() & " L " clone()
+				for (j in 1..this dataSeries count)
+					result = result & (scaling x * this dataSeries[j] x) toString() & " " clone() & (- scaling y * this dataSeries[j] y) toString() & " " clone()
+				result = result >> "' "
+				match (this lineStyle) {
+					case LineStyle Solid =>
+						// do not do anything, will result in solid line
```
davidhesselbom: This case is superfluous, then.

source/plot/LinePlotData2D.ooc
```diff
@@ -2,39 +2,75 @@ use ooc-collections
...
+					case LineStyle Dashed =>
+						result = result >> "stroke-dasharray='" & (this lineWidth * 5) toString() >> "," & (this lineWidth * 5) toString() >> "'"
+					case LineStyle Dotted =>
+						result = result >> "stroke-dasharray='" & this lineWidth toString() >> "," & this lineWidth toString() >> "'"
+					case =>
+						// do not do anything, will result in solid line
+				}
+				result = result >> "/>\n"
```
davidhesselbom: This section is indented one step too far.

source/collections/VectorList.ooc
```diff
@@ -74,7 +74,7 @@ VectorList: class <T> {
 	}
 	pop: func -> T {
 		result := this[0]
-		this remove(0)
+		this removeAt(0)
```
davidhesselbom: What motivated this change?

source/math/FloatExtension.ooc
```diff
@@ -134,7 +134,7 @@ extend Float {
 		(coefficient, radix) := This decomposeToCoefficientAndRadix(value, 1)
 		power := log10(radix) as Int
 		result := ""
-		if (coefficient != 1.0f)
+		if (Float absolute(coefficient - 1.0f) > 0.000001f)
```
davidhesselbom: ooc supports scientific notation for Float literals, like this:
```ooc
if (Float absolute(coefficient - 1.0f) > 1.0E-6f
```
This makes constants like these easier to read.

source/collections/VectorList.ooc
alexandersundin: pop() leaked memory before since remove() allocates and returns a T object. Another solution is to use remove() and assign the returned object to result in pop().

alexandersundin:
Updated

source/collections/VectorList.ooc
davidhesselbom: Do you mean 
```ooc
pop: func -> T {
    this remove(0)
}
```
or
```ooc
pop: func -> T {
    result := this remove(0)
    result
}
```
would work, too, and not leak memory? Then why aren't we doing that instead?

source/collections/Queue.ooc
```diff
@@ -0,0 +1,51 @@
...
+        this _head = (this _head + 1) % this _capacity
+      this _count -= 1
+    }
+    return this _queue[tempHead]
+  }
+
+  peek: func -> T {
+    return this _queue[this _head]
+  }
+}
```
davidhesselbom: Nice, but now this whole file is using spaces instead of tabs.

source/math/FloatExtension.ooc
```diff
@@ -110,7 +110,7 @@ extend Float {
 				value /= 10.0f
 				radix *= 10.0f
 			}
-			while (Float absolute(value) < pow(10.0f, valueDigits-1)) {
+			while (Float absolute(value) - pow(10.0f, valueDigits-1) < -0.000001) {
```
davidhesselbom: You can use `1.0E-6f` here, as well. But why 1.0E-6, specifically?

davidhesselbom: You can use `1.0E-6f` here, as well. But why 1.0E-6, specifically?

davidhesselbom: You can use `1.0E-6f` here, as well. But why 1.0E-6, specifically?

source/plot/LinePlotData2D.ooc
```diff
@@ -27,14 +40,35 @@ LinePlotData2D: class extends PlotData2D {
...
-	getSvgLegend: func (legendCount: Int) -> String {
-		boundaryOffset := 5
-		result := "<text id='" clone() & this label clone() & "' x='" clone() & boundaryOffset toString() & "' y='" clone() & (this fontSize * legendCount + boundaryOffset) toString() & "' font-size='" clone() & this fontSize toString() & "' fill='" clone() & this color clone() & "' fill-opacity='" clone() & this opacity toString() & "'> ▬ " clone() & this label clone() & "</text>\n" clone()
+	getSvgLegend: func (legendCount, fontSize: Int) -> String {
+		result: String
+		start := FloatPoint2D new(this legendOffset, this legendOffset + (fontSize * legendCount - fontSize / 2) as Float)
+		end := FloatPoint2D new(this legendOffset + fontSize, start y)
+		match (this lineStyle) {
+			case LineStyle Solid =>
+				result = Shapes line(start, end, this lineWidth, this opacity, this color)
```
davidhesselbom: Same here, this case is the same as the default case. Leave a comment in the default case `case => // Solid` if you want.

source/plot/ScatterPlotData2D.ooc
```diff
@@ -50,18 +50,20 @@ ScatterPlotData2D: class extends PlotData2D {
...
 		match (this shape) {
 			case Shape Circle =>
-				symbol = "•"
+				result = result & Shapes circle(FloatPoint2D new(start x + halfLineHeight, start y), size / 2.0f, this opacity, this color)
 			case Shape Square =>
-				symbol = "■"
+				result = result & Shapes rect(FloatPoint2D new(start x, start y - halfLineHeight), FloatPoint2D new(size, size), this opacity, this color)
 			case =>
-				symbol = ""
+				result = ""
```
davidhesselbom: This line shouldn't be necessary, just don't do anything in the default case.

source/plot/svg/Shapes.ooc
```diff
@@ -1,28 +1,31 @@
...
+		"<line x1='" << start x toString() >> "' y1='" & start y toString() >> "' x2='" & end x toString() >> "' y2='" & end y toString() >> "' stroke-width='" & lineWidth toString() >> "' stroke='" >> color >> "' stroke-opacity='" & opacity toString() >> "' stroke-dasharray='" & dashFloatPair toString() >> "'/>\n"
 	}
-	circle: static func (x, y, r, opacity: Float, color: String) -> String {
-		"<circle cx='" clone() & x toString() & "' cy='" clone() & y toString() & "' r='" clone() & r toString() & "' fill='" clone() & color clone() & "' fill-opacity='" clone() & opacity toString() & "'/>\n" clone()
+	circle: static func (center: FloatPoint2D, r, opacity: Float, color: String) -> String {
+		This ellipse(center, r, r, opacity, color)
 	}
 	ellipse: static func (center: FloatPoint2D, rx, ry, opacity: Float, color: String) -> String {
 		"<ellipse cx='" clone() & center x toString() & "' cy='" clone() & center y toString() & "' rx='" clone() & rx toString() & "' ry='" clone() & ry toString() & "' fill='" clone() & color clone() & "' fill-opacity='" clone() & opacity toString() & "'/>\n" clone()
```
davidhesselbom: You shouldn't need to use `clone()` anywhere at all in this file now that you have the new operators...

source/collections/VectorList.ooc
alexandersundin: The second example was my thought but the first example was cleaner. Don't know if that will leak or not. I'll check that out.

source/plot/Axis.ooc
```diff
@@ -8,13 +8,14 @@ Orientation: enum {
...
 	min: Float { get set }
 	max: Float { get set }
 	tick: Float { get set (length) {
 			(value, radix) := Float decomposeToCoefficientAndRadix(length, 1)
 			this tick = radix
-			if (value != 0) {
+			if (value != 0.0f) {
 				if (length / this tick < 1.1f) {
 					this tick /= 10.0f
 				} else if (length / this tick < 4.0f) {
```
davidhesselbom: I see 4 unnecessary brackets here.

source/plot/Axis.ooc
```diff
@@ -42,56 +43,63 @@ Axis: class {
...
+					tickValue += this tick
+					position x += scaling x * tick
+				}
+			} else {
+				result = result >> "<g desc='Y-axis data'>\n"
+				labelOffset := FloatPoint2D new(- (log10(Float maximum(this max, Float absolute(this min)) / radix) + 3.0f) * 0.6f * (this fontSize - 4.0f), - plotAreaSize y / 2.0f)
+				numberOffset := FloatPoint2D new(- 0.6f * (this fontSize - 4.0f), (this fontSize - 4.0f) / 3.0f)
+				tickMarkerEndOffset := FloatPoint2D new(5.0f, 0.0f)
+				rightTickMarkerStartOffset := FloatPoint2D new(plotAreaSize x, 0.0f)
+				result = result & "<g desc='rotated Y-axis' transform='rotate(-90," clone() & (position x + labelOffset x) toString() & "," clone() & (position y + labelOffset y) toString() & ")'>\n" clone()
```
davidhesselbom: More unnecessary `clone()` calls.

source/collections/VectorList.ooc
alexandersundin: Works perfectly with the first example.

source/plot/LinePlotData2D.ooc
```diff
@@ -27,14 +40,35 @@ LinePlotData2D: class extends PlotData2D {
 			result = result & "<path stroke='" clone() & this color clone() & "' stroke-opacity='" clone() & this opacity toString() & "' fill='none' stroke-width='" clone() & this lineWidth toString() & "' d='M " clone() & (scaling x * this dataSeries[0] x) toString() & " " clone() & (- scaling y * this dataSeries[0] y) toString() & " L " clone()
 			for (j in 1..this dataSeries count)
 				result = result & (scaling x * this dataSeries[j] x) toString() & " " clone() & (- scaling y * this dataSeries[j] y) toString() & " " clone()
-			result = result & "'/>\n" clone()
+			result = result >> "' "
```
davidhesselbom: More `clone()` calls.

source/plot/LinePlotData2D.ooc
```diff
@@ -2,22 +2,35 @@ use ooc-collections
...
 
-	init: func ~dataSeries(dataSeries: VectorList<FloatPoint2D>, label := "", colorBgra := ColorBgra new()) {
+	init: func ~dataSeries(dataSeries: VectorList<FloatPoint2D>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
+		this lineStyle = lineStyle
 		super(dataSeries, label, colorBgra)
 	}
 
-	init: func ~color(dataSeries: VectorList<FloatPoint2D>, colorBgra: ColorBgra) {
+	init: func ~color(dataSeries: VectorList<FloatPoint2D>, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
+		this lineStyle = lineStyle
```
davidhesselbom: In this case, you can call `this init(dataSeries, "", colorBgra, lineStyle)` and skip `super`.

source/plot/LinePlotData2D.ooc
```diff
@@ -2,22 +2,35 @@ use ooc-collections
...
 		super(dataSeries, "", colorBgra)
 	}
 
-	init: func ~twoFloatSeries(xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new()) {
+	init: func ~twoFloatSeries(xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
+		this lineStyle = lineStyle
 		super(xSeries, ySeries, label, colorBgra)
```
davidhesselbom: Generally speaking, in constructors, always call `super` before you do anything else, unless there's a reason you can't.

source/math/FloatExtension.ooc
alexandersundin: Ah, missed that line. 1.0E-6 was chosen to achieve good precision and at the same time prevent truncation errors.

davidhesselbom: I see. 

What does this second loop do, exactly, and why was it changed?

alexandersundin: It loops while value is smaller than 10^(N-1) and adjusts value and radix so that the value has N digits before the decimal point. Radix represents the factor that, if multiplied with decomposed value, equals the original value.

It was modified because the behaviour wasn't correct in some cases. I don't remember the numbers exactly but an example of such a case would be:

Say that the number 0.100001 is actually represented as 0.09999999 in a Float and that we want 1 value digit. Then we would expect the returned value to be 1.00001. But the case if using value < pow in the while loop is that it will be executed two times and have a value of 9.999999 after the loop.  

If we instead use value - pow < threshold:
After the first iteration of the loop value will be 0.9999999. 
0.9999999 - 1 = -0.0000001 which is not smaller than -1.0E-6 and consequently the result will be 0.9999999. This is much closer to the intended 1.00001 than in the first case.

alexandersundin:
Updated

source/math/FloatExtension.ooc
davidhesselbom: Right, epsilon is > 1E-7 for 32-bit `Float`. Thanks for the fix.

We should probably have a test for this...

source/plot/SVGPlot.ooc
```diff
@@ -14,7 +14,6 @@ SVGPlot: class {
 	yAxis: Axis { get set }
 	colorList: VectorList<String> { get set }
 	colorCount := 0
-	plotAreaPercentage: Float
```
davidhesselbom: The test still uses this variable...

source/collections/Queue.ooc
```diff
@@ -0,0 +1,51 @@
...
+	capacity ::= this _capacity
+	_count: Int
+	count ::= this _count
+	_head: Int
+	_tail: Int
+
+	init: func (=_capacity) {
+		this _queue = gc_calloc(capacity as SizeT, T size)
+		this _head = 0
+		this _tail = 0
```
davidhesselbom: You should be able to skip these lines if you do 
```ooc
_head: 0
_tail: 0
```
above.

source/collections/Queue.ooc
```diff
@@ -0,0 +1,51 @@
...
+
+	isEmpty: func -> Bool {
+		this _count == 0
+	}
+
+	isFull: func -> Bool {
+		this _count == this _capacity
+	}
+
+	add: func (item: T) {
```
davidhesselbom: These are the add/remove function pairs I've seen for queues in other programming languages:
- C#: enqueue/dequeue
- C++: push/pop
- Java: add/remove
... and I think we shouldn't mix the pairs. The original, C# version of Kean uses the C# names. @simonmika?

davidhesselbom: OK, C# style it is.

source/collections/Queue.ooc
```diff
@@ -0,0 +1,51 @@
...
+		super()
+	}
+
+	isEmpty: func -> Bool {
+		this _count == 0
+	}
+
+	isFull: func -> Bool {
+		this _count == this _capacity
+	}
```
davidhesselbom: `isEmpty` and `isFull` should be properties, like `count` and `capacity` above.

Also, by order from above, we don't put `is` in property names if it can be avoided, so
```ooc
empty ::= this _count == 0
full ::= this _count == this _capacity
```

source/math/FloatExtension.ooc
```diff
@@ -110,7 +110,7 @@ extend Float {
 				value /= 10.0f
 				radix *= 10.0f
 			}
-			while (Float absolute(value) < pow(10.0f, valueDigits-1)) {
+			while (Float absolute(value) - pow(10.0f, valueDigits-1) < -1.0E-6f) {
```
davidhesselbom: Sorry I didn't think of this sooner, but you should probably use `Float epsilon` instead of `1.0E-6f` - I think that's what it's for, after all.

alexandersundin: Nice!

alexandersundin:
Updated

source/collections/Queue.ooc
```diff
@@ -0,0 +1,44 @@
...
+		gc_free(_queue)
+		super()
+	}
+
+	enqueue: func (item: T) {
+		if (!this full) {
+			this _queue[this _tail] = item
+			this _tail = (this _tail + 1) % this _capacity
+			this _count += 1
+		}
```
davidhesselbom: What happens if the queue is full?

source/collections/Queue.ooc
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	dequeue: func -> T {
+		tempHead := this _head
+		if (!this empty) {
+			if (this _head != this _tail || this full)
+				this _head = (this _head + 1) % this _capacity
+			this _count -= 1
+		}
+		return this _queue[tempHead]
```
davidhesselbom: What happens if the queue is empty?

source/collections/Queue.ooc
```diff
@@ -0,0 +1,44 @@
+
+Queue: class <T> {
+	_queue: T*
+	_capacity: Int
+	_count: Int
```
davidhesselbom: These should also be set to 0.

source/collections/Queue.ooc
alexandersundin: Nothing is added. It is implemented as a fixed size queue.

source/collections/Queue.ooc
alexandersundin: The logical thing to do would be to return null right? In that case I have to add that to <code>dequeue</code>

davidhesselbom: And if T is a cover type?

source/collections/Queue.ooc
davidhesselbom: So whoever adds things to a queue is responsible for making sure that it's not already full? How about adding
```ooc
else
    raise("Trying to enqueue something on a full queue")
```

source/math/FloatExtension.ooc
```diff
@@ -134,7 +134,7 @@ extend Float {
 		(coefficient, radix) := This decomposeToCoefficientAndRadix(value, 1)
 		power := log10(radix) as Int
 		result := ""
-		if (coefficient != 1.0f)
+		if (Float absolute(coefficient - 1.0f) > Float epsilon)
```
davidhesselbom: Another `clone()` call here.

source/plot/svg/Shapes.ooc
```diff
@@ -1,28 +1,31 @@
...
+		"<ellipse cx='" << center x toString() >> "' cy='" & center y toString() >> "' rx='" & rx toString() >> "' ry='" & ry toString() >> "' fill='" >> color >> "' fill-opacity='" & opacity toString() >> "'/>\n"
 	}
-	rect: static func ~FloatPoint2D(upperRight: FloatPoint2D, width, height: Float, opacity: Float, color: String) -> String {
-		"<rect x='" clone() & upperRight x toString() & "' y='" clone() & upperRight y toString() & "' width='" clone() & width toString() & "' height='" clone() & height toString() & "' fill='" clone() & color clone() & "' fill-opacity='" clone() & opacity toString() & "'/>\n" clone()
+	rect: static func ~FloatPoint2D(upperRight, size: FloatPoint2D, opacity: Float, color: String) -> String {
+		This rect(upperRight x, upperRight y, size x, size y, opacity, color)
+	}
+	rect: static func ~positionFloatPoint2D(upperRight: FloatPoint2D, width, height: Float, opacity: Float, color: String) -> String {
```
davidhesselbom: `width, height, opacity: Float`

source/plot/Axis.ooc
```diff
@@ -42,56 +42,63 @@ Axis: class {
 		result
 	}
 
-	getSVG: func (plotAreaSize, axisAreaSize, position, translationToOrigo, scaling: FloatPoint2D) -> String {
+	getSVG: func (plotAreaSize, margin, translationToOrigo, scaling: FloatPoint2D, fontSize: Int) -> String {
```
davidhesselbom: By convention, acronyms with more than 3 letters only capitalize the first letter, so `getSvg`.

source/plot/Axis.ooc
```diff
@@ -42,56 +42,63 @@ Axis: class {
...
+				tickValue := this getFirstTickValue()
+				position y += translationToOrigo y - plotAreaSize y - scaling y * tickValue
+				while (tickValue <= this max) {
+					result = result & this getTickSVG(tickValue, radix, position, numberOffset, rightTickMarkerStartOffset, tickMarkerEndOffset, "end")
+					tickValue += this tick
+					position y -= scaling y * tick
+				}
 			}
+			result = result >> "</g>\n"
 		}
```
davidhesselbom: This function is too long and contains a lot of duplicate code. Instead of a huge if/else block, consider adding the functions `getHorizontalSvg` and `getVerticalSvg` or something and try to move any duplicate code into their own functions.

alexandersundin:
Updated

source/collections/Queue.ooc
```diff
@@ -0,0 +1,46 @@
...
+		super()
+	}
+
+	enqueue: func (item: T) {
+		if (!this full) {
+			this _queue[this _tail] = item
+			this _tail = (this _tail + 1) % this _capacity
+			this _count += 1
+		} else
+    	raise("Trying to enqueue something on a full queue")
```
davidhesselbom: tabs

source/plot/SVGWriter2D.ooc
```diff
@@ -3,30 +3,63 @@ use ooc-math
...
 import PlotData2D
 import Axis
+import io/File
 import io/FileWriter
 import SVGPlot
 import LinePlotData2D
 import svg/Shapes
+import math
 
 SVGWriter2D: class {
```
davidhesselbom: Another TLA: `SvgWriter2D`

source/plot/SVGWriter2D.ooc
```diff
@@ -44,21 +77,26 @@ SVGWriter2D: class {
...
-				numPlotsX = 3
-			else
-				numPlotsX = 1
+			if (this numberOfPlotsHorizontally == 0) {
+				if (this svgPlots count > 1 && this svgPlots count < 5)
+					numPlotsX = 2
+				else if (this svgPlots count >= 5)
+					numPlotsX = 3
+				else
+					numPlotsX = 1
```
davidhesselbom:
```ooc
if (this svgPlots count == 1)
    numPlotsX = 1
else if (this svgPlots count < 5)
    numPlotsX = 2
else
    numPlotsX = 3
```

source/collections/Queue.ooc
```diff
@@ -0,0 +1,46 @@
...
+	_capacity := 0
+	_count := 0
+	_head := 0
+	_tail := 0
+	capacity ::= this _capacity
+	count ::= this _count
+	empty ::= this _count == 0
+	full ::= this _count == this _capacity
+
+
```
davidhesselbom: Double empty line

ooc-collections.use
```diff
@@ -2,4 +2,5 @@ Name: OOC Collections
 Description: Collections Library
 SourcePath: source/collections
 Requires: ooc-unit, ooc-base, ooc-sdk-lite
-Imports: Vector, FloatVector, VectorList, PointerVector, PointerVectorList, SortedVectorList
+Imports: Vector, FloatVector, VectorList, PointerVector, PointerVectorList, SortedVectorList, Queue
+
```
davidhesselbom: Extra empty line?

alexandersundin:
Updated

PR #167, opened 2015-03-31
Added function that returns a random Float value between 0 an 1. Added a...

bettinaselig:
... function converting FloatBox2D into IntBox2D

### DIFF
--- a/source/math/IntBox2D.ooc
+++ b/source/math/IntBox2D.ooc
@@ -42,7 +42,6 @@ IntBox2D: cover {
 	bottomCenter ::= IntPoint2D new(this center x, this bottom)
 	empty ::= this size empty
 	init: func@ (=leftTop, =size)
-	init: func@ ~fromFloatBox2D (floatBox2D: FloatBox2D) {this init(floatBox2D left round() as Int, floatBox2D top round() as Int, floatBox2D width round() as Int, floatBox2D height round() as Int)}
 	init: func@ ~fromFloats (left, top, width, height: Int) { this init(IntPoint2D new(left, top), IntSize2D new(width, height)) }
 	//init: func@ ~fromSize (size: IntSize2D) { this init(IntPoint2D new(), size) }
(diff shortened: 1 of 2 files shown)


source/sdk/math/Random.ooc
```diff
@@ -25,7 +25,7 @@ Random: class {
 	state := static Time microtime()
 
 	/**
-	   :return: a pseudo-random number between INT_MIN and INT_MAX, generated using
+	   :return: a pseudo-random number between 0 and INT_MAX, generated using
```
davidhesselbom: Was this comment wrong all along, or was there a change in behavior?

source/math/IntBox2D.ooc
```diff
@@ -42,6 +42,7 @@ IntBox2D: cover {
 	bottomCenter ::= IntPoint2D new(this center x, this bottom)
 	empty ::= this size empty
 	init: func@ (=leftTop, =size)
+	init: func@ ~fromFloatBox2D (floatBox2D: FloatBox2D) {this init(floatBox2D left round() as Int, floatBox2D top round() as Int, floatBox2D width round() as Int, floatBox2D height round() as Int)}
```
davidhesselbom: Leave a space after `{` and before `}`, please.

davidhesselbom: Is the `round()` call necessary? What happens if you leave it out, will the resulting `Int` be floored, ceiled, or rounded?

source/sdk/math/Random.ooc
bettinaselig: The comment was wrong before. I checked it up and tested the function.

source/math/IntBox2D.ooc
bettinaselig: We did not see the function toIntBox2D(). That was the only reason why we added this constructor. 
Shall I take them away again? Or shall I add a corresponding constructor in the FloatBox2D class?

davidhesselbom: I don't really see the point in having both functions, so I think you should remove this one and use the existing one instead, yes.

source/sdk/math/Random.ooc
davidhesselbom: Great, thanks!

bettinaselig:
Done.

PR #172, opened 2015-04-09
Added Queue class and improved some code

emilwestergren:


source/draw/gpu/android/AndroidContext.ooc
```diff
@@ -146,7 +121,7 @@ AndroidContext: class extends OpenGLES3Context {
 	createPacker: func (size: IntSize2D, bytesPerPixel: UInt) -> GpuPacker {
 		result := this _packerBin find(size, bytesPerPixel)
 		if (result == null) {
-			Debug print("Could not find a recycled GpuPacker in list with size " + this _packerBin _packers size toString() + " with size " + size toString())
+			version(debugGLv) { Debug print("Could not find a recycled GpuPacker in list with size " + this _packerBin _packers size toString() + " with size " + size toString()) }
```
davidhesselbom:
```ooc
Debug print("Could not find a recycled GpuPacker in list with size " << this _packerBin _packers size toString() >> " with size " & size toString())
```
to prevent memory leaks.

PR #174, opened 2015-04-13
Made a "noRecycle" version flag to disable recycling of GPU images.

davidpiuva:
This can be used to detect when something crash from memory recycling on the GPU. InMotion only works with the -DnoRecycle flag active.


source/draw/gpu/GpuImage.ooc
```diff
@@ -54,7 +54,10 @@ GpuImage: abstract class extends Image {
 	recycle: func {
 		if (this _canvas != null)
 			this _canvas onRecycle()
-		this _context recycle(this)
+		version(safe)
+			free()
```
davidhesselbom: `this free()`

PR #175, opened 2015-04-14
FloatBox2D, IntBox2D and gaussian distributed random numbers

bettinaselig:


source/math/FloatBox2D.ooc
```diff
@@ -50,16 +56,16 @@ FloatBox2D: cover {
 	intersection: func (other: This) -> This {
 		left := this left > other left ? this left : other left
 		top := this top > other top ? this top : other top
-		width := ((this right < other right ? this right : other right) - left) maximum(0.0f)
-		height := ((this bottom < other bottom ? this bottom : other bottom) - top) maximum(0.0f)
+		width := Float maximum(0, (this right < other right ? this right : other right) - left)
+		height := Float maximum(0, (this bottom < other bottom ? this bottom : other bottom) - top)
```
davidhesselbom:
```ooc
this bottom < other bottom ? this bottom : other bottom
```
This is exactly what `maximum` does:
```ooc
    // FloatExtension.ooc
	maximum: static func(first: This, second: This) -> This {
		first > second ? first : second
	}
```
So it would be better to write
```ooc
Float maximum(0, (Float maximum(this right, other right) - left))
```

This goes for the rest of the file as well. It's probably necessary to add a suffix `~two` to `FloatExtension`'s `maximum` and `minimum` functions so that it looks just like in `IntExtension`, and the reason I didn't use these `maximum` and `minimum` functions to begin with is probably that they didn't work and I didn't know why, but that's no problem anymore, so let's use those functions and make this code a little neater.

bettinaselig: Done.

bettinaselig:
And one has to call the functions Int maximum~two() with prefix and Float maximum() without prefix. 
If I change maximum() in FloatExtensions to maximum~two(), errors somewhere else in the code occur.

davidhesselbom:
Then we shall have to fix those errors. I will look into it.

source/sdk/math/Random.ooc
```diff
@@ -39,6 +40,16 @@ Random: class {
...
 		return (rand() as Float) / INT_MAX
 	}
+	/**
+	   :return: a pseudo-random float number with a normal distribution
+	 */
+	randn: static func (sigma := 1.0f) -> Float {
+		result := 0.0f
+		u1 := randFloat()
+		u2 := randFloat()
+		result = sqrt(-2.0f * log(u1)) * cos(2 * PI * u2)
```
davidhesselbom: Due to a bug in rock , `2 * PI * u2` will always result in an `Int` because the first term, `2`, is an `Int`. Change this to `2.0f * PI * u2`.

bettinaselig: Done.

source/sdk/math/Random.ooc
```diff
@@ -39,6 +40,16 @@ Random: class {
 	randFloat: static func -> Float {
 		return (rand() as Float) / INT_MAX
 	}
+	/**
+	   :return: a pseudo-random float number with a normal distribution
+	 */
+	randn: static func (sigma := 1.0f) -> Float {
```
davidhesselbom: We should rewrite how Random numbers work, but that's a larger project. For now, rename this function to `generateGaussianFloat`. There should also exist a version of the function that takes `my` as an argument.

bettinaselig: Done.

source/sdk/math/Random.ooc
```diff
@@ -39,6 +40,17 @@ Random: class {
 	randFloat: static func -> Float {
 		return (rand() as Float) / INT_MAX
 	}
+	/**
+	   Box-Muller transform, http://en.wikipedia.org/wiki/Box-Muller_transform
+	   :return: a pseudo-random float number with a normal distribution
```
davidhesselbom: `:return: a pseudo-random float number with a normal distribution, centered about mu`. for clarity.

source/sdk/math/Random.ooc
```diff
@@ -39,6 +40,17 @@ Random: class {
...
+	 */
+	generateGaussianFloat: static func (sigma := 1.0f, mu := 0.0f) -> Float {
+		result := 0.0f
+		u1 := randFloat()
+		u2 := randFloat()
```
davidhesselbom: `randFloat()` can return 0. According to the [wikipedia page for Box-Muller](http://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform), 
> The basic form as given by Box and Muller takes two samples from the uniform distribution on the interval (0, 1]

and `log(0)` is probably something you don't want to happen. I'm afraid you'll have to do 
```ooc
u1 := Float maximum(randFloat(), Float minimumValue)
```
to be on the safe side.

PR #178, opened 2015-04-17
Cleaned some code and minor fixes. Fixed deadlock bug in GpuImage recycling.

emilwestergren:


ooc-collections.use
```diff
@@ -2,4 +2,4 @@ Name: OOC Collections
 Description: Collections Library
 SourcePath: source/collections
 Requires: ooc-unit, ooc-base, ooc-sdk-lite
-Imports: Vector, FloatVector, VectorList, PointerVector, PointerVectorList, SortedVectorList
```
davidhesselbom: Why was FloatVector removed?

emilwestergren: My bad.

emilwestergren:
Updated.

PR #185, opened 2015-04-20
Improved the VectorList class and adapted code to changes.

emilwestergren:


davidhesselbom:
Can't merge until we solve the compiler issues.

emilwestergren:
Readded removeAt so we can still prevent memory leaks until we can fix the remove(i) -> T function.

PR #195, opened 2015-04-24
Made the syntax not crash the compiler. Extended box with scaling operands

davidpiuva:
Needed for auto zoom.


davidhesselbom:
This didn't crash my compiler...

PR #197, opened 2015-04-24
Move lists from collections to math and renamed them to ...VectorList

bettinaselig:


source/math/FloatComplexVectorList.ooc
```diff
@@ -0,0 +1,133 @@
...
+	sum: func -> FloatComplex {
+		result := FloatComplex new()
+		for (i in 0..this _count)
+			result = result + this[i]
+		result
+	}
+	mean: func -> FloatComplex {
+		sum() / this _count
+	}
+	getReal: func -> FloatVectorList {
```
davidhesselbom: No but seriously, shouldn't this be a property?

davidhesselbom:
Needs rebasing.

source/math/FloatComplexVectorList.ooc
bettinaselig: Updated.

bettinaselig:
Done.

PR #202, opened 2015-04-28
Made a transformed box drawing method.

davidpiuva:
To be used by rectangle overlays for correct perspective.

### DIFF
--- a/source/draw/gpu/GpuYuv420Semiplanar.ooc
+++ b/source/draw/gpu/GpuYuv420Semiplanar.ooc
@@ -16,6 +16,7 @@
 
 use ooc-math
 use ooc-draw
+use ooc-collections
 import GpuMonochrome, GpuCanvas, GpuPlanar, GpuUv, GpuContext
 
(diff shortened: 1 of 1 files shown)


source/draw/gpu/GpuYuv420Semiplanar.ooc
```diff
@@ -66,4 +67,14 @@ GpuYuv420Semiplanar: abstract class extends GpuPlanar {
 		result := RasterYuv420Semiplanar new(y as RasterMonochrome, uv as RasterUv)
 		result
 	}
+	drawBox: func (box: FloatBox2D, transform: FloatTransform2D) {
```
davidhesselbom: Is this better than `OpenGLES3Canvas drawBox` for any reason?

simonmika:
The method should be on canvas, on the canvas that belongs to the YUV image and, at least soon, draw on all channels at once using a color argument. There should never be any access from image too canvas, only the other way around, with exception for the property that creates the canvas.

davidhesselbom:
@simonmika, so this pull request is a no-go?

PR #204, opened 2015-04-28
Added __onheap__  for Cell value.

fredrikbryntesson:


davidhesselbom:
Will require cogneco rock 1.05.

PR #212, opened 2015-05-04
Added indexing operator to get item in queue.

simonmika:


source/collections/Queue.ooc
```diff
@@ -89,4 +89,7 @@ VectorQueue: class <T> extends Queue<T> {
 		this _capacity = newCapacity
 	}
 	toString: func -> String { "Queue: count=" << this _count toString() >> " capacity=" & this _capacity toString() }
+	operator [] (index: Int) -> T {
+		this _backend[this _tail - index % this _capacity]
```
davidhesselbom: Not sure I understand the math here. Why are you indexing from the tail? If I want queue[0], do I want the tail or the head? And what happens when `index > tail`?

simonmika: Index 0 will return the last enqueued object. I considered the safe flag but failed to find usage examples so I didn't do it.

davidhesselbom: A simple search for "version(safe)" in the code base should find you a few examples.

Keep in mind that if `index > tail` then `tail - index` will give you a negative value, and the `%` operator will not make this a positive value, so even if `index < count` (which I see you run a check for in the code where you actually use this operator), chances are you will be trying to access `this _backend[something negative]`.

simonmika: I will be generalizing this and writing a couple of tests on top.

PR #216, opened 2015-05-05
Added indexing operator to get item in queue.

simonmika:
Replaces #212.

### DIFF
--- a/source/collections/Queue.ooc
+++ b/source/collections/Queue.ooc
@@ -1,3 +1,4 @@
+import math
 
 Queue: abstract class <T> {
 	_count := 0
@@ -89,4 +90,11 @@ VectorQueue: class <T> extends Queue<T> {
 		this _capacity = newCapacity
 	}
 	toString: func -> String { "Queue: count=" << this _count toString() >> " capacity=" & this _capacity toString() }
+	operator [] (index: Int) -> T {
+		version(safe)
+			if (index < -this count || index >= this count)
+				raise("Indexing in get accessor of VectorQueue outside of range.")
+		position := (index >= 0 ? this _head + index : this _tail + index) modulo(this _capacity)
+		this _backend[position]
+	}
 }
--- a/test/collections/VectorQueueTest.ooc
+++ b/test/collections/VectorQueueTest.ooc
@@ -86,6 +86,49 @@ VectorQueueTest: class extends Fixture {
 			}
 			queue free()
 		})
+		this add("Queue positive indexing", func {
+			queue := this _createQueue(10, 5)
+			version(debugTests)
+				for (i in 0..queue count)
+					(i toString() + " -> " + queue[i] toString()) println()
+			for (i in 0..5)
+				expect(queue[i], is equal to(i))
+		})
+		this add("Queue negative indexing", func {
+			queue := this _createQueue(10, 5)
+			version(debugTests)
+				for (i in 1..queue count + 1)
+					((-i) toString() + " -> " + queue[-i] toString()) println()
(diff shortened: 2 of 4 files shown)


test/collections/VectorQueueTest.ooc
```diff
@@ -88,6 +88,42 @@ VectorQueueTest: class extends Fixture {
...
+			queue := VectorQueue<Int> new(10)
+			queue enqueue(0)
+			queue enqueue(1)
+			queue enqueue(2)
+			queue enqueue(3)
+			queue enqueue(4)
+			queue enqueue(5)
+			version(debugTests)
+				for (i in 0..queue count)
+					queue[queue count - 1 - i] toString() println()
```
davidhesselbom: This will only test 1 negative index: -1. The other indices will be positive.

source/collections/Queue.ooc
```diff
@@ -89,4 +89,8 @@ VectorQueue: class <T> extends Queue<T> {
 		this _capacity = newCapacity
 	}
 	toString: func -> String { "Queue: count=" << this _count toString() >> " capacity=" & this _capacity toString() }
+	operator [] (index: Int) -> T {
+		position := index > 0 ? this _head + index : this _tail + index
```
davidhesselbom: What should the index be if I want to access the tail element?

source/collections/Queue.ooc
```diff
@@ -89,4 +89,8 @@ VectorQueue: class <T> extends Queue<T> {
...
+		this _backend[position % this _capacity]
```
davidhesselbom: If `index > this _tail`, this will access `this _backend[something negative]`.

simonmika:
Thanks for the comments, fixed it.

source/collections/Queue.ooc
```diff
@@ -89,4 +89,8 @@ VectorQueue: class <T> extends Queue<T> {
 		this _capacity = newCapacity
 	}
 	toString: func -> String { "Queue: count=" << this _count toString() >> " capacity=" & this _capacity toString() }
+	operator [] (index: Int) -> T {
```
davidhesselbom: add
```version(safe)
    if (index >= this _count || index < -this _count)
        raise("Index out of bounds in Queue [] operator call")
```
or similar.

source/collections/Queue.ooc
```diff
@@ -89,4 +89,8 @@ VectorQueue: class <T> extends Queue<T> {
...
+		position := index >= 0 ? this _head + index : this _tail + index
+		this _backend[position % this _capacity]
```
davidhesselbom: To reiterate, you need to use `Int modulo`, because position be a negative number.

test/collections/VectorQueueTest.ooc
```diff
@@ -86,6 +86,42 @@ VectorQueueTest: class extends Fixture {
...
+			version(debugTests)
+				for (i in 1..queue count + 1)
+					queue[-i] toString() println()
+			expect(queue[-1], is equal to(5))
+			expect(queue[-2], is equal to(4))
+			expect(queue[-3], is equal to(3))
+			expect(queue[-4], is equal to(2))
+			expect(queue[-5], is equal to(1))
+			expect(queue[-6], is equal to(0))
+		})
```
davidhesselbom: It would be nice to have a test that ensures the Queue will still work even if you go across the "seam" of the bounding buffer, both with positive and negative indexing. To do that, you will need to add a number of things to the queue that is larger than its capacity, but remove items along way so that the count never exceeds capacity, or the Queue will just expand to accommodate more items.

PR #228, opened 2015-05-07
Added some necessary methods.

kmartinpettersson:
I must copy to save updates of a FloatComplexVectorList class object from Meta. Should i also add this method in FloatVectorList?

### DIFF
--- a/source/math/FloatVectorList.ooc
+++ b/source/math/FloatVectorList.ooc
@@ -48,9 +48,10 @@ FloatVectorList: class extends VectorList<Float> {
 	}
 	maxValue: Float {
 		get {
-			result: Float
+			result := Float negativeInfinity
 			for (i in 0..this count)
-				result = Float maximum(result, this[i])
+				if (result < this[i])
+					result = this[i]
 			result
 		}
 	}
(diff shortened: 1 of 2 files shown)


source/math/FloatVectorList.ooc
```diff
@@ -46,6 +46,14 @@ FloatVectorList: class extends VectorList<Float> {
 			result
 		}
 	}
+	maxValue: Float {
+		get {
+			result: Float
```
davidhesselbom: Set `result` to `Float negativeInfinity` instead, or `result` will be uninitialized in first iteration of the `for` loop.

source/math/FloatVectorList.ooc
```diff
@@ -46,6 +46,14 @@ FloatVectorList: class extends VectorList<Float> {
...
+			for (i in 0..this count)
+				result = Float maximum(result, this[i])
```
davidhesselbom: This is unnecessary work,
`if (result < this[i]) result = this[i]`
should be more effective.

kmartinpettersson:
Added Copy to FloatVectorList and updated maxValue.

PR #231, opened 2015-05-08
OpenGL maps now use a virtual use function (API BREAKING)

emilwestergren:
OpenGL maps now use a virtual use function instead of passing closure to base class. This breaks the Kean API when creating shader classes outside of Kean.


davidhesselbom:
Doesn't build:
> source/vidhance/vidflow/Gpu/Overlay.ooc:108:3 error waiting for all args to resolve
        super(This fragmentSource, context, false,

PR #233, opened 2015-05-08
Added align function that handles any positive alignment

emilwestergren:


source/math/IntExtension.ooc
```diff
@@ -64,8 +64,16 @@ extend Int {
 		this * this
 	}
 	align: static func (x: Int, align: Int) -> This {
-		align > 0 ? (x + align - 1) & ~(align - 1) : x
+		result := x
+		if (align > 0) {
+			remainder := x % align
+			if (remainder > 0) {
+				result = x + align - remainder
+			}
```
davidhesselbom: We don't need these brackets...

emilwestergren:
Updated.

PR #234, opened 2015-05-11
Add method getFirstElements() for VectorList.

bettinaselig:


bettinaselig:
Rewriting it once more.

PR #236, opened 2015-05-13
Add adaptTo() to FloatBox2D.

bettinaselig:


source/math/FloatBox2D.ooc
```diff
@@ -112,6 +112,13 @@ FloatBox2D: cover {
 	operator != (other: This) -> Bool { !(this == other) }
 	toIntBox2D: func -> IntBox2D { IntBox2D new(this left, this top, this width, this height) }
 	operator as -> String { this toString() }
+	adaptTo: func (other: This, weight: Float) -> This {
+		weightedAverage := func (first, second: Float) -> Float { (1 - weight) * first + weight * second }
```
davidhesselbom: Does this create a new `Closure`? Did you check if this leaks memory? Wouldn't it be better to make `weightedAverage` a static function in FloatExtension? Speaking of which, what does `weightedAverage` do, compared to the one named `linearInterpolation`?

bettinaselig: It's like `linearInterpolation`. Thanks, did not see the function.

PR #239, opened 2015-05-13
Add functions for FloatBox2D and getElements() in VectorList

bettinaselig:


source/math/FloatBox2D.ooc
```diff
@@ -80,9 +87,19 @@ FloatBox2D: cover {
...
+	contains: func ~FloatPoint2DVectorList (list: FloatPoint2DVectorList) -> VectorList<Bool> {
+		result := VectorList<Bool> new()
+		for (i in 0..list count) {
+			if (this contains(list[i]))
+				result add(true)
+			else
+				result add(false)
+		}
+		result
+	}
```
davidhesselbom: Instead of a list of `Bool` for each index in the list, wouldn't it be better to have list of `Int` with "valid" indices?

bettinaselig: I wanted to be consistent with the other contains functions. That is why I chose to do it like this.
In VectorList getElements() I use the outcome of this function.

davidhesselbom: How many elements do you expect there to be in the list?

bettinaselig: Usually 25 (if complexity is set to max, it is 100). And I expect that at least 2/3 should be ´true´ in the outcome of ´contains´.

davidhesselbom: So then you create a list of Booleans that you directly afterwards send into a function that extracts the "true" values from the "false" values? That seems like a lot of unnecessary work. I would suggest doing the work done in `FloatBox2D contains~FloatPoint2DVectorList` directly where you use it (which is in only one place, as far as I can tell, and it shouldn't need to be longer than 4 lines if you only add the "true" elements).

bettinaselig: I need the indices, not the actual values. So either I give back a list with indices or a list with booleans. And since the other contains functions give `Bool` back, I thought I do the same with this function. If you anyway think it's better to give back a list of indices, I change that.

bettinaselig: Done.

davidhesselbom: Looks like the test you made still needs to be updated.

bettinaselig:
Done and rebased.

source/math/FloatBox2D.ooc
```diff
@@ -53,10 +55,15 @@ FloatBox2D: cover {
 	pad: func ~fromFloat (pad: Float) -> This { this pad(pad, pad, pad, pad) }
 	pad: func ~fromSize (pad: FloatSize2D) -> This { this pad(pad width, pad width, pad height, pad height) }
 	pad: func ~fraction (pad: Float) -> This {
-		padX := pad * this width
-		padY := pad * this height
+		padX := (pad * this width) / 2
+		padY := (pad * this height) / 2
 		this pad(padX, padX, padY, padY)
```
davidhesselbom: Simpler:
`this pad(pad * this size / 2)`

source/math/FloatBox2D.ooc
```diff
@@ -53,10 +55,15 @@ FloatBox2D: cover {
...
 	}
+	shrink: func ~fraction (margin: Float) -> This {
+		marginX := (margin * this width) / 2
+		marginY := (margin * this height) / 2
+		this pad(-marginX, -marginX, -marginY, -marginY)
```
davidhesselbom: And here

source/math/FloatBox2D.ooc
bettinaselig: Smooth.

PR #247, opened 2015-05-24
Add init~square function for IntSize2D.

bettinaselig:


source/math/IntSize2D.ooc
```diff
@@ -27,6 +27,7 @@ IntSize2D: cover {
 	basisX: static This { get { This new(1, 0) } }
 	basisY: static This { get { This new(0, 1) } }
 	init: func@ (=width, =height)
+	init: func@ ~square (length: Float) { this width = this height = length }
```
davidhesselbom: Nice, but `length` should be an `Int`.

bettinaselig: Oops... of course... typical copy-paste-error.

davidhesselbom: It's funny, because the exact opposite fix was made in pull request #244.

PR #253, opened 2015-06-03
Fixed cleaning of recycled resources and some minor cleaning and bug fixes

emilwestergren:


source/draw/gpu/GpuImageBin.ooc
```diff
@@ -14,63 +14,64 @@
...
-		this _uv = FreeArrayList<GpuImage> new()
-		this _yuv422 = FreeArrayList<GpuImage> new()
+		this _monochrome = VectorList<GpuImage> new()
+		this _bgr = VectorList<GpuImage> new()
+		this _bgra = VectorList<GpuImage> new()
+		this _uv = VectorList<GpuImage> new()
+		this _yuv422 = VectorList<GpuImage> new()
+	}
+	_cleanList: func (list: VectorList<GpuImage>) {
+		list apply(|image| image _recyclable = false)
```
davidhesselbom: Did you check the `apply` call for memory leaks?

source/draw/gpu/GpuImageBin.ooc
```diff
@@ -14,63 +14,64 @@
...
-		this _bgra = FreeArrayList<GpuImage> new()
-		this _uv = FreeArrayList<GpuImage> new()
-		this _yuv422 = FreeArrayList<GpuImage> new()
+		this _monochrome = VectorList<GpuImage> new()
+		this _bgr = VectorList<GpuImage> new()
+		this _bgra = VectorList<GpuImage> new()
+		this _uv = VectorList<GpuImage> new()
+		this _yuv422 = VectorList<GpuImage> new()
+	}
+	_cleanList: func (list: VectorList<GpuImage>) {
```
davidhesselbom: This is a static function, yeah?

source/draw/gpu/GpuImageBin.ooc
```diff
@@ -82,19 +83,19 @@ GpuImageBin: class {
...
 		this _mutex unlock()
 	}
-	_search: func (size: IntSize2D, arrayList: FreeArrayList<GpuImage>) -> GpuImage {
+	_search: func (size: IntSize2D, list: VectorList<GpuImage>) -> GpuImage {
 		result := null
-		index := 0
-		for (i in 0..arrayList size) {
-			image := arrayList[i]
+		index := -1
+		count := list count
```
davidhesselbom: Why store `count` in a separate variable?

source/draw/gpu/GpuImageBin.ooc
davidhesselbom: Looks to me like this creates a Func, which you then have no way of freeing afterwards.

emilwestergren: Shame, I like the syntax but I guess we can't use it unless we fix the generated C code.

davidhesselbom: You could modify the `apply` function to have a Boolean argument that tells whether to free the Func when it's done applying stuff. Someone has to tell someone whether the Func should be freed. If you want the program itself to figure out when it's time to free something you've allocated, you use a garbage collector.

emilwestergren: I'll just make a for loop instead, it's only two lines of code anyway.

davidhesselbom: Thanks

emilwestergren: Well, you're very welcome sir.

emilwestergren:
Updated.

PR #256, opened 2015-06-08
Optimized shaders for packing and unpacking from Android RGBA buffer.

emilwestergren:


source/draw/gpu/opengl/OpenGLES3Context.ooc
```diff
@@ -192,4 +192,13 @@ OpenGLES3Context: class extends GpuContext {
 	}
 	update: func { this _backend swapBuffers() }
 	setViewport: func (viewport: Viewport) { Fbo setViewport(viewport offset width, viewport offset height, viewport resolution width, viewport resolution height) }
+	packToRgba: func (source: GpuImage, target: GpuBgra, viewport: Viewport) {
+		map := match(source) {
+			case image: GpuMonochrome => this _packMonochrome
+			case image: GpuUv => this _packUv
```
davidhesselbom: This gives me warnings about `image` being unused.
```ooc
			case source: GpuMonochrome => this _packMonochrome
			case source: GpuUv => this _packUv
```
seems to work just as well, and it gives no warnings.

emilwestergren:
Updated.

PR #258, opened 2015-06-10
Memory alignment is now defined by stride and offsets.

emilwestergren:


davidhesselbom:
Looks like this broke ImageFileTest. To run the test locally, go to your `ooc-kean` folder and run `./test.sh draw/ImageFile`

emilwestergren:
Fixed.

PR #276, opened 2015-06-18
Some small functions

bettinaselig:


source/math/FloatTransform2D.ooc
```diff
@@ -107,6 +107,18 @@ FloatTransform2D: cover {
...
+		a := Float linearInterpolation(first a, second a, weight)
+		b := Float linearInterpolation(first b, second b, weight)
+		c := Float linearInterpolation(first c, second c, weight)
+		d := Float linearInterpolation(first d, second d, weight)
+		e := Float linearInterpolation(first e, second e, weight)
+		f := Float linearInterpolation(first f, second f, weight)
+		g := Float linearInterpolation(first g, second g, weight)
+		h := Float linearInterpolation(first h, second h, weight)
+		i := Float linearInterpolation(first i, second i, weight)
+		This new(a, b, c, d, e, f, g, h, i)
```
davidhesselbom: As we discussed, this won't work for rotations, as they are not linear - we need to use quaternions or [slerp](https://en.wikipedia.org/wiki/Slerp)

Also, we should add a test to make sure the improved implementation *does* work.

PR #281, opened 2015-07-07
Copy function and Add to FloatComplexVectorList and FloatVectorList

kmartinpettersson:


source/math/FloatComplexVectorList.ooc
```diff
@@ -44,11 +45,8 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 	mean: func -> FloatComplex {
 		sum() / this _count
 	}
-	copy: func -> This {
-		result := This new(this _count)
-		for (i in 0..this _count)
-			result add(this[i])
-		result
+	copy: override func () -> This {
+		super() as This
```
davidhesselbom: Is this function actually needed in this class? What happens if you remove it?

source/math/FloatComplexVectorList.ooc
```diff
@@ -70,6 +68,11 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 			result
 		}
 	}
+	addInto: func (other: This) {
+		minimumCount := Int minimum(this count, other count)
+		for (i in 0..minimumCount)
+			this[i] = this[i] + other[i]
+	}
```
davidhesselbom: Can this be an operator instead of a function?

source/math/FloatVectorList.ooc
```diff
@@ -86,11 +86,8 @@ FloatVectorList: class extends VectorList<Float> {
 		}
 		result
 	}
-	copy: func -> This {
-		result := This new(this _count)
-		for (i in 0..this _count)
-			result add(this[i])
-		result
+	copy: override func () -> This {
+		super() as This
```
davidhesselbom: Is this function actually needed in this class? What happens if you remove it?

source/math/FloatComplexVectorList.ooc
davidhesselbom: Actually, nevermind, we'll save that for later.

PR #283, opened 2015-07-14
Quaternion class and tests

thomasfanell:


test/math/QuaternionTest.ooc
```diff
@@ -0,0 +1,88 @@
...
+	//
+	isWithinTolerance: func (target, actualValue, tolerance: Float) -> Bool {
+		actualValue >= target - tolerance abs() && actualValue <= target + tolerance abs()
+	}
+    toRadians: func (angle: Float) -> Float {
+		Float pi / 180.0f * angle
+	}
+	toDegrees: func (angle: Float) -> Float {
+		180.0f / Float pi * angle
+	}
```
davidhesselbom: `toRadians` and `toDegrees` should be added in `source/sdk/math.ooc` if they don't already exist.

thomasfanell: I can add them

source/math/Quaternion.ooc
```diff
@@ -21,27 +21,43 @@ import math
...
-	init: func@ ~floats (w: Float, x: Float, y: Float, z: Float) { this init(w, FloatPoint3D new(x, y, z)) }
+	init: func@ ~floats (w, x, y, z: Float) { this init(w, FloatPoint3D new (x, y, z)) }
 	init: func@ ~default { this init(0, 0, 0, 0) }
-	apply: func(vector: FloatPoint3D) -> FloatPoint3D {
- 		vectorQuaternion := Quaternion new(0.0f, vector)
+	apply: func (vector: FloatPoint3D) -> FloatPoint3D {
+ 		vectorQuaternion := Quaternion new (0.0f, vector)
 		result := hamiltonProduct(hamiltonProduct(this, vectorQuaternion), this inverse)
-		FloatPoint3D new(result x, result y, result z)
+		FloatPoint3D new (result x, result y, result z)
```
davidhesselbom: I would argue that `new` is a function, not a keyword, so no space after `new`.

source/math/Quaternion.ooc
```diff
@@ -91,7 +107,79 @@ Quaternion: cover {
 		else
 			roll = asin((upYawPitch scalarProduct(upRotated) * upYawPitch z - upRotated z) / rightYawPitch z)
 
-		FloatRotation3D new(pitch, -yaw, roll)
+		FloatRotation3D new (pitch, -yaw, roll)
+	}
+	norm: Float { get { (this real squared() + (this imaginary norm) squared()) sqrt() } }
```
davidhesselbom: A shorter syntax for `function: ReturnType { get { code } }` is `function ::= code` (see lines 26-29, for example)

source/math/Quaternion.ooc
```diff
@@ -50,14 +66,14 @@ Quaternion: cover {
...
+		
 		w := a1 * a2 - b1 * b2 - c1 * c2 - d1 * d2;
 		x := a1 * b2 + b1 * a2 + c1 * d2 - d1 * c2;
 		y := a1 * c2 - b1 * d2 + c1 * a2 + d1 * b2;
 		z := a1 * d2 + b1 * c2 - c1 * b2 + d1 * a2;
-		return Quaternion new(w, x, y, z);
+		return Quaternion new (w, x, y, z);
 	}
-	getEulerAngles: func() -> FloatRotation3D {
+	getEulerAngles: func () -> FloatRotation3D {
```
davidhesselbom: Functions with no arguments don't need an arguement list, so remove the `()`.

source/math/Quaternion.ooc
```diff
@@ -99,17 +187,39 @@ Quaternion: cover {
...
+	/*operator + (other: This) -> This {
+		This new (this w + other w, this x + other x, this y + other y, this z + other z)
+	}*/
 	operator + (other: This) -> This {
-		This new(this w + other w, this x + other x, this y + other y, this z + other z)
+		This new (this real + other real, this imaginary + other imaginary)
 	}
+	/*operator - (other: This) -> This {
+		This new (this w - other w, this x - other x, this y - other y, this z - other z)
+	}*/
```
davidhesselbom: Why keep the old code?

thomasfanell: Did not want to remove it until I was sure it worked - I am not used to this whole Git thing ;-)

source/math/Quaternion.ooc
```diff
@@ -91,7 +107,79 @@ Quaternion: cover {
...
+			result real = norm log()
+			result imaginary = (this imaginary / point3DNorm) * ((this real / norm) acos())
+		}
+		else {
+			result real = norm
+			result imaginary = FloatPoint3D new()
+		}
+		result
+	}
+	calculateExponential: func -> Quaternion {
```
davidhesselbom: You have a bunch of other functions named `rotationZ`, `distance` etc. Why not name these `logarithm` and `exponential`? They don't affect the original instance, so they should be considered properties.

thomasfanell: Simon said to keep those as methods

thomasfanell: I asked when to prefer properties and methods and vice versa... seems to be rather abstract when to use the one or the other... :-)

davidhesselbom: OK, so - 
 * if you need an argument, it's a function. 
 * if 
     * you don't need an argument and
     * you don't affect the original instance,

    it's a property.

source/math/Quaternion.ooc
```diff
@@ -99,17 +187,39 @@ Quaternion: cover {
...
+		imaginaryResult := this real * other imaginary + this imaginary * other real + this imaginary vectorProduct(other imaginary)
+		This new (realResult, imaginaryResult)
 	}
-	toString: func -> String { "Real:" + "%8f" format(this real) + " Imaginary: " + "%8f" format(this imaginary x) + " " + "%8f" format(this imaginary y) + " " + "%8f" format(this imaginary z)}
+	operator * (value: FloatPoint3D) -> FloatPoint3D {
+		(this * Quaternion new (0.0f, value) * this inverse) imaginary
+	}
+	toString: func -> String {
+		"Real:" + "%8f" format(this real) +
+		" Imaginary: " + "%8f" format(this imaginary x) + " " + "%8f" format(this imaginary y) + " " + "%8f" format(this imaginary z)}
```
davidhesselbom: This is going to leak memory.

```ooc
"Real: " << "%8f" formatFloat(this real) >>
" Imaginary: " & "%8f" formatFloat(this imaginary x) >> " " & "%8f" formatFloat(this imaginary y) >> " " & "%8f" formatFloat(this imaginary z)
```
should do the trick. The operators aren't pretty, but they work well when you get them right. For reference,
`<<` keeps the left argument, frees the right one, and returns the concatenation of both arguments
`>>` keeps the right argument, frees the left one, and returns the concatenation of both arguments
`&` frees both arguments and returns the concatenation of both arguments
`+` keeps both arguments and returns the concatenation of both arguments

source/math/Quaternion.ooc
```diff
@@ -91,7 +107,79 @@ Quaternion: cover {
...
+		norm := this norm
+		point3DNorm := this imaginary norm
+		if (point3DNorm != 0) {
+			result real = norm log()
+			result imaginary = (this imaginary / point3DNorm) * ((this real / norm) acos())
+		}
+		else {
+			result real = norm
+			result imaginary = FloatPoint3D new()
+		}
```
davidhesselbom: Shorter: 

```ooc
if (point3DNorm != 0)
	result = Quaternion new(norm log(), (this imaginary / point3DNorm) * ((this real / norm) acos()))
else
	result = Quaternion new(norm, FloatPoint3D new())
```

source/math/Quaternion.ooc
```diff
@@ -91,7 +114,73 @@ Quaternion: cover {
...
+	}
+	rotationY: Float {
+		get {
+			result: Float
+			value := this w * this y - this z * this x
+			if ((value abs() - 0.5f) abs() > this precision)
+				result = ((2.0f * (this w * this y - this z * this x)) clamp(-1, 1)) asin()
+			else
+				result = Float sign(value) * (Float pi / 2.0f)
```
davidhesselbom: If you switch this for
```ooc
			if ((value abs() - 0.5f) abs() < this precision)
				result = Float sign(value) * (Float pi / 2.0f)
			else
				result = ((2.0f * (this w * this y - this z * this x)) clamp(-1, 1)) asin()
```
do the tests still pass?

source/math/Quaternion.ooc
```diff
@@ -91,7 +114,73 @@ Quaternion: cover {
...
+	}
+	direction: FloatPoint3D {
+		get {
+			quaternionLogarithm := this logarithm
+			quaternionLogarithm imaginary / quaternionLogarithm imaginary norm
+		}
+	}
+	logarithm: Quaternion {
+		get {
+			result: Quaternion
```
davidhesselbom: You should be able to replace every occurence of `Quaternion` (except the first one) with `This`, including constructor calls, argument types, local variables, and return types. The word `Quaternion` should only occur once in this entire file, unless I'm mistaken.

source/math/Quaternion.ooc
thomasfanell: They do, yes

source/math/Quaternion.ooc
thomasfanell: Nice, will try

source/math/Quaternion.ooc
davidhesselbom: Then do that instead. Having `if ((value abs() - 0.5f) abs() > this precision)` in 2 places and `if ((value abs() - 0.5f) abs() < this precision)` in 1 place makes the latter look like a mistake, even though it isn't.

source/math/Quaternion.ooc
```diff
@@ -91,7 +114,73 @@ Quaternion: cover {
...
-		FloatRotation3D new(pitch, -yaw, roll)
+		FloatRotation3D new (pitch, -yaw, roll)
+	}
+	distance: func (other: Quaternion) -> Float {
+		(this - other) norm
+	}
+	rotationX: Float {
+		get {
+			result: Float
+			value := this w * this y - this z * this x
```
davidhesselbom: Any idea what `value` is? It's used in all 3 rotation[XYZ] functions. It might be better to have a property for this, if we can figure out what it actually means.

source/math/Quaternion.ooc
```diff
@@ -91,7 +114,73 @@ Quaternion: cover {
...
+				result = (2.0f * (this w * this x + this y * this z)) atan2(1.0f - 2.0f * (this x squared() + this y squared()))
+			result
+		}
+	}
+	rotationY: Float {
+		get {
+			result: Float
+			value := this w * this y - this z * this x
+			if ((value abs() - 0.5f) abs() > this precision)
+				result = ((2.0f * (this w * this y - this z * this x)) clamp(-1, 1)) asin()
```
davidhesselbom: Here, you compute `value` again. `result = ((2.0f * value) clamp(-1, 1)) asin()` looks cleaner.

thomasfanell: Wow, can't believe I missed that...

source/math/Quaternion.ooc
thomasfanell: Not a single clue. I'll promote this to a property when I find out.

source/math/Quaternion.ooc
```diff
@@ -100,10 +189,13 @@ Quaternion: cover {
 		this z == other z
```
davidhesselbom: This looks really wrong.

source/math/Quaternion.ooc
```diff
@@ -111,5 +203,19 @@ Quaternion: cover {
...
 		This new(this w * value, this x * value, this y * value, this z * value)
 	}
-	toString: func -> String { "Real:" + "%8f" format(this real) + " Imaginary: " + "%8f" format(this imaginary x) + " " + "%8f" format(this imaginary y) + " " + "%8f" format(this imaginary z)}
+	operator * (other: This) -> This {
+		realResult := this real * other real - this imaginary scalarProduct(other imaginary)
+		imaginaryResult := this real * other imaginary + this imaginary * other real + this imaginary vectorProduct(other imaginary)
+		This new(realResult, imaginaryResult)
+	}
+	operator * (value: FloatPoint3D) -> FloatPoint3D {
+		(this * Quaternion new (0.0f, value) * this inverse) imaginary
```
davidhesselbom: Can this `Quaternion` be replaced with `This`?

thomasfanell: Amended and pushed.

PR #284, opened 2015-07-14
Added median and movingMedian methods to FloatVectorList

alexandersundin:


source/math/FloatVectorList.ooc
```diff
@@ -212,4 +212,33 @@ FloatVectorList: class extends VectorList<Float> {
 		previous free()
 		result
 	}
+	median: func -> Float {
+		result: Float
+		tempVector := this copy()
+		tempVector sort()
+		if(Int odd(this count))
```
davidhesselbom: There should be a space between `if` and `(`. Our style checker will fix things like these, but for your own sake, make a habit out of being consistent.

source/math/FloatVectorList.ooc
```diff
@@ -212,4 +212,33 @@ FloatVectorList: class extends VectorList<Float> {
 		previous free()
 		result
 	}
+	median: func -> Float {
+		result: Float
+		tempVector := this copy()
+		tempVector sort()
```
davidhesselbom: `tempVector` is never freed, so this will cause a memory leak.

source/math/FloatVectorList.ooc
```diff
@@ -212,4 +212,33 @@ FloatVectorList: class extends VectorList<Float> {
...
+		else
+			result = (tempVector[this count / 2 - 1] + tempVector[this count / 2]) / 2
+		result
+	}
+	movingMedianFilter: func (windowSize: Int) -> This {
+		result := FloatVectorList new()
+		indices := VectorList<Int> new()
+		for (i in 0..(windowSize - 1) / 2) {
+			indices add(i)
+		}
```
davidhesselbom: We don't need these brackets.

source/math/FloatVectorList.ooc
davidhesselbom: Also, there are ways of finding the median that do not require sorting the data, but I don't expect you to have time to implement one. Here's one in C++, though: http://discuss.codechef.com/questions/1489/find-median-in-an-unsorted-array-without-sorting-it

source/math/FloatVectorList.ooc
```diff
@@ -212,4 +212,33 @@ FloatVectorList: class extends VectorList<Float> {
...
+		for (i in 0..this count) {
+			if (i + (windowSize - 1) / 2 < this count)
+				indices add(i + (windowSize - 1) / 2)
+			if (i > (windowSize - 1) / 2)
+				indices removeAt(0)
+
+			elementsInWindow := this getElements(indices)
```
davidhesselbom: This is not a very efficient way of getting a slice of a VectorList. Add a `getRange` function in VectorList and use that instead. The `getRange` function should take a `Range` cover as its argument (this will let us call it as `this getElements(start .. stop).

davidhesselbom:
Fixed according to the comments myself in #288.

PR #285, opened 2015-07-16
implemented more operators and some more tests

thomasfanell:
* implemented more comparison operators and operator []
* implemented conjugate property and toArray()
* added more tests
* fixed a minor error in a comment in FloatTransform3D

### DIFF
--- a/source/math/FloatTransform3D.ooc
+++ b/source/math/FloatTransform3D.ooc
@@ -20,7 +20,7 @@ import FloatPoint3D
 import text/StringTokenizer
 import structs/ArrayList
 
-// The 2D transform is a 3x3 homogeneous coordinate matrix.
+// The 3D transform is a 4x3 homogeneous coordinate matrix.
 // The element order is:
 // A D G J
 // B E H K
--- a/source/math/Quaternion.ooc
+++ b/source/math/Quaternion.ooc
@@ -16,6 +16,7 @@
 import FloatPoint3D
 import FloatRotation3D
 import FloatTransform2D
+import FloatTransform3D
 import math
 
 Quaternion: cover {
@@ -45,6 +46,7 @@ Quaternion: cover {
 	init: func@ (=real, =imaginary)
 	init: func@ ~floats (w, x, y, z: Float) { this init(w, FloatPoint3D new (x, y, z)) }
 	init: func@ ~default { this init(0, 0, 0, 0) }
+	init: func@ ~floatArray (source: Float[]) { this init(source[0], source[1], source[2], source[3]) }
 	apply: func (vector: FloatPoint3D) -> FloatPoint3D {
  		vectorQuaternion := This new(0.0f, vector)
 		result := hamiltonProduct(hamiltonProduct(this, vectorQuaternion), this inverse)
@@ -184,10 +186,22 @@ Quaternion: cover {
 		}
 	}
 	operator == (other: This) -> Bool {
-		this w == other w &&
-		this x == other x &&
-		this y == other y &&
-		this z == other z
+		this w == other w && this x == other x && this y == other y && this z == other z
+	}
+	operator != (other: This) -> Bool {
+		!(this == other)	
+	}
+	operator < (other: This) -> Bool {
+		this w < other w && this x < other x && this y < other y && this z < other z
+	}
+	operator > (other: This) -> Bool {
+		this w > other w && this x > other x && this y > other y && this z > other z
+	}
+	operator <= (other: This) -> Bool {
+		this w <= other w && this x <= other x && this y <= other y && this z <= other z
+	}
+	operator >= (other: This) -> Bool {
+		this w >= other w && this x >= other x && this y >= other y && this z >= other z
 	}
 	operator + (other: This) -> This {
 		This new(this real + other real, this imaginary + other imaginary)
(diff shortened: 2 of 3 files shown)


davidhesselbom:
`QuaternionTest` is still failing on Travis:
```
Quaternion ..........f... failed

Failed test: ./test/math/QuaternionTest.ooc
```
As before, it looks like `LogarithmExponential` is where things go wrong. The test framework currently doesn't let us see exactly which `expect` causes the test to fail, so can you please try to divide that section up into smaller pieces?

source/math/FloatTransform3D.ooc
```diff
@@ -20,7 +20,7 @@ import FloatPoint3D
 import text/StringTokenizer
 import structs/ArrayList
 
-// The 2D transform is a 3x3 homogeneous coordinate matrix.
+// The 3D transform is a 4x3 homogeneous coordinate matrix.
```
davidhesselbom: Oh, dear. Thanks!

thomasfanell:
I'll split the test up and see if I can see what the problem is.

source/math/Quaternion.ooc
```diff
@@ -210,8 +224,51 @@ Quaternion: cover {
...
+			case 2 => result = this y
+			case 3 => result = this z
+			case => raise("Invalid index: #{index}, valid indices are 0-3.")
+		}
+		result
+	}
+	operator as -> String { this toString() }
+	toArray: func -> Float[] {
+		result := [this w, this x, this y, this z]
+		result
```
davidhesselbom: `result` is useful in multiple return functions, but here, it just adds an extra line.
```ooc
	toArray: func -> Float[] {
		[this w, this x, this y, this z]
```
should suffice.

thomasfanell: Tried that actually, did not compile. Hence the temp variable

test/math/QuaternionTest.ooc
```diff
@@ -28,12 +47,35 @@ QuaternionTest: class extends Fixture {
 		this add("norm", func() {
 			expect(this quaternion0 norm, is equal to(65.5991592f))
 		})
+		this add("conjugate", func() {
+			conjugate := quaternion5 conjugate
+			result := conjugate w == quaternion5 w &&
+				conjugate[1] == -quaternion5[1] &&
+				conjugate[2] == -quaternion5[2] &&
+				conjugate[3] == -quaternion5[3]
+			expect(result)
```
davidhesselbom: Wouldn't this be better as 4 `expect`s?

source/math/Quaternion.ooc
```diff
@@ -210,8 +224,51 @@ Quaternion: cover {
 		This new(realResult, imaginaryResult)
 	}
 	operator * (value: FloatPoint3D) -> FloatPoint3D {
-		(this * Quaternion new (0.0f, value) * this inverse) imaginary
+		(this * This new (0.0f, value) * this inverse) imaginary
+	}
+	operator [] (index: Int) -> Float {
+		result: Float
+		match(index) {
```
davidhesselbom: `match` is a keyword, so there should be a space between it and `(`.

test/math/QuaternionTest.ooc
```diff
@@ -8,11 +8,30 @@ QuaternionTest: class extends Fixture {
...
+			expect(this quaternion0 <= this quaternion0)
+			expect(this quaternion0 >= this quaternion0)
+			expect(this quaternion6 < this quaternion5)
+			expect(this quaternion5 > this quaternion6)
+		})
+		this add("indexing", func() {
+			expect(this quaternion0[0] == 33.0f)
+			expect(this quaternion0[1] == 10.0f)
+			expect(this quaternion0[2] == -12.0f)
+			expect(this quaternion0[3] == 54.5f)
```
davidhesselbom: Also add a try - catch for indexing out of bounds.

source/math/Quaternion.ooc
```diff
@@ -210,8 +224,51 @@ Quaternion: cover {
...
+			case 0 => result = this w
+			case 1 => result = this x
+			case 2 => result = this y
+			case 3 => result = this z
+			case => raise("Invalid index: #{index}, valid indices are 0-3.")
```
davidhesselbom: It might be helpful to mention, in the error message, in which class the exception occurred.

source/math/Quaternion.ooc
davidhesselbom: Interesting. I'll give it a try.

test/math/QuaternionTest.ooc
thomasfanell: Do you mean I should index out of bounds and catch it?

source/math/FloatTransform3D.ooc
davidhesselbom: Next time, though, this type of thing should be in its own commit, because it's unrelated to the rest.

test/math/QuaternionTest.ooc
davidhesselbom: Yes, make sure the exception is thrown as desired.

thomasfanell: I added this test in the amended version

source/math/Quaternion.ooc
davidhesselbom: I'm pretty sure this was fixed in https://github.com/fasterthanlime/rock/commit/1b88d707e8959bcba9da34a8cba5267f4b0e5f7b, but our version of rock doesn't have that fix.

PR #291, opened 2015-07-20
Added API functions and classes for sensor sampling

emilwestergren:


davidhesselbom:
This broke DateTimeTest. If we no longer need that test, remove it. Otherwise, fix it.

emilwestergren:
Updated.

PR #293, opened 2015-07-22
Added getOnes, getZeros and divideByMaxValue to FloatVectorList

kmartinpettersson:


source/math/FloatVectorList.ooc
```diff
@@ -141,6 +141,21 @@ FloatVectorList: class extends VectorList<Float> {
 			result = result >> this[i] toString() >> "\n"
 		result
 	}
+	divideByMaxValue: func () -> This {
+		this maxValue != 0 ? (this / this maxValue) copy() : this copy()
```
davidhesselbom: `(this / this maxValue)` creates a new list, so there's no reason to do `(this / this maxValue) copy()`, right?

kmartinpettersson: Yes, that the * operator does, deleted!

PR #295, opened 2015-07-23
Replaced __destroy__ with proper free() methods.

thomasfanell:
* replaced __destroy__ with proper free() methods.
* minor typographic improvements

### DIFF
--- a/source/collections/PointerVector.ooc
+++ b/source/collections/PointerVector.ooc
@@ -23,10 +23,13 @@ PointerVector: abstract class {
 
 	init: /* protected */ func ~preallocated (=_backend, =_count, freeContent := false) {}
 	init: /* protected */ func (=_count, freeContent := false) {
-		this _allocate(this _count)
 		this _freeContent = freeContent
 	}
-	__destroy__: func {	this _free(0, this count) }
+	free: override func {
+		this _free(0, this count)
+		gc_free(this _backend)
+		super()
+	}
 	_free: func ~range (start: Int, end: Int) {
 		if (this _freeContent) {
 			for (i in start..end) {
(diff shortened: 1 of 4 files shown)


source/collections/PointerVector.ooc
```diff
@@ -72,43 +75,31 @@ PointerVector: abstract class {
...
-	init: func(data: Pointer*, count: Int) {
+	init: func (data: Pointer*, count: Int) {
 		super(data, count)
 	}
-
-	// TODO: Why does this function exist here?
-	_allocate: func(count: Int)
-
-	resize: override func(count: Int) {
+	_allocate: func (count: Int)
```
davidhesselbom: Did you figure out the purpose of this function?

davidhesselbom: I'm not merging this. Remove this function here and in the base class, since it's not used in either place. Keep the function in `PointerHeapVector` where it's actually used.

thomasfanell: Fixed.

thomasfanell: I just noticed that this issue is present in Vector as well, will amend.

source/collections/Vector.ooc
```diff
@@ -23,31 +23,32 @@ Vector: abstract class <T> {
...
-						old free()
+				old := this[i] as Object
+				if (old != null)
+					old free()
 			}
 		}
 	}
-	_allocate: abstract func (capacity: Int)
-
+	//_allocate: abstract func (capacity: Int)
```
davidhesselbom: ?

davidhesselbom:
Removing empty lines is nice, but should go in their own commit.

PR #296, opened 2015-07-23
What's up with fancy_backtrace?

davidhesselbom:


zhaihj:
Do you mean the message "[lang/Backtrace] Falling back on execinfo.. (build extension if you want fancy backtraces)"?

Several month ago I found that rock may print extra(unnecessary?) messages (especially under windows). It may be helpful?

Related commit: https://github.com/zhaihj/rock/commit/60d4b44a126a953c65ab662f64a97286ffbb66ba

davidhesselbom:
How do you mean it's "unnecessary"? Does the message ever print even when a backtrace is not requested (which, as far as I know, only happens when the program crashes)?

davidhesselbom:
The problem I'm trying to solve, btw, is that I can't seem to get backtraces to work on Travis when using my pre-compiled binaries for rock and fancy_backtrace.

zhaihj:
To be honest, I don't really remember all the details of this commit. The only thing I remeber is that under windows, `sam test` prints lots of `[lang/Backtrace]` before `[passed]` messagess appear. Just like this:
```
IntSize2D ......... done
[lang/Backtrace] Falling back on execinfo.. (build extension if you want fancy backtraces)
Quaternion ............................ done
```

I will try to reproduce your problem and find out the reason ...

davidhesselbom:
Oh, I see. Travis doesn't run on Windows, though, and we're using our own test framework, not `sam test` (but I'm not sure why... As I remember it, sam was too limited when we started using ooc).

If you find any reason why it happens, let me know! It turns out backtrace actually does work, but I still get the messages that it hasn't been built: https://travis-ci.org/cogneco/ooc-kean/builds/72277903

PR #301, opened 2015-07-27
base for TimeSpan and DateTime tests

sebastianbaginski:
Moved TimeSpan to separate file, created classes for DateTime and TimeSpan tests


source/base/DateTime.ooc
```diff
@@ -14,22 +14,34 @@
...
 */
-TimeSpan: cover {
-	_ticks: Int64
-	ticks ::= this _ticks
-	init: func@ (=_ticks)
-	kean_base_timeSpan_getTicks: unmangled func -> Int64 { this _ticks }
-	new: unmangled(kean_base_timeSpan_new) static func ~API (ticks: UInt64) -> This { This new(ticks) }
-}
+
+import TimeSpan
```
davidhesselbom: No, `use ooc-base` instead.

source/base/TimeSpan.ooc
```diff
@@ -0,0 +1,31 @@
...
+* This software is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with this software. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import DateTime
```
davidhesselbom: No, `use ooc-base` instead.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,34 @@
...
 	new: unmangled(kean_base_dateTime_new) static func ~API (ticks: UInt64) -> This { This new(ticks) }
 
+	TicksPerMillisecond : static const Int = 1000
+	TicksPerSecond : static const Int = TicksPerMillisecond*1000
+	/* number of days in year ( non-leap ) */
+	DaysInYear : static const Int = 365
+
+	/* validate argument ranges for hour/minutes/seconds vaules */
+	isHourValid : static func (hour : Int, minute : Int, second : Int) -> Bool {
+		(hour>=0 && hour < 24 && minute >=0 && minute < 60 && second >=0 && second < 60)
```
davidhesselbom: `hour >= 0`, `minute >= 0`, `second >= 0` (be consistent about how you use spaces)

source/base/DateTime.ooc
```diff
@@ -14,22 +14,34 @@
...
+
+import TimeSpan
 
 DateTime: cover {
 	/* Number of 100 ns intervals since 00.00 1/1/1 */
 	_ticks: UInt64
 	ticks ::= this _ticks
-
 	init: func@ (=_ticks)
+	init: func ~fromYearMonthDay (year : Int, month: Int, day: Int) {}
```
davidhesselbom: This constructor doesn't appear to do anything at all.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,34 @@
...
+	init: func ~fromHourMinuteSec (hour : Int, minute : Int, second : Int){
```
davidhesselbom: `) {`

davidhesselbom: Also,
`init: func ~fromHourMinuteSec (hour, minute, second: Int)`

davidhesselbom: I'm pretty sure it needs to be `func@` for all the `init` functions in `cover`s.

source/base/TimeSpan.ooc
```diff
@@ -0,0 +1,31 @@
...
+
+TimeSpan: cover {
+
+	timeToTicks : static func(hours: Int, minutes: Int, seconds : Int) -> Int64 {
```
davidhesselbom: `func (`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,34 @@
...
+		if ( isHourValid(hour,minute,second) ){
+			this _ticks = TimeSpan timeToTicks(hour,minute,second)
+		} else {
+			raise ("invalid input specified for constructor(hour,minute,second)")
+		}
```
davidhesselbom: Don't add braces `{ }` if you don't need them.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,34 @@
...
 DateTime: cover {
 	/* Number of 100 ns intervals since 00.00 1/1/1 */
 	_ticks: UInt64
 	ticks ::= this _ticks
-
 	init: func@ (=_ticks)
+	init: func ~fromYearMonthDay (year : Int, month: Int, day: Int) {}
+	init: func ~fromHourMinuteSec (hour : Int, minute : Int, second : Int){
+		if ( isHourValid(hour,minute,second) ){
+			this _ticks = TimeSpan timeToTicks(hour,minute,second)
```
davidhesselbom: Space after comma: `hour, minute, second`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,34 @@
...
 	kean_base_dateTime_getTicks: unmangled func -> UInt64 { this _ticks }
 	new: unmangled(kean_base_dateTime_new) static func ~API (ticks: UInt64) -> This { This new(ticks) }
 
+	TicksPerMillisecond : static const Int = 1000
+	TicksPerSecond : static const Int = TicksPerMillisecond*1000
+	/* number of days in year ( non-leap ) */
+	DaysInYear : static const Int = 365
+
+	/* validate argument ranges for hour/minutes/seconds vaules */
+	isHourValid : static func (hour : Int, minute : Int, second : Int) -> Bool {
```
davidhesselbom: `isHourValid` sounds like a question... `hourIsValid` or `isValidHour` would be better.

source/base/TimeSpan.ooc
```diff
@@ -0,0 +1,31 @@
...
+		hours * 3600 * minutes * 60 * seconds * DateTime TicksPerSecond
```
davidhesselbom: Pretty sure that should be `(hours * 3600 + minutes * 60 + seconds) * DateTime TicksPerSecond`...

source/base/TimeSpan.ooc
davidhesselbom: Any particular reason why this function belongs in `TimeSpan` instead of `DateTime`?

source/base/TimeSpan.ooc
```diff
@@ -0,0 +1,76 @@
...
+	operator > (other: TimeSpan) -> Bool {
+		this compareTo(other) == Order greater
+	}
+	operator < (other: TimeSpan) -> Bool {
+		this compareTo(other) == Order less
+	}
+	operator >= (other: TimeSpan) -> Bool {
+		! (this < other)
+	}
+	operator <= (other: TimeSpan) -> Bool {
```
davidhesselbom: The word `TimeSpan` should only be needed once: `TimeSpan: cover {`. All the other occurrences can be replaced by `This` (with a capital 'T').

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
-TimeSpan: cover {
-	_ticks: Int64
-	ticks ::= this _ticks
-	init: func@ (=_ticks)
-	kean_base_timeSpan_getTicks: unmangled func -> Int64 { this _ticks }
-	new: unmangled(kean_base_timeSpan_new) static func ~API (ticks: UInt64) -> This { This new(ticks) }
+
+use ooc-base
+
+DateTimeData: class {
```
davidhesselbom: Is there any reason why `DateTimeData` is a class and not a cover?

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+		}
+		month := 0
+		for (m in 1 .. 13) {
+			t := This ticksInMonth(year, m)
+			if (ticksLeft < t) {
+				month = m
+				break
+			} else {
+				ticksLeft -= t
+			}
```
davidhesselbom: You / we don't need the `else` brackets.

davidhesselbom: ... I think.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+		result day = days + 1
+		result hour = hour
+		result minute = minute
+		result second = second
+		result millisecond = millisecond
+		result
+	}
+
+	/* returns number of ticks for given hours, minutes and seconds */
+	timeToTicks: static func(hours, minutes, seconds, millisecond: Int) -> Int64 {
```
davidhesselbom: `func (`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+
+	ticksInMonth: static func (year, month: Int) -> UInt64 {
+		This TicksPerDay * This daysInMonth(year, month)
+	}
+	/* validate argument ranges for hour/minutes/seconds vaules */
+	timeIsValid: static func (hour, minute, second, millisecond: Int) -> Bool {
+		hour >= 0 && hour < 24 && minute >= 0 && minute < 60 && second >= 0 && second < 60 && millisecond >= 0 && millisecond < 1000
+	}
+	/* validate argument ranges for year/month/day values */
+	dateIsValid: static func (year, month, day : Int) -> Bool {
```
davidhesselbom: `day: Int`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+		year >= 1 && month >= 1 && month <= 12 && day >= 1 && day <= daysInMonth(year, month)
```
davidhesselbom: Should we also check that `year` isn't too large?

source/base/TimeSpan.ooc
```diff
@@ -0,0 +1,115 @@
...
+* along with this software. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use ooc-base
+
+TimeSpan: cover {
+	_ticks: Int64
+	ticks ::= this _ticks
+	init: func@ (=_ticks)
+	init: func@ ~fromHourMinuteSec (hour, minute, second, millisecond : Int) {
```
davidhesselbom: `millisecond: Int`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+			}
+			fourYearBlocks := (year - 1) / 4
+			year_start := fourYearBlocks * 4
+			for (y in year_start + 1 .. year)
+				totalDays += This daysInYear(y)
+			totalDays += fourYearBlocks * This DaysInFourYears
+			totalDays * This TicksPerDay
+		} else {
+			0
+		}
```
davidhesselbom: Another few extra brackets here.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+	DaysInFourYears: static const Int = 3 * 365 + 366
+	TicksPerMillisecond: static const Int64 = 1000
+	TicksPerSecond: static const Int64 = TicksPerMillisecond * 1000
+	TicksPerMinute: static const Int64 = TicksPerSecond * 60
+	TicksPerHour: static const Int64 = TicksPerMinute * 60
+	TicksPerDay: static const Int64 = TicksPerHour * 24
+	TicksPerWeek: static const UInt64 = TicksPerDay * 7
+	TicksPerFourYears: static const Int64 = DaysInFourYears * TicksPerDay
+	/* default date/time printing format */
+	DefaultFormat: static const String = "%yyyy-%MM-%dd %hh:%mm:%ss::%zzzz"
```
davidhesselbom: For no particular reason, constants like these are not named with a capital in the rest of the code, so they shouldn't be that way here, either.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+	}
+	operator >= (other: This) -> Bool {
+		this compareTo(other) != Order less
+	}
+
+	/* number of days in year ( non-leap ) */
+	DaysInYear: static const Int = 365
+	DaysInFourYears: static const Int = 3 * 365 + 366
+	TicksPerMillisecond: static const Int64 = 1000
+	TicksPerSecond: static const Int64 = TicksPerMillisecond * 1000
```
davidhesselbom: `This ticksPerMillisecond`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,246 @@
...
+		ticksLeft -= second * This TicksPerSecond
+		millisecond := ticksLeft / This TicksPerMillisecond
+		result year = year
+		result month = month
+		result day = days + 1
+		result hour = hour
+		result minute = minute
+		result second = second
+		result millisecond = millisecond
+		result
```
davidhesselbom: If you give `DateTimeData` a constructor, you won't need to write all this on separate lines.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,253 @@
...
+		result = result replaceAll("%yy", "%d" format(data year % 100))
+		result = result replaceAll("%MM", (data month < 10 ? "0%d" : "%d") format(data month))
+		result = result replaceAll("%dd", (data day < 10 ? "0%d" : "%d") format(data day))
+		result = result replaceAll("%M", "%d" format(data month))
+		result = result replaceAll("%d", "%d" format(data day))
+		result = result replaceAll("%hh", "%d" format(data hour))
+		result = result replaceAll("%mm", "%d" format(data minute))
+		result = result replaceAll("%ss", "%d" format(data second))
+		result = result replaceAll("%zzzz", "%d" format(data millisecond))
+		result
```
davidhesselbom: This is going to be fun to leak-check... :)

source/base/DateTime.ooc
```diff
@@ -14,22 +14,253 @@
...
+	operator > (other: This) -> Bool {
+		this compareTo(other) == Order greater
+	}
+	operator >= (other: This) -> Bool {
+		this compareTo(other) != Order less
+	}
+
+	/* number of days in year ( non-leap ) */
+	DaysInYear: static const Int = 365
+	DaysInFourYears: static const Int = 3 * 365 + 366
```
davidhesselbom: This could be more correct... Also, it doesn't use `This daysInYear`.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,253 @@
...
+	TicksPerMillisecond: static const Int64 = 1000
+	TicksPerSecond: static const Int64 = This TicksPerMillisecond * 1000
+	TicksPerMinute: static const Int64 = TicksPerSecond * 60
```
davidhesselbom: `This ticksPerSecond` and so on. Every time you use static things from the same class, prepend `This`.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,253 @@
...
+	TicksPerHour: static const Int64 = TicksPerMinute * 60
+	TicksPerDay: static const Int64 = TicksPerHour * 24
+	TicksPerWeek: static const UInt64 = TicksPerDay * 7
+	TicksPerFourYears: static const Int64 = DaysInFourYears * TicksPerDay
+	/* default date/time printing format */
+	DefaultFormat: static const String = "%yyyy-%MM-%dd %hh:%mm:%ss::%zzzz"
+
+	now: static DateTime {
+      get {
```
davidhesselbom: Spaces

source/base/DateTime.ooc
```diff
@@ -14,22 +14,253 @@
...
+					DateTime new(Time year(), Time month(), Time day(), Time hour(), Time min(), Time sec(), Time microsec()/1000)
```
davidhesselbom: Hmm. What happens if the time changes while we run this function, if, say, we call this function the microsecond before the new year? Can we somehow ask the OS for the current time *once* and then build our `DateTime` from that one timestamp instead?

source/base/DateTime.ooc
davidhesselbom: For clarity, this function should use a `return` variable.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,253 @@
...
+		result
+	}
+
+	/* returns number of ticks for given hours, minutes and seconds */
+	timeToTicks: static func(hours, minutes, seconds, millisecond: Int) -> Int64 {
+		(hours * 3600 + minutes * 60 + seconds) * This TicksPerSecond + millisecond * This TicksPerMillisecond
+	}
+
+	/* returns number of ticks for given date at 0:00*/
+	dateToTicks: static func(year, month, day : Int) -> UInt64 {
```
davidhesselbom: `func (`, `day: Int`

source/base/DateTime.ooc
sebastianbaginski: Yes, I'm gonna refactor it to use the system call only once. Should be more efficient as well.

source/base/DateTime.ooc
sebastianbaginski: I was thinking about it too, as well as about possible overflows in TimeSpan class. I asked Simon and he said "you shouldn't care about that too much, unless you can prove me that such problems will happen very often".

davidhesselbom: I guess they won't.

davidhesselbom: By the same rationale, the `DaysInFourYears` constant also shouldn't cause much trouble since it's true for the years 1900-2100...

source/base/DateTime.ooc
```diff
@@ -14,22 +14,251 @@
...
+	second: Int { get set }
+	millisecond: Int { get set }
+	init: func@ (year, month, day, hour, minute, second, millisecond: Int) {
+		this year = year
+		this month = month
+		this day = day
+		this hour = hour
+		this minute = minute
+		this second = second
+		this millisecond = millisecond
```
davidhesselbom: I'm not 100% sure this is going to work with properties, but, it might be possible to write this as 
```ooc
DateTimeData: class {
	year: Int { get set }
	month: Int { get set }
	day: Int { get set }
	hour: Int { get set }
	minute: Int { get set }
	second: Int { get set }
	millisecond: Int { get set }
	init: func@ (=year, =month, =day, =hour, =minute, =second, =millisecond)
}
```
Give it a try!

source/base/DateTime.ooc
```diff
@@ -14,22 +14,251 @@
...
-	_ticks: Int64
-	ticks ::= this _ticks
-	init: func@ (=_ticks)
-	kean_base_timeSpan_getTicks: unmangled func -> Int64 { this _ticks }
-	new: unmangled(kean_base_timeSpan_new) static func ~API (ticks: UInt64) -> This { This new(ticks) }
+
+use ooc-base
+import os/Time
+
+DateTimeData: class {
```
davidhesselbom: Did you try making this a cover? Didn't it work, or why is this still a class?

source/base/DateTime.ooc
```diff
@@ -14,22 +14,251 @@
...
+	year: Int { get set }
+	month: Int { get set }
+	day: Int { get set }
+	hour: Int { get set }
+	minute: Int { get set }
+	second: Int { get set }
+	millisecond: Int { get set }
```
davidhesselbom: I guess it doesn't really matter, but, do we need `Int` for all these values? Most of them will never have values above 255.

source/base/DateTime.ooc
davidhesselbom: By the way, what is the purpose of this class?

source/base/DateTime.ooc
```diff
@@ -14,22 +14,251 @@
...
+	daysPerFourYears: static const Int = 3 * This daysPerYear + 366
+	ticksPerMillisecond: static const Int64 = 1000
+	ticksPerSecond: static const Int64 = This ticksPerMillisecond * 1000
+	ticksPerMinute: static const Int64 = This ticksPerSecond * 60
+	ticksPerHour: static const Int64 = This ticksPerMinute * 60
+	ticksPerDay: static const Int64 = This ticksPerHour * 24
+	ticksPerWeek: static const UInt64 = This ticksPerDay * 7
+	ticksPerFourYears: static const Int64 = This daysPerFourYears * This ticksPerDay
+	/* default date/time printing format */
+	DefaultFormat: static const String = "%yyyy-%MM-%dd %hh:%mm:%ss::%zzzz"
```
davidhesselbom: `defaultFormat`, without capitalizing the first letter.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,251 @@
...
+	ticksInMonth: static func (year, month: Int) -> UInt64 {
+		This ticksPerDay * This daysInMonth(year, month)
+	}
+	/* validate argument ranges for hour/minutes/seconds vaules */
+	timeIsValid: static func (hour, minute, second, millisecond: Int) -> Bool {
+		hour >= 0 && hour < 24 && minute >= 0 && minute < 60 && second >= 0 && second < 60 && millisecond >= 0 && millisecond < 1000
+	}
+	/* validate argument ranges for year/month/day values */
+	dateIsValid: static func (year, month, day: Int) -> Bool {
+		year >= 1 && month >= 1 && month <= 12 && day >= 1 && day <= daysInMonth(year, month)
```
davidhesselbom: There's a function `in?` for ranges, so `month >= 1 && month <= 12` could have been written `month in?(1..12)` etc. Unfortunately, `1..12` means the values 1 to 12, not 1 through 12. Still, I think we should consider using or making a function that can check whether a value is within some boundary so we don't have to do `value > number && value < otherNumber` checks like this. Something like `value isWithin(1..12)` looks much cleaner, imo.

source/base/DateTime.ooc
sebastianbaginski: It works, thanks!

source/sdk/os/Time.ooc
```diff
@@ -135,7 +135,7 @@ Time: class {
         }
         return -1
     }
-	
+
```
davidhesselbom: Long-term, we intend to make the SDK completely independent from the official one, but until then, I think we should avoid making any changes to it that we can avoid, in order to make it more compatible with the official one so that we can benefit from any improvements made there. Whitespace changes qualify as changes we can avoid. I guess your text editor made this change for you, but when it happens, don't commit the change.

source/base/DateTime.ooc
sebastianbaginski: Making it a 'cover' seems to be working, thanks.
I found it useful when implementing the _ticksToDateTimeHelper method, just to hold the result data. It also simplifies the implementation of 'toString' as all the computations needed to get the year, month, day etc. are performed only once and the result is stored in this 'structure'. 
It could also be useful in holding the results of current date/time returned by Time class, as I'm not sure if I can use 'DateTime' module directly in the 'sdk/os/Time' class. Time could simply return this 'raw' date/time data so we could use it in DateTime class.

davidhesselbom: No, you shouldn't use `DateTime` in the `Time` class. You could extend the `Time` class, though, and use `DateTime` there, if it helps.

davidhesselbom: Also, glad it can be a cover. In case no one told you, covers are like structs, i.e. stack allocated, so you don't have to free them yourself, which is nice. ooc isn't as good at cleaning things up when they go out of scope as, say, C++ is.

source/base/DateTime.ooc
sebastianbaginski: Ok, thanks.

source/sdk/os/Time.ooc
sebastianbaginski: I see, so I guess you won't like my latest commit :)

source/base/DateTime.ooc
sebastianbaginski: btw. what about spacing convention with the 'in?' operator ?
should it be "month in?(1..12)" or "month in? (1 .. 12)" (is 'double dot' considered an operator ? because i used it in the 'for' loops with spaces between double-dot and the operands and now Iäm just a bit confused)

davidhesselbom: I haven't decided yet, so leave it for now. Double dot is a bit of a special case - it seems to create a `Range` cover implicitly, and I haven't really bothered to figure out how.

source/sdk/os/Time.ooc
davidhesselbom: Additions are fine, just avoid modifications. As for moving `DateTimeData` into `Time`, I think you should be able to `extend` `Time` instead. Move back `DateTimeData` into the same file where `DateTime` is located, and add something like
```ooc
extend Time {
    dateTimeData: static func -> DateTimeData {
...
```
This is done in several places with e.g. `String`, but I'm not sure it's going to work with static functions. Worth a try though.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,233 @@
...
+	}
+	init: func@ ~fromHourMinuteSec (hour, minute, second, millisecond: Int) {
+		if (timeIsValid(hour, minute, second, millisecond))
+			this _ticks = This timeToTicks(hour, minute, second, millisecond)
+		else
+			raise ("invalid input specified for constructor(hour,minute,second)")
+	}
+	init: func@ ~fromDateTime (year, month, day, hour, minute, second, millisecond: Int) {
+		if (dateIsValid(year, month, day) && timeIsValid(hour, minute, second, millisecond))
+			this _ticks = This timeToTicks(hour, minute, second, millisecond) + This dateToTicks(year, month, day)
```
davidhesselbom: Wouldn't this make more sense? `this _ticks = This dateToTicks(year, month, day) + This timeToTicks(hour, minute, second, millisecond)`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,233 @@
...
+			raise ("invalid input specified for constructor(year,month,day)")
+	}
+	init: func@ ~fromHourMinuteSec (hour, minute, second, millisecond: Int) {
+		if (timeIsValid(hour, minute, second, millisecond))
+			this _ticks = This timeToTicks(hour, minute, second, millisecond)
+		else
+			raise ("invalid input specified for constructor(hour,minute,second)")
+	}
+	init: func@ ~fromDateTime (year, month, day, hour, minute, second, millisecond: Int) {
+		if (dateIsValid(year, month, day) && timeIsValid(hour, minute, second, millisecond))
```
davidhesselbom: `dateIsValid` and `timeIsValid` are both static functions from this cover, so `This dateIsValid` and `This timeIsValid`.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,232 @@
...
+	ticksPerWeek: static const UInt64 = This ticksPerDay * 7
+	ticksPerFourYears: static const Int64 = This daysPerFourYears * This ticksPerDay
+	/* default date/time printing format */
+	defaultFormat: static const String = "%yyyy-%MM-%dd %hh:%mm:%ss::%zzzz"
+
+	now: static DateTime {
+		get {
+					data := Time dateTimeData()
+					DateTime new(data year, data month, data day, data hour, data minute, data second, data millisecond)
+		}
```
davidhesselbom: Something's wrong with the indentation here.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,232 @@
...
+		else
+			month % 2 ? 30 : 31
+	}
+
+	ticksInMonth: static func (year, month: Int) -> UInt64 {
+		This ticksPerDay * This daysInMonth(year, month)
+	}
+	/* validate argument ranges for hour/minutes/seconds vaules */
+	timeIsValid: static func (hour, minute, second, millisecond: Int) -> Bool {
+		hour in?(0..24) && minute in?(0..60) && second in?(0..60) && millisecond in?(0..1000)
```
davidhesselbom: This looks much cleaner, yes!

source/base/DateTime.ooc
sebastianbaginski: ok, same for isLeapYear calls below

source/sdk/os/Time.ooc
sebastianbaginski: ok, thanks, I will try that

source/base/DateTime.ooc
```diff
@@ -14,22 +14,232 @@
...
+		DateTimeData new(year, month, days + 1, hour, minute, second, millisecond)
+	}
+
+	/* returns number of ticks for given hours, minutes and seconds */
+	timeToTicks: static func (hours, minutes, seconds, millisecond: Int) -> Int64 {
+		(hours * 3600 + minutes * 60 + seconds) * This ticksPerSecond + millisecond * This ticksPerMillisecond
+	}
+
+	/* returns number of ticks for given date at 0:00*/
+	dateToTicks: static func(year, month, day : Int) -> UInt64 {
```
davidhesselbom: `day: Int`

davidhesselbom: `func (`

source/base/DateTime.ooc
```diff
@@ -14,22 +14,232 @@
...
+		result := 0 as UInt64
+		if (This dateIsValid(year, month, day)) {
+			totalDays := day - 1
+			for (m in 1 .. month)
+				totalDays += This daysInMonth(year, m)
+			fourYearBlocks := (year - 1) / 4
+			year_start := fourYearBlocks * 4
```
davidhesselbom: I'm not sure what this variable signifies, but it should be `startYear` or `yearStart`.

sebastianbaginski: sorry, looks like my old C coding habits are still running strong

source/sdk/os/Time.ooc
sebastianbaginski: it worked.

source/base/DateTime.ooc
davidhesselbom: I figured as much :)

source/sdk/os/Time.ooc
davidhesselbom: Cool!

source/base/DateTime.ooc
sebastianbaginski: Right, updated it.

sebastianbaginski:
I'm wondering about the "Travis Cl build failed" message - sometimes it fails for the DateTime module, while the same test launched on this machine completes correctly. What could be wrong and should I be concerned about it ?

davidhesselbom:
Some tests sometimes fail on Travis, but not locally. This means something doesn't work reliably, which is a problem, yes. The whole point of running tests on Travis is to make sure tests are actually run (as we're likely to forget otherwise), and ignoring failing tests makes Travis pointless. There are a few tests that fail randomly, though, and we haven't been able to figure out why yet.

Try to figure out why the test fails and let me know if you get stuck.

sebastianbaginski:
Ok, thanks.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,262 @@
...
+		this compareTo(other) == Order greater
+	}
+	operator >= (other: This) -> Bool {
+		this compareTo(other) != Order less
+	}
+
+	/* number of days in year ( non-leap ) */
+	daysPerYear: static const Int = 365
+	daysPerFourYears: static const Int = 3 * This daysPerYear + 366
+	ticksPerMillisecond: static const Int64 = 1000
```
davidhesselbom: Hmm... isn't it `10_000`? Pretty sure a 'tick' was decided to be 100 nanoseconds.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,262 @@
...
+				st: SystemTime
+				GetLocalTime(st&)
+				return DateTimeData new(st wYear, st wMonth, st wDay, st wHour, st wMinute, st wSecond, st wMilliseconds)
+		}
+		version(!windows) {
+				tt := time(null)
+				tv : TimeVal
+				val := localtime(tt&)
+				gettimeofday(tv&, null)
+				return DateTimeData new(val@ tm_year+1900, val@ tm_mon+1, val@ tm_mday, val@ tm_hour, val@ tm_min, val@ tm_sec, tv tv_usec / 1000)
```
davidhesselbom: The system gives you the time with microsecond accuracy, and you want to store it in 100-nanosecond accuracy, but here you reduce it to millisecond accuracy. Bit of a shame, right?

source/base/DateTime.ooc
sebastianbaginski: Yes, it should be 10^4

source/base/DateTime.ooc
sebastianbaginski: Right. I think the same could be achieved with QueryPerformanceCounter on Windows. If you would like to have it.

davidhesselbom: I doubt we'll need it on Windows, but we will *need* it on Android. Millisecond precision isn't good enough.

sebastianbaginski:
Ok, so looks like "Travis build" fails if I change the datatypes to Unsigned for the DateTimeData struct members. I will go through each data member changing it one by one to see which is the one offending, maybe I can find the exact cause of the failure.

davidhesselbom:
It might be because Travis is using a different version of gcc (Try adding `gcc -v` to the Travis script and check), or has somehow configured it differently, or is running 32-bit Linux while we are running 64-bit...

sebastianbaginski:
Great, I can reproduce it locally if I pass the 32bit architecture flag to gcc. It fails exactly on the same test as Travis build in this case.

davidhesselbom:
Alright. We need our stuff to work reliably on both 32-bit and 64-bit systems. I'm not sure what the best way to fix this is...

sebastianbaginski:
I found something weird, this snippet:

"\n8: %d" format(1 as UInt8) println()
"16: %d" format(1 as UInt16) println()
"32: %d" format(1 as UInt32) println()

produces following output:

8: 1
16: 140732899393537
32: 1

Something is wrong either with my code or with the 16 bit unsigned int type implementation.
Tested on both 32 and 64 bit compilator modes.

davidhesselbom:
It might have something to do with the flags that rock (the ooc compiler) passes to gcc, but I have no idea. I do remember having trouble with unsigned integers being converted into very large numbers, though, and `UInt8` was one example (but one that didn't happen to you just now).

Try doing some arithmetic with `UInt`s of various sizes and see if you can nail down when things fail. It's entirely possible that whatever code is responsible for formatting those numbers into strings always reads 32 or 64 bytes and that you just got lucky with the `UInt8` and `UInt32`s because that the surrounding bits just happened to be all 0.

source/base/DateTime.ooc
```diff
@@ -14,22 +14,263 @@
...
+	//	%ss - second
+	//	%zzzz - millisecond
+	// <param name="format">output format specification</param>
+	toStringFormat: func (format: String) -> String {
+		result := format
+		data := This _ticksToDateTimeHelper(this ticks)
+		result = result replaceAll("%yyyy", "%d" format(data year))
+		result = result replaceAll("%yy", "%d" format(data year % 100))
+		result = result replaceAll("%MM", (data month < 10 ? "0%d" : "%d") format(data month))
+		result = result replaceAll("%dd", (data day < 10 ? "0%d" : "%d") format(data day))
```
davidhesselbom: It won't help with memory leaks, but isn't it possible to do this as
```ooc
result = result replaceAll("%MM", "%02d" format(data month)
result = result replaceAll("%dd", "%02d" format(data day))
```
?

source/base/DateTime.ooc
```diff
@@ -14,22 +14,263 @@
...
+		result = result replaceAll("%M", "%d" format(data month))
+		result = result replaceAll("%d", "%d" format(data day))
+		result = result replaceAll("%hh", "%d" format(data hour))
+		result = result replaceAll("%mm", "%d" format(data minute))
+		result = result replaceAll("%ss", "%d" format(data second))
```
davidhesselbom: Hour, date, and minute should always be written as 2 digits, imo. Use `"%02d"`, as I mentioned above (assuming it actually works!).

sebastianbaginski: ok, it is corrected

sebastianbaginski:
This UInt issue is probably related to code that formats the output, as basic arithmetic seems to be working on the unsigned types

x := 1 as UInt8
y := 2 as UInt16
w := 3 as UInt32
z := 4 as Int
expect(z+w+x+y == 10)
"%d %d %d %d" format(x, y, w, z) println()

output:
1 140725224210434 3 4

sebastianbaginski:
Problem exists also for signed 16 bit integer type.

davidhesselbom:
Interesting. We should check if this happens in the official version of rock as well.

sebastianbaginski:
Looks like this issue does not exist in the official 'rock' release, I just cloned the repo from "https://github.com/fasterthanlime/rock.git", this is the output for the above snippet:
'1 2 3 4'
Same for signed 16 bit integer.

davidhesselbom:
Alright. Another reason for us to get our fork up to date, then - there's a long list of fixes in it that we don't have. Updating our fork isn't trivial, unfortunately.

sebastianbaginski:
Ok, so I'll just leave these types as Int for now, as it works on Travis build too. Just need to remember not to print the 16 bit types:)

source/base/DateTime.ooc
```diff
@@ -14,22 +14,266 @@
...
+		version(windows) {
+				st: SystemTime
+				GetLocalTime(st&)
+				result = DateTime new(st wYear, st wMonth, st wDay, st wHour, st wMinute, st wSecond, st wMilliseconds)
+		}
+		version(!windows) {
+				tt := time(null)
+				tv : TimeVal
+				val := localtime(tt&)
+				gettimeofday(tv&, null)
```
davidhesselbom: Doesn't the current time change between the `localtime` and the `gettimeofday` calls?

davidhesselbom: It seems `gettimeofday` is marked obsolete anyway. Maybe we can use `clock_gettime` instead?
http://stackoverflow.com/questions/3756323/getting-the-current-time-in-milliseconds

sebastianbaginski:
I think it is done.

davidhesselbom:
I know the messages are confusing but to me it looks like `clock_gettime` is missing on Travis:
```ooc
.libs/ooc/ooc-base/DateTime.o: In function `os_Time__Time_currentDateTime': /home/travis/build/cogneco/ooc-kean/source/base/DateTime.ooc:44: undefined reference to `clock_gettime' collect2: ld returned 1 exit status
```

sebastianbaginski:
According to this, looks like Travis build needs to link with the 'librt.so' library (-lrt) :
https://code.google.com/p/pymssql/issues/detail?id=105

davidhesselbom:
Interesting. Someone there suggested that one shouldn't link `librt` unless necessarly, though (it isn't in Ubuntu, evidently), and I agree. Not sure how to figure out whether we need to or not, though...

davidhesselbom:
Importantly, do we need to on Android?

sebastianbaginski:
One could figure out if 'clock_gettime' is in librt.so by using 'dlopen' and 'dlsym' functions to open the shared library and locate the symbol 'clock_gettime' inside. This is a C library so the function name should not be mangled in .so file.
I don't really know much about the Android, so I can't really help with that.

davidhesselbom:
Alright...

sebastianbaginski:
So looks like you will need another pre-build script to locate this librt.so on *nix systems and check if this function is inside. There could be more than one version of this library on the filesystem, so you will need to use the same LD_LIBRARY_PATH which 'rock' passes to gcc to locate the correct file. This sounds at least like a non-trivial task to me.
I'm not sure what is the best approach, you know this build system a lot better, so maybe you can come up with something.
Maybe in the meantime I will look for alternatives to this function.

davidhesselbom:
Yeah, that does sound complicated. Maybe we can just import `librt` for now... but I'm not sure.

davidhesselbom:
@simonmika, what do you think? Always link `librt`, even if `clock_gettime` isn't located there, or try to figure out whether `clock_gettime` needs `librt` first?

davidhesselbom:
So, as per Simon's recommendation, let's just have `DateTime now` give the current time in *second* precision for now. `DateTime` and `TimeSpan` still need to handle 100-nanosecond precision, but if we don't need to ask the system for anything beyond the current time in second precision, we don't need `clock_gettime` and the issue with `librt` can be dealt with later. You can probably revert most of the changes in `Time.ooc` that way, too.

sebastianbaginski:
Looks like Travis is complaining again.
"The command "./test.sh nogpu -Dnogpu" exited with 0."
but
"Done. Your build exited with 1." and "All checks have failed" ?

sebastianbaginski:
Should I cleanup the commit history ?

davidhesselbom:
> I guess I can assume it's not my fault :)

No, that's unrelated. Sorry...

> Should I cleanup the commit history ?

Normally, I would say yes, but you have more important things to do. Do revert the changes you made in `Time.ooc`, though.

sebastianbaginski:
They are reverted, at least most of them. I left only the addition of data member in windows date structure, as they were missing in original file.

davidhesselbom:
... and the whitespace change we talked about earlier.

davidhesselbom:
Pull request #317 is a cleanup of this pull request.

PR #303, opened 2015-07-28
Added 4th row to FloatTransform3D

emilwestergren:

PR #307, opened 2015-07-28
Replaced and removed incorrect usage of FloatTransform2D in Quaternion.

niklasborwell:


source/math/Quaternion.ooc
```diff
@@ -78,7 +79,10 @@ Quaternion: cover {
 		z := a1 * d2 + b1 * c2 - c1 * b2 + d1 * a2;
 		return This new(w, x, y, z);
 	}
-	fromRotationMatrix: static func (matrix: FloatTransform2D) -> This {
+	// TODO: This cannot be used with FloatTransform2D, only with a 3D rotation matrix
+	// for which we don't have a data type. We could do this for FloatTransform3D but that would
+	// make the calculations more complex than we have here.
```
davidhesselbom: Once #303 has been merged, what will be the difference between `FloatTransform3D` and a "3D rotation matrix?

davidhesselbom:
Is this going to break when #303 is merged?

source/math/Quaternion.ooc
niklasborwell: A 3D rotation matrix is 3x3 and contains only rotation, but FloatTransform3D is 4x4 and contains translations as well as 3 more parameters that represents something else.

niklasborwell:
Yes, it will need some adjustments.

source/math/Quaternion.ooc
davidhesselbom: Ah, right. So a 3D rotation matrix can be a `FloatTransform3D` that looks like this?
```
a b c 0
d e f 0
g h i 0
0 0 0 1
```

davidhesselbom:
So, should I merge this now, before #303?

niklasborwell:
Ehrm...If you want Emil to do the dirty work, yes. If you want me to do it, merge his first.

niklasborwell:
Either way is fine.

source/math/Quaternion.ooc
niklasborwell: Yes. If it looks like that we could use the above code. But if we add translations and the other values whatever they mean it won't be correct anymore.

davidhesselbom: So... a Quaternion can only represent rotations, not translations. So, should it be possible to verify that the incoming matrix is a rotation matrix first and then perform the conversion...?

niklasborwell: Possibly, but that doesn't seem very robust. What if it isn't? Throw exception? Also, we don't really have any use for this function currently.

davidhesselbom: So why keep the code for it?

niklasborwell: Because it's pretty. And we might want to use it in the future. We haven't implemented the quaternion stabilizer yet, and we might need it there somewhere.

davidhesselbom:
#303 was merged, go ahead.

source/math/Quaternion.ooc
davidhesselbom: I can't say I like having commented out code hanging around because "we might need it later", but have it your way.

niklasborwell: We can delete it if you want, I don't really care. But I didn't want to be rude to the person who wrote it. I can dig through the repository if I need it later.

niklasborwell:
Rebased and fixed. Also removed the commented code.

niklasborwell:
I guess this is just a random travis fail?

Unable to establish SSL connection.
The command "wget https://github.com/cogneco/rock/releases/$tmp" failed and exited with 4 during .
Your build has been stopped.

niklasborwell:
Also, hold on, cause I might have made a stupid mistake. Checking now.

niklasborwell:
Updated.

source/math/FloatRotation3D.ooc
```diff
@@ -50,6 +50,7 @@ FloatRotation3D: cover {
 	init: func@ ~full(=_roll, =_pitch, =_yaw) { this _updateQuaternion() }
 	init: func@ ~default { this init(0.0f, 0.0f, 0.0f) }
 	init: func@ ~fromPoint(point: FloatPoint3D) { this init(point x, point y, point z) }
+	init: func@ ~fromQuaternion (=_quaternion)
```
davidhesselbom: What happens if someone tries this:
```ooc
// q is a Quaternion from somewhere
t := FloatTransform3D new(q)
value := t x 
```
`_roll`, `_pitch_`, `_yaw` are not updated when the `_quaternion` is set.

davidhesselbom: Maybe you can use `Quaternion toFloatTransform3D()` here, somehow?

niklasborwell: Yeah, we don't care about that now. The entire euler angle system will be removed and this is just a necessary step for us while we test the new solution. I figured it doesn't matter anyway since the class is broken.

davidhesselbom: We seriously need a development branch...

niklasborwell:
Updated with slightly better tests and fixed error.

PR #308, opened 2015-07-30
Initial version of Stringbuilder

jonathanudd:


source/base/StringBuilder.ooc
```diff
@@ -0,0 +1,64 @@
...
+	prepend: func (other: This) {
+		for (i in 0..other count)
+			this _stringList insert(0, other[i])
+	}
+
+	toString: func -> String {
+		result := ""
+		for (i in 0..this _stringList count) {
+			result += _stringList[i]
+		}
```
davidhesselbom: Don't use brackets unless you actually need them.

test/base/StringBuilderTest.ooc
```diff
@@ -0,0 +1,63 @@
...
+				a append(b)
+				expect(a count, is equal to(2))
+				expect(a toString(), is equal to("Hello World Dlrow olleh"))
+		})
+		this add("prepend", func {
+				a := StringBuilder new("Hello World")
+				b := StringBuilder new("Dlrow olleh ")
+				a prepend(b)
+				expect(a count, is equal to(2))
+				expect(a toString(), is equal to("Dlrow olleh Hello World"))
```
davidhesselbom: Indentation...

test/collections/VectorListTest.ooc
```diff
@@ -133,6 +133,46 @@ VectorTest: class extends Fixture {
...
+			list add(0)
+			list add(1)
+			list add(2)
+
+			list modify(|value| value += 1)
+			c := 1
+			list apply(|value|
+				expect(value, is equal to(c))
+				c += 1)
+			})
```
davidhesselbom: Indentation

test/base/StringBuilderTest.ooc
```diff
@@ -0,0 +1,63 @@
...
+* along with this software. If not, see <http://www.gnu.org/licenses/>.
+*/
+use ooc-base
+use ooc-unit
+
+StringBuilderTest: class extends Fixture {
+	init: func {
+		super("StringBuilder")
+		this add("constructor", func {
+				a := StringBuilder new("Hello")
```
davidhesselbom: `s` seems like a better name than `a`.

source/base/StringBuilder.ooc
```diff
@@ -0,0 +1,64 @@
...
+	_stringList := VectorList<String> new()
+
+	count ::= this _stringList count
+
+	init: func ~default
+	init: func ~string (value: String) {
+		this init()
+		this _stringList add(value)
+	}
+	init: func ~this (orig: This) {
```
davidhesselbom: Don't needlessly shorten names like this - there's plenty of room for `original`.

source/base/StringBuilder.ooc
```diff
@@ -0,0 +1,64 @@
...
+	}
+	prepend: func (other: This) {
+		for (i in 0..other count)
+			this _stringList insert(0, other[i])
+	}
+
+	toString: func -> String {
+		result := ""
+		for (i in 0..this _stringList count) {
+			result += _stringList[i]
```
davidhesselbom: Here, `result` will be replaced with a new string, each time leaking memory, because we will lose our only reference to what `result` is currently pointing to. In `String`:

```ooc
operator >> (left, right: String) -> String {
	result := left + right
	left free()
	result
}
```
so 
```ooc
result += _stringList[i]
```
could be changed into
```ooc
result = result >> _stringList[i]
```
and the leak would be fixed.

source/base/StringBuilder.ooc
```diff
@@ -0,0 +1,64 @@
...
+		this init()
+		orig _stringList apply( func (value: String) { this _stringList add(value) })
+	}
+	copy: func -> This {
+		This new(this)
+	}
+
+	add: func (value: String) {
+		this _stringList add(value)
+	}
```
davidhesselbom: So, `Strings` are added to the list. When the `StringBuilder` is freed, it should also free its list. But what happens to the contents in the list? Should they be freed? If you want to keep the `String` you add to the `StringBuilder`, should you give the `StringBuilder` a copy of the `String`?

```ooc
sb := StringBuilder new()
myString := "Hello World"
sb add(myString)
sb toString() println().free()
sb free() // the list in s is also free, and so are all the strings in the list
myString println() // this is going to crash
```
so should we instead do
```ooc
sb := StringBuilder new()
myString := "Hello World"
sb add(myString copy())
sb toString() println().free() // free the String created by the `toString()` call
sb free() // the list in s is also freed, and so are all the Strings in the list
myString println() // this is should work
```
? Having to type `copy()` in each `add()` call is going to get messy. If we modify `add()` like this
```ooc
	add: func (value: String) {
		this _stringList add(value copy())
	}
```
we don't have to type
```ooc
sb add(myString copy())
```
but can do
```ooc
sb add(myString)
```
, so that's nice. But what if we don't want to keep `myString`? Then we'll have to do
```ooc
sb add(myString)
myString free()
```
and that's going to get messy, too. In addition, we copy a `String` and then free the original immediately afterwards, which is unnecessary work.

One way to fix this is by adding a second argument to `add`:
```ooc
	add: func (value: String, copy := true) {
		this _stringList add(value copy())
	}
```
but
```ooc
sb add(myString, false)
```
is also not very pretty and people are likely to make mistakes.

Another solution is to give `StringBuilder` operators for appending and freeing (or not) the argument, but then we haven't gained anything by adding `StringBuilder`, because the whole point is that the person using it shouldn't have to worry about memory leaks, or else they're just as well off using `String` operators directly.

PR #309, opened 2015-07-30
added magic ignore file

thomasfanell:
Added magic ignore file

### DIFF
--- a/.magicignore
+++ b/.magicignore
@@ -1,0 +1,5 @@
+.git
+.libs
+rock_tmp
+source/sdk
+test


.magicignore
```diff
@@ -0,0 +1,5 @@
+.git
+.libs
+rock_tmp
```
davidhesselbom: There are no ooc files in these folders, afaik. Does adding them here improve magic's performance?

thomasfanell: No, no ooc files there, so it is completely unnecessary for magic to look there. Actually, these folders should be excluded by default, I just haven't got around to that yet. I suggest we keep those in there for now. Excluding them will boost performance a bit.

PR #315, opened 2015-08-04
Failed tests are now printed by name

thomasfanell:
* fixed so that the testing script continues to run tests even if one or more failed
* fixed so any failed tests are also printed by name

Sample output:
```
...
IntBox2D ....... done
FloatTransform3D ff................ failed
  -> 'equality'
  -> 'determinant'
Failed test: ./test/math/FloatTransform3DTest.ooc
FloatBox2D ......... done
...
```

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -30,6 +30,7 @@ Fixture: abstract class {
 		this add(test)
 	}
 	run: func {
+		failures := ArrayList<TestFailedException> new()
 		result := true
 		(this name + " ") print()
 		this tests each(|test|
@@ -38,11 +39,17 @@ Fixture: abstract class {
 				test run()
 			} catch (e: TestFailedException) {
 				e test = test
+				e message = test name
 				result = r = false
+				failures add(e)
 			}
 			(r ? "." : "f") print()
 		)
 		(result ? " done" : " failed") println()
+		if (!result)
+			for (f in failures)
+				"  -> '%s'" printfln(f message)
+		failures free()
 		exit(result ? 0 : 1)
 	}
 	is ::= static IsConstraints new()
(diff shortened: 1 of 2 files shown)


thomasfanell:
Apparently it's a bit touchy how one tests for booleans in bash. Amended and working now.

source/unit/Fixture.ooc
```diff
@@ -38,11 +39,17 @@ Fixture: abstract class {
 				test run()
 			} catch (e: TestFailedException) {
 				e test = test
+				e message = test name
 				result = r = false
+				failures add(e);
```
davidhesselbom: Semicolon?

source/unit/Fixture.ooc
```diff
@@ -70,7 +77,10 @@ TestFailedException: class extends Exception {
 	test: Test
 	value: Object
 	constraint: Constraint
-	init: func (=value, =constraint)
+	//message: String
+	init: func (=value, =constraint, message := "") {
+		this message = message;
```
davidhesselbom: And here

source/unit/Fixture.ooc
```diff
@@ -30,6 +30,7 @@ Fixture: abstract class {
 		this add(test)
 	}
 	run: func {
+		failures := ArrayList<TestFailedException> new();
```
davidhesselbom: And here

thomasfanell:
Superior habits die hard. Will fix...

source/unit/Fixture.ooc
davidhesselbom: You should probably run magic on this pull request... :D

source/unit/Fixture.ooc
thomasfanell: I'll also remove // message: String?

davidhesselbom: why is it commented out?

source/unit/Fixture.ooc
thomasfanell: Touché :+1:

source/unit/Fixture.ooc
thomasfanell: The base class Exception already has that field

davidhesselbom: Ahhh... Alright.

PR #321, opened 2015-08-11
fixed style violations

thomasfanell:


source/math/Quaternion.ooc
```diff
@@ -109,11 +109,11 @@ Quaternion: cover {
...
-		w := a1 * a2 - b1 * b2 - c1 * c2 - d1 * d2;
-		x := a1 * b2 + b1 * a2 + c1 * d2 - d1 * c2;
-		y := a1 * c2 - b1 * d2 + c1 * a2 + d1 * b2;
-		z := a1 * d2 + b1 * c2 - c1 * b2 + d1 * a2;
-		return This new(w, x, y, z);
+		w := a1 * a2 - b1 * b2 - c1 * c2 - d1 * d2
+		x := a1 * b2 + b1 * a2 + c1 * d2 - d1 * c2
+		y := a1 * c2 - b1 * d2 + c1 * a2 + d1 * b2
+		z := a1 * d2 + b1 * c2 - c1 * b2 + d1 * a2
+		return This new(w, x, y, z)
```
davidhesselbom: `return` is unnecessary. Should we keep it for clarify?

davidhesselbom:
I've been trying to avoid fixing style violations in `math` and `draw` until we've merged Emil's giant refactoring, because they're likely to cause a lot of conflicts. Maybe we can keep this one for later?

thomasfanell:
Oh ok, did not know that... :-)

source/math/Quaternion.ooc
thomasfanell: I vote no, we should take advantage of that fact that we don't have to use ```return``` in cases like this. It's quite clear without it.

davidhesselbom: Alright then.

davidhesselbom: Hehe, guess who wrote the semicolons? Simon...

thomasfanell: We should have an annual "magic award"... ;-)

davidhesselbom: Only for code written after `magic` was released, then.

davidhesselbom: Or for things that `magic` should have caught, but didn't!

thomasfanell: I'll need a separate storage room for all the trophies i'll win...

PR #323, opened 2015-08-12
added constants for memory management in vector class

sebastianbaginski:
Simple boolean flags added for readability

### DIFF
--- a/source/collections/Vector.ooc
+++ b/source/collections/Vector.ooc
@@ -21,8 +21,8 @@ Vector: abstract class <T> {
 	capacity ::= this _capacity
 	_freeContent: Bool
 
-	init: /* protected */ func ~preallocated (=_backend, =_capacity, freeContent := true)
-	init: /* protected */ func (=_capacity, freeContent := true) {
+	init: /* protected */ func ~preallocated (=_backend, =_capacity, freeContent := This FreeContent)
+	init: /* protected */ func (=_capacity, freeContent := This FreeContent) {
 		this _freeContent = freeContent
 	}
 	free: override func {
@@ -92,6 +92,9 @@ Vector: abstract class <T> {
 		}
 		this _backend[index] = item
 	}
+
+	FreeContent: static const Bool = true
+	DoNotFreeContent: static const Bool = false
 }
 
 HeapVector: class <T> extends Vector<T> {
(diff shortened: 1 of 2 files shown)


davidhesselbom:
So, the point of this is that 
```ooc
myVector := VectorList<Float> new(32, VectorList doNotFreeContent)
```
demonstrates intent more clearly than does
```ooc
myVector := VectorList<Float> new(32, true)
```
but it also means longer typing more words to get the same result, and if you need to use the constants, you have to look them up somewhere. I'm conflicted about whether this is a good idea.

@simonmika, once again, your input would be appreciated.

simonmika:
No constants like this. Use enums, stick with Bool or use a special constructor.

davidhesselbom:
Gotcha, thanks. Closing.

sebastianbaginski:
I agree that this means longer typing. On the other hand, I'd have to look up what the 'false' means in the constructor call too.
Maybe the Vector class is not the best place to keep these constants, so maybe something like
```C++
enum MemoryPolicy {
    TakeContent,
    CopyContent,
    ...
};

vec := Vector<...> new(MemoryPolicy CopyContent)
```
so we can use it in other places too ?

davidhesselbom:
Maybe. I don't like the idea of `this freeContent = (memoryPolicy == MemoryPolicy copyContent)` in classes that use this, though.

sebastianbaginski:
Yeah that looks bad.
Maybe the Vector class could have a "memoryAllocator" object instead of the Boolean flag, so we can move the memory management code elsewhere ? Just wondering.

simonmika:
Wy complicate things? We have had this discussion about nine months ago on the very same classes. Either inherit Vector and create a Vector that clears its children or go for one of bool, enum or hide bool behind specially named constructor.

The PR did make things worse by not enforcing one way but rather allowing both. We DO have a type system, it is NOT c or c++.

davidhesselbom:
> We have had this discussion about nine months ago on the very same classes.

I don't remember that discussion, but I don't doubt your word. Apparently things still aren't clear enough, or the discussion wouldn't resurface... Anyway, I'll try to keep what you said just now in mind. Thanks!

> We DO have a type system, it is NOT c or c++.

I'm not getting this last part. What do you mean?

PR #327, opened 2015-08-13
bug in "VectorList sort"

sebastianbaginski:
I'm creating a pull request so that you can see this branch.
There is a bug in VectorList::sort( func greaterThan ) method, rock generates incorrect C code for the arguments passed to sorting function. Addresses are passed instead of argument values.
Definition for the simple sorting function:
```ooc
sortFunc: static func (a,b: Int) -> Bool {
a > b
}
```
Here is the C code generated for the sorting function definition:
```C
lang_types__Bool ___test_collections_VectorListTest__VectorTest_sortFunc(lInt a, Int b){ ...
```
Everything is ok so far.
Now, here is the C code generated for the VectorList sort method, for the fragment which calls the comparison operator:
```ooc
((lang_types__Bool (*)(uint8_t*, uint8_t*, void*)) greaterThan.thunk)((uint8_t*) a, (uint8_t*) b, greaterThan.context)) { ...
```
You can clearly see that addresses are passed to the greaterThan() function.
Here is example output when running the "VectorListTest" from this branch:

> values in VectorList: 2 0
> addresses in VectorList: 348047904 348047872
> arguments to sortFunc 348047904, 348047872

```ooc
// from VectorList class
sort: func (greaterThan: Func (T, T) -> Bool) {
	inOrder := false
	while (!inOrder) {
		inOrder = true
		for (i in 0 .. count - 1) {
			a := this[i]
			b := this[i + 1]
			"values in VectorList: %i %i" format(a as Int, b as Int) println()
			"addresses in VectorList: %i %i" format(a& as Int, b& as Int) println()
			if (greaterThan(a,b)) {
				inOrder = false
				tmp := this[i]
				this[i] = this[i + 1]
				this[i + 1] = tmp
			}
		}
	}
}

// from VectorListTest
sortFunc: static func (a,b: Int) -> Bool {
	result := a > b
	"arguments to sortFunc %i, %i : %i" format(a,b,result) println()
	result
}
```

Maybe I'm wrong, but I think there is a bug here.

### DIFF
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -87,7 +87,11 @@ VectorList: class <T> {
 		while (!inOrder) {
 			inOrder = true
 			for (i in 0 .. count - 1) {
-				if (greaterThan(this[i], this[i + 1])) {
+				a := this[i]
+				b := this[i + 1]
+				"values in VectorList: %i %i" format(a as Int, b as Int) println()
+				"addresses in VectorList: %i %i" format(a& as Int, b& as Int) println()
+				if (greaterThan(a,b)) {
 					inOrder = false
 					tmp := this[i]
 					this[i] = this[i + 1]
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -89,8 +89,8 @@ VectorList: class <T> {
 			for (i in 0 .. count - 1) {
 				a := this[i]
 				b := this[i + 1]
-				"values in VectorList: %i %i" format(a as Int, b as Int) println()
-				"addresses in VectorList: %i %i" format(a& as Int, b& as Int) println()
+				//"values in VectorList: %i %i" format(a as Int, b as Int) println()
+				//"addresses in VectorList: %i %i" format(a& as Int, b& as Int) println()
 				if (greaterThan(a,b)) {
 					inOrder = false
 					tmp := this[i]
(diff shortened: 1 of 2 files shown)


sebastianbaginski:
@davidhesselbom what do you think about it ?

simonmika:
This is a design mistake in how generics work. It would be interesting to test with the newest version of rock.

davidhesselbom:
We've had a lot of issues with generics and closures being passed around, yeah. And yes, let's see if the new version of rock that Fredrik is working on solves any of this, and get back to it then. If the bug persists, and if it also happens in the official rock, we can open an issue in its repository.

sebastianbaginski:
Ok, thank you.
@davidhesselbom about your question, it does not seem to work with classes. Running the VectorListTest from latest commit will result in segmentation fault.

sebastianbaginski:
Ok thanks.
This seems to be broken the other way around as well - extending the generic class for a specific type will allow all other instantiations of this generic to use the extension. Code below compiles without any warnings and passes the test:
```ooc
extend VectorList <Float> {
	newMethod: func -> Int {
		return 1
	}
}
...
list := VectorList<String> new()
expect(list newMethod() == 1)
```
Thanks @davidhesselbom for the suggestion.

davidhesselbom:
So, is this meant to be merged, thereby fixing the bug, or is it only here to demonstrate the bug?

sebastianbaginski:
There is no fix for this bug in this pull request, I just wanted to demonstrate it. I will keep this branch anyway as useful test case collection, so I guess you can just close it.

davidhesselbom:
There are a lot of pull requests open right now, so I'm closing this for now.

PR #330, opened 2015-08-19
FloatVectorList: improved sorting performance and added fast median algorithm

sebastianbaginski:
Implementation of quicksort and quickselect for float vectors.
It is not a general implementation (based on generics or iterators, with a custom comparator function) because of #327 .
Profiling and testing is on my branch https://github.com/sebastianbaginski/ooc-kean/tree/quickselectProfiling  (run *./test.sh math/FloatVectorList*)

Example profiling results:
> start testing new: 10:20:22
> done testing new: 10:20:26
> start testing old: 10:20:26
> done testing old: 10:26:20
> number of tests: 20000

calculating median and sorting for 20.000 random float vectors of length 1000 took ~4 seconds with new implementation vs. ~6 minutes with old.

### DIFF
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -121,9 +121,14 @@ VectorList: class <T> {
 	}
 	getSlice: func (range: Range) -> This<T> {
 		result := This<T> new(range count)
-		result _count = range count
-		source := (this pointer as Long + (range min * (T size))) as Pointer
-		memcpy(result pointer, source, range count * (T size))
+		this getSliceInto(range, result&)
 		result
+	}
+	getSliceInto: func(range: Range, buffer: This<T>@) {
+		if (buffer _vector capacity < range count)
+			buffer _vector resize(range count)
+		buffer _count = range count
+		source := (this pointer as Long + (range min * (T size))) as Pointer
+		memcpy(buffer pointer, source, range count * (T size))
 	}
 }
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -124,7 +124,7 @@ VectorList: class <T> {
 		this getSliceInto(range, result&)
 		result
 	}
-	getSliceInto: func(range: Range, buffer: This<T>@) {
+	getSliceInto: func (range: Range, buffer: This<T>@) {
 		if (buffer _vector capacity < range count)
 			buffer _vector resize(range count)
 		buffer _count = range count
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -128,7 +128,7 @@ VectorList: class <T> {
 		if (buffer _vector capacity < range count)
 			buffer _vector resize(range count)
 		buffer _count = range count
-		source := (this pointer as Long + (range min * (T size))) as Pointer
+		source := (this _vector _backend + (range min * (T size))) as Pointer
 		memcpy(buffer pointer, source, range count * (T size))
 	}
 }
--- a/source/math/FloatVectorList.ooc
+++ b/source/math/FloatVectorList.ooc
@@ -249,8 +249,8 @@ FloatVectorList: class extends VectorList<Float> {
 		result := This new(this count)
 		windowBuffer := This new(windowSize)
 		start := -((windowSize - 1) / 2)
-		for (i in 0..this count) {
-			range := ((start..(start + windowSize - 1)) + i) clamp(0, this count-1)
+		for (i in 0 .. this count) {
+			range := ((start .. (start + windowSize - 1)) + i) clamp(0, this count-1)
 			this getSliceInto(range, (windowBuffer as VectorList<Float>)&)
 			result add((windowBuffer as This) fastMedian(0, range count - 1))
 		}
@@ -270,7 +270,7 @@ FloatVectorList: class extends VectorList<Float> {
 		pivotValue := array[pivot]
 		This _swap(array, pivot, end)
 		result := start
-		for (i in start..end)
+		for (i in start .. end)
 			if (array[i] < pivotValue) {
 				This _swap(array, result, i)
 				++result
--- a/source/math/FloatVectorList.ooc
+++ b/source/math/FloatVectorList.ooc
@@ -251,7 +251,7 @@ FloatVectorList: class extends VectorList<Float> {
 		start := -((windowSize - 1) / 2)
 		for (i in 0..this count) {
 			range := ((start..(start + windowSize - 1)) + i) clamp(0, this count-1)
-			this getSliceInto(range, windowBuffer&)
+			this getSliceInto(range, (windowBuffer as VectorList<Float>)&)
 			result add((windowBuffer as This) fastMedian(0, range count - 1))
 		}
 		windowBuffer free()
--- a/source/math/FloatVectorList.ooc
+++ b/source/math/FloatVectorList.ooc
@@ -238,22 +238,23 @@ FloatVectorList: class extends VectorList<Float> {
 	/* calculate median without copying and sorting the vector
 			WARNING: this function can partially rearange the elements in vector
 	*/
-	fastMedian: func -> Float {
-		data := this _vector _backend as Float*
-		result := This _nthElement(this _vector _backend as Float*, 0, this count - 1, this count / 2)
-		if (Int even(this count))
-			result = (result + This _nthElement(this _vector _backend as Float*, 0, this count - 1, this count / 2 - 1)) / 2
+	fastMedian: func (start := 0, end := this count - 1) -> Float {
+		count := end - start + 1
+		result := This _nthElement(this _vector _backend as Float*, start, end, count / 2)
+		if (Int even(count))
+			result = (result + This _nthElement(this _vector _backend as Float*, start, end, count / 2 - 1)) / 2
 		result
 	}
 	movingMedianFilter: func (windowSize: Int) -> This {
 		result := This new(this count)
+		windowBuffer := This new(windowSize)
 		start := -((windowSize - 1) / 2)
-		for (i in 0 .. this count) {
-			range := (start .. (start + windowSize - 1)) + i
-			elementsInWindow := this getSlice(range clamp(0, this count-1))
-			result add((elementsInWindow as This) fastMedian())
-			elementsInWindow free()
+		for (i in 0..this count) {
+			range := ((start..(start + windowSize - 1)) + i) clamp(0, this count-1)
+			this getSliceInto(range, windowBuffer&)
+			result add((windowBuffer as This) fastMedian(0, range count - 1))
 		}
+		windowBuffer free()
 		result
 	}
 	_swap: static func (array: Float*, i, j: Int) {
--- a/source/math/FloatVectorList.ooc
+++ b/source/math/FloatVectorList.ooc
@@ -65,19 +65,7 @@ FloatVectorList: class extends VectorList<Float> {
 	}
 	standardDeviation ::= this variance sqrt()
 	sort: func {
-		inOrder := false
-		vector := this _vector _backend as Float*
-		while (!inOrder) {
-			inOrder = true
-			for (i in 0 .. count - 1) {
-				if (vector[i] > vector[i + 1]) {
-					inOrder = false
-					tmp := vector[i]
-					vector[i] = vector[i + 1]
-					vector[i + 1] = tmp
-				}
-			}
-		}
+		This _quicksort(this _vector _backend as Float*, 0, this count - 1)
 	}
 	accumulate: func -> This {
 		result := This new(this _count)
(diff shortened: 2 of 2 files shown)


davidhesselbom:
6 minutes > 4 seconds is a pretty amazing improvement :)

As far as I know though, we never calculate the median for `VectorList`s with more than, say, 25 elements or so. Can you investigate? I'm not sure all this hard-to-understand code is actually worth it in the end.

sebastianbaginski:
Sure, I'll check that.

source/math/FloatVectorList.ooc
```diff
@@ -247,15 +235,93 @@ FloatVectorList: class extends VectorList<Float> {
...
+		end should be the last valid array index in the range
+	*/
+	_partition: static func (array: Float*, start, end, pivot: Int) -> Int {
+		pivotValue := array[pivot]
+		This _swap(array, pivot, end)
+		result := start
+		for (i in start..end)
+			if (array[i] < pivotValue) {
+				This _swap(array, result, i)
+				++result
```
davidhesselbom: What? I didn't even know the `++` operator existed in ooc! :O

sebastianbaginski: Nice:D Looks like *magic* doesn't really like this operator:
> [288, 17].....missing space before and after operator '+'
> [288, 18].....missing space before and after operator '+'

davidhesselbom: Well, I didn't think it was actually part of the language, or we would have used it in lots of places. I have to check something...

davidhesselbom: Oh so `++i` works but `i++` doesn't. Fascinating...

davidhesselbom: Must be because the guy who made ooc doesn't like C++ ;)

sebastianbaginski: To me it feels so natural to use it that I just used it, without even thinking about it. I just assumed that it exsits in ooc, because, well, why it shouldn't :)

source/math/FloatVectorList.ooc
```diff
@@ -247,15 +235,93 @@ FloatVectorList: class extends VectorList<Float> {
...
+		result
+	}
 	movingMedianFilter: func (windowSize: Int) -> This {
 		result := This new(this count)
 		start := -((windowSize - 1) / 2)
 		for (i in 0..this count) {
 			range := (start..(start + windowSize - 1)) + i
 			elementsInWindow := this getSlice(range clamp(0, this count-1))
-			result add((elementsInWindow as This) median())
+			result add((elementsInWindow as This) fastMedian())
```
davidhesselbom: A `median` function that takes a range (start and stop indexes) instead of a temporary `VectorList` would improve performance a lot, here, even for tiny `VectorList`s. We create a new `VectorList` called `elementsInWindow` `this count` times and free it immediately after computing its median.

source/math/FloatVectorList.ooc
davidhesselbom: Because `i++` doesn't (and it's what I'm used to, so once I realized `i++` didn't work, I never even bothered to check whether `++i` would.

davidhesselbom: You can go ahead and open an issue on https://github.com/cogneco/magic about this leading ++ operator.

sebastianbaginski:
I did some profiling on short lists (15-45 elements). Results for sorting 1000 times (time in ms., I suppose):

15 elements:
> new version Time: 1.975 Average: 1.975
> old version Time: 3.231 Average: 3.231

25 elements:
> new version Time: 3.176 Average: 3.176
> old version Time: 7.421 Average: 7.421

35 elements:
> new version Time: 4.372 Average: 4.372
> old version Time: 13.579 Average: 13.579

45 elements:
> new version Time: 5.608 Average: 5.608
> old version Time: 21.196 Average: 21.196

Simultaneously I am working on the "sharpness" problem, where I use the median filter for vectors as big as 'feature count' for given frame. I don't have a lot of real-life data, but in my test videos vectors with size > 25 are quite common. Maybe the performance improvement is not enormous, but it's measurable.
I'd like to introduce as less computing overhead per frame as possible, that's why I started working on the FloatVectorList in the first place.

davidhesselbom:
Alright, it looks like the improvement is definitely still worth it, then.

davidhesselbom:
Just to be sure - by "old" version, which commit do you refer to?

sebastianbaginski:
By "old version" I mean the code from the 'master' branch.

davidhesselbom:
Alright, so the "old" version already has the `Float` array usage, and we still get this much better performance. Amazing...

sebastianbaginski:
Exactly, I was comparing to the code improved with your *Float** idea from yesterday.
I think I could make it even faster by using bubble sort or something for really short lists instead of recurring into *quicksort* on lower levels. We could see if it is worth to avoid the function call and use *slower* algorithm. But *now* I do have doubts if it is needed :)

davidhesselbom:
Yeah, it doesn't look like it. Short lists should be fast enough to sort with quicksort anyway to avoid having two different sort functions (which will have to be added to each kind of list we want to sort...).

davidhesselbom:
I'd still like to see an implementation of `movingMedianFilter` that doesn't create a new list more than once, though.

sebastianbaginski:
Yeah I agree, that's what I'm gonna work on now.

sebastianbaginski:
About the sorting, too bad that generics in ooc are broken. Normally (read: *in c++* :D) this would be a template method accepting an abstract *iterator* type and comparator function (or object), so it would simply work as long as the sorted sequence had *random access* and two elements could be compared in some way...

davidhesselbom:
> too bad that generics in ooc are broken

This is the biggest issue with what we do. All of our codebase has been written to avoid this, and so many things could have been so much better if this was fixed. I know Amos is working on it, but I have no idea when we'll see it happen.

source/math/FloatVectorList.ooc
```diff
@@ -247,15 +235,97 @@ FloatVectorList: class extends VectorList<Float> {
...
 		result
 	}
+	/* calculate median without copying and sorting the vector
+			WARNING: this function can partially rearange the elements in vector
+	*/
+	fastMedian: func (start := 0, end := -1) -> Float {
+		if (end < 0)
+			end := this count - 1
+		count := end - start + 1
+		data := this _vector _backend as Float*
```
davidhesselbom: This variable is never used.

sebastianbaginski: Thanks, fixed.

source/math/FloatVectorList.ooc
```diff
@@ -247,15 +235,97 @@ FloatVectorList: class extends VectorList<Float> {
 		tempVector free()
 		result
 	}
+	/* calculate median without copying and sorting the vector
+			WARNING: this function can partially rearange the elements in vector
+	*/
+	fastMedian: func (start := 0, end := -1) -> Float {
+		if (end < 0)
+			end := this count - 1
```
davidhesselbom: This should be `end = this count - 1` - there's no need to change `end`'s type, and it results in a warning that `end` is unused. Anyway, why would you even get the median for the range (`x`..`y`) where `y` is a negative number?

davidhesselbom: Oh, you consider all negative numbers to be the end, whatever that is. Gotcha.

davidhesselbom: You should probably explain that behavior in the comment describing the function, though.

davidhesselbom:
I get this warning:
```
/home/dhesselbom/versioned/ooc/ooc-kean/source/math/FloatVectorList.ooc:257:13: warning: passing argument 3 of ‘VectorList__VectorList_getSliceInto’ from incompatible pointer type [enabled by default]
    this getSliceInto(range, windowBuffer&)
             ^
In file included from .libs/ooc/ooc-math/FloatVectorList-fwd.h:27:0,
                 from .libs/ooc/ooc-math/FloatVectorList.h:7,
                 from rock_tmp/ooc/ooc-math/FloatVectorList.c:3:
.libs/ooc/ooc-collections/VectorList-fwd.h:83:6: note: expected ‘struct VectorList__VectorList **’ but argument is of type ‘struct FloatVectorList__FloatVectorList **’
 void VectorList__VectorList_getSliceInto(VectorList__VectorList* this, lang_Numbers__Range range, VectorList__VectorList** buffer);
```
You probably need to cast something explicitly, somewhere.

source/math/FloatVectorList.ooc
sebastianbaginski: I should just remove it, for some reason the vector test failed when the 'end' defaulted to 'this count - 1'. I just assumed that this is just another weird ooc compiler behaviour. Now after running 'rock -x' and compiling again it worked ok. I will correct this.

davidhesselbom: Ok, thanks.

source/collections/VectorList.ooc
```diff
@@ -121,9 +121,14 @@ VectorList: class <T> {
 	}
 	getSlice: func (range: Range) -> This<T> {
 		result := This<T> new(range count)
-		result _count = range count
-		source := (this pointer as Long + (range min * (T size))) as Pointer
-		memcpy(result pointer, source, range count * (T size))
+		this getSliceInto(range, result&)
 		result
 	}
+	getSliceInto: func(range: Range, buffer: This<T>@) {
```
davidhesselbom: Looks like you missed this one in magic.

sebastianbaginski: My *magic* is probably outdated, it does not report anything for this file:

>sbaginski@mosquito:~/versioned2/ooc/magic$ magic ../ooc-kean/source/collections/VectorList.ooc 
> -> magic 0.1.9-alpha

davidhesselbom:
```
/home/dhesselbom/versioned/ooc/ooc-kean/source/collections/VectorList.ooc
-------------------------------------------------------------------------
[127, 23].....missing space after keyword 'func'

-> magic 0.1.9-alpha
```

davidhesselbom: Maybe you fixed it but forgot to commit the change?

sebastianbaginski: OK, it is corrected now.

source/collections/VectorList.ooc
```diff
@@ -121,9 +121,14 @@ VectorList: class <T> {
...
-		source := (this pointer as Long + (range min * (T size))) as Pointer
-		memcpy(result pointer, source, range count * (T size))
+		this getSliceInto(range, result&)
 		result
 	}
+	getSliceInto: func (range: Range, buffer: This<T>@) {
+		if (buffer _vector capacity < range count)
+			buffer _vector resize(range count)
+		buffer _count = range count
+		source := (this pointer as Long + (range min * (T size))) as Pointer
```
davidhesselbom: Hmm, what if we're on a 32-bit system?

sebastianbaginski: Should work anyway, 32 bit *pointer* value will be cast to bigger 64 bit datatype - we do not lose anything here. As long as 'range min * T size' won't exceed 2^32 we should be ok with casting this value back to 32 bit pointer type.
Btw. code for 32-bit machine can be forced with '-m32' flag passed to gcc, so we can check that.

sebastianbaginski: It works with *--m32* flag in test.sh script.
Assuming that 'rock --m32 ...' will cause rock to pass the *-m32* flag to gcc. I can't find this now, but I'm pretty sure it is in the docs.

sebastianbaginski: Actually shouldn't this cast use *(unsigned) long long* to be sure that we won't cast to a smaller type on 64 bit systems ? sizeof(long) is not guaranteed to be 8 bytes.
I assumed Long to be 8 bytes, but I'm not sure anymore. I don't understand this code:
```ooc
Long:  cover from signed long  extends LLong
```
does it mean that Long has the size of *signed long* (because it is a *cover* from this type) or that Long has the size of *LLong* (because it *extends* this type) ?

davidhesselbom: `LLong: cover from signed long long {` and every signed integer type seems to extend `LLong`, and they are not all the same size. Also, it seems they extend `LLong` to get the functions implemented for `LLong`, but I was honestly not aware covers could extend other covers. I'm very confused now.

Anyway, each type is also a `cover from` a C type, and as far as I know, those types are not guaranteed to have any size. That's why I'm a little worried about the cast to `Long`.

davidhesselbom: At any rate, I'm not sure you have to cast the pointer to any integer type, I think you can just add pointers and integers together.

sebastianbaginski:
I agree, it is not clear. And looks like the cast is needed only if you access the pointer through the property. This code
```ooc
source := (this _vector _backend + (range min * (T size))) as Pointer
```
maps directly to a simple pointer arithmetic in C
```C
lang_types__Pointer source = ((lang_types__Pointer) ((this->_vector->_backend + (range.min * (this->T->size)))));
```

davidhesselbom:
Yeah, I seem to remember having problems accessing pointer members by using properties. I don't know why exactly. If you have time, feel free to look into it. Probably another bug in rock.

sebastianbaginski:
```ooc
source := (this pointer + (range min * (T size))) as Pointer
```
Apparently rock wants to perform a string concatenation:
```C
    lang_types__Pointer source = ((lang_types__Pointer) ((lang_String____OP_ADD_String_LLong__String(VectorList__VectorList___getpointer__(this), (lang_Numbers__LLong) ((range.min * (this->T->size)))))));
```
The best part is that it compiles without any warning (and crashes at runtime, of course).
I think this one is my favourite so far.

davidhesselbom:
Nice one, yes.

davidhesselbom:
Have you tried this with [rock 1.0.8](https://github.com/cogneco/rock/releases/tag/rock_1.0.8)?

sebastianbaginski:
No, this could be interesting, I'll try that.

davidhesselbom:
If it still doesn't work, you can try it with the official rock version, make a minimal example, and file an issue at the [official rock repository](https://github.com/fasterthanlime/rock). Good practice :)

davidhesselbom:
Here's an example of an issue I made, btw: https://github.com/fasterthanlime/rock/issues/903

sebastianbaginski:
OK, thanks. 
I don't know what's wrong, but even after compiling new version of rock and accessing it manually in the *bin* folder it still prints the old version:
> sbaginski@mosquito:~/tests/official_rock/rock$ bin/rock --version
> rock 0.9.11-head codename sapporo, built on Mon Aug 24 14:58:56 2015

and I am inside the 'official_rock folder', freshly cloned from the rock repository...

davidhesselbom:
The official rock is at 0.9.11, our version is at 1.0.8. Confusing, I know.

sebastianbaginski:
Looks like this bug exists in official rock. Here is a simple test case
```ooc
Test: class {
	_backend: Float*
	pointer ::= this _backend as Pointer
	init: func
	test: func {
		source:= (this pointer + 8) as Pointer
	}
}
Test new() test()
```
rock generates the following code for the test() function
```C
void testPointerProperty__Test_test(testPointerProperty__Test* this) {
    lang_types__Pointer source = ((lang_types__Pointer) ((lang_String____OP_ADD_String_LLong__String(testPointerProperty__Test___getpointer__(this), (lang_Numbers__LLong) (8)))));
}
```
Can you double-check that ?

davidhesselbom:
It just hit me that it's probably a bad idea to return `Pointer` since it doesn't say anything about the pointer type, and so we don't know what happens if we increment it, for instance. I tried changing the example to

```ooc
Test: class {
    _backend: Float*
    pointer ::= this _backend as Float*
    init: func
    test: func {
        source:= (this pointer + 8) as Float*
    }
}
Test new() test()
```
and the C code is
```c
void testPointerproperty__Test_test_impl(testPointerproperty__Test* this) {
    #line 6 "/home/dhesselbom/versioned/rock/testPointerproperty.ooc"
    lang_Numbers__Float* source = ((lang_Numbers__Float*) ((testPointerproperty__Test___getpointer__(this) + 8)));
}
```
which seems a lot less insane.

sebastianbaginski:
Yeah now it looks ok, so I guess I won't report anything in the official repo :)
Btw. I think my work on this issue is done.

davidhesselbom:
It might even be possible to remove at least one of those casts and get working code that compiles without warnings...

PR #336, opened 2015-08-25
Create BoolVectorList and extended FloatVectorList ans VectorList.

bettinaselig:
Goes together with pull request in vidhance.


source/collections/VectorList.ooc
```diff
@@ -119,9 +124,12 @@ VectorList: class <T> {
...
 	}
-	getSlice: func (range: Range) -> This<T> {
-		result := This<T> new(range count)
-		this getSliceInto(range, result&)
+	getSlice: func (start, end: Int) -> This<T> {
+		resultCount := end - start + 1
+		result := This<T> new(resultCount)
+		result _count = resultCount
+		source := (this pointer as Long + (start * (T size))) as Pointer
+		memcpy(result pointer, source, resultCount * (T size))
```
davidhesselbom: This modification shouldn't be necessary anymore.

source/collections/VectorList.ooc
```diff
@@ -75,7 +75,12 @@ VectorList: class <T> {
 		this _vector free()
 		super()
 	}
-
+	reverse: func -> This<T> {
+		result := This<T> new(this _count)
+		for (i in 1..(this _count + 1))
+			result add(this[this _count - i])
```
davidhesselbom: This looks clearer, at least to me:
```ooc
for (i in 0..this _count)
    result add(this[(this _count - 1) - i])

File: source/collections/VectorList.ooc [hunk 3]
```diff
@@ -119,9 +124,12 @@ VectorList: class <T> {
 			result add(this[indices[i]])
 		result
 	}
-	getSlice: func (range: Range) -> This<T> {
-		result := This<T> new(range count)
-		this getSliceInto(range, result&)
+	getSlice: func (start, end: Int) -> This<T> {
```
davidhesselbom: We could still use this signature, though. Keep the old function, but also keep this line and have it call the other, like this:
```ooc
getSlice: func (range: Range) -> This<T> {
	result := This<T> new(range count)
	this getSliceInto(range, result&)
}
getSlice: func (start, end: Int) -> This<T> { this getSlice(start .. end+1) }
```

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+		result
+	}
+	and: func (other: This) -> This {
+		minimumCount := this count < other count ? this count : other count
+		result := This new(minimumCount)
+		for (i in 0..minimumCount)
+			result add(this[i] && other[i])
+		result
+	}
+}
```
davidhesselbom: Check this file for style violations with magic: https://github.com/cogneco/magic

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+		for (i in 0..minimumCount)
+			result add(this[i] || other[i])
+		result
+	}*/
+	reverse: func -> This {
+		result := This new(this _count)
+		for (i in 1..(this _count + 1))
+			result add(this[this _count - i])
+		result
+	}
```
davidhesselbom: You shouldn't have to implement this function anywhere but in `VectorList`.

source/math/FloatVectorList.ooc
```diff
@@ -325,4 +364,10 @@ FloatVectorList: class extends VectorList<Float> {
 				This _quicksort(array, pivot + 1, end)
 		}
 	}
+	clamp: func (floor, ceiling: Float) -> This {
+		result := This new(this _count)
+		for (i in 0..this _count)
+			result add(this[i] clamp(floor, ceiling))
+		result
+	}
```
davidhesselbom: This file is getting way too long...

test/math/FloatVectorListTest.ooc
```diff
@@ -123,7 +123,8 @@ FloatVectorListTest: class extends Fixture {
 			list add(2.0f)
 			list add(3.0f)
 			list add(4.0f)
-			slice := list getSlice(1 .. 2)
+			slice := list getSlice(1, 2)
```
davidhesselbom: OK, if you keep this line, and fix the `getSlice` function above like I said, this call will "test" both versions, which is nice.

source/collections/VectorList.ooc
bettinaselig: I still think that it is awkward to use `range` here, since the last element is usually excluded. But here (as seen in `FloatVectorListTest.ooc `) 
`slice := list getSlice(1 .. 2)` is equivalent to `slice := list getSlice(1, 2)`.

I suggest, I remove `getSlice: func ~range (range: Range)` and replace it with `getSlice: func (start, end: Int)`.

source/math/BoolVectorList.ooc
bettinaselig: The thing is that I need to cast, if I use reverse() from VectorList.
I would have liked to write something like:
```ooc
reverse: override func -> This {
    super() as This
```

source/math/FloatVectorList.ooc
bettinaselig: Don't blame me! :-P

test/math/FloatVectorListTest.ooc
bettinaselig: As I said above... I think we should get rid of using `range` here.

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+		halfSize := round((elementSize - 1) / 2) as Int
+		currentSequence := This new()
+		for (i in 0 .. this _count) {
+			currentSequence = this getSlice((i - halfSize) clamp(0, this count - 1), (i + halfSize) clamp(0, this count - 1))
+			result add(currentSequence totalTrues() > 0)
+		}
+		result
+	}
+	opening: func (elementSize: Int) -> This {
+		erosion(elementSize) dilation(elementSize)
```
davidhesselbom: `this erosion`

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+			result add(this[i] || other[i])
+		result
+	}*/
+	reverse: func -> This {
+		result := This new(this _count)
+		for (i in 1 .. (this _count + 1))
+			result add(this[this _count - i])
+		result
+	}
+	erosion: func (elementSize: Int) -> This {
```
davidhesselbom: `elementSize`? Do you mean `numberOfElements`?

bettinaselig: You need a _structuring element_ for applying erosion/dilation/opening/closing. I thought `structuringElementSize` would be too long. How about `windowSize`? The name would not be completely correct, but maybe less misleading?

davidhesselbom: Alright. In that case, I would suggest `structuringElementSize` to keep the same terminology, even though it is a bit long.

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+*/
+use ooc-collections
+import math
+import FloatVectorList
+
+BoolVectorList: class extends VectorList<Bool> {
+	init: func ~default {
+		this super()
+	}
+	init: func ~heap (capacity: Int){
```
davidhesselbom: `) {`

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+	}
+	totalTrues: func -> Int {
+		result := 0
+		for (i in 0 .. this count)
+			result += (this[i] ? 1 : 0)
+		result
+	}
+	totalFalses: func -> Int {
+		this _count - totalTrues()
+	}
```
davidhesselbom: I don't like these 2 function names, and besides, they should be properties, not functions. Here's two ideas:
```ooc
// Returns both numbers, use _ in the result variable to ignore either
	tally: func -> (Int, Int) {
		trues := 0
		for (i in 0 .. this count)
			trues += (this[i] ? 1 : 0)
		(trues, this _count - trues)
	}
```

```
// call with tally(true) to get the number of trues, tally(false) to get the number of falses
	tally: func (tallyTrues: Bool) -> Int {
		trues := 0
		for (i in 0 .. this count)
			trues += (this[i] ? 1 : 0)
		tallyTrues ? trues : this _count - trues
	}
```

davidhesselbom: Correction: the second version is not a property, because it takes an argument.

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+		result
+	}
+	opening: func (elementSize: Int) -> This {
+		erosion(elementSize) dilation(elementSize)
+	}
+	closing: func (elementSize: Int) -> This {
+		dilation(elementSize) erosion(elementSize)
+	}
+	toFloatVectorList: func (floatForFalse := 0.0f, floatForTrue := 1.0f) -> FloatVectorList {
+			result := FloatVectorList new(this _count)
```
davidhesselbom: Indentation

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,119 @@
...
+		for (i in 1 .. (this _count + 1))
+			result add(this[this _count - i])
+		result
+	}
+	erosion: func (elementSize: Int) -> This {
+		result := This new(this _count)
+		halfSize := round((elementSize - 1) / 2) as Int
+		currentSequence := This new()
+		for (i in 0 .. this _count) {
+			currentSequence = this getSlice((i - halfSize) clamp(0, this count - 1), (i + halfSize) clamp(0, this count - 1))
```
davidhesselbom: `getSlice` will create a new `VectorList` every time. This loop will leak lots of memory! Create the `currentSequence` variable once, then use the `getSliceInto` function to modify `currentSequence` in each iteration instead of creating a new list. Look at the `movingMedianFilter` function in `FloatVectorList`.

davidhesselbom: Also, remember to `free` `currentSequence` after the loop.

bettinaselig:
Updated.

bettinaselig:
Rebased.

bettinaselig:
Updated, once more.

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,120 @@
...
+		result := This new(minimumCount)
+		for (i in 0 .. minimumCount)
+			result add(this[i] || other[i])
+		result
+	}*/
+	reverse: func -> This {
+		result := This new(this _count)
+		for (i in 1 .. (this _count + 1))
+			result add(this[this _count - i])
+		result
```
davidhesselbom: This function is already implemented in `VectorList`. Does it really need to be implemented here, as well?

source/math/BoolVectorList.ooc
```diff
@@ -0,0 +1,120 @@
...
+			result add(this[i] && other[i])
+		result
+	}
+	operator || (other: This) -> This {
+		minimumCount := this count < other count ? this count : other count
+		result := This new(minimumCount)
+		for (i in 0 .. minimumCount)
+			result add(this[i] || other[i])
+		result
+	}*/
```
davidhesselbom: These are commented out because they don't work, right? Try using rock 1.0.8 and see if that fixes anything.

source/math/BoolVectorList.ooc
bettinaselig: Didn't we talk about this before? I wanted to know if you can call `VectorList reverse()` in `BoolVectorList reverse()` with `super()` or similar? Otherwise one needs to casts whenever you call `BoolVectorList reverse()`.

source/math/BoolVectorList.ooc
bettinaselig: Yes. Trying rock 1.0.8. Did not know that the new version is available.

source/math/BoolVectorList.ooc
davidhesselbom: Alright. Would it be possible to write the function as
```ooc
reverse: func -> This {
    (this as VectorList<Bool>) reverse()
}
?

**Comment by bettinaselig** [comment]
Worked. Exchanged it here and in Vidhance.

File: source/math/BoolVectorList.ooc (hunk 14 above)
Comment (bettinaselig) [inline source/math/BoolVectorList.ooc]: It works with `(this as VectorList<Bool>) reverse() as This`, since it needs to return a `BoolVectorList`.

**Comment by davidhesselbom** [comment]
Great, fix it in `FloatVectorList` while you're at it!

File: source/collections/VectorList.ooc [hunk 16]
```diff
@@ -135,12 +140,19 @@ VectorList: class <T> {
...
 	}
-	getSlice: func (range: Range) -> This<T> {
-		result := This<T> new(range count)
-		this getSliceInto(range, result&)
-		result
+	getSlice: func (start, end: Int) -> This<T> {
+		result := This<T> new(end - start + 1)
+		this getSliceInto(start, end, result&)
+	}
+	getSliceInto: func ~indices (start, end: Int, buffer: This<T>@) {
```
davidhesselbom: Again, 
```ooc
getSliceInto: func ~indices (start, end: Int, buffer: This<T>@) {
    this getSliceInto(start .. end, buffer)
}
```
would be a much better implementation.

source/collections/VectorList.ooc
```diff
@@ -135,12 +140,19 @@ VectorList: class <T> {
...
 		result
 	}
-	getSlice: func (range: Range) -> This<T> {
-		result := This<T> new(range count)
-		this getSliceInto(range, result&)
-		result
+	getSlice: func (start, end: Int) -> This<T> {
+		result := This<T> new(end - start + 1)
+		this getSliceInto(start, end, result&)
+	}
```
davidhesselbom: Put `getSlice: func (range: Range) -> This<T> {` back and have `getSlice: func (start, end: Int) -> This<T> {` do this instead:
```ooc
getSlice: func (start, end: Int) -> This<T> {
    this getSliceInto(start .. end)
}
```

bettinaselig:
Done.

bettinaselig:
I give up... I use `range` instead.

bettinaselig:
Rebased.

davidhesselbom:
Does VectorListTest work on your machine?

bettinaselig:
No.

davidhesselbom:
Can you find out why?

bettinaselig:
The method `getSlice()` is broken somehow.

This works fine: 
```ooc
list := VectorList<Float> new()
list add(1.0f)
list add(2.0f)
list add(3.0f)
list add(4.0f)
slice := list getSlice(1, 2)
```
But when I try to access `slice`, it crashes.

davidhesselbom:
OK. Does `getSlice(1 .. 2)` work better?

bettinaselig:
No.

bettinaselig:
Two suggestions: Either I leave it like this and someone else (Sebastian) fixes that also the test goes through, or I restore my version. 
This method is killing me...

bettinaselig:
Sebastian helped me. Now everything is working.

davidhesselbom:
OK, what did you change? And did you push the change?

bettinaselig:
Not pushed yet. `getSlice~range()` was missing a return statement.

davidhesselbom:
Ohhh... of course. I feel dumb now. Shouldn't there be a warning for that when compiling, though?

davidhesselbom:
The return statement was actually removed in an earlier commit in this pull request. Probably my fault.

source/collections/VectorList.ooc
davidhesselbom: This is where I told you to remove the return statement, by mistake. Sorry about that...

bettinaselig:
There should be a warning. But there was no warning. Sebastian and I discussed some more things. So I updated `getSlice()` and `getSliceInto()` even more.

davidhesselbom:
Well, that last commit was certainly a nice improvement. I don't know what that @ was doing there to begin with.

sebastianbaginski:
> There should be a warning.

I'd say a missing return statement should be a compilation error. Can we hack our version of rock somehow ?

>  I don't know what that @ was doing there to begin with.

That's my fault. I was making sure the input is really passed by reference.

bettinaselig:
Exactly, class objects are always passed as references in ooc. It's only covers that are passed as copies.

davidhesselbom:
> I'd say a missing return statement should be a compilation error.

It should be. I'm not sure why that isn't the case, and I thought it *was*. I'll see if we can't do something about that.

PR #337, opened 2015-08-28
fixed bug related to default function value in floatvectorlist

sebastianbaginski:


sebastianbaginski:
>  default values are probably computed only once, maybe even at compile time

you're right, here is the C code from *previous* version (with end parameter set to `this count - 1` by default):
```CConstraints__IsConstraints___getequal__(Fixture__Fixture___getis__()), FloatVectorList__FloatVectorList_fastMedian(list, 0, -1)), this->tolerance));
```

PR #340, opened 2015-08-31
Major revision of the draw API

emilwestergren:


davidhesselbom:
If it's not too much trouble, please try rebasing this on `master` one last time.

source/draw/gpu/GpuSurface.ooc
```diff
@@ -16,14 +16,110 @@
...
 	clear: abstract func
-	recycle: abstract func
+	draw: func ~general (action: Func) {
+		this _bind()
+		this _context setViewport(this viewport)
+		this _context enableBlend(this blend)
+		action()
+		this _unbind()
+	}
+	draw: func ~WithoutBind (destination: IntBox2D, map: GpuMap) {
```
davidhesselbom: This function causes a lot of leaks, according to valgrind. My guess is the `func` needs to be freed after use.

davidhesselbom: OK, so, as we agreed, fix the memory leaks by giving the closure its own variable and free it after use.

emilwestergren: Fixed.

emilwestergren:
Updated and rebased on upstream/master.

PR #344, opened 2015-09-02
Removed compilation warning

sebastianbaginski:
This fixes the compilation warning
> ShaderProgram.ooc:160:5: warning: passing argument 3 of ‘glShaderSource’ from incompatiblepointer type [enabled by default]
>   glShaderSource(shaderID, 1, (source toCString())&, null)

### DIFF
--- a/source/draw/gpu/opengl/OpenGLES3/ShaderProgram.ooc
+++ b/source/draw/gpu/opengl/OpenGLES3/ShaderProgram.ooc
@@ -18,6 +18,8 @@
 use ooc-math
 import include/gles, DebugGL
 
+CONST_CHAR_CAST: extern func (CString*) -> const CString*
+
 ShaderProgram: class {
 	_backend: UInt
 	init: func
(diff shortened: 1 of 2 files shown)


davidhesselbom:
@simonmika, this is a fix for the warning
```
[CC] Map/OpenGLES3MapPack
/home/dhesselbom/versioned/ooc/ooc-kean/source/draw/gpu/opengl/OpenGLES3/ShaderProgram.ooc: In function ‘OpenGLES3_ShaderProgram__ShaderProgram__compileShader’:
/home/dhesselbom/versioned/ooc/ooc-kean/source/draw/gpu/opengl/OpenGLES3/ShaderProgram.ooc:160:5: warning: passing argument 3 of ‘glShaderSource’ from incompatible pointer type [enabled by default]
   glShaderSource(shaderID, 1, ((source toCString())&) as const Char *, null)
     ^
In file included from .libs/ooc/ooc-opengl/OpenGLES3/include/gles-fwd.h:7:0,
                 from .libs/ooc/ooc-opengl/OpenGLES3/ShaderProgram-fwd.h:50,
                 from .libs/ooc/ooc-opengl/OpenGLES3/ShaderProgram.h:7,
                 from rock_tmp/ooc/ooc-opengl/OpenGLES3/ShaderProgram.c:3:
/usr/include/GLES3/gl3.h:475:29: note: expected ‘const GLchar * const*’ but argument is of type ‘lang_Character__Char *’
 GL_APICALL void GL_APIENTRY glShaderSource (GLuint shader, GLsizei count, const GLchar *const*string, const GLint *length);
```
The keyword `const` exists in ooc, but it seems to be ignored when generating the C code, so there doesn't appear to be a "clean" (i.e. ooc-only) solution to the problem. What do you think?

PR #345, opened 2015-09-03
An alternative version of PR #344.

davidhesselbom:
The obvious backdraw here is that a static function is used. It would be much better off as a member function, but I don't know how the macro should be written, then. Notably, the result of the macro must be used immediately, because if it passes through an ooc function on the way, it becomes a normal char ** again. @sebastianbaginski, any ideas?

### DIFF
--- a/source/draw/gpu/opengl/OpenGLES3/ShaderProgram.ooc
+++ b/source/draw/gpu/opengl/OpenGLES3/ShaderProgram.ooc
@@ -157,7 +157,7 @@ ShaderProgram: class {
 	_glCompileShader: func (shaderID: UInt) { glCompileShader(shaderID) }
 	_compileShader: func (source: String, shaderID: UInt) -> Bool {
 		version(debugGL) { validateStart() }
-		glShaderSource(shaderID, 1, (source toCString())&, null)
+		glShaderSource(shaderID, 1, String toConstCharPointerConstPointer(source toCString()&), null)
 		this _glCompileShader(shaderID)
 
 		success: Int
--- a/source/sdk/lang/String.ooc
+++ b/source/sdk/lang/String.ooc
@@ -1,3 +1,11 @@
+include ./StringExtension
+
+extend String {
+    // This function only exists to silence a warning when calling OpenGL's glShaderSource function.
+    // It requires a const char * const * argument, but since ooc doesn't have const, we need to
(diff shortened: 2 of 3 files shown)


source/draw/gpu/opengl/OpenGLES3/ShaderProgram.ooc
```diff
@@ -157,7 +157,7 @@ ShaderProgram: class {
 	_glCompileShader: func (shaderID: UInt) { glCompileShader(shaderID) }
 	_compileShader: func (source: String, shaderID: UInt) -> Bool {
 		version(debugGL) { validateStart() }
-		glShaderSource(shaderID, 1, (source toCString())&, null)
+		glShaderSource(shaderID, 1, (String toConstCharPointerConstPointer((source toCString())&)), null)
```
davidhesselbom: The obvious backdraw here is that a static function is used. It would be much better off as a member function, but I don't know how the macro should be written, then. Notably, the result of the macro must be used immediately, because if it passes through an ooc function on the way, it becomes a normal `char **` again. @sebastianbaginski, any ideas?

source/sdk/lang/String.ooc
```diff
@@ -1,3 +1,11 @@
+include ./StringExtension
+
+extend String {
+    // This function only exists to silence a warning when calling OpenGL's glShaderSource function.
+    // It requires a const char * const * argument, but since ooc doesn't have const, we need to
+    // cast the char * into a const char * const * by using this C macro, defined in StringExtension.h.
+    toConstCharPointerConstPointer: static extern(CONST_CHAR_POINTER_CONST_POINTER_CAST) func (CString*) -> const CString*
```
davidhesselbom: Actually, this function name doesn't agree with the parameter types. Hang on...

sebastianbaginski:
> any ideas?

Not really, if I'm correct, pure ooc function can't return *const char*** and the opengl api expects a *const* parameter, so I guess you have to pass it through a macro / extern function like that.
Can a member function of a class/cover be extern ? I mean, technically it could be possible - assuming that first argument of this function is a pointer to class instance. I just don't know if it is possible in ooc to declate a member function as *extern*. Class/cover names are mangled when rock generates the C code anyway, so it would require you to hard-code some ugly class name.
I don't think there is a *nice-looking* solution to this.

PR #347, opened 2015-09-03
Random generator class with tests

sebastianbaginski:
Core generator code (random state updating) is copied from @davidpiuva ContrastAnalyzer.ooc.

### DIFF
--- a/source/math/FloatRandomGenerator.ooc
+++ b/source/math/FloatRandomGenerator.ooc
@@ -1,0 +1,76 @@
+//
+// Copyright (c) 2011-2014 Simon Mika <simon@mika.se>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+import math
+import os/Time
+
+FloatRandomGenerator: abstract class {
+	next: abstract func -> Float
+	next: func ~withCount (count: Int) -> Float[] {
+		result := Float[count] new()
+		for (i in 0 .. count)
+			result[i] = this next()
+		result
+	}
+}
+
+FloatUniformRandomGenerator: class extends FloatRandomGenerator {
+	_state: Int
+	_min := 0.0f
+	_max := 1.0f
+	init: func {
+		this _state = Time microtime()
+	}
+	init: func ~withSeed (seed: Int) {
+		this _state = seed
+	}
+	init: func ~withParameters (=_min, =_max, seed := Time microtime()) {
+		this _state = seed
+	}
+	free: func
+	next: func -> Float {
+		this _state = 214013 * this _state + 2531011
+		value := ((((this _state >> 16) & 0x7fff) % 10000) as Float) / 10000.0f
+		Float linearInterpolation(this _min, this _max, value)
+	}
+}
+
+FloatGaussianRandomGenerator: class extends FloatRandomGenerator {
+	_backend: FloatUniformRandomGenerator
+	_mu := 0.0f
+	_sigma := 1.0f
+	init: func {
+		this _backend = FloatUniformRandomGenerator new()
+	}
+	init: func ~withSeed (seed: Int) {
+		this _backend = FloatUniformRandomGenerator new(seed)
+	}
+	init: func ~withParameters (=_mu, =_sigma, seed := Time microtime()) {
+		this _backend = FloatUniformRandomGenerator new(seed)
+	}
+	init: func ~withBackend (=_backend)
+	init: func ~withBackendAndParameters (=_mu, =_sigma, =_backend)
+	free: func {
+		this _backend free()
+	}
+	next: func -> Float {
+		first := this _backend next()
+		if (first == 0.0f)
+			first = Float minimumValue
+		second := this _backend next()
+		value := sqrt(-2.0f * log(first)) * cos(2.0f * PI * second)
+		value * this _sigma + this _mu
+	}
+}
--- a/ooc-math.use
+++ b/ooc-math.use
@@ -1,4 +1,4 @@
 Name: ooc Math
 Description: Math functionality library.
 SourcePath: source/math
-Imports: FloatSize2D, FloatSize3D, FloatPoint2D, FloatPoint3D, FloatBox2D, FloatShell2D, FloatTransform2D, FloatTransform3D, IntSize2D, IntSize3D, IntPoint2D, IntPoint3D, IntBox2D, IntShell2D, IntTransform2D, FloatMatrix, FloatEuclidTransform, Quaternion, FloatComplex, FloatRotation3D, FloatVectorList, FloatPoint2DVectorList, FloatComplexVectorList, FloatConvexHull2D, BoolVectorList
+Imports: FloatSize2D, FloatSize3D, FloatPoint2D, FloatPoint3D, FloatBox2D, FloatShell2D, FloatTransform2D, FloatTransform3D, IntSize2D, IntSize3D, IntPoint2D, IntPoint3D, IntBox2D, IntShell2D, IntTransform2D, FloatMatrix, FloatEuclidTransform, Quaternion, FloatComplex, FloatRotation3D, FloatVectorList, FloatPoint2DVectorList, FloatComplexVectorList, FloatConvexHull2D, BoolVectorList, FloatRandomGenerator, IntRandomGenerator
--- a/source/math/IntRandomGenerator.ooc
+++ b/source/math/IntRandomGenerator.ooc
@@ -1,0 +1,68 @@
+//
+// Copyright (c) 2011-2014 Simon Mika <simon@mika.se>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use ooc-math
+import os/Time
+
+IntRandomGenerator: class {
+	_backend: FloatRandomGenerator
+	init: func ~withBackend (=_backend)
+	free: func {
+		this _backend free()
+	}
+	next: func -> Int {
+		this _backend next() as Int
+	}
+	next: func ~withCount (count: Int) -> Int[] {
+		result := Int[count] new()
+		for (i in 0 .. count)
+			result[i] = this next()
+		result
+	}
+}
+
+IntUniformRandomGenerator: class extends IntRandomGenerator {
+	init: func {
+		super(FloatUniformRandomGenerator new(0, 100))
+	}
+	init: func ~withSeed (seed: Int) {
+		super(FloatUniformRandomGenerator new(0, 100, seed))
+	}
+	init: func ~withParameters (min, max: Int, seed := Time microtime()) {
+		super(FloatUniformRandomGenerator new(min, max, seed))
+	}
+}
+
+IntGaussianRandomGenerator: class extends IntRandomGenerator {
+	init: func {
+		super(FloatGaussianRandomGenerator new(0.0f, 10.0f))
+	}
+	init: func ~withSeed (seed: Int) {
+		super(FloatGaussianRandomGenerator new(0.0f, 10.0f, seed))
+	}
+	init: func ~withParameters (mu, sigma: Float, seed := Time microtime()) {
+		super(FloatGaussianRandomGenerator new(mu, sigma, seed))
+	}
+	init: func ~withBackend (backend: FloatGaussianRandomGenerator) {
+		super(backend)
+	}
+	init: func ~withUniformBackend (backend: FloatUniformRandomGenerator) {
+		super(FloatGaussianRandomGenerator new~withBackendAndParameters(0.0f, 10.0f, backend))
+	}
+	init: func ~withUniformBackendAndParameters (mu, sigma: Float, backend: FloatUniformRandomGenerator) {
+		super(FloatGaussianRandomGenerator new~withBackendAndParameters(mu, sigma, backend))
+	}
+}
--- a/test/math/FloatRandomGeneratorTest.ooc
+++ b/test/math/FloatRandomGeneratorTest.ooc
@@ -1,0 +1,105 @@
+use ooc-unit
+use ooc-math
+import math
+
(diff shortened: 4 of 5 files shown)


sebastianbaginski:
So how do we handle *Int* data type here ? I just used *Float* for now. Sorry if I just misunderstood, my swedish isn't good yet :(

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,76 @@
...
+UniformRandomGenerator: class extends RandomGenerator {
+	_min := 0.0f
+	_max := 1.0f
+	init: func {
+		super()
+	}
+	init: func ~withSeed (seed: Int) {
+		super(seed)
+	}
+	init: func ~withParameters (=_min, =_max, seed := Time microtime()) {
```
davidhesselbom: Is `seed` going to be a different value each time you call this function, or the same one every time?

sebastianbaginski:
I decided to copy @davidpiuva s solution, as it is more efficient than the other one I had in mind (https://en.wikipedia.org/wiki/Blum_Blum_Shub , this one requires to compute x^2 and modulo with large number)

source/math/RandomGenerator.ooc
sebastianbaginski: Good point, I think it is the same situation we had with the vector default value. I'll check that

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+UniformRandomGenerator: class extends RandomGenerator {
+	_min := 0.0f
+	_max := 1.0f
+	init: func {
+		super()
+	}
+	init: func ~withSeed (seed: Int) {
```
davidhesselbom: You could also replace this function with
```ooc
init: func ~withSeed (seed := Time microtime(), min := 0.0f, max := 1.0f) {
   super(seed)
   this _min = min
   this _max = max
}
```
so that the order of the parameters is always the same. Downside is, you *could* call the function with a specified `min` parameter and leave out the `max` one, which would be weird.

source/math/RandomGenerator.ooc
davidhesselbom: Right, that's why I asked.

sebastianbaginski: Looks like the parameter is re-calculated each time, I tried to create three UniformRandomGenerator objects in a row, here is the printout:
> seed is 24542660
> seed is 24542673
> seed is 24542680

source/math/RandomGenerator.ooc
sebastianbaginski: You're right. But another downside is that each user who just wants to get uniform random distribution on a given interval will be forced to give the seed parameter as well.

davidhesselbom:
> So how do we handle Int data type here ? I just used Float for now.

Because neither namespaces nor generics work very well, you'll need to prepend `Float` to the name of all 3 classes, then create more or less identical copies of those classes, but prepend `Int` to the names instead. Lots of duplicate code, I know, but there it is.

source/math/RandomGenerator.ooc
davidhesselbom: That's interesting. How come that function is computed each time, and the other thing you tried to do was only computed once?

sebastianbaginski: Maybe because this one is global/extern function and the other one was a class member function ? I'm only guessing, I will check the generated code.

sebastianbaginski: Line from RandomGeneratorTest..c
```C
generator = RandomGenerator__UniformRandomGenerator_new_withParameters(15.0f, 20.0f, os_Time__Time_microtime());
```
so maybe the problem is that here the default value is a function call, and previously we had a member variable (*this count* or something) ? But shouldn't a member access through a property like this translate to a C function call each time ?

sebastianbaginski:
It is updated. What do you think about naming convention:
*RandomFloatGenerator* vs. *FloatRandomGenerator*
*UniformRandomFloatGenerator* vs. *FloatUniformRandomGenerator*
..
I'd prefer the first one ( <distribution>Random<type>Generator ), but the other one is ok too. Your call I guess.

davidhesselbom:
We've always put the type first in the name so far, so we may as well stick with that. I agree the other names look better, though.

source/math/RandomGenerator.ooc
davidhesselbom: Ah, that's true.

source/math/RandomGenerator.ooc
davidhesselbom: A property should be a function call under the hood, yes, but maybe they get special treatment by rock? I can't say.

sebastianbaginski:
> We've always put the type first in the name so far, so we may as well stick with that.

Updated. I will add more tests tomorrow, will try to make the tests *smarter* as well.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,134 @@
...
+	next: abstract func -> Float
+	next: func ~withCount (count: Int) -> Float[] {
+		result := Float[count] new()
+		for (i in 0 .. count)
+			result[i] = this next()
+		result
+	}
+	_next: func -> Float {
+		this _state = (214013 * this _state) + 2531011
+		((((this _state >> 16) & 0x7fff) % 10000) as Float) / 10000.0f
```
davidhesselbom: Yay lisp

sebastianbaginski:
Corrected Int generator classes and added better tests. *magic* reports *0* errors.
I think I'm done here.

test/math/RandomGeneratorTest.ooc
```diff
@@ -0,0 +1,134 @@
+use ooc-unit
+use ooc-math
+import math
+
+RandomGeneratorTest: class extends Fixture {
+	init: func {
+		super("RandomGenerator")
+		this add ("test uniform", func {
```
davidhesselbom: There's no need to put the word "test" in the string, here.

sebastianbaginski: True. Updated.

test/math/RandomGeneratorTest.ooc
```diff
@@ -0,0 +1,134 @@
...
+			b := 934
+			expectedMean := (b + a) / 2
+			tolerance := expectedMean / 10
+			intGenerator := IntUniformRandomGenerator new(a, b)
+			values := intGenerator next(1_000_000)
+			mean := 0
+			for (i in 0 .. values length)
+				mean += values[i]
+			mean /= values length
+			expect( Int absolute(mean - expectedMean) < tolerance )
```
davidhesselbom: Here, and in the next `expect`, you've put spaces inside the parentheses.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,132 @@
...
+	init: func ~withSeed (seed: Int) {
+		super(seed)
+	}
+	init: func ~withParameters (=_mu, =_sigma, seed := Time microtime()) {
+		super(seed)
+	}
+	next: func -> Float {
+		u1 := this _next()
+		if (u1 == 0) u1 = FLT_MIN
+		u2 := this _next()
```
davidhesselbom: `u1` and `u2` are funny looking names. How about `first` and `second`?

test/math/RandomGeneratorTest.ooc
sebastianbaginski: Updated.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,132 @@
...
+	}
+	init: func ~withSeed (seed: Int) {
+		super(seed)
+	}
+	init: func ~withParameters (=_mu, =_sigma, seed := Time microtime()) {
+		super(seed)
+	}
+	next: func -> Float {
+		u1 := this _next()
+		if (u1 == 0) u1 = FLT_MIN
```
davidhesselbom: `Float minimumValue` should work.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,132 @@
...
+		u2 := this _next()
+		result := sqrt(-2.0f * log(u1)) * cos(2.0f * PI * u2)
```
davidhesselbom: Technically, `result` isn't the result of this function...

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,132 @@
...
+		return result * this _sigma + this _mu
```
davidhesselbom: `result` needs not be written explicitly.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,132 @@
...
+	init: func ~withSeed (=_state)
+	next: abstract func -> Float
+	next: func ~withCount (count: Int) -> Float[] {
+		result := Float[count] new()
+		for (i in 0 .. count)
+			result[i] = this next()
+		result
+	}
+	_next: func -> Float {
+		this _state = (214013 * this _state) + 2531011
```
davidhesselbom: The parentheses should be necessary on this line.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,132 @@
...
+	next: abstract func -> Int
+	next: func ~withCount (count: Int) -> Int[] {
+		result := Int[count] new()
+		for (i in 0 .. count)
+			result[i] = this next()
+		result
+	}
+	_next: func -> Float {
+		this _state = (214013 * this _state) + 2531011
+		(((((this _state >> 16) & 0x7fff) % 10000) as Float) / 10000.0f)
```
davidhesselbom: Pretty sure you don't need all 5 parentheses at the start of this line.

sebastianbaginski:
Updated.

sebastianbaginski:
Base class added. This lack of generics makes things quite messed up here.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,140 @@
...
+		super(seed)
+	}
+	next: func -> Int {
+		first := this _next()
+		if (first == 0.0f)
+			first = Float minimumValue
+		second := this _next()
+		value := sqrt(-2.0f * log(first)) * cos(2.0f * PI * second)
+		(value * this _sigma + this _mu) as Int
+	}
```
davidhesselbom: This function is identical to the one in `FloatGaussianRandomGenerator`, but casts the result to an Int. @simonmika, I know you were against having function names like `nextGaussian`, but what if we 

- create a function `_nextGaussian` in the `RandomGenerator` base class, 
- move the code in `FloatGaussianRandomGenerator next` there, and
- let both `IntGaussianRandomGenerator` and `FloatGaussianRandomGenerator` return the value from `RandomGenerator _nextGaussian` instead? The API will still look the same to anyone outside this module, but there'll be less duplicate code.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,140 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+import math
+use ooc-collections
+import os/Time
+
+RandomGenerator: abstract class {
+	_state: Int
```
simonmika: Class needs to be called FloatRandomGenerator to allow for IntegerRandomGenerator.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,140 @@
...
+	init: func {
+		this _state = Time microtime()
+	}
+	init: func ~withSeed (=_state)
+	_next: func -> Float {
```
simonmika: Method should be public so that we can use a FloatRandomGenerator in the first place. Else this type is useless.

source/math/RandomGenerator.ooc
sebastianbaginski: Class FloatRandomGenerator is below, this one is a common base for Float and Int generators.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,140 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+import math
+use ooc-collections
+import os/Time
+
+RandomGenerator: abstract class {
```
simonmika: This class is useless as it can not do anything. There is no method that can be used.

source/math/RandomGenerator.ooc
```diff
@@ -0,0 +1,140 @@
...
+	}
+	init: func ~withSeed (seed: Int) {
+		super(seed)
+	}
+	init: func ~withParameters (=_min, =_max, seed := Time microtime()) {
+		super(seed)
+	}
+	next: func -> Int {
+		this _next() * (this _max - this _min) + this _min
+	}
```
simonmika: Use a FloatRandomGenerator as backend instead of inheritance.

source/math/RandomGenerator.ooc
sebastianbaginski: Method `_next` is used. I don't think this is a good design, though.

sebastianbaginski:
Updated. Now `FloatUniformGenerator` is basically used as a backend everywhere.
Integer generator classes are merely a set of constructors creating different variants of the backend generator.

source/math/FloatRandomGenerator.ooc
```diff
@@ -0,0 +1,78 @@
...
+		this _state = seed
+	}
+	init: func ~withParameters (=_min, =_max, seed := Time microtime()) {
+		this _state = seed
+	}
+	free: func
+	next: func -> Float {
+		Float linearInterpolation(this _min, this _max, this _next())
+	}
+	_next: func -> Float {
```
simonmika: I suggest moving this into next.

test/math/RandomGeneratorTest.ooc
```diff
@@ -0,0 +1,150 @@
+use ooc-unit
+use ooc-math
+import math
+
+RandomGeneratorTest: class extends Fixture {
+	init: func {
+		super("RandomGenerator")
+		this add("creating int generators with backends", func {
+			valuesCount := 1_000_000
+			generator := IntRandomGenerator new~withBackend(FloatUniformRandomGenerator new(0.0f, 20.0f))
```
simonmika: Why do we need `~withtBackend`?

test/math/RandomGeneratorTest.ooc
```diff
@@ -0,0 +1,150 @@
...
+			for (i in 0 .. valuesCount)
+				expect(Int absolute(generator next()) <= 20)
+			generator free()
+			generator = IntGaussianRandomGenerator new~withUniformBackend(FloatUniformRandomGenerator new(0.0f, 5.0f))
```
simonmika: Again.

test/math/RandomGeneratorTest.ooc
```diff
@@ -0,0 +1,150 @@
+use ooc-unit
+use ooc-math
+import math
+
+RandomGeneratorTest: class extends Fixture {
```
simonmika: Can we separate Integer from Float tests?

sebastianbaginski: Sure.

sebastianbaginski:
Updated, separate test files for Float and Int generators.

source/math/FloatRandomGenerator.ooc
```diff
@@ -0,0 +1,76 @@
...
+	init: func ~withBackendAndParameters (=_mu, =_sigma, =_backend)
+	free: func {
+		this _backend free()
+	}
+	next: func -> Float {
+		first := this _backend next()
+		if (first == 0.0f)
+			first = Float minimumValue
+		second := this _backend next()
+		value := sqrt(-2.0f * log(first)) * cos(2.0f * PI * second)
```
thomasfanell: ```(-2.0f * first log()) sqrt() * (2.0f * Float pi * second) cos()```

PR #351, opened 2015-09-09
Initial version of new Text cover

sebastianbaginski:
Initial version, so we have something to comment on.


sebastianbaginski:
Btw. how do I declate a *real* string literal in ooc ? `"this"` is automatically converted to a String object.

source/base/Text.ooc
```diff
@@ -0,0 +1,102 @@
...
+* This software is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with this software. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+MemoryOwnership: enum {
```
simonmika: Call this just Owner and move to a separate file. It should be usable in more places in the future.

source/base/Text.ooc
```diff
@@ -0,0 +1,102 @@
...
+	Stack,
+	Literal,
+	Caller,
+	Callee
+}
+
+Text: cover {
+	_buffer: Char*
+	_length: Int
+	_memoryMode: MemoryOwnership
```
simonmika: `_owner`

source/base/Text.ooc
```diff
@@ -0,0 +1,102 @@
...
+
+MemoryOwnership: enum {
+	Stack,
+	Literal,
+	Caller,
+	Callee
+}
+
+Text: cover {
+	_buffer: Char*
```
simonmika: CString ?

source/base/Text.ooc
```diff
@@ -0,0 +1,102 @@
...
+		this _buffer = string
+		this _memoryMode = MemoryOwnership Literal
+	}
+	init: func@ ~fromString (string: String) {
+		this _length = string length()
+		this _buffer = gc_malloc(this length)
+		memcpy(this _buffer, string _buffer data, this length)
+		this _memoryMode = MemoryOwnership Caller
+	}
+	empty?: func -> Bool {
```
simonmika: Do we like ? in identifiers?

source/base/Text.ooc
```diff
@@ -0,0 +1,102 @@
...
+	}
+	endsWith?: func ~char (c: Char) -> Bool {
+		if (this length > 0)
+			this _buffer[this length - 1] == c
+		false
+	}
+	take: func@ {
+		this _changeOwnership(MemoryOwnership Caller)
+	}
+	give: func@ {
```
simonmika: give should return a new struct so that the copy you have is unchanged or?

source/base/Text.ooc
sebastianbaginski: I don't know, I just copied this api from `String` class.

source/base/Text.ooc
sebastianbaginski: Ok, I was under the impression that `give` should *give* away the ownership of the original object.

source/base/Text.ooc
```diff
@@ -0,0 +1,102 @@
...
+		this _changeOwnership(MemoryOwnership Callee)
+	}
+	callerFree: func@ {
```
simonmika: This should only be free without any ownership test.

source/base/Text.ooc
```diff
@@ -0,0 +1,102 @@
...
+		if (this _memoryMode == MemoryOwnership Caller) {
+			gc_free(this _buffer)
+			this _length = 0
+		}
+	}
+	calleeFree: func@ {
```
simonmika: Better name?

test/base/TextTest.ooc
```diff
@@ -0,0 +1,29 @@
+use ooc-base
+use ooc-unit
+
+TextTest: class extends Fixture {
+	init: func {
+		super("Text")
+		this add("literals", func {
+			t := Text new~fromStringLiteral("test string" as CString, 5)
```
simonmika: The way I know of to create CStrings as literals is: `c"test string"`.

source/base/Text.ooc
sebastianbaginski: `safeFree`
`safeDelete`
...
?

sebastianbaginski: Not that good names though, it should reflect that the *Calee* is done using this object and it can be safely released.

source/base/Text.ooc
```diff
@@ -0,0 +1,275 @@
...
+		}
+		result
+	}
+	toFloat: func -> Float {
+		this toLDouble() as Float
+	}
+	toDouble: func -> Double {
+		this toLDouble() as Double
+	}
+	toLDouble: func -> LDouble {
```
simonmika: Should it be `toLongDouble`, or?

source/base/Text.ooc
```diff
@@ -0,0 +1,275 @@
...
+		if (this startsWith(This new(c"0x", 2)))
+			result = 16
+		result
+	}
+	/*empty ::= static This new()*/
+}
+
+operator [] (string: Text, index: Int) -> Char { string _buffer [index] }
+operator [] (string: Text, range: Range) -> Text { string slice(range min, range count) }
+operator == (str1: String, str2: Text) -> Bool { str2 equals(str1) }
```
simonmika: Call the arguments left and right.

source/base/Text.ooc
sebastianbaginski: You prefer it like that ? I just copied the api from String, again.

source/base/Text.ooc
```diff
@@ -0,0 +1,275 @@
...
+	_detectNumericBase: func -> Int {
+		result := 10
+		if (this startsWith(This new(c"0x", 2)))
+			result = 16
+		result
+	}
+	/*empty ::= static This new()*/
+}
+
+operator [] (string: Text, index: Int) -> Char { string _buffer [index] }
```
simonmika: Can and therefore should be a part of class instead.

source/base/Text.ooc
```diff
@@ -0,0 +1,275 @@
...
+operator [] (string: Text, range: Range) -> Text { string slice(range min, range count) }
+operator == (str1: String, str2: Text) -> Bool { str2 equals(str1) }
+operator == (str1: Text, str2: String) -> Bool { str1 equals(str2) }
```
simonmika: Duplicate of `equals`. Maybe skip `equals`, or?

source/base/TextBuilder.ooc
```diff
@@ -0,0 +1,140 @@
...
+*/
+use ooc-base
+use ooc-collections
+
+TextBuilder: class {
+	_data := VectorList<Text> new(32, false)
+	count ::= this _data count
+
+	init: func ~default
+	init: func ~string (value: Text) {
```
simonmika: `~text`

source/base/TextBuilder.ooc
```diff
@@ -0,0 +1,140 @@
...
+	}
+	init: func ~this (original: This) {
+		this init()
+		for (i in 0 .. original count) {
+			this append(original[i] copy())
+		}
+	}
+	free: func {
+		for (i in 0 .. this count)
+				this _data[i] free()
```
simonmika: `free(Owner Callee)`?

source/base/TextBuilder.ooc
```diff
@@ -0,0 +1,140 @@
...
+			prepend(other[other count -1 -i] copy())
+	}
+	toString: func -> String {
+		result := ""
+		for (i in 0 .. this _data count)
+			result = result << this _data[i] toString()
+		result
+	}
+	toText: func -> Text {
+		this join(Text new())
```
simonmika: `Text empty`?

sebastianbaginski: It won't compile with static property.

source/base/TextBuilder.ooc
```diff
@@ -0,0 +1,140 @@
...
+				offset += this[i] length
+				if (separator length > 0 && i < this count - 1) {
+					memcpy(buffer + offset, separator raw, separator length)
+					offset += separator length
+				}
+			}
+		}
+		Text new(buffer, totalLength, Owner Caller)
+	}
+	println: func {
```
simonmika: Should we introduce a better name?

source/base/Text.ooc
```diff
@@ -0,0 +1,298 @@
...
+		if (_isNumeric(c, base)) {
+			if (base != 16)
+				result = (c - '0') as Int
+			else {
+				if (c >= 'a' && c <= 'f')
+					result = (c - 'a') as Int + 10
+				else if (c >= 'A' && c <= 'F')
+					result = (c - 'A') as Int + 10
+				else if (c >= '0' && c <= '9')
+					result = (c - '0') as Int
```
davidhesselbom: The digit case is arguably the most common one, so shouldn't that be the first case?

source/base/Text.ooc
```diff
@@ -0,0 +1,298 @@
...
+		result := (c >= '0') && (c <= '9')
+		if (!result && base == 16)
+			result = (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')
+		result
+	}
+	_toInt: static func (c: Char, base: Int) -> Int {
+		result := 0
+		if (_isNumeric(c, base)) {
+			if (base != 16)
+				result = (c - '0') as Int
```
davidhesselbom: Are we assuming that no one is going to try to use this function with, say, base 12?

source/base/Text.ooc
```diff
@@ -0,0 +1,298 @@
...
+					}
+				}
+				if (this[index] == 'e' || this[index] == 'E') {
+					exponent := this slice(index + 1, this count - index) toInt~inBase(10)
+					result = result * pow(10, exponent)
+				}
+			}
+			result *= sign
+		}
+		result
```
davidhesselbom: This is a pretty long function. Is there any way you could split it up somewhat?

source/base/Text.ooc
```diff
@@ -0,0 +1,298 @@
...
+	}
+	empty: static This { get { This new() } }
+}
+
+operator == (left: String, right: Text) -> Bool { Text new(left) == right }
+operator != (left: String, right: Text) -> Bool { !(left == right) }
+operator != (left: Text, right: String) -> Bool { !(left == right) }
+operator != (left: Text, right: Text) -> Bool { !(left == right) }
```
davidhesselbom: This is a very long file... and some function are located far away from where they are called, only because they're private and the calling functions are not. I don't see an easy way to fix that, though.

source/base/Text.ooc
sebastianbaginski: Yes, for now.
Most common bases (2,8,10,16) are supported, so we can just put a comment about not fully supporting numeric bases for which single digit can occupy more than one character.
For simplicity this class assumes `sizeof(single character)==1`, this will have to be updated anyway when UTF-16 charset is introduced, someday. Then we can revisit the conversion functions.

davidhesselbom: And no error handling in case someone tries to use an unsupported base?

source/base/Text.ooc
davidhesselbom: Unless we change `LDouble` to `LongDouble`, I don't think `toLongDouble` makes a lot of sense.

source/base/TextBuilder.ooc
```diff
@@ -0,0 +1,141 @@
...
+	}
+	init: func ~cstring (value: CString, length: Int) {
+		this init()
+		this append(value, length)
+	}
+	init: func ~this (original: This) {
+		this init()
+		for (i in 0 .. original count) {
+			this append(original[i] copy())
+		}
```
davidhesselbom: Remove these brackets.

source/base/Text.ooc
sebastianbaginski: Fixed.

davidhesselbom: I don't see the fix...

source/base/TextBuilder.ooc
```diff
@@ -0,0 +1,140 @@
...
+	copy: func -> This {
+		This new(this)
+	}
+	append: func ~cstring (value: CString, length: Int) {
+		this append(Text new(value, length))
+	}
+	append: func ~text (value: Text) {
+		this _data add(value)
+	}
+	append: func ~This (other: This) {
```
davidhesselbom: Having no suffix when the argument is of type `This`, as you did in `Text`, seems like a good idea to me.

source/base/Text.ooc
sebastianbaginski:
```ooc
_isNumeric: static func (c: Char, base: Int) -> Bool {
		version(safe) {
			if (base < 2 || (base > 10 && base != 16))
				raise("Unsupported numeric base in Text")

```

davidhesselbom: Ah, haha. Gotcha. Thanks.

source/base/TextBuilder.ooc
sebastianbaginski: Sure, fixed.

source/base/Text.ooc
```diff
@@ -0,0 +1,302 @@
...
+		if (this beginsWith(This new(c"0x", 2)))
+			result = 16
+		result
+	}
+	_isNumeric: static func (c: Char, base: Int) -> Bool {
+		version(safe) {
+			if (base < 2 || (base > 10 && base != 16))
+				raise("Unsupported numeric base in Text")
+		}
+		result := (c >= '0') && (c <= '9')
```
davidhesselbom: If `base` is 4, is `'9'` numeric?

davidhesselbom: `result := (c >= '0') && (c <= '9') && (c < base)` will mean an unnecessary check in the event that `base` is 16, but it should still be more correct.

sebastianbaginski: :) Fixed. Updated tests.

davidhesselbom:
I merged something else, and now there's a conflict in `ooc-base.use`.

sebastianbaginski:
> I merged something else, and now there's a conflict in ooc-base.use.

Rebased, should be ok now.

source/base/TextBuilder.ooc
```diff
@@ -0,0 +1,140 @@
...
+		this append(value, length)
+	}
+	init: func ~this (original: This) {
+		this init()
+		for (i in 0 .. original count)
+			this append(original[i] copy())
+	}
+	free: func {
+		for (i in 0 .. this count)
+				this _data[i] free(Owner Callee)
```
davidhesselbom: Indentation

sebastianbaginski: Fixed. Is it a possible catch for Magic ?

davidhesselbom: Magic doesn't understand much about the code yet, but yes, it should be able to detect when the number of leading tabs on a line is more than (the corresponding number for the above line + 1). Will you make an issue for this?

Remember that the intentation sometimes decreases by more than one tab from one line to the next because of omitted brackets, though, so the rule should only apply when the number of leading tabs *increases* by more than one from one line to the next.

source/base/TextBuilder.ooc
simonmika: Lets call it witeLine.

simonmika:
Can we get this ready?

source/base/TextBuilder.ooc
davidhesselbom: I assume you mean `writeLine`. Works for me, but we could keep also this one and let one call the other.

davidhesselbom:
It was merged before you even wrote that.

PR #352, opened 2015-09-09
Added blend function with opacity as parameter.

emilwestergren:


source/draw/gpu/GpuSurface.ooc
```diff
@@ -81,10 +81,10 @@ GpuSurface: abstract class {
 	_bind: virtual func
 	_unbind: virtual func
 	clear: abstract func
+	blend: func (factor: Float) { this _context blend(factor) }
```
simonmika: Blend should be a state on the surface.

emilwestergren: OK, I just copied the code from the C# version of Kean where it is a function. Also by doing it this way I didn't have to adapt anything in Vidproc and the old blend hack for monochromes still works. So I thought it was a pretty clean solution for now but I can fix everything in a larger PR if you want me to.

emilwestergren:
Updated.

emilwestergren:
Updated.

source/draw/gpu/GpuContext.ooc
```diff
@@ -56,6 +56,7 @@ GpuContext: abstract class {
 	drawPoints: abstract func (pointList: VectorList<FloatPoint2D>, transform: FloatTransform3D)
 	drawQuad: abstract func
 	enableBlend: abstract func (blend: Bool)
+	blend: abstract func (factor: Float)
```
simonmika: Is GpuContext accessable outside this compontent?

emilwestergren: Yes

simonmika: Then this should not be here. It should be in the backend not in the frontend. The same goes for the other methods as well.

emilwestergren: True

emilwestergren:
Updated.

PR #354, opened 2015-09-10
Changed use files to one statement per line.

niklasborwell:
I suggest changing how we write statements in use files to this. 
Partly to make it more readable in general, and partly to avoid the unnecessary conflicts that arise when all the information is on the same row.


ooc-math.use
```diff
@@ -1,4 +1,30 @@
...
+Imports: Quaternion
+Imports: FloatComplex
+Imports: FloatRotation3D
+Imports: FloatVectorList
+Imports: FloatPoint2DVectorList
+Imports: FloatComplexVectorList
+Imports: FloatConvexHull2D
+Imports: BoolVectorList
+Imports: FloatRandomGenerator
+Imports: IntRandomGenerator
```
davidhesselbom: As per POSIX standards, all files need to end with a newline. This should be handled by your editor. Check your settings and amend.

ooc-base.use
```diff
@@ -1,4 +1,23 @@
...
+Imports: Synchronized
+Imports: ReferenceCounter
+Imports: Event
+Imports: ThreadPool
+Imports: Container
+Imports: Timer
+Imports: Debug
+Imports: Profiler
+Imports: CLIParser
+Imports: StringBuilder
```
davidhesselbom: Would it be a good idea to sort this list alphabetically?

ooc-sdk-lite.use
```diff
@@ -1,4 +1,17 @@
...
+Imports: lang/Exception
+Imports: lang/Format
+Imports: lang/IO
+Imports: lang/Iterators
+Imports: lang/Memory
+Imports: lang/Numbers
+Imports: lang/String
+Imports: lang/stdlib
+Imports: lang/types
+Imports: lang/VarArgs
```
davidhesselbom: Hmm. If all of these classes are in `lang`, can't we change the `SourcePath` to `source/sdk/lang`?

ooc-math.use
niklasborwell: Fixed.

ooc-sdk-lite.use
niklasborwell: I don't know..

ooc-base.use
niklasborwell: I don't really think it matters.

davidhesselbom: Okay, so if someone adds another import here, where should it be added? At the bottom?

niklasborwell: I don't think it matters that much either. At the bottom, or rather at the end, is the way it was before.

ooc-sdk-lite.use
simonmika: We should investigate a change of source paths.

davidhesselbom:
Looks good to me, aside from the questions above. @simonmika, should I merge this?

PR #356, opened 2015-09-10
fixed an error in map()

thomasfanell:
```result[i] = function(this[i])``` does not update the list count, thus causing the resulting vector list to be in an invalid state.


thomasfanell:
Eh ok, my CSV experimentation file apparently invited itself in here... i'll remove them for now

davidhesselbom:
I don't want the CSV commits. Create a new branch at the "fixed an error in map()" commit, push that branch to your `origin`, and a open a new pull request using that new branch.

PR #359, opened 2015-09-14
fixed an error in map()

thomasfanell:
```result[i] = function(this[i])``` does not update the list count, thus causing the resulting vector list to be in an invalid state.


davidhesselbom:
Is there a test for this?

thomasfanell:
I found it when messing about with my CSV reader/writer API, which makes use of VectorList. I noticed it when I tried to iterate my CSV reader after performing a map().

thomasfanell:
There is a test, but it is commented out, now we know why? :-)

PR #361, opened 2015-09-14
Attempting to use Ubuntu 14.04 on Travis.

davidhesselbom:


davidhesselbom:
It doesn't look like this is going anywhere. We're going to migrate our tests to Circle CI instead.

PR #363, opened 2015-09-15
Added methods to copy row and column of raster image to given buffer

sebastianbaginski:
Fixed naming conventions.
Fixed possible bug - it was possible to access non existing row or column in *safe* version.

### DIFF
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -189,7 +189,7 @@ RasterMonochrome: class extends RasterPacked {
 
 	getValue: func (x, y: Int) -> UInt8 {
 		version(safe) {
-			if (x > this size width || y > this size height || x < 0 || y < 0)
+			if (x >= this size width || y >= this size height || x < 0 || y < 0)
 				raise("Accessing RasterMonochrome index out of range in getValue")
 		}
 		(this buffer pointer[y * this stride + x])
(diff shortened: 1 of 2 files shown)


source/draw/RasterMonochrome.ooc
```diff
@@ -189,7 +189,7 @@ RasterMonochrome: class extends RasterPacked {
 
 	getValue: func (x, y: Int) -> UInt8 {
 		version(safe) {
-			if (x > this size width || y > this size height || x < 0 || y < 0)
+			if (x >= this size width || y >= this size height || x < 0 || y < 0)
```
davidhesselbom: Oh dear. Nice catch...

source/draw/RasterMonochrome.ooc
```diff
@@ -197,30 +197,35 @@ RasterMonochrome: class extends RasterPacked {
...
+	getRowInto: func (row: Int, vector: FloatVectorList) {
 		version(safe) {
-			if (y > this size height || y < 0)
+			if (row >= this size height || row < 0)
 				raise("Accessing RasterMonochrome index out of range in getRow")
 		}
-		for (x in 0 .. (this size width))
-				result add(this buffer pointer[y * this stride + x] as Float)
+		for (x in 0 .. this size width)
+				vector add(this buffer pointer[row * this stride + x] as Float)
```
davidhesselbom: I'm tempted to suggest making a new function `FloatVectorList` that takes an `Int*` and inserts each element without using generics (as we did before for sorting and other things), but I guess performance isn't an issue here anyway.

source/draw/RasterMonochrome.ooc
```diff
@@ -197,30 +197,35 @@ RasterMonochrome: class extends RasterPacked {
...
+	}
+	getRowInto: func (row: Int, vector: FloatVectorList) {
 		version(safe) {
-			if (y > this size height || y < 0)
+			if (row >= this size height || row < 0)
 				raise("Accessing RasterMonochrome index out of range in getRow")
 		}
-		for (x in 0 .. (this size width))
-				result add(this buffer pointer[y * this stride + x] as Float)
+		for (x in 0 .. this size width)
```
davidhesselbom: If you're going to call `y` `row`, you may as well call `x` `column` as well. Same thing in `getColumnInto`.

source/draw/RasterMonochrome.ooc
sebastianbaginski: This is probably a good idea, but maybe in a separate pull request. I wrote these functions because the code I'm working on called 'getColumn' and 'getRow' a lot, allocating and destroying large buffers over and over again, causing really bad performance.

davidhesselbom: Sure, fix it later.

source/draw/RasterMonochrome.ooc
sebastianbaginski: Corrected.

davidhesselbom:
There's no test for this, is there?

sebastianbaginski:
> There's no test for this, is there?

There is a RasterMonochromeTest, but `getRow` and `getColumn` are not tested. Do you want me to write the missing tests ?

davidhesselbom:
Please do. This would be a good time to do that.

sebastianbaginski:
OK, no problem.

davidhesselbom:
You may add the tests in a new commit, btw.

davidhesselbom:
But thanks for amending when you made the other fixes :)

sebastianbaginski:
> But thanks for amending when you made the other fixes :)

Sure;)

sebastianbaginski:
Updated with test for *getColumn* and *getRow*.

PR #366, opened 2015-09-15
Added two test cases for HeapVector that both fails.

davidpiuva:
The first test require the ooc compiler to be rewritten to allow passing elements by reference. As it is now, you have to create a new cover and assign it by value over the old cover.

The second test is nesting of heap vectors. More tests can be added to it once initial assignment of the inner vectors can be done without segmentation fault.


davidpiuva:
This PR is not ready but I can't access the labels for some reason.

PR #370, opened 2015-09-16
Fixed memory leak in FFT

sebastianbaginski:
Also few small formatting corrections.

### DIFF
--- a/source/math/FloatComplexVectorList.ooc
+++ b/source/math/FloatComplexVectorList.ooc
@@ -99,14 +99,14 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 	}
 	discreteFourierTransform: static func (input: This) -> This {
 		result := This createDefault(input count)
-		for (i in 0 .. (input count))
-			for (j in 0 .. (input count))
+		for (i in 0 .. input count)
+			for (j in 0 .. input count)
 				result[i] = result[i] + input[j] * FloatComplex rootOfUnity(input count, -i * j)
 		result
 	}
 	inverseDiscreteFourierTransform: static func (input: This) -> This {
-		conjugates := This new()
-		for (i in 0 .. (input count))
+		conjugates := This new(input count)
+		for (i in 0 .. input count)
 			conjugates add(input[i] conjugate)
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
`oddInput` and `evenInput` temporary vectors were never *free*d .
I know you said something about reimplementing it, but here is a fix anyway.

source/math/FloatComplexVectorList.ooc
```diff
@@ -134,18 +133,20 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 				result[i] = evenOutput[i] + root * oddOutput[i]
 				result[halfLength + i] = evenOutput[i] - root * oddOutput[i]
 			}
+			evenInput free()
+			oddInput free()
```
davidhesselbom: I see a potential for doing this work in-place, which would greatly reduce memory allocation and deallocation... probably shouldn't spend time on that right now though, but it would be something to consider in the future.

sebastianbaginski: I am working on this now.

sebastianbaginski: > probably shouldn't spend time on that right now though
If I can't make it today then I'll skip it.

davidhesselbom: Alright, give it a shot.

davidhesselbom:
I didn't say to reimplement the FFT algorithm, only that the blur test thing shouldn't use a 1D DFT for a 2D FFT task. It may well be possible to make our own FFT implementation that's fast enough, it just seemed like unnecessary work when there are highly specialized libraries available. Also, even though this pull request is a big improvement, it doesn't give us 2D FFT, which, again, I don't know how hard it would be to implement.

sebastianbaginski:
Ok, I was under the impression that you are not happy with the FFT code from Kean.

davidhesselbom:
No, that wasn't my problem, sorry for the confusion.

sebastianbaginski:
Ok.
I guess this pull request is ready, I will create another one if I come up with something for this FFT.

PR #372, opened 2015-09-16
Restructured GPU API

emilwestergren:


source/draw/gpu/GpuCanvas.ooc
```diff
@@ -18,82 +18,41 @@ use ooc-math
...
+GpuCanvasYuv420Semiplanar: class extends GpuSurface {
+	_target: GpuYuv420Semiplanar
 	transform: FloatTransform3D {
 		get { this _view }
 		set(value) {
 			this _view = this _toLocal * value * this _toLocal
-			this target y canvas _view = this _view
-			this target uv canvas _view = FloatTransform3D createTranslation(-this _view m / 2.0f, -this _view n / 2.0f, -this _view o / 2.0f) * this _view
+			this _target y canvas _view = this _view
+			this _target uv canvas _view = FloatTransform3D createTranslation(-this _view m / 2.0f, -this _view n / 2.0f, -this _view o / 2.0f) * this _view
```
davidhesselbom: Can't you do
```ooc
this _target uv canvas _view = FloatTransform3D createTranslation(-this _view m, -this _view n, -this _view o) / 2.0f * this _view
``` instead?
```

davidhesselbom: Or use a FloatSize or something... I dunno.

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -18,82 +18,41 @@ use ooc-math
...
+			this _target uv canvas focalLength = value / 2
 		}
 	}
 
-	init: func (image: GpuYuv420Semiplanar, context: GpuContext) {
-		super(image, context)
-		this target uv canvas clearColor = ColorBgra new(128, 128, 128, 128)
+	init: func (=_target, context: GpuContext) {
+		super(this _target size, context, context defaultMap, IntTransform2D identity)
+		this _target uv canvas clearColor = ColorBgra new(128, 128, 128, 128)
```
davidhesselbom: `ColorBgra(128, 128)` should do it.

source/draw/gpu/pc/Window.ooc
```diff
@@ -23,7 +23,7 @@ use ooc-base
 import X11/X11Window
 import X11/include/x11
 
-Window: class extends GpuSurface {
+Window: class extends OpenGLSurface {
```
davidhesselbom: Is `Window` always going to be an `OpenGLSomething`?

source/draw/gpu/GpuCanvas.ooc
emilwestergren: No, that operator is not defined for FloatTransform3D and even if it was it would not be correct.

source/draw/gpu/pc/Window.ooc
emilwestergren: Kind of, I have another PR which sorts out the Window mess where this class will be named OpenGLWindow instead and be a part of the OpenGL module.

davidhesselbom: Okay, looking forward to it. I'd like to get input working at some point and it doesn't make sense to have it as part of the OpenGL stuff.

source/draw/gpu/GpuYuv420Semiplanar.ooc
```diff
@@ -16,23 +16,29 @@
...
 GpuYuv420Semiplanar: class extends GpuImage {
-	_y: GpuMonochrome
+	_y: GpuImage
 	y ::= this _y
-	_uv: GpuUv
+	_uv: GpuImage
 	uv ::= this _uv
-
+	filter: Bool {
+		get
```
davidhesselbom: ... I feel stupid now, but, what does this return?

emilwestergren: Probably false, but it should return `this _y filter`

davidhesselbom: And how does it know it should return `this _y filter`?

source/draw/gpu/opengl/backend/gles3/Renderer.ooc
```diff
@@ -0,0 +1,27 @@
...
+import include/gles3
+Renderer: class {
+	_quad: Quad
+	init: func { this _quad = Quad new() }
+	free: override func {
+		this _quad free()
+		super()
+	}
+	drawQuad: func { this _quad draw() }
+	drawLines: func (positions: Float*, count: Int, dimensions: Int, lineWidth: Float) {
```
davidhesselbom: `count, dimensions: Int`

source/draw/gpu/opengl/backend/gles3/Renderer.ooc
```diff
@@ -0,0 +1,27 @@
...
+		glLineWidth(lineWidth)
+		glBindVertexArray(0)
+		glEnableVertexAttribArray(0)
+		glVertexAttribPointer(0, dimensions, GL_FLOAT, 0, 0, positions)
+		glDrawArrays(GL_LINE_STRIP, 0, count)
+		glDisableVertexAttribArray(0)
+	}
+	drawPoints: func (positions: Float*, count: Int, dimensions: Int) {
```
davidhesselbom: `count, dimensions: Int`

source/draw/gpu/GpuYuv420Semiplanar.ooc
emilwestergren: It doesn't, I will fix it.

davidhesselbom: Okay, good. For a moment there, I thought rock figured it out by itself and it seemed a little crazy.

source/draw/gpu/opengl/backend/gles3/Fbo.ooc
```diff
@@ -65,7 +64,7 @@ Fbo: class {
 		if (status != GL_FRAMEBUFFER_COMPLETE) {
 			statusMessage := getErrorMessage(status)
 			errorMessage := "Framebuffer Object creation failed with status: " + statusMessage + " for texture of size " +
-			texture width toString() + " x " + texture height toString()
+			texture size width toString() + " x " + texture size height toString()
```
davidhesselbom: This will leak memory...

emilwestergren: Do we care? This error should not occur and we probably only want to check for it in debug builds.

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -18,82 +18,41 @@ use ooc-math
...
+			this _target uv canvas focalLength = value / 2
 		}
 	}
 
-	init: func (image: GpuYuv420Semiplanar, context: GpuContext) {
-		super(image, context)
-		this target uv canvas clearColor = ColorBgra new(128, 128, 128, 128)
+	init: func (=_target, context: GpuContext) {
+		super(this _target size, context, context defaultMap, IntTransform2D identity)
+		this _target uv canvas clearColor = ColorBgra new(128, 128)
```
davidhesselbom: I messed up, this isn't going to work.

emilwestergren: Yeah I noticed it.

emilwestergren:
Updated.

source/draw/gpu/opengl/backend/gles3/Fbo.ooc
davidhesselbom: Okay, so it's only supposed to happen once, if it happens at all?

emilwestergren: Yes, the error throws an exception so unless we catch it (which we don't) it will crash the application.

davidhesselbom: Alright then.

davidhesselbom:
Fails magic.

emilwestergren:
Updated.

emilwestergren:
Updated.

PR #376, opened 2015-09-16
Added methods to calculate FFT in place and with preallocated buffer

sebastianbaginski:
`fastFourierTransform` method will now allocate the needed buffer once instead of allocating and deleting smaller chunks multiple times.
It is now possible to provide a buffer used when calculating FFT. Convenience function to allocate a buffer with correct size is added, so it is possible to allocate it once and reuse between FFT calls.
Added method to process complex array in place (and with optional buffer as well).
Added tests to validate new functionality.

### DIFF
--- a/source/math/FloatComplexVectorList.ooc
+++ b/source/math/FloatComplexVectorList.ooc
@@ -116,29 +116,42 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 	}
 	fastFourierTransform: static func (input: This) -> This {
 		result := This createDefault(input count)
-		if (input count == 1)
-			result[0] = input[0]
-		else {
-			halfLength: Int = input count / 2
-			evenInput := This new(halfLength)
-			oddInput := This new(halfLength)
-			for (i in 0 .. halfLength) {
-				evenInput add(input[2 * i])
-				oddInput add(input[2 * i + 1])
-			}
-			evenOutput := This fastFourierTransform(evenInput)
-			oddOutput := This fastFourierTransform(oddInput)
-			for (i in 0 .. halfLength) {
-				root := FloatComplex rootOfUnity(input count, -i)
-				result[i] = evenOutput[i] + root * oddOutput[i]
-				result[halfLength + i] = evenOutput[i] - root * oddOutput[i]
-			}
-			evenInput free()
-			oddInput free()
-			evenOutput free()
-			oddOutput free()
+		if (input count > 0) {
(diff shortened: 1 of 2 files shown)


source/math/FloatComplexVectorList.ooc
```diff
@@ -161,4 +174,32 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 			result add(value)
 		result
 	}
+	_fastFourierTransformBufferSize: static func (inputSize: Int) -> Int {
+		logValue := inputSize as Float log() / 2.0f log()
```
sebastianbaginski: Ok, so how do I calculate `log2(value)` ? :)

davidhesselbom: There's doesn't appear to be a function called `log2`. Maybe we should add one in `math.ooc`?

sebastianbaginski:
I think with current implementation we can't escape allocating a buffer for even and odd items from the complex array. I saw implementations of FFT done without any buffering, but they were for very specific array sizes (when `vector count` is a power of two).
Anyway this should speed up things a little in some cases. It certainly did with the motion evaluator filter I'm working on, it is possible to use it now.

source/math/FloatComplexVectorList.ooc
sebastianbaginski: I guess it could be useful. Maybe a separate pull request with missing *Math* functions (log2, log10) ? Can you think about any more functions which are missing and could be useful ?

davidhesselbom:
Right, things are going to get messy when the size isn't a power of 2, and many libraries don't even support any other sizes, while others are limited to powers of low primes. Even CUFFT only allows sizes on the form 2<sup>a</sup> × 3<sup>b</sup> × 5<sup>c</sup> × 7<sup>d</sup>, in-place or not, so don't beat yourself up over it.

source/math/FloatComplexVectorList.ooc
davidhesselbom: Yes, put them in a separate pull request. If you want, I can merge that one first so you can edit this one to use them. I don't have a list of functions I'd like to add, no. but I've seen functions lying around in various classes that only deal with `Float`s and that should probably be moved there.

sebastianbaginski: Alright, you can merge this. I will take a look at updating the *Math* tomorrow.

sebastianbaginski:
Updated with `log2`.

PR #379, opened 2015-09-16
CSV API

thomasfanell:


thomasfanell:
Apparently I have a different version of Rock installed here at home, as this compiles just fine here. I'm also having some issues with generics when attempting to use map(), ```Missing info for type argument S. Have you forgotten to qualify Iterator map<T, S> (function : Func (T) -> S :Func (T) -> S) -> Iterator<S>, e.g. List<Int>?```

ooc-io.use
```diff
@@ -0,0 +1,7 @@
+Name: ooc IO
+Description: IO library
+SourcePath: source/io
+Requires: ooc-base
+Requires: ooc-sdk
```
davidhesselbom: Try `ooc-sdk-lite` instead.

thomasfanell:
I think this is ready for basic use. There are some bells and whistles that are yet to be implemented, as i'm still having some issues with generics. Other than that, is there anything else that we really **need**, before we can merge this? @simonmika @davidhesselbom

source/io/CsvReader.ooc
```diff
@@ -0,0 +1,80 @@
...
+	_fileReader: FileReader
+	init: func (=_fileReader)
+	free: func {
+		if (this _fileReader != null) {
+			this _fileReader close()
+			this _fileReader free()
+		}
+		super()
+	}
+	open: static func (filename: Text) -> This {
```
simonmika: Put static fields, properties and methods at the end of the file.

source/io/CsvWriter.ooc
```diff
@@ -0,0 +1,31 @@
...
+		super()
+	}
+	open: static func (filename: Text) -> This {
+		result: This
+		file := File new(filename toString())
+		result = This new(FileWriter new(file))
+		file free()
+		result
+	}
+	writeRow: func (row: VectorList<Text>) {
```
simonmika: Just `write`?

source/io/CsvReader.ooc
```diff
@@ -0,0 +1,80 @@
...
+				textBuilder append(readCharacter)
+			result = this _parseRow(textBuilder toText())
+			textBuilder free()
+		}
+		result
+	}
+	_parseRow: func (rowData: Text) -> VectorList<Text> {
+		result := VectorList<Text> new()
+		rowDataLength := rowData count
+		position := 0
```
davidhesselbom: This variable doesn't appear to be used anywhere.

source/io/CsvReader.ooc
```diff
@@ -0,0 +1,80 @@
...
+		}
+		rowData free()
+		result
+	}
+	_extractStringLiteral: func (rowData: Text, position: Int*) -> Text {
+		result: Text
+		readCharacter: Char
+		textBuilder := TextBuilder new()
+		rowDataLength := rowData count
+		while (position@ < rowDataLength && (readCharacter = rowData[position@]) != '"') {
```
davidhesselbom: Why `position@`?

davidhesselbom: Oh wait. Nevermind...

source/io/CsvReader.ooc
```diff
@@ -0,0 +1,80 @@
...
+		result = textBuilder toText()
+		textBuilder free()
+		result
+	}
+	open: static func (filename: Text) -> This {
+		result: This
+		file := File new(filename toString())
+		if (file exists?())
+			result = This new(FileReader new(file))
+		file free()
```
davidhesselbom: So... this can return `null`, then?

source/io/CsvReader.ooc
```diff
@@ -0,0 +1,80 @@
...
+			textBuilder append(rowData[position@])
+			++position@
+		}
+		++position@
+		result = textBuilder toText()
+		textBuilder free()
+		result
+	}
+	open: static func (filename: Text) -> This {
+		result: This
```
davidhesselbom: Might be a good idea to set this: `result : This = null`

source/io/CsvWriter.ooc
```diff
@@ -0,0 +1,31 @@
...
+			if (i < row count - 1)
+				this _fileWriter write(";")
+		}
+		this _fileWriter write("\n")
+	}
+	open: static func (filename: Text) -> This {
+		result: This
+		file := File new(filename toString())
+		result = This new(FileWriter new(file))
+		file free()
```
davidhesselbom: The whole function could be written as 
```ooc
This new(FileWriter new(filename toString()))
```
unless that causes a `String` leak, of course.

source/sdk/lang/Iterators.ooc
```diff
@@ -4,101 +4,100 @@ import structs/[List, ArrayList] /* for Iterable<T> toArrayList */
...
+		for(elem in this)
+			if(!f(elem)) break
+	}
+	each: func ~withIndex (f: Func (T, Int)) {
+		index := 0
+		for(elem in this) {
+			f(elem, index)
+			index += 1
+		}
+	}
```
davidhesselbom: No reason to change spaces to tabs in the SDK, is there?

test/io/CsvReaderTest.ooc
```diff
@@ -0,0 +1,43 @@
...
+use ooc-collections
+use ooc-unit
+use ooc-io
+
+CsvReaderTest: class extends Fixture {
+	init: func {
+		super("CsvReader")
+		this add("verify entries", func {
+			filename := Text new(c"test/io/input/3x3.csv", 21)
+			reader := CsvReader open(filename)
```
davidhesselbom: And if the file doesn't exist?

test/io/input/strings.csv
```diff
@@ -0,0 +1,2 @@
+"mary had ";"a";" little lamb"
+"hello";" from";" row";" #2"
```
davidhesselbom: Hello missing newline...

thomasfanell: That is the point. EOF instead of \n there.

davidhesselbom: Got it.

davidhesselbom: I'm just pretty sure that someone is going to open this file in their editor, a newline will be added and commited... but I guess tests will catch that.

test/io/CsvReaderTest.ooc
thomasfanell: Then we can expect a segfault, like with every other test in Kean. The input files should never be removed, and if they are, the tests should be updated accordingly.

davidhesselbom: Alright then ...

source/io/CsvReader.ooc
thomasfanell: I thought null is the default value?

source/io/CsvReader.ooc
thomasfanell: Yes, if the file can't be opened for some reason.

test/io/input/strings.csv
thomasfanell: Hmm, well you are right about that. But I don't think that will matter, since the reader accepts both \n or \0 as line "terminators".

source/sdk/lang/Iterators.ooc
thomasfanell: I did it because I didn't want to have mixed styles in the file, by editor settings or by mistake. We're going to add some more stuff to that file later on.

source/io/CsvReader.ooc
thomasfanell: Will fix.

source/io/CsvReader.ooc
davidhesselbom: It seems to depend on the platform...

PR #381, opened 2015-09-17
Added resize() and move() to Vector2D.

tomasmahlberg:


test/collections/Vector2DTest.ooc
```diff
@@ -43,6 +43,73 @@ Vector2DTest: class extends Fixture {
...
+					vector2D[i, j] = i * j
+
+			vector2D resize(20, 20)
+			for (i in 0 .. 10)
+				for (j in 0 .. 10)
+					vector2D move(i, i, i + 1, i + 1)
+			for (i in 0 .. 10)
+				for (j in 0 .. 10)
+					expect(vector2D[i, j], is equal to(0))
+
```
davidhesselbom: It would probably be a good idea to split this test function up into several smaller test functions, if possible. Start out with simple things and get more advanced as you go. You can make `vector2D` a member variable in `Vector2DTest` so you don't have to create a new one for each test - look at `FloatBox2DTest` for an example.

Also, consider putting the numbers 5, 10, 20 and so on in variables so they can easily modified without having to change the code in lots of different places.

source/collections/Vector2D.ooc
```diff
@@ -38,6 +38,52 @@ Vector2D: class <T> {
 	_elementPosition: func (row, column: Int, columnCount := this columnCapacity) -> Int {
 		result := columnCount * row + column
 	}
+	resize: func (newRowCapacity, newColumnCapacity: Int) {
+		temporaryResult: T*
+
+		maximumRowCapacity := this rowCapacity > newRowCapacity ? newRowCapacity : this rowCapacity
+		maximumColumnCapacity := this columnCapacity > newColumnCapacity ? newColumnCapacity : this columnCapacity
```
davidhesselbom: Use `Int maximum` instead.

source/collections/Vector2D.ooc
```diff
@@ -38,6 +38,52 @@ Vector2D: class <T> {
...
+
+		if (newRowCapacity == this rowCapacity && newColumnCapacity == this columnCapacity)
+			return
+
+		temporaryResult = gc_calloc(newRowCapacity * newColumnCapacity, T size)
```
sebastianbaginski: If the `newRowCapacity` is the same as `this rowCapacity ` you should be able to use `realloc` directly on the old memory block, without the need to allocate temporary buffer.

source/collections/Vector2D.ooc
```diff
@@ -38,6 +38,52 @@ Vector2D: class <T> {
...
+
+		for (r in 0 .. maximumRowCapacity)
```
davidhesselbom: Avoid single letter variable names. How about `row`?

source/collections/Vector2D.ooc
tomasmahlberg: True. On the same note, if both newRowCapacity and newColumnCapacity are smaller than rowCapacity and columnCapacity respectively one could use malloc instead of calloc since the zero-initialization won't be needed. Should be slightly faster.

test/collections/Vector2DTest.ooc
```diff
@@ -20,30 +20,107 @@ use ooc-collections
...
-
-			for (i in 0 .. 10)
-				for (j in 0 .. 10)
+		this add("capacity", func {
+			expect(vector2D rowCapacity, is equal to(sizeM))
+			expect(vector2D columnCapacity, is equal to(sizeM))
+		})
+		this add("get/set operators", func {
+			for (i in 0 .. sizeM)
+				for (j in 0 .. sizeM)
```
davidhesselbom: I should have told you this before, but, if you want to iterate over all elements of `vector2D`, why not use `vector2D rowCapacity` and `vector2D columnCapacity` instead of `sizeM` as the loop bound?

source/collections/Vector2D.ooc
```diff
@@ -38,6 +40,58 @@ Vector2D: class <T> {
 	_elementPosition: func (row, column: Int, columnCount := this columnCapacity) -> Int {
```
sebastianbaginski: I know that you have probably discussed this with @simonmika but to me this naming seems very confusing.
Let me explain. `columnCount` sounds like a *number of the columns*, while `columnCapacity` sounds like a *capacity of a single column* or *size of a single column*, which is in fact exactly the same as `rowCount` (*number of rows*)...
I suggest using *count* instead of *capacity* in this class, but this is only my opinion.
@davidhesselbom , @simonmika what do you think about it ?

source/collections/Vector2D.ooc
```diff
@@ -38,6 +40,58 @@ Vector2D: class <T> {
...
 		result := columnCount * row + column
 	}
+	resize: func (newRowCapacity, newColumnCapacity: Int) {
+		if (newColumnCapacity == this columnCapacity)
+			if (newRowCapacity == this rowCapacity)
+				return
+			else
+				_allocate(newRowCapacity, newColumnCapacity)
```
sebastianbaginski: Again I'm confused by the names. Or maybe I just don't understand the basics - is the data layout in this class *row-major* or *column-major* ?

davidhesselbom: Also, it should be `this _allocate`.

source/collections/Vector2D.ooc
```diff
@@ -38,6 +40,58 @@ Vector2D: class <T> {
...
+				temporaryResult = gc_calloc(newRowCapacity * newColumnCapacity, T size)
+			else
+				temporaryResult = gc_malloc(newRowCapacity * newColumnCapacity * T size)
+
+			for (row in 0 .. minimumRowCapacity)
+				memcpy(temporaryResult[T size * this _elementPosition(row, 0, newRowCapacity)]&,
+					this _backend[T size * this _elementPosition(row, 0)]&, minimumColumnCapacity * T size)
+
+			this free()
+			this init(temporaryResult, newRowCapacity, newColumnCapacity)
```
davidhesselbom: This certainly looks interesting...

source/collections/Vector2D.ooc
davidhesselbom: Well, it depends on whether the count is the same as the capacity. In a `VectorList`, they are not the same - you can have capacity for elements you don't have. In this case, it looks like they are the same, in which case I agree `count` would be much less ambiguous.

tomasmahlberg: The confusion is justified. `rowCapacity` here was referring to the capacity of a single row, i.e. the column count. The naming was inherited from the `Vector` class. I will switch to using the term count instead and make changes in the code accordingly.

davidhesselbom: It's still going to be confusing, if `Vector` has capacity and `Vector2D` has count, and I'm not sure what the correct terminology would be in `Vector` then. Still, yes, go ahead and change to `count` here and leave `Vector` as is for now.

source/collections/Vector2D.ooc
tomasmahlberg: Interesting as in?

davidhesselbom: I don't recall ever seeing anything free itself and then run its own `init` function afterwards, and I hadn't even considered it a possibility. It's not a problem at all!

test/collections/Vector2DTest.ooc
```diff
@@ -20,30 +20,107 @@ use ooc-collections
...
-			vector2D free()
+			for (i in 0 .. vector2D rowCapacity)
+				for (j in 0 .. vector2D columnCapacity)
+					vector2D[i, j] = sizeM
+			for (i in 0 .. vector2D rowCapacity)
+				for (j in 0 .. vector2D columnCapacity)
+					expect(vector2D[i, j], is equal to(sizeM))
+		})
+		this add("resize increase", func {
+			// Increase array size to 20x20
```
davidhesselbom: It wouldn't be a bad idea to make the test function name more descriptive instead of adding comments to elaborate. When a test fails, the name is printed, but your comments won't be.

source/collections/Vector2D.ooc
```diff
@@ -35,19 +36,74 @@ Vector2D: class <T> {
 	free: override func {
 		gc_free(this _backend)
 	}
-	_elementPosition: func (row, column: Int, columnCount := this columnCapacity) -> Int {
+	_elementPosition: func (row, column: Int, columnCount := this columnCount) -> Int {
 		result := columnCount * row + column
```
davidhesselbom: This doesn't return anything...

tomasmahlberg: It seems to be working fine:
```
lang_Numbers__Int Vector2D__Vector2D__elementPosition(Vector2D__Vector2D* this, lang_Numbers__Int row, lang_Numbers__Int column, lang_Numbers__Int columnCount) {
    
    {
        lang_Numbers__Int result = columnCount * row + column;
        return result;
    }
}
```
But I will change the line to just `columnCount * row + column`since the assignment to a vairable seems to be redundant?

thomasfanell: Weird. I got a compile error, and yes, the variable is redundant.

davidhesselbom: Maybe, sometimes, rock returns the last created variable...? Anyway, yeah, the `result` variable is unnecessary.

source/collections/Vector2D.ooc
```diff
@@ -14,40 +14,94 @@
...
+		if (targetColumnStart + columnCount > this columnCount)
+			columnCount = this columnCount - targetColumnStart
+
+		for (row in 0 .. rowCount) {
+			if (sourceRowStart > targetRowStart) {
+				// Move row-wise from top to bottom.
+				sourceRowIndex = sourceRowStart + row
+				targetRowIndex = sourceRowIndex - sourceRowStart + targetRowStart
+			}
+			else {
```
davidhesselbom: `} else {`

source/collections/Vector2D.ooc
```diff
@@ -14,40 +14,94 @@
...
+		super()
 	}
-	_elementPosition: func (row, column: Int, columnCount := this columnCapacity) -> Int {
+	_elementPosition: func (row, column: Int, columnCount := this columnCount) -> Int {
 		columnCount * row + column
 	}
+	resize: func (newRowCount, newColumnCount: Int) {
+		if (newColumnCount == this columnCount) {
+			if (newRowCount == this rowCount)
+				return
```
davidhesselbom: Don't return from more than one place - replace this by negating the `if` clause above.

source/collections/Vector2D.ooc
```diff
@@ -14,40 +14,94 @@
...
+			this _allocate(newRowCount, newColumnCount)
+			this _rowCount = newRowCount
+			this _columnCount = newColumnCount
+		}
+		else {
```
davidhesselbom: `} else {`

source/collections/Vector2D.ooc
```diff
@@ -14,40 +14,92 @@
...
 		columnCount * row + column
 	}
+	resize: func (newRowCount, newColumnCount: Int) {
+		if (newColumnCount == this columnCount) {
+			if (newRowCount != this rowCount) {
+				this _allocate(newRowCount, newColumnCount)
+				this _rowCount = newRowCount
+				this _columnCount = newColumnCount
+			}
+		} else {
```
davidhesselbom: Actually, the brackets around the outer `if` are not necessary.

tomasmahlberg: Hmm... without them, the compiler will think that the else belongs to the inner if statement, no?

davidhesselbom: Yes, you're right. So why not
```ooc
if (newColumnCount == this columnCount && newRowCount != this rowCount) {
```
?

tomasmahlberg: That's logically not the same though, is it? This would cause us to jump into the else statement if both `newColumnCount == this columnCount` and `newRowCount == this rowCount` holds true. In this case, we don't want any code to execute.

davidhesselbom: Well, this is embarrassing :-/ You're right, again. The only way to "simplify" this seems to be
```ooc
if (newColumnCount != this columnCount) {
	[ do the complicated stuff ]
else if (newRowCount != this rowCount) {
	this _allocate(newRowCount, newColumnCount)
	this _rowCount = newRowCount
	this _columnCount = newColumnCount
}
```
but this doesn't exactly make anything clearer.

This was my last note on this pull request, so, merging.

PR #387, opened 2015-09-18
Added parentheses around conditional assignment

thomasfanell:
Rock (and a few other) compilers seems to prefer it this way.

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -104,7 +104,7 @@ Text: cover {
 	}
 	free: func@ -> Bool {
 		result: Bool
-		if (result = (this _owner != Owner Literal && this _owner != Owner Stack && this _buffer != null)) {
+		if ((result = (this _owner != Owner Literal && this _owner != Owner Stack && this _buffer != null))) {
 			gc_free(this _buffer)
 			this _buffer = null
 			this _count = 0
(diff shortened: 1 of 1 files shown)


davidhesselbom:
Seems unnecessary. What gave you the idea that rock prefers it this way?

davidpiuva:
I noticed that too for if statements. A variable is fine but an operand will require ( ) like in C.

davidhesselbom:
How annoying. Did it generate a warning before, or what?

thomasfanell:
Yes, rock provided a hint about it, right before it crashed and burned.

davidpiuva:
I think the parser had no idea of what I tried to write.

davidhesselbom:
Hmm, but the test works, doesn't it?

sebastianbaginski:
> Yes, rock provided a hint about it, right before it crashed and burned.

How can I reproduce this? I'd love to see it :)

thomasfanell:
@sebastianbaginski it's a warning, and is apparently provided when rock crashes when performing something related to the function in question. I don't remember exactly what I did to cause it, but if it happens again, i'll ping you about it. Below is the warning.
```
In function ‘Text__Text_free’: ..ooc-kean/source/base/Text.ooc:107:5: warning: suggest parentheses around assignment used as truth value [-Wparentheses] if (result = (this _owner != Owner Literal && this _owner != Owner Stack && this _buffer != null))
```

PR #389, opened 2015-09-18
added fixture support for Text, added corresponding fixture test

thomasfanell:


source/unit/Fixture.ooc
```diff
@@ -15,6 +15,7 @@
  * along with this software. If not, see<http://www.gnu.org/licenses/>.
  */
 
+use ooc-base
 import ./Constraints
```
davidhesselbom: Is `./` necessary?

davidhesselbom:
Neat!

source/unit/Fixture.ooc
thomasfanell: Probably not, hold on, i'll find out.

PR #391, opened 2015-09-18
Window cleanup

emilwestergren:
Separated Window classes from OpenGL and could clean some code because of it.


ooc-x11.use
```diff
@@ -0,0 +1,5 @@
+Name: ooc UI
```
davidhesselbom: You probably shouldn't use the same name as in `ooc-ui.use`...

ooc-x11.use
```diff
@@ -0,0 +1,5 @@
...
+Description: User interface stuff
+SourcePath: source/ui/X11
```
davidhesselbom: Shouldn't the folder be named with a lower-case `x`?

emilwestergren:
Updated.

emilwestergren:
Updated.

PR #394, opened 2015-09-21
Added simple 'resizeTo' implementation to RasterMonochrome

sebastianbaginski:


test/draw/RasterMonochromeTest.ooc
```diff
@@ -57,6 +57,17 @@ RasterMonochromeTest: class extends Fixture {
...
 			}
 		})
+		this add("resize", func {
+			size := IntSize2D new(13, 5)
+			image := RasterMonochrome open(this sourceFlower)
+			image2 := image resizeTo(image size / 2)
+			expect(image2 size == image size / 2)
+			image2 = image2 resizeTo(image size)
+			expect(image2 size == image size)
+			expect(image distance(image2) < image size width / 10)
```
davidhesselbom: How was `10` chosen?

sebastianbaginski: I just thought that the difference threshold of 10% of the image width is *good enough* to consider the images similar.

davidhesselbom: Okay. I don't know what a good threshold is for this comparison, so I'll take your word for it!

PR #396, opened 2015-09-21
Fix memory leaks in mathematical morphology methods in BoolVectorList.

bettinaselig:


davidhesselbom:
I don't think there are any tests for these functions. Do you think you could write some in a later pull request?

Issue #397, opened 2015-09-21
ooc-unit: Print expected and actual value when tests fail


davidhesselbom:
```ooc
			expect(rasterFromGpu distance(correctImage), is equal to(0.0f))
```
If `rasterFromGpu distance(correctImage)` > 0, we have no idea what the actual value is or what a suitable threshold for a `within` might be. 

Modify `ooc-unit` so that when an `expect` fails, it prints both the expected value and the actual value, not just the name of the test, e.g.
```
GpuImageTranslationTest ..f failed
  -> 'GPU translate Z (BGRA), expected 0.0, was 0.7543456'
```

davidhesselbom:
Partially fixed by #405.

marcusnaslund:
What more remains?

davidhesselbom:
I think this has been fixed by now, yes.

davidhesselbom:
Nope &mdash; it doesn't work for `String`, and I think it should.

marcusnaslund:
Why? Don't we want to replace all of that with `Text`?

davidhesselbom:
Mmm. A lot of work, by comparison, but I suppose  it *is* a better long-term solution.

PR #400, opened 2015-09-21
Print actual value when expects fail.

davidhesselbom:
This is an attempt to get somewhere with #397. Only prints the actual value, not the expected one, and doesn't do anything for `Object`.


davidhesselbom:
Replaced by #405.

PR #404, opened 2015-09-21
Implemented cancel method for threads

sebastianbaginski:
Implemented missing functions to cancel a running thread

### DIFF
--- a/test/sdk/ThreadTest.ooc
+++ b/test/sdk/ThreadTest.ooc
@@ -1,0 +1,57 @@
+use ooc-unit
+use ooc-base
+import threading/Thread
+import threading/native/ConditionUnix
+import os/Time
+
+ThreadTest: class extends Fixture {
+	init: func {
+		super("Thread")
+		this add("starting thread", This _testStartingThread)
+		this add("canceling thread", This _testCancelation)
+	}
+	_testStartingThread: static func {
+		threadStarted := Cell<Int> new(0)
+		job := func {
+			threadStarted set(1)
+		}
+		thread := Thread new(|| job())
+		expect(threadStarted get() == 0)
+		expect(thread start())
+		expect(thread wait())
+		expect(threadStarted get() == 1)
+		thread free()
+	}
+	_testCancelation: static func {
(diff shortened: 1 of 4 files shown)


sebastianbaginski:
Not ready yet, I want to revisit the code and ugly formatting bugs tomorrow.

sebastianbaginski:
It seems to work, but I don't know if every platform enables cancelling a thread by default. Maybe we should add http://man7.org/linux/man-pages/man3/pthread_setcanceltype.3.html call before starting the actual thread job.
I guess this could be handled in a separate pull request.

test/sdk/ThreadTest.ooc
```diff
@@ -0,0 +1,77 @@
+use ooc-unit
+use ooc-base
+import threading/Thread
+import threading/native/ConditionUnix
+import os/Time
+
+IntObject: class {
```
sebastianbaginski: Closures didn't work with integers, it seems like a separate local copy was created instead of modifying the original integer, so I created this helper class.

test/sdk/ThreadTest.ooc
```diff
@@ -0,0 +1,77 @@
...
+		expect(thread start())
+		expect(thread wait())
+		expect(threadStarted == 1)
+		thread free()
+	}
+	_testCancelation: static func {
+		expectedValue := 1_000
+		value := IntObject new(0)
+		mutex := Mutex new()
+		startedCondition := ConditionUnix new()
```
sebastianbaginski: https://github.com/cogneco/ooc-kean/issues/407

test/sdk/ThreadTest.ooc
davidhesselbom: Is that it? We've had lots of problems with Closures, but I never realized it had to do with covers... Might explain a few things!

davidhesselbom: Would it be possible to use a `Cell<Int>` instead?

sebastianbaginski: It works, thanks.

PR #405, opened 2015-09-21
Fixture prints tested and expected values on failed tests

thomasfanell:


davidhesselbom:
The style fix has already been fixed...

thomasfanell:
Should be safe to merge then...?

davidhesselbom:
Yes. I should make the test merge with master first and run magic on that instead.

PR #411, opened 2015-09-22
Added LessThan/GreaterThan constraints

thomasfanell:
Failed unit tests now produce friendlier output when the values are numeric types


test/unit/DoubleTest.ooc
```diff
@@ -35,6 +35,8 @@ DoubleTest: class extends Fixture {
 		isNotEqualTo42 := is not equal to(4.2)
 		this add("0.0 is not equal to 4.2", func { expect(0.0, isNotEqualTo42) })
 		this add("0.0 is equal to -0.0", func { expect(0.0, is equal to(-0.0)) })
+		this add("0.0 is less than 0.00000001", func { expect(0.0, is less than(0.00000001)) })
+		this add("0.0 is greater than -0.00000001", func { expect(0.0f, is greater than(-0.00000001f)) })
```
davidhesselbom: There's an `f` on the expected value...

thomasfanell: yep I caught that, fixed :-)

PR #413, opened 2015-09-23
Include guard around ConditionUnix to prevent compilation on windows

sebastianbaginski:


sebastianbaginski:
We ran into this while testing kean on windows with @jonathanudd 
Will be needed in https://github.com/cogneco/ooc-kean/issues/407

PR #414, opened 2015-09-23
updated Fixture to flush output buffer for interactive display

sebastianbaginski:
I think it looks nice for tests which takes some time to run, like gpu, image or random generator tests. The *dots* will be displayed one after another, instead of having the string `Class ..... done` displayed at once after the test is done.
@davidhesselbom @thomasfanell


thomasfanell:
Nice :+1:

davidhesselbom:
Sweet! For most tests, all it does it show how long it takes to build the test compared to how long it takes to run it, but it really helps in the other cases. Thanks!

PR #418, opened 2015-09-23
Added some basic operations to FloatMatrix.

tomasmahlberg:
Going to need these in a bit.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -94,6 +94,19 @@ FloatMatrix : cover {
 	}
 
 	// <summary>
+	// Calculates the trace of a square matrix.
+	// </summary>
+	// <returns>The trace of the matrix.</returns>
+	trace: func -> Float {
+		if (!this isSquare)
+			InvalidDimensionsException new() throw()
+		result := 0.0f
+		for (i in 0 .. this dimensions height)
+			result += this get(i, i)
+		result
+	}
+
+	// <summary>
 	// Swaps the position of two rows
 	// </summary>
 	// <param name="row1">First row</param>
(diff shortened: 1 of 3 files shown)


source/math/FloatMatrix.ooc
```diff
@@ -288,6 +301,58 @@ operator * (left: FloatMatrix, right: FloatMatrix) -> FloatMatrix {
 result
 }
 
+// <summary>
+// Multiplication of scalar and matrix.
+// </summary>
+// <param name="left">Left matrix in the subtraction.</param>
+// <param name="right">Right matrix in the subtraction.</param>
+// <returns>Difference of left and right matrices.</returns>
```
sebastianbaginski: :)

davidhesselbom:
This isn't going to stop this PR, but, do we need both `Vector2D` and `FloatMatrix`?

source/math/FloatMatrix.ooc
tomasmahlberg: Haha, I think these function descriptions are being thrown away but I shall update it anyway.

tomasmahlberg:
I don't know, but `FloatMatrix` has many specific functions for the float type like the solver for linear systems. How would these be implemented in the generic `Vector2D`?

davidhesselbom:
Not sure. In `ooc-math` we have things like `FloatVectorList` that inherit from `VectorList` and contain specific functions, so `FloatMatrix` *could* inherit from `Vector2D`. I'm not suggesting you make the change now, though.

source/math/FloatMatrix.ooc
```diff
@@ -288,6 +301,58 @@ operator * (left: FloatMatrix, right: FloatMatrix) -> FloatMatrix {
...
 }
 
+// <summary>
+// Addition of matrices.
+// </summary>
+// <param name="left">Left matrix in the addition.</param>
+// <param name="right">Right matrix in the addition.</param>
+// <returns>Sum of left and right matrices.</returns>
+operator + (left: FloatMatrix, right: FloatMatrix) -> FloatMatrix {
+	if (left dimensions width != right dimensions width || left dimensions height != right dimensions height)
```
davidhesselbom: `if (left dimensions != right dimensions)`

source/math/FloatMatrix.ooc
```diff
@@ -94,6 +94,19 @@ FloatMatrix : cover {
 	}
 
 	// <summary>
+	// Calculates the trace of a square matrix.
+	// </summary>
+	// <returns>The trace of the matrix.</returns>
+	trace: func -> Float {
+		if (!this isSquare)
+			InvalidDimensionsException new() throw()
```
davidhesselbom: It would be better to get rid of `InvalidDimensionsException` and use `raise("Invalid dimensions in FloatMatrix trace: not square!")` or something.

source/math/FloatMatrix.ooc
```diff
@@ -288,6 +301,58 @@ operator * (left: FloatMatrix, right: FloatMatrix) -> FloatMatrix {
 result
 }
 
+// <summary>
+// Addition of matrices.
+// </summary>
+// <param name="left">Left matrix in the addition.</param>
+// <param name="right">Right matrix in the addition.</param>
+// <returns>Sum of left and right matrices.</returns>
+operator + (left: FloatMatrix, right: FloatMatrix) -> FloatMatrix {
```
davidhesselbom: That's odd, magic should have caught this...

`operator - (left, right: FloatMatrix) -> FloatMatrix {`

source/math/FloatMatrix.ooc
```diff
@@ -288,6 +301,58 @@ operator * (left: FloatMatrix, right: FloatMatrix) -> FloatMatrix {
...
+result
+}
+
+// <summary>
+// Multiplication of scalar and matrix.
+// </summary>
+// <param name="left">Scalar in the multiplication.</param>
+// <param name="right">Matrix in the multiplication.</param>
+// <returns>Product of the sclar and matrix.</returns>
+operator * (left: Float, right: FloatMatrix) -> FloatMatrix {
```
davidhesselbom: All of the operators where the first parameter is a `FloatMatrix` should be moved into the cover. This here operator is the only one I see that needs to be a stand-alone operator (since it takes a `Float` as the first argument).

tomasmahlberg:
Updated.

source/math/FloatMatrix.ooc
```diff
@@ -264,28 +277,49 @@ FloatMatrix : cover {
...
+	result
+	}
+
+	operator + (other: This) -> This {
+		if (this dimensions != other dimensions)
+			raise("Invalid dimensions in FloatMatrix + operator: dimensions must match!")
+		result := This new (this dimensions width, this dimensions height)
+		for (x in 0 .. this dimensions width)
+			for (y in 0 .. this dimensions height)
+				result elements[x + result dimensions width * y] = this elements[x + this dimensions width * y] + other elements[x + other dimensions width * y]
```
davidhesselbom: You've already established that the two are the same dimension, which means that `x + result dimensions width * y == x + this dimensions width * y` so why not put `x + this dimensions width * y` in a variable named, say, `index` and use that as the index all 3 times?

source/math/FloatMatrix.ooc
```diff
@@ -264,28 +277,49 @@ FloatMatrix : cover {
...
 			}
-			result elements[x + result dimensions width * y] = temp
 		}
+	result
+	}
+
+	operator + (other: This) -> This {
+		if (this dimensions != other dimensions)
+			raise("Invalid dimensions in FloatMatrix + operator: dimensions must match!")
+		result := This new (this dimensions width, this dimensions height)
```
davidhesselbom: I was going to suggest
```ooc
result := This new(this dimensions)
```
but for some reason, that constructor doesn't allocate the array :fearful: That really should be fixed. I'll send you a pull request.

source/math/FloatMatrix.ooc
davidhesselbom: Actually I'll fix this in my pull request, too.

davidhesselbom:
Waiting for my own pull request to be merged into this one.

PR #421, opened 2015-09-23
Ignore test/io output folder

thomasfanell:


test/io/CsvWriterTest.ooc
```diff
@@ -30,7 +33,21 @@ CsvWriterTest: class extends Fixture {
 			}
 			reader free()
 			outputFilename free()
+			// This needs to be moved if any more tests are implemented
+			this _removeOutputDirectory()
```
davidhesselbom: Why do we need to remove the output folder?

thomasfanell: Instead of having to remove each file after each test, we wipe the entire directory. You just have to make sure that ```this removeOutputDirectory``` is the last instruction in the last test. There won't be that many tests for this particular class anyway.

davidhesselbom: But why remove the output files at all?

thomasfanell: And there is a likelyhood that someone will forget to remove their test file.

thomasfanell: Because they are temporary files, so why not?

davidhesselbom: In `.gitignore`:
```
# Test output
/test/draw/output
```
Just add  `/test/io/output`, or maybe use a wildcard: `/test/*/output` to ignore all output files in all tests.

thomasfanell: Well ok, that does sound better.

PR #423, opened 2015-09-23
Integration of Promises with Threadpool

marcusnaslund:
Big changes, but ThreadPool add(...) behaves as before - nothing in vidproc has any objections.

### DIFF
--- a/source/base/ThreadPool.ooc
+++ b/source/base/ThreadPool.ooc
@@ -1,111 +1,203 @@
 use ooc-collections
+use ooc-base
 
 import structs/LinkedList
 import threading/Thread
+import threading/native/ConditionUnix
+import os/Time
 
-ThreadJob: class {
-	_body: Func
-	init: func (=_body)
-	_execute: func { this _body() }
-	run: virtual func {
-		this _execute()
-		this free()
+_TaskState: enum {
+	Unfinished
+	Finished
+	Cancelled
+}
+
+_Task: class {
+	_mutex: Mutex
+	_action: Func
+	_state : _TaskState
+	_freeDirectly: Bool
+	init: func (=_action, =_mutex) {
+		_state = _TaskState Unfinished
+		_mutex = Mutex new()
+	}
+	free: override func {
+		if (this _state != _TaskState Unfinished)
+			this _freeDirectly = true
+		else
+			super()
+	}
+	run: virtual func (=_mutex) {
+		this _action()
+		_mutex lock()
+		if (this _state != _TaskState Cancelled)
+			this _state = _TaskState Finished
+		_mutex unlock()
+		if (this _freeDirectly)
+			this free()
 	}
 }
 
-SynchronizedThreadJob: class extends ThreadJob {
-	_finishedCondition := WaitCondition new()
-	_mutex := Mutex new()
-	_finished := false
-	finished ::= this _finished
-	_freeOnCompletion := false
-	init: func (body: Func) { super(body) }
-	free: override func {
-		this _mutex lock()
-		if (!this _finished) {
-			this _freeOnCompletion = true
-			this _mutex unlock()
-		} else {
-			this _mutex unlock()
-			this _mutex destroy()
-			this _finishedCondition free()
+_ResultTask: class <T> extends _Task {
+	_result: Cell<T>
+	_resultAction: Func -> T
+	init: func (=_resultAction, =_mutex) {
+		this _state = _TaskState Unfinished
+	}
+	free: func {
+		if (this _state != _TaskState Unfinished)
+			this _freeDirectly = true
+		else
 			super()
-		}
 	}
-	run: override func {
-		this _execute()
+	run: override func (=_mutex) {
+		temporary := Cell<T> new(this _resultAction())
 		this _mutex lock()
-		this _finished = true
-		if (this _freeOnCompletion) {
-			this _mutex unlock()
+		this _result = temporary
+		if (this _state != _TaskState Cancelled)
+			this _state = _TaskState Finished
+		if (this _freeDirectly)
 			this free()
-		} else {
-			this _mutex unlock()
-			this _finishedCondition broadcast()
-		}
-	}
-	wait: func {
-		this _mutex lock()
-		if (!this _finished)
-			this _finishedCondition wait(this _mutex)
 		this _mutex unlock()
+	}
+}
+
+_TaskPromise: class extends Promise {
+	_task: _Task
+	init: func (=_task)
+	free: override func {
+		this _task free()
+		super()
+	}
+	wait: override func -> Bool {
+		status := false
+		while (true) {
+			this _task _mutex lock()
+			if (this _task _state != _TaskState Unfinished) {
+				status = (this _task _state == _TaskState Finished)
+				this _task _mutex unlock()
+				break
+			}
+			this _task _mutex unlock()
+			Time sleepMilli(10)
+		}
+		status
+	}
+	cancel: override func -> Bool {
+		status := false
+		this _task _mutex lock()
+		if (this _task _state == _TaskState Unfinished) {
+			this _task _state = _TaskState Cancelled
+			status = true
+		}
+		this _task _mutex unlock()
+		status
+	}
+}
+
+_ResultTaskPromise: class <T> extends ResultPromise<T> {
+	_resultTask: _ResultTask<T>
+	init: func (=_resultTask)
+	free: override func {
+		this _resultTask free()
+		super()
+	}
+	wait: func -> Bool {
+		status := false
+		while (true) {
+			this _resultTask _mutex lock()
+			if (this _resultTask _state != _TaskState Unfinished) {
+				status = (this _resultTask _state == _TaskState Finished)
+				this _resultTask _mutex unlock()
+				break
+			}
+			this _resultTask _mutex unlock()
+			Time sleepMilli(100)
+		}
+		status
+	}
+	wait: func ~default (defaultValue: T) -> T {
+		status := this wait()
+		status ? this _resultTask _result[T] : defaultValue
+	}
+	cancel: override func -> Bool {
+		status := false
+		this _resultTask _mutex lock()
+		if (this _resultTask _state == _TaskState Unfinished) {
+			this _resultTask _state = _TaskState Cancelled
+			status = true
+		}
+		this _resultTask _mutex unlock()
+		status
 	}
 }
 
 ThreadPool: class {
+	_globalmutex := Mutex new()
 	_threads: Thread[]
-	_jobs := BlockedQueue<ThreadJob> new()
-	_mutex := Mutex new()
+	_threadMutexes: Mutex[]
+	_tasks := BlockedQueue<_Task> new()
 	_allFinishedCondition := WaitCondition new()
 	_activeJobs: Int
 	_threadCount: Int
 	threadCount ::= this _threadCount
-
 	init: func (threadCount := 4) {
 		this _threadCount = threadCount
 		this _threads = Thread[threadCount] new()
+		this _threadMutexes = Mutex[threadCount] new()
 		for (i in 0 .. threadCount) {
-			this _threads[i] = Thread new(|| threadLoop())
+			this _threadMutexes[i] = Mutex new()
+			this _threads[i] = Thread new(|| threadLoop(i))
 			this _threads[i] start()
 		}
 	}
 	free: override func {
-		for (i in 0 .. this _threadCount)
+		for (i in 0 .. this _threadCount) {
 			this _threads[i] free()
+			this _threadMutexes[i] destroy()
+		}			
 		gc_free(this _threads data)
-		this _jobs free()
+		this _tasks free()
 		this _allFinishedCondition free()
-		this _mutex destroy()
+		this _globalmutex destroy()
 		super()
 	}
-	threadLoop: func {
+	threadLoop: func (threadId: Int) {
 		while (true) {
-			job := this _jobs wait()
-			job run()
-			this _mutex lock()
+			job := this _tasks wait()
+			job run(this _threadMutexes[threadId])
+			this _globalmutex lock()
 			this _activeJobs -= 1
-			if (this _activeJobs == 0)
+			if (this _activeJobs == 0) {
 				this _allFinishedCondition broadcast()
-			this _mutex unlock()
+			}
+			this _globalmutex unlock()
 		}
 	}
-	_add: func (job: ThreadJob) {
-		this _mutex lock()
+	_add: func (task: _Task) -> Void {
+		this _globalmutex lock()
 		this _activeJobs += 1
-		this _mutex unlock()
-		this _jobs enqueue(job)
+		this _globalmutex unlock()
+		this _tasks enqueue(task)
 	}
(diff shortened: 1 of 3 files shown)


davidhesselbom:
I messed up and merged this even though it broke the build, hence #428. Apparently it needs some more reviewing. @emilwestergren and @simonmika, can you take a look at this?

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
 use ooc-collections
+use ooc-base
 
 import structs/LinkedList
```
emilwestergren: Not used.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
 
-SynchronizedThreadJob: class extends ThreadJob {
-	_finishedCondition := WaitCondition new()
-	_mutex := Mutex new()
-	_finished := false
-	finished ::= this _finished
-	_freeOnCompletion := false
-	init: func (body: Func) { super(body) }
+_Task: class {
+	_mutex: Mutex
```
emilwestergren: `_mutex := Mutex new()`

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+	_action: Func
+	_state : _TaskState
```
emilwestergren: _state := _TaskState Unfinished

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
 	free: override func {
-		this _mutex lock()
-		if (!this _finished) {
-			this _freeOnCompletion = true
-			this _mutex unlock()
-		} else {
-			this _mutex unlock()
-			this _mutex destroy()
-			this _finishedCondition free()
+		if (this _state != _TaskState Unfinished)
```
davidhesselbom: What happened to `this _mutex destroy()`?

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-	_freeOnCompletion := false
-	init: func (body: Func) { super(body) }
+_Task: class {
+	_mutex: Mutex
+	_action: Func
+	_state : _TaskState
+	_freeDirectly: Bool
+	init: func (=_action, =_mutex) {
+		_state = _TaskState Unfinished
+		_mutex = Mutex new()
```
davidhesselbom: `this _mutex`

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-	finished ::= this _finished
-	_freeOnCompletion := false
-	init: func (body: Func) { super(body) }
+_Task: class {
+	_mutex: Mutex
+	_action: Func
+	_state : _TaskState
+	_freeDirectly: Bool
+	init: func (=_action, =_mutex) {
+		_state = _TaskState Unfinished
```
davidhesselbom: `this _state`

source/base/ThreadPool.ooc
davidhesselbom: Both of these calls can be moved up:
```ooc
    _mutex: Mutex new()
    _state: _TaskState Unfinished
```

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-SynchronizedThreadJob: class extends ThreadJob {
-	_finishedCondition := WaitCondition new()
-	_mutex := Mutex new()
-	_finished := false
-	finished ::= this _finished
-	_freeOnCompletion := false
-	init: func (body: Func) { super(body) }
+_Task: class {
+	_mutex: Mutex
+	_action: Func
```
emilwestergren: This should not be in this class. My suggestion is to create an abstract class without any function pointers, and then you inherit two classes from it: _ActionTask and _ResultTask where the different function pointers are located.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+	_state : _TaskState
+	_freeDirectly: Bool
+	init: func (=_action, =_mutex) {
```
emilwestergren: With the other changes this can be simplified to `init: func`

source/base/ThreadPool.ooc
davidhesselbom: Wait, something's wrong here. 
```ooc
init: func (=_action, =_mutex) {
```
will assign `this _mutex` to the variable `_mutex` used as an argument in the `_Task new` call. 2 lines later, you reassign `_mutex` to `Mutex new()`. Why?

source/base/ThreadPool.ooc
emilwestergren: Shouldn't this be `this _state != _TaskState Finished`?

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-			this _freeOnCompletion = true
-			this _mutex unlock()
-		} else {
-			this _mutex unlock()
-			this _mutex destroy()
-			this _finishedCondition free()
+		if (this _state != _TaskState Unfinished)
+			this _freeDirectly = true
+		else
 			super()
```
emilwestergren: You need to free the Mutex before you call super.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-		}
 	}
-	run: override func {
-		this _execute()
-		this _mutex lock()
-		this _finished = true
-		if (this _freeOnCompletion) {
-			this _mutex unlock()
+	run: virtual func (=_mutex) {
```
davidhesselbom: This again reassigns `this _mutex` without destroying the current one. Why?

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-	_mutex := Mutex new()
-	_finished := false
-	finished ::= this _finished
-	_freeOnCompletion := false
-	init: func (body: Func) { super(body) }
+_Task: class {
+	_mutex: Mutex
+	_action: Func
+	_state : _TaskState
+	_freeDirectly: Bool
```
emilwestergren: `_freeDirectly := false`

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+	init: func (=_action, =_mutex) {
+		_state = _TaskState Unfinished
+		_mutex = Mutex new()
+	}
 	free: override func {
```
emilwestergren: This function needs to be thread-safe, use the mutex.

source/base/ThreadPool.ooc
emilwestergren: Change this to `run: abstract func` and implement in concrete classes.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
 			this _threads[i] start()
 		}
 	}
 	free: override func {
-		for (i in 0 .. this _threadCount)
+		for (i in 0 .. this _threadCount) {
 			this _threads[i] free()
+			this _threadMutexes[i] destroy()
+		}
 		gc_free(this _threads data)
```
davidhesselbom: I think `this _threads free()` should work, too.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
 	}
-	wait: func {
+}
+
+_ResultTask: class <T> extends _Task {
+	_result: Cell<T>
+	_resultAction: Func -> T
+	init: func (=_resultAction, =_mutex) {
+		this _state = _TaskState Unfinished
+	}
```
emilwestergren: With the other changes, this can be simplified to `init: func(=_resultAction) { super() }`

source/base/ThreadPool.ooc
davidhesselbom: I don't see `_this threadMutexes` getting freed anywhere...

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+	free: func {
```
emilwestergren: This function is not needed since it is already implemented in the base class.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+		if (this _state != _TaskState Unfinished)
+			this _freeDirectly = true
+		else
+			super()
+	}
+	run: override func (=_mutex) {
```
emilwestergren: this should be `run: override func {`

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-			this _mutex lock()
+			job := this _tasks wait()
+			job run(this _threadMutexes[threadId])
+			this _globalmutex lock()
 			this _activeJobs -= 1
-			if (this _activeJobs == 0)
+			if (this _activeJobs == 0) {
 				this _allFinishedCondition broadcast()
-			this _mutex unlock()
+			}
```
davidhesselbom: These brackets are unnecessary.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
 }
 
+_TaskPromise: class extends Promise {
+	_task: _Task
+	init: func (=_task)
+	free: override func {
+		this _task free()
+		super()
+	}
+	wait: override func -> Bool {
```
emilwestergren: This is not a very efficient way of synchronizing. Use Conditions like in the earlier ThreadPool implementation.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+	_add: func (task: _Task) -> Void {
+		this _globalmutex lock()
 		this _activeJobs += 1
-		this _mutex unlock()
-		this _jobs enqueue(job)
+		this _globalmutex unlock()
+		this _tasks enqueue(task)
+	}
+	add: func (action: Func) {
+		task := _Task new(action, this _globalmutex)
```
davidhesselbom: Okay, so... tasks are given the `globalmutex`...? So that's why they're never freed in `_Task`.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+			this _resultTask _state = _TaskState Cancelled
+			status = true
+		}
+		this _resultTask _mutex unlock()
+		status
+	}
+}
+
 ThreadPool: class {
+	_globalmutex := Mutex new()
```
davidhesselbom: `_globalMutex`

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
+				status = (this _task _state == _TaskState Finished)
+				this _task _mutex unlock()
+				break
+			}
+			this _task _mutex unlock()
+			Time sleepMilli(10)
+		}
+		status
+	}
+	cancel: override func -> Bool {
```
emilwestergren: This currently has no effect on the Task, is this something for the future?

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
 	}
-	run: override func {
-		this _execute()
-		this _mutex lock()
-		this _finished = true
-		if (this _freeOnCompletion) {
-			this _mutex unlock()
+	run: virtual func (=_mutex) {
+		this _action()
+		_mutex lock()
```
davidhesselbom: `this _mutex`

source/base/ThreadPool.ooc
davidhesselbom: Okay, I'm starting to see what's going on here. Still, if you create a new `Mutex` here in the constructor and don't destroy it later, it will leak.

source/base/ThreadPool.ooc
emilwestergren: Why not have private mutexes for the tasks?

source/base/ThreadPool.ooc
marcusnaslund: This was very unfortunate. There should not be a new mutex created here.

source/base/ThreadPool.ooc
marcusnaslund: Since we have the new (in the future platform independent?) WaitCondition class, absolutely.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,203 @@
...
-		job := SynchronizedThreadJob new(body)
-		this _add(job)
-		job
+	addWaitResult: func ~result <T> (action: Func -> T) -> ResultPromise<T> {
+		task := _ResultTask<T> new(action, this _globalmutex)
+		this _add(task)
+		_ResultTaskPromise new(task)
 	}
-	add: func (body: Func) { this _add(ThreadJob new(body)) }
-	wait: func (body: Func) { this addSynchronized(body) wait() . free() }
```
emilwestergren: Why did you remove this? It's nice to have when you want to add something to the pool and synchronize immediately after, without having to deal with waiting on and freeing promises.

source/base/ThreadPool.ooc
marcusnaslund: To avoid the cost of creating a new mutex for each task, since there may be many.

emilwestergren: Don't worry about that, we can solve this in the long run by recycling Mutexes.

source/base/ThreadPool.ooc
marcusnaslund: Good point, I removed it because it was related to the now removed SynchronizedThreadJob. I will make it work again.

source/base/ThreadPool.ooc
simonmika: That is the default value. There is no difference.

source/base/ThreadPool.ooc
simonmika: Again that is the default value already, no need for unnecessary assignments.

source/base/ThreadPool.ooc
simonmika: It is enough to have a mutex per thread so why ever create more? Even if you recycle.

source/base/ThreadPool.ooc
simonmika: I don't think that it should be there. Wait for the task you added.

source/base/ThreadPool.ooc
davidhesselbom: We've had trouble with uninitialized values before, `value: Int` not being 0 before being assigned etc. On the other hand, the case might be different with data type members of classes, since Objects are instantiated with a `malloc` call...

source/base/ThreadPool.ooc
emilwestergren: It causes unnecessary waiting. Why share a mutex when they are independent of eachother?

source/base/ThreadPool.ooc
marcusnaslund: Ideally, this would interrupt/cancel the thread and have it move on to the next task in the queue (if any). Haven't yet figured out a nice way to do this though.

marcusnaslund: For now, it allows wait() to return directly (although the task will still run).

source/base/ThreadPool.ooc
emilwestergren: If you say so. Just remember to adapt the Android code in ooc-vidproc because this breaks the kean API.

PR #425, opened 2015-09-24
 Implemented QUEST for weighted quaternion averaging.

tomasmahlberg:
Depends on cogneco/ooc-kean#418.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -331,7 +331,9 @@ FloatMatrix : cover {
 			if (this width != 3)
 				raise("Determinant implemented only for 3x3 matrices in FloatMatrix")
 		}
-		result := this[0, 0] * (this[1, 1] * this[2, 2] - this[2, 1] * this[1, 2]) - this[1, 0] * (this[0, 1] * this[2, 2] - this[2, 1] * this[0, 2]) + this[2, 0] * (this[0, 1] * this[1, 2] - this[1, 1] * this[0, 2])
+		result := this[0, 0] * (this[1, 1] * this[2, 2] - this[2, 1] * this[1, 2]) -
+			this[1, 0] * (this[0, 1] * this[2, 2] - this[2, 1] * this[0, 2]) +
+			this[2, 0] * (this[0, 1] * this[1, 2] - this[1, 1] * this[0, 2])
 		this free(Owner Receiver)
 		result
 	}
@@ -405,16 +407,16 @@ FloatMatrix : cover {
 			this elements[i] -= other elements[i]
 		other free(Owner Receiver)
 	}
+
+	operator * (other: Float) -> This {
+		result := This new(this dimensions)
+		for (i in 0 .. this dimensions area)
+			result elements[i] = other * this elements[i]
+		this free(Owner Receiver)
+		result
+	}
 }
 
 operator * (left: Float, right: FloatMatrix) -> FloatMatrix {
-	result := FloatMatrix new(right dimensions)
-	for (i in 0 .. right dimensions area)
-		result elements[i] = left * right elements[i]
-	right free(Owner Receiver)
-	result
-}
-
-operator * (left: FloatMatrix, right: Float) -> FloatMatrix {
 	right * left
 }
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -289,6 +289,53 @@ FloatMatrix : cover {
 		upper free(Owner Receiver)
 		result
 	}
+
+	cofactors: func -> This {
+		// TODO: At some point, when needed, implement for the general case.
+		version(safe) {
+			if (!this isSquare)
+				raise("Matrix must be square in FloatMatrix cofactors")
+			if (this width != 3)
+				raise("Cofactors implemented only for 3x3 matrices in FloatMatrix")
+		}
+		result := This new(this dimensions)
+		result[0, 0] = this[1, 1] * this[2, 2] - this[2, 1] * this[1, 2]
+		result[1, 0] = - (this[0, 1] * this[2, 2] - this[2, 1] * this[0, 2])
+		result[2, 0] = this[0, 1] * this[1, 2] - this[1, 1] * this[0, 2]
+		result[0, 1] = - (this[1, 0] * this[2, 2] - this[2, 0] * this[1, 2])
+		result[1, 1] = this[0, 0] * this[2, 2] - this[2, 0] * this[0, 2]
+		result[2, 1] = - (this[0, 0] * this[1, 2] - this[1, 0] * this[0, 2])
+		result[0, 2] = this[1, 0] * this[2, 1] - this[2, 0] * this[1, 1]
+		result[1, 2] = - (this[0, 0] * this[2, 1] - this[2, 0] * this[0, 1])
+		result[2, 2] = this[0, 0] * this[1, 1] - this[1, 0] * this[0, 1]
+		this free(Owner Receiver)
+		result
+	}
+
+	adjugate: func -> This {
+		// TODO: At some point, when needed, implement for the general case.
+		version(safe) {
+			if (!this isSquare)
+				raise("Matrix must be square in FloatMatrix adjugate")
+			if (this width != 3)
+				raise("Adjugate implemented only for 3x3 matrices in FloatMatrix")
+		}
+		this cofactors() transpose()
+	}
+
+	determinant: func -> Float {
+		// TODO: At some point, when needed, implement for the general case.
+		version(safe) {
+			if (!this isSquare)
+				raise("Matrix must be square in FloatMatrix determinant")
+			if (this width != 3)
+				raise("Determinant implemented only for 3x3 matrices in FloatMatrix")
+		}
+		result := this[0, 0] * (this[1, 1] * this[2, 2] - this[2, 1] * this[1, 2]) - this[1, 0] * (this[0, 1] * this[2, 2] - this[2, 1] * this[0, 2]) + this[2, 0] * (this[0, 1] * this[1, 2] - this[1, 1] * this[0, 2])
+		this free(Owner Receiver)
+		result
+	}
+
 	take: func -> This { // call by value -> modifies copy of cover
 		this _elements = this _elements take()
 		this
@@ -366,4 +413,8 @@ operator * (left: Float, right: FloatMatrix) -> FloatMatrix {
 		result elements[i] = left * right elements[i]
 	right free(Owner Receiver)
 	result
+}
+
+operator * (left: FloatMatrix, right: Float) -> FloatMatrix {
+	right * left
 }
--- a/source/math/Quaternion.ooc
+++ b/source/math/Quaternion.ooc
@@ -13,10 +13,11 @@
 //
 // You should have received a copy of the GNU Lesser General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
+use ooc-collections
 import FloatPoint3D
+import FloatVectorList
 import FloatTransform3D
 import FloatMatrix
-import VectorList
 import math
 
 Quaternion: cover {
@@ -262,7 +263,7 @@ Quaternion: cover {
 		}
 		result
 	}
-	weightedQuaternionMean: static func (quaternions: VectorList<This>, weights: Float[]) -> This {
+	weightedQuaternionMean: static func (quaternions: VectorList<This>, weights: FloatVectorList) -> This {
 		// Implementation of the QUEST algorithm. Original publication:
 		// M.D. Shuster and S.D. Oh, "Three-Axis Attitude Determination from Vector Observations", 1981
 		// [http://www.malcolmdshuster.com/Pub_1981a_J_TRIAD-QUEST_scan.pdf]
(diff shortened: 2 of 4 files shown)


tomasmahlberg:
Suggestions on renaming and/or relocating the helper functions I put in the Quaternion cover are welcome.

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,80 @@ Quaternion: cover {
...
+
+		B := FloatMatrix new(3, 3)
+		w := FloatMatrix new(1, 3)
+		v := FloatMatrix new(1, 3)
+		for (n in 0 .. vectorCount) {
+			for (i in 0 .. referenceVectors dimensions height) {
+				v set(0, i, referenceVectors get(n, i))
+				w set(0, i, observationVectors get(n, i))
+			}
+			B += weights[n / 3] / 3.0f * w * v transpose()
```
davidhesselbom: `B += ` replaces `B` with a new `FloatMatrix` and doesn't free the original `B`. This will cause a memory leak.

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,80 @@ Quaternion: cover {
...
+		}
+		S := B + B transpose()
```
davidhesselbom: `B transpose()` creates a new `FloatMatrix`, which is not freed. This will cause a memory leak.

source/math/Quaternion.ooc
davidhesselbom: `v transpose()` creates a new `FloatMatrix` and so do the operators `/` and `*`. This will cause memory leaks.

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,80 @@ Quaternion: cover {
...
+		w free()
+		v free()
+
+		temporaryZ := FloatPoint3D new()
+		for (n in 0 .. vectorCount)
+			temporaryZ += weights[n / 3] / 3.0f * FloatPoint3D new(observationVectors get(n, 0), observationVectors get(n, 1), observationVectors get(n, 2)) vectorProduct(FloatPoint3D new(referenceVectors get(n, 0), referenceVectors get(n, 1), referenceVectors get(n, 2)))
```
davidhesselbom: `get` and `set` is unique to `FloatMatrix`, I think, and this is probably the result of bugs in the compiler that have since been fixed. Could you try implementing the `[]` and `[]=` operators for `FloatMatrix` and use those instead? Keep the old `get` and `set` methods and let them use the `[]` and `[]=` operators under the hood until we've made sure they're not used anywhere else (as I did with the `free` function`).

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,80 @@ Quaternion: cover {
...
+		// M.D. Shuster and S.D. Oh, "Three-Axis Attitude Determination from Vector Observations", 1981
+		// [http://www.malcolmdshuster.com/Pub_1981a_J_TRIAD-QUEST_scan.pdf]
+		// TODO: Fix singularity problem when the angle is close to PI, using sequential rotations
+		vectorCount := 3 * quaternions count
+		referenceVectors := FloatMatrix new(vectorCount, 3)
+		observationVectors := FloatMatrix new(vectorCount, 3)
+		This _createVectorMeasurementsForQuest(quaternions, referenceVectors, observationVectors)
+		This _normalizeFloatArray(weights, quaternions count)
+
+		B := FloatMatrix new(3, 3)
```
davidhesselbom: Variable names shouldn't be capitalized, normally. I take it you're doing it to agree with the original formulae?

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,80 @@ Quaternion: cover {
...
+			B += weights[n / 3] / 3.0f * w * v transpose()
+		}
+		S := B + B transpose()
+		w free()
+		v free()
+
+		temporaryZ := FloatPoint3D new()
+		for (n in 0 .. vectorCount)
+			temporaryZ += weights[n / 3] / 3.0f * FloatPoint3D new(observationVectors get(n, 0), observationVectors get(n, 1), observationVectors get(n, 2)) vectorProduct(FloatPoint3D new(referenceVectors get(n, 0), referenceVectors get(n, 1), referenceVectors get(n, 2)))
+		This _copyFloatPoint3DToFloatMatrixColumn(Z := FloatMatrix new(1, 3), temporaryZ, 0, 0)
```
davidhesselbom: I don't get it, what happens to `Z` here? It's created here, and not returned. What's the point?

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,80 @@ Quaternion: cover {
...
+
+		// TODO: Implement Newton-Raphson for better estimation of max(eigenvalue) (Eq. 70 in the article).
+		// For this, the adjugate and determinant matrix operations are needed
+		maximumEigenvalue := 1.0f
+		Y := ((maximumEigenvalue + B trace()) * FloatMatrix identity(3) - S) solve(Z)
```
davidhesselbom: `Z` actually exists here? Fascinating. I thought it'd gone out of scope.

davidhesselbom: More memory leaks.

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,80 @@ Quaternion: cover {
...
+
+		constant := 1.0f / sqrt(1.0f + ((Y transpose() * Y) get(0, 0) as Float))
```
davidhesselbom: More memory leaks.

tomasmahlberg:
I will work on these memory leaks. In some cases though, maybe it will be better to try to not use the `FloatMatrix` functions and use nested for loops and operate on the elements instead? Otherwise the code will get bloated with instantiations and frees of temporary variables. I will think about this.

source/math/Quaternion.ooc
tomasmahlberg: That's correct. Is this OK?

sebastianbaginski:
Maybe you could introduce a helper class for this that will keep all the variables used in calculations ? Looks like each time you need `B`, `S`, `w`, `v` etc. and they always seem to have the same size. This would allow you to hide the memory management code inside this helper class and use only nice public interface in Quaternion to do the interesting work.

davidhesselbom:
> Otherwise the code will get bloated with instantiations and frees of temporary variables

That's already the case in many places, but yeah, nobody likes it that way, so I like Sebastian's idea.

tomasmahlberg:
Agreed. I will look into this.

source/math/Quaternion.ooc
tomasmahlberg: I heard that @marcusnaslund might be working on this?

source/math/Quaternion.ooc
marcusnaslund: How often do we do this? There is certainly a more efficient way of adding a transpose than creating a new matrix, shifting values around, and then adding (and freeing).

source/math/Quaternion.ooc
marcusnaslund: I was discussing it but haven't done it, but I can look into it. Is there a way of defining the []-operator to take two arguments? Otherwise, how should it be done?

davidhesselbom: Yes, look at `FloatTransform2D` for an example.

source/math/Quaternion.ooc
davidhesselbom: If you need to read some paper to understand what's going on here, it has benefits. Personally, I would prefer more descriptive names... @simonmika ?

source/math/Quaternion.ooc
marcusnaslund: Perfect, I'll fix it.

source/math/Quaternion.ooc
tomasmahlberg: Not often. Is there a term for the sum of a square matrix and its transpose? If so, it could be implemented as function in the `FloatMatrix` cover. Otherwise I think it mostly just reduces the readability and compactness of the code to do these kind of operations with nested for loops. It's a trade-off I guess...

davidhesselbom: Especially when it's a 3x3 matrix, yes.

marcusnaslund: Don't think so because it has no special properties. The product of a matrix and its transpose does, though (it's symmetric) and is used in solve(), but I can't find a name for it either. There is more to be gained there, but as you say it's still a specialization. Depends on how much performance we want out of this.

source/math/Quaternion.ooc
simonmika: Use more descriptive names, and capital initial letter is reserved for types. Write a mapping between names and the formula in the comment.

source/math/Quaternion.ooc
marcusnaslund: #439

simonmika:
We should solve the memory management issue with a owner flag, much the same way at is done in Text. I will take a look at it.

source/math/Quaternion.ooc
davidhesselbom: Okay, that works too.

simonmika:
I have a solution on its way.

tomasmahlberg:
:thumbsup:

tomasmahlberg:
Updated and rebased.

source/math/FloatMatrix.ooc
```diff
@@ -367,3 +414,7 @@ operator * (left: Float, right: FloatMatrix) -> FloatMatrix {
 	right free(Owner Receiver)
 	result
 }
+
+operator * (left: FloatMatrix, right: Float) -> FloatMatrix {
+	right * left
+}
```
davidhesselbom: No, this should be defined inside the class, like the other operators (preferably next to the other `*` operator implementation):
```ooc
operator * (factor: Float) -> This {
    factor * left
}
```

source/math/FloatMatrix.ooc
```diff
@@ -289,6 +289,53 @@ FloatMatrix : cover {
...
+
+	determinant: func -> Float {
+		// TODO: At some point, when needed, implement for the general case.
+		version(safe) {
+			if (!this isSquare)
+				raise("Matrix must be square in FloatMatrix determinant")
+			if (this width != 3)
+				raise("Determinant implemented only for 3x3 matrices in FloatMatrix")
+		}
+		result := this[0, 0] * (this[1, 1] * this[2, 2] - this[2, 1] * this[1, 2]) - this[1, 0] * (this[0, 1] * this[2, 2] - this[2, 1] * this[0, 2]) + this[2, 0] * (this[0, 1] * this[1, 2] - this[1, 1] * this[0, 2])
```
davidhesselbom: Please split this up on at least 2 lines.

source/math/Quaternion.ooc
```diff
@@ -15,6 +15,8 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 import FloatPoint3D
 import FloatTransform3D
+import FloatMatrix
+import VectorList
```
davidhesselbom: No, import `ooc-collections`.

source/math/Quaternion.ooc
```diff
@@ -260,6 +262,105 @@ Quaternion: cover {
 		}
 		result
 	}
+	weightedQuaternionMean: static func (quaternions: VectorList<This>, weights: Float[]) -> This {
```
davidhesselbom: Why a `VectorList` for the `Quaternion`s but an array for the `Float`s?

tomasmahlberg: Actually, there is no good reason... I will change `Float[]` to `FloatVectorList`. Then I can also get rid of the function `_normalizeFloatArrayForQuest` and just use `weights / weights sum`. Better?

davidhesselbom: Yes. Less code is often a good thing.

tomasmahlberg:
Updated.

source/math/Quaternion.ooc
```diff
@@ -260,6 +263,99 @@ Quaternion: cover {
...
+		constantB := sigma squared() + temporary[0, 0]
+		temporary free()
+		temporary = vectorQuantityZ transpose() * matrixQuantityS * vectorQuantityZ
+		constantC := matrixQuantityS determinant() + temporary[0, 0]
+		temporary free()
+		temporary = vectorQuantityZ transpose() * matrixQuantityS * matrixQuantityS * vectorQuantityZ
+		constantD := temporary[0, 0]
+		temporary free()
+
+		for (_ in 0 .. maximumIterationCount) {
```
davidhesselbom: Huh, haven't seen that one before!

source/math/Quaternion.ooc
```diff
@@ -260,6 +263,99 @@ Quaternion: cover {
 		}
 		result
 	}
+	weightedQuaternionMean: static func (quaternions: VectorList<This>, weights: FloatVectorList) -> This {
```
davidhesselbom: How about `weightedMean`?

tomasmahlberg: Sounds good to me.

source/math/Quaternion.ooc
tomasmahlberg: Haha, yes, does it look weird? I thought it made sense since I am only interested in limiting the number of iterations, not to actually access the iteration variable.

davidhesselbom:
There is plenty of room for improvement here, as far as both readability and performance goes, but I'll leave for a later pull request. I guess we'll find out, when the time comes, whether it leaks memory or not...

tomasmahlberg:
Yes, I made an issue for this already - #617.

source/math/Quaternion.ooc
davidhesselbom: Oh, I thought this was fixed already, or I wouldn't have merged... oh well.

source/math/Quaternion.ooc
davidhesselbom: No, I like it, just didn't know you could do that.

PR #426, opened 2015-09-24
WaitCondition implementation for Windows

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/407 .
WaitCondition implementation for win32 based on event objects.
It could be easier to use condition variables (https://msdn.microsoft.com/en-us/library/windows/desktop/ms682052%28v=vs.85%29.aspx) but they are supported only from Vista and above, and they didn't work with msys2 (outdated headers and libraries I guess).
This implementation requires some manual bookkeeping. We need to keep the list of events that are waiting on the wait condition and then manually manipulate this list when we broadcast a *wake up* signal.
I do not have a good idea about when to free created events. We cannot do that after sending the *wake up* signal, as we can't know when it will be delivered, and destroying an event while it is still processed is not what we want (possible deadlock). Maybe we could test the sent events with `WaitForSingleObject` with very short timeout before we queue new events.
Right now event objects are freed when the wait condition is released, so it is not a good idea to destroy the wait condition while threads are still waiting on it.
Anyway, it seems to be working on windows (thanks @jonathanudd for all the help with testing :)), so I guess it is good enough for now.

### DIFF
--- a/source/sdk/threading/native/ConditionWin32.ooc
+++ b/source/sdk/threading/native/ConditionWin32.ooc
@@ -1,0 +1,87 @@
+version(windows) {
+
+	import ../Thread
+	import native/win32/[types, errors]
+	import MutexWin32
+	import structs/ArrayList
+
+	include windows
+
+	CreateEvent: extern func (...) -> Handle
+	SetEvent: extern func (Handle) -> Bool
+	CloseHandle: extern func (Handle) -> Bool
+	Infinite: extern (INFINITE) Long
+	WaitSuccess: extern (WAIT_OBJECT_0) Long
+
+	ConditionWin32: class extends WaitCondition {
+		_mutex := Mutex new()
+		_waitingEvents := ArrayList<Handle> new()
+		_waitingForRelease := ArrayList<Handle> new()
+		init: func
+		free: override func {
+			this destroy()
+			this _waitingEvents free()
+			this _waitingForRelease free()
+			this _mutex destroy()
+			super()
+		}
+		wait: func (mutex: Mutex) -> Bool {
+			eventId := CreateEvent(null, false, false, null)
+			this _mutex lock()
+			this _waitingEvents add(eventId)
+			this _mutex unlock()
+			mutex unlock()
+			(WaitForSingleObject(eventId, Infinite) == WaitSuccess)
+		}
+		signal: func -> Bool {
+			result := false
+			toSignal := 0 as Handle
+			this _mutex lock()
+			if (this _waitingEvents size > 0) {
+				toSignal = this _waitingEvents first()
+				this _waitingEvents removeAt(0)
+				this _waitingForRelease add(toSignal)
+			}
+			this _mutex unlock()
+			if (toSignal != 0)
+				result = SetEvent(toSignal)
+			result
+		}
+		broadcast: func -> Bool {
+			result := false
+			toSignal := ArrayList<Handle> new()
+			this _mutex lock()
+			while (this _waitingEvents size > 0) {
+				eventId := this _waitingEvents first()
+				this _waitingEvents removeAt(0)
+				this _waitingForRelease add(eventId)
+				toSignal add(eventId)
+			}
+			this _mutex unlock()
+			if (toSignal size > 0) {
+				result = true
+				for (i in 0 .. toSignal size)
+					result = result && SetEvent(toSignal[i])
+			}
+			toSignal free()
+			result
+		}
+		// do not destroy wait condition while threads are still waiting on it
+		destroy: func -> Bool {
+			this _mutex lock()
+			//waiting threads should have a chance to wake up
+			for (i in 0 .. this _waitingEvents size) {
+				SetEvent(this _waitingEvents[i])
+				Thread yield()
+			}
+			for (i in 0 .. this _waitingForRelease size)
+				CloseHandle(this _waitingForRelease[i])
+			for (i in 0 .. this _waitingEvents size)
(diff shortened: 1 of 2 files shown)


davidhesselbom:
So, if you don't version block both entire files, it won't build on any system?

sebastianbaginski:
I don't know if the entire file needs to be enclosed into a *version* block, I'm not very familiar with this feature of ooc. If I understand it correctly, it maps to a `#ifdef` block in C sources.
I don't think that moving anything more than `import ../Thread` out of this block makes any sense, as almost everything in this file is specific to windows (and same for unix in ConditionUnix).

davidhesselbom:
I'm okay with it, I was just wondering.

sebastianbaginski:
OK. In MutexWin32 we have `import native/win32/[types, errors]` out of the version block. It looks strange to me, if the platform is not windows, then why including the data types native to windows ?

source/sdk/threading/native/ConditionWin32.ooc
```diff
@@ -0,0 +1,87 @@
...
+	import MutexWin32
+	import structs/ArrayList
+
+	include windows
+
+	CreateEvent: extern func (...) -> Handle
+	SetEvent: extern func (Handle) -> Bool
+	CloseHandle: extern func (Handle) -> Bool
+	INFINITE: extern Long
+	WAIT_OBJECT_0: extern Long
```
davidhesselbom: Give these ooc names:
```ooc
Infinite: extern (INFINITE) Long
WaitObject0: extern (WAIT_OBJECT_0) Long
```
Actually, you can probably think of a better name than `WaitObject0`.

source/sdk/threading/native/ConditionWin32.ooc
```diff
@@ -0,0 +1,87 @@
...
+	ConditionWin32: class extends WaitCondition {
+		_mutex := Mutex new()
+		_waitingEvents := ArrayList<Handle> new()
+		_waitingForRelease := ArrayList<Handle> new()
+		init: func
+		free: override func {
+			this destroy()
+			this _waitingEvents free()
+			this _waitingForRelease free()
+			this _mutex destroy()
```
davidhesselbom: Since we seem to have given up the idea of compatibility with the original SDK, what do you think about renaming `destroy` to `free` for consistency? It doesn't need doing in this pull request, of course, just asking.

source/sdk/threading/native/ConditionWin32.ooc
```diff
@@ -0,0 +1,87 @@
...
+				Thread yield()
+			}
+			for (i in 0 .. this _waitingForRelease size)
+				CloseHandle(this _waitingForRelease[i])
+			for (i in 0 .. this _waitingEvents size)
+				CloseHandle(this _waitingEvents[i])
+			this _waitingEvents clear()
+			this _waitingForRelease clear()
+			this _mutex unlock()
+			true
```
davidhesselbom: Is this function ever supposed to be able to return false?

source/sdk/threading/native/ConditionWin32.ooc
```diff
@@ -0,0 +1,87 @@
...
+			if (toSignal size > 0) {
+				result = true
+				for (i in 0 .. toSignal size)
+					result = result && SetEvent(toSignal[i])
+			}
+			toSignal free()
+			result
+		}
+		// do not destroy wait condition while threads are still waiting on it
+		destroy: func -> Bool {
```
davidhesselbom: Who will call this function?

davidhesselbom:
So... how do I test this? How do I check for memory leaks?

source/sdk/threading/native/ConditionWin32.ooc
sebastianbaginski: I agree, there should be only one way to dispose of the object.

source/sdk/threading/native/ConditionWin32.ooc
sebastianbaginski: This is only a cosmetic change, can I do that on separate pull request, or at least when I have access to a windows machine ? I know that this code worked, and if I change something then I will have to ask Jonathan to test it again.

source/sdk/threading/native/ConditionWin32.ooc
sebastianbaginski: The unix version, yes. This one is just implementing the abstract interface. There should be only the `free` method, anyway.

source/sdk/threading/native/ConditionWin32.ooc
sebastianbaginski: As above, `free` should be responsible for it.

sebastianbaginski:
Get a windows machine, install rock, run `Thread` test. I would be happy to do that for you, but I don't have access to a windows workstation.

sebastianbaginski:
I asked @jonathanudd to help me with testing again today, so I will update this PR after we run the tests.

source/sdk/threading/native/ConditionWin32.ooc
davidhesselbom: Okay, no problem. I just don't want to litter the code with UGLY_ALL_CAPS_CONSTANTS if it can be avoided :)

sebastianbaginski:
I updated the ugly names for windows constants and ran it again on windows. It seems to be working fine, I don't have anything more to add here.
Remarks about `free`/`delete` are still valid, though, but I think they deserve a separate issue / pull request.

davidhesselbom:
Sure, don't fix `free/destroy` here.

PR #427, opened 2015-09-24
Making yet another attempt at fixing the queue.

davidpiuva:
Implemented dequeue and peek with only one return value since both previous implementations were too complex for the compiler. Instead of returning false on failure, there is a default value that should be null for reference types and identity for value types.

Added comments about things that were not so obvious when I started reading the code. The head usually goes first.

The problem with the old interfaces is that both pointers and tuples are not so tested features in the compiler. Especially when combined with overloaded pattern matching of generic types and automatic referencing. Anyone wanting to use the new dequeue and peek should use the ~functional postfix to be explicit about which to use until the overloading mess is solved.


davidhesselbom:
Could you write a test for `Queue` with a reference type? The current test only tests covers.

davidpiuva:
I have added the task to my to do list next to testing synchronize.

davidhesselbom:
My point is, you have made changes for fixing the Queue when using reference types, but you nothing to show for it. How do I know this will even help anything?

davidhesselbom:
I will leave this open until someone (not necessarily you) writes a test to show your fixes actually work.

davidpiuva:
I make some tests now then.

davidpiuva:
Tests have been created and passed.

source/collections/Queue.ooc
```diff
@@ -12,17 +12,20 @@ Queue: abstract class <T> {
 	dequeue: abstract func -> (T, Bool)
 	peek: abstract func -> (T, Bool)
 	//TODO: This works much better and will do for now
-	dequeue: abstract func ~out (result: T*) -> Bool
+	dequeue: abstract func ~out (result: T*) -> Bool // Also crashing the compiler
```
davidhesselbom: What is the purpose of this comment?

davidhesselbom:
@simonmika, please take a look at this.

source/collections/Queue.ooc
davidpiuva: I suggest slowly obsoleting it over a cycle of pull requests.
Removing it right away might break something if someone can't come up with a good default value for a value type.

davidhesselbom: But it only works for value types, or what?

davidpiuva: The new methods works for both classes and covers but covers need a reserved value when using the thread safe queue since you can't ask if it is empty and then assume that no other thread came before you.

davidhesselbom: This sounds complicated enough to deserve an explanation in the file itself. `// Also crashing the compiler` is not helpful. Also, what do you mean by "functional"? That it works, as opposed to the existing function?

davidpiuva: It is hard to understand why it crashed.

"Functional" means returning by value instead of returning by reference to target. What name do you suggest?

source/collections/Queue.ooc
```diff
@@ -11,18 +11,21 @@ Queue: abstract class <T> {
 	//TODO: This is how we want it to look but we get problems compiling
 	dequeue: abstract func -> (T, Bool)
 	peek: abstract func -> (T, Bool)
-	//TODO: This works much better and will do for now
+	// Obsoleted from using template pointers that are not reliable in the compiler's type safety
 	dequeue: abstract func ~out (result: T*) -> Bool
```
simonmika: The argument `result` should be called `default` or something similar.

source/collections/Queue.ooc
```diff
@@ -11,18 +11,21 @@ Queue: abstract class <T> {
...
 	peek: abstract func ~out (result: T*) -> Bool
-
+	// Yet another interface to go around compiler bugs
+	// T must be nullable or have a reserved value to know when it failed
+	dequeue: abstract func ~default (onEmpty: T) -> T
```
simonmika: Call `onEmpty` `fallback` instead.

source/collections/Queue.ooc
```diff
@@ -77,6 +88,12 @@ VectorQueue: class <T> extends Queue<T> {
 		success := this peek(result&)
 		(result, success)
 	}
+	peek: func ~default (onEmpty: T) -> T {
+		if (this empty)
```
simonmika:
```
this empty ? onEmpty : this _backend[this _head]

**Comment by simonmika** [comment]
Can we remove all old `peek` and `dequeue` soon?

**Comment by davidpiuva** [comment]
BufferedPlayer, DelayModifier and VidhanceContextAndroid will have to be remade with fallback. Anyone already working on those should be given the task before someone apply it to the rest to avoid conflicts.

**Comment by davidhesselbom** [comment]
Why do callers have to be updated to work after you changed the name, not the type, of the parameter?

**Comment by davidpiuva** [comment]
In some cases like natural numbers and floats, it would make sense to specify -1 or NaN as fallback instead of 0. It is however weird to specify a fallback for a complex cover at all when we just checked if a local queue is empty or not and we might need another raising an exception on empty.

**Comment by davidpiuva** [comment]
The callers are using the obsolete functions returning bool and getting result by reference.

File: source/collections/Queue.ooc [hunk 5]
```diff
@@ -77,6 +88,9 @@ VectorQueue: class <T> extends Queue<T> {
 		success := this peek(result&)
 		(result, success)
 	}
+	peek: func ~default (fallback: T) -> T {
+		this empty ? fallback : this _backend[this _head]
+	}
```
davidhesselbom:
```[CC] Queue
/home/dhesselbom/versioned/ooc/ooc-kean/source/collections/Queue.ooc: In function ‘Queue__VectorQueue_peek_default’:
/home/dhesselbom/versioned/ooc/ooc-kean/source/collections/Queue.ooc:92:105: warning: pointer/integer type mismatch in conditional expression [enabled by default]
   this empty ? fallback : this _backend[this _head]
```
Maybe it needs to be `this _backend[this _head] as T` or something?

davidpiuva: Only pointers can cast generics according to the compiler.

davidpiuva:
Value types will need standard names for functions creating an empty value equivalent to null or Nan and a function checking if the value is empty.

source/collections/Queue.ooc
```diff
@@ -77,6 +88,12 @@ VectorQueue: class <T> extends Queue<T> {
 		success := this peek(result&)
 		(result, success)
 	}
+	peek: func ~default (fallback: T) -> T {
+		if (this empty)
+			fallback
+		else
+			this _backend[this _head]
+	}
```
davidhesselbom: This has multiple returns, which we normally don't accept.
```ooc
	peek: func ~default (fallback: T) -> T {
		result: T
		if (this empty)
		 	result = fallback
		else
			result = this _backend[this _head]
		result
	}
```
solves this, but it's longer...

simonmika:
Or be wrapped in a Cell.

PR #429, opened 2015-09-24
Support for missing numeric types in Fixture

sebastianbaginski:
Added missing compare constraints for LDouble, Long and ULong.
Updated Text test to use compare constraints.


source/sdk/lang/types.ooc
```diff
@@ -137,12 +137,18 @@ Cell: class <T> {
 		match T {
 			case Int =>
 				(this val as Int) toString()
+			case Long =>
+				(this val as Long) toString()
 			case UInt =>
 				(this val as UInt) toString()
+			case ULong =>
+				(this val as ULong) toString()
```
davidhesselbom: Was it not possible to call `toString()` on `ULong` and `Long`, or was the result truncated or otherwise incorrect?

source/unit/Constraints.ooc
```diff
@@ -131,6 +139,10 @@ EqualModifier: class extends Modifier {
 		f := func (value, c: Object) -> Bool { value as Cell<Double> get() == c as Cell<Double> get() }
 		CompareWithinConstraint new(this, Cell<Double> new(correct), f)
 	}
+	to: func ~ldouble (correct: LDouble) -> CompareWithinConstraint {
+		f := func (value, c: Object) -> Bool { value as Cell<LDouble> get() == c as Cell<LDouble> get() }
+		CompareWithinConstraint new(this, Cell<LDouble> new(correct), f)
+	}
```
davidhesselbom: The duplicate code here makes me sad. Say, would this type of thing be avoidable in e.g. C++ by using e.g. templates?

source/sdk/lang/types.ooc
sebastianbaginski: Exception was thrown.

source/unit/Constraints.ooc
sebastianbaginski: Exactly. This would be something like
```C++
template <typename T>
CompareWithConstraint to(const T correct) {
    auto f = [&](const T value, Object& c){ return (Cell&)(value) . get<T>() == (Cell&)(c) . get<T>(); }
   return CompareWithinConstraint new(this, new Cell<T>(correct), f);
}
```

source/unit/Fixture.ooc
```diff
@@ -94,12 +94,21 @@ Fixture: abstract class {
...
 	}
 	expect: static func ~double (value: Double, constraint: Constraint) {
 		This expect(Cell new(value), constraint)
 	}
+	expect: static func ~ldouble (value: LDouble, constraint: Constraint) {
+		This expect(Cell new(value), constraint)
+	}
 	expect: static func ~isTrue (value: Bool) {
 		This expect(Cell new(value), is true)
 	}
```
davidhesselbom: Hmm. So we need one overload for each kind of Cover we want to use? Would it be possible to instead something like this?
```ooc
expect: static func (value: T, constraint: Constraint) {
    if (T inheritsFrom?(Object))
        value = Cell new(value)
    if (!constraint verify(value))
        TestFailedException new(value, constraint) throw()
}
```
But maybe generics aren't reliable enough for this sort of thing...

source/unit/Constraints.ooc
davidhesselbom: Is there really no way we can do the same thing with `T` in ooc?

source/sdk/lang/types.ooc
davidhesselbom: Alright, that's all the justification I need.

source/unit/Constraints.ooc
sebastianbaginski: I think you know ooc a lot better than I do, I don't even know if function templates exist in ooc (I know class templates are available).
I don't know, can we have a template class with static `to` method, something like
```ooc
CompareCreator: class <T> {
    to: static func (value: T)-> CompareWithinConstraint {
    to: func ~ulong (correct: ULong) -> CompareConstraint {
		f := func (value, c: Object) -> Bool { value as Cell<T> get() == c as Cell<T> get() }
		CompareConstraint new( ???? , Cell<T> new(correct), f, ComparisonType Equal)
	}
    }
}
```
but what should be in place of `this` then ? I can look into it later, though.

thomasfanell: A complete rewrite of ooc-unit is planned, though not yet started.

davidhesselbom: That's true, there's no point in spending more time on cleaning this up at the moment since the whole thing is getting thrown out soon. I welcome the fixes in this PR, though!

test/base/TextTest.ooc
```diff
@@ -92,49 +92,50 @@ TextTest: class extends Fixture {
...
+			expect(Text new("BEEF") toInt~inBase(16), is equal to(Text new("beef") toInt~inBase(16)))
+			expect(Text new("0xff") toInt(), is equal to(255))
+			expect(Text new("0x11") toInt(), is equal to(17))
+			expect(Text new("0xAA") toInt(), is equal to(170))
+			expect(Text new("0xffZZZ") toInt(), is equal to(255))
+			expect(Text new("0xDEADBEEF") toULong(), is equal to(3735928559))
+			expect(Text new(INT_MAX toString()) toLong(), is equal to(INT_MAX))
+			expect(Text new("-9") toULong(), is equal to(0))
+			expect(Text new("-9") toLLong(), is equal to(-9))
+			expect(Text new("-9") toLong(), is equal to(-9))
```
davidhesselbom: If any of these expects fail, it's going to be hard to figure out which one. Split this up into smaller test functions with a smaller number of expects. I think you can figure out how to split it up logically.

test/base/TextTest.ooc
```diff
@@ -92,49 +92,56 @@ TextTest: class extends Fixture {
...
+			expect(Text new("654") toInt~inBase(7), is equal to(4 + 5 * 7 + 6 * 7 * 7))
+		})
+		this add("convert to long and ulong", func {
+			expect(Text new(INT_MAX toString()) toLong(), is equal to(INT_MAX))
+			expect(Text new("0xDEADBEEF") toULong(), is equal to(3735928559))
+			expect(Text new("-9") toULong(), is equal to(0))
+			expect(Text new("-9") toLLong(), is equal to(-9))
+			expect(Text new("-9") toLong(), is equal to(-9))
 		})
 		this add("converting to float", func {
```
davidhesselbom: Some of these are `convert`, some `converting`.

Also, there's no reason to keep the function names in all lower-case. It's `Float`, not `float`, for instance.

sebastianbaginski: Fixed.

test/base/TextTest.ooc
```diff
@@ -91,50 +91,57 @@ TextTest: class extends Fixture {
...
+			expect(Text new("1") toFloat(), is equal to(1.0f) within(tolerance))
+			expect(Text new("-1.0") toFloat(), is equal to(-1.0f) within(tolerance))
+			expect(Text new("-1.") toFloat(), is equal to(-1.0f) within(tolerance))
+			expect(Text new("22.5") toFloat(), is equal to(22.5f) within(tolerance))
+			expect(Text new("123.763") toFloat(), is equal to(123.763f) within(tolerance))
+			expect(Text new("1e0") toFloat(), is equal to(1.0f) within(tolerance))
+			expect(Text new("5E-2") toFloat(), is equal to(0.05f) within(tolerance))
+			expect(Text new("2E12") toLDouble(), is equal to(2.0 * pow(10, 12) as LDouble) within(tolerance as LDouble))
+			expect(Text new("6.5E5") toFloat(), is equal to(6.5f * pow(10, 5) as Float) within(tolerance))
+			expect(Text new("-34.5E-2") toFloat(), is equal to(-0.345f) within(tolerance))
```
davidhesselbom: Since there are so many cases for `Float`, how about putting the ones with `E` in them in their own test function to keep the number of expects down?

test/base/TextTest.ooc
```diff
@@ -91,50 +91,57 @@ TextTest: class extends Fixture {
...
+			expect(Text new("-9") toLong(), is equal to(-9))
+		})
+		this add("Convert to Float", func {
+			tolerance := 0.001f
+			expect(Text new("1") toFloat(), is equal to(1.0f) within(tolerance))
+			expect(Text new("-1.0") toFloat(), is equal to(-1.0f) within(tolerance))
+			expect(Text new("-1.") toFloat(), is equal to(-1.0f) within(tolerance))
+			expect(Text new("22.5") toFloat(), is equal to(22.5f) within(tolerance))
+			expect(Text new("123.763") toFloat(), is equal to(123.763f) within(tolerance))
+			expect(Text new("1e0") toFloat(), is equal to(1.0f) within(tolerance))
```
davidhesselbom: `e` and `E` are both supposed to work, then?

sebastianbaginski: Yes, both should work.

test/base/TextTest.ooc
sebastianbaginski: Ok, I will update it.

sebastianbaginski: It is done, I think this is ready.

davidhesselbom:
screenshot from 2015-09-25 17 29 07
Great, thanks GitHub.

PR #430, opened 2015-09-24
Random Generator classes rewritten for performance

marcusnaslund:
IntUniformRandomGenerator next() is now on par with the sdk Random function, and next~fast() is much faster for range sizes < 32767. Gaussian generators for int and float now use both of the values the Box-Muller transform actually provides.
Float RandomGenerators are still slower than the sdk Random but much better than the Int version was.


source/math/IntRandomGenerator.ooc
```diff
@@ -15,17 +15,35 @@
...
-	_backend: FloatRandomGenerator
-	init: func ~withBackend (=_backend)
-	free: override func {
-		this _backend free()
-		super()
+IntUniformRandomGenerator: class {
+	_state, _min, _max, _range: Int
+	init: func (seed := Time microtime()) {
+		this _min = 0
+		this _max = 32767
```
davidhesselbom: Why `32767`?

source/math/FloatRandomGenerator.ooc
```diff
@@ -50,14 +50,16 @@ FloatGaussianRandomGenerator: class extends FloatRandomGenerator {
 	_backend: FloatUniformRandomGenerator
 	_mu := 0.0f
 	_sigma := 1.0f
+	_secondValue : Float
+	_hasSecond := false
 	init: func {
-		this _backend = FloatUniformRandomGenerator new()
+		this _backend = FloatUniformRandomGenerator new(Float minimumValue, 1.0f - Float minimumValue)
```
davidhesselbom: This program
```ooc
import math

println((1.0f - Float minimumValue < 1.0f) toString())
```
prints `false`, so what is the point of this subtraction?

source/math/FloatRandomGenerator.ooc
```diff
@@ -66,11 +68,20 @@ FloatGaussianRandomGenerator: class extends FloatRandomGenerator {
...
-			first = Float minimumValue
-		second := this _backend next()
-		value := (-2.0f * first log()) sqrt() * (2.0f * Float pi * second) cos()
-		value * this _sigma + this _mu
+		result : Float
+		if (this _hasSecond) {
+			result = this _secondValue
+			this _hasSecond = false
+		}
+		else {
```
davidhesselbom: } else {

source/math/IntRandomGenerator.ooc
```diff
@@ -35,35 +53,47 @@ IntRandomGenerator: class {
...
+		}
+		else {
+			scale := (-2.0f * (this _backend next()) log()) sqrt()
+			trigValue := (2.0f * Float pi * (this _backend next()))
+			value := scale * trigValue cos()
+			secondValue := scale * trigValue sin()
+			result = value * this _sigma + this _mu
+			this _secondValue = secondValue * this _sigma + this _mu
+			this _hasSecond = true
+		}
```
davidhesselbom: This looks identical to the `FloatGaussianRandomGenerator` version.

test/math/IntRandomGeneratorTest.ooc
```diff
@@ -5,22 +5,6 @@ import math
 IntRandomGeneratorTest: class extends Fixture {
 	init: func {
 		super("IntRandomGenerator")
```
davidhesselbom: Why did you remove the test?

davidhesselbom: Oh wait. Nevermind.

source/math/IntRandomGenerator.ooc
```diff
@@ -15,17 +15,35 @@
...
+	}
+	next: func ~fast -> Int {
+		value := this _generate()
+		this _min + (value % this _range)
 	}
 	next: func -> Int {
-		this _backend next() as Int
+		valuea := this _generate()
+		valueb := this _generate()
+		valuec := this _generate()
```
davidhesselbom: `valueA`, at the very least.

source/math/IntRandomGenerator.ooc
marcusnaslund: This is 0x7FFF (line 35) giving the maximum range that next~fast() can return, but I guess we could go with something else?

davidhesselbom: So... the highest random number we can generate for `Int`, which has a maximum value of `0x7FFFFFFF`, is `0x7FFF`? That doesn't seem right.

thomasfanell: Suggestion: create an extension in sdk/math.ooc for ```Short``` and set 0x7FFF as max value.

marcusnaslund: No, that's why the default next() generates more values.

davidhesselbom: > Suggestion: create an extension in sdk/math.ooc for Short and set 0x7FFF as max value.

```ooc
extend Short {
    maximum: static SHRT_MAX
...
```
would be nice, yeah. But, if `this _min` is 0 and `this _range` is `32767` by default, and the result of the default `next` function returns `this _min + (result % this _range)`, doesn't that mean that the returned value will always be < 32767?

marcusnaslund: Ah, I see now, you're right. We could have the default value of _max be INT_MAX.
Is it possible (easy) to extend Short in a nice way so that next~fast() returns Short but works well (casts automatically) with Int variables?
Otherwise I'll make next() the only function, and it will use the faster version if range < 32767, to avoid all confusion.

davidhesselbom: Having a fast way to return values in the `Short` range is nice, but I don't think it should be the **default** behavior. If you create an `IntRandomGenerator`, you should be able to assume it will return values in the `Int` range unless otherwise specified, which is what the other constructor is for.

marcusnaslund:
@davidhesselbom My tests all run fine. What went wrong here?

davidhesselbom:
Could be due to bad luck (race conditions, unfortunate random values) or maybe bitness issues (Circle runs 32-bit, you probably run 64-bit). That test has been having trouble for a long time.

marcusnaslund:
Since this update has no relevance with the queue, I will mark this as ready.

source/math/FloatRandomGenerator.ooc
```diff
@@ -27,22 +27,30 @@ FloatRandomGenerator: abstract class {
...
 		this _state = seed
+		this updateRange(0.0f, 1.0f)
 	}
-	init: func ~withParameters (=_min, =_max, seed := Time microtime()) {
+	init: func ~withParameters (min, max: Float, seed := Time microtime()) {
 		this _state = seed
+		this updateRange(min, max)
+	}
+	updateRange: func (=_min, =_max) {
+		this _rangeCoefficient = (1.0f / 4294967295.0f) * (this _max - this _min)
```
davidhesselbom: To the uninitiated reader, it would be nice to have some form of explanation for this magic number.

source/math/FloatRandomGenerator.ooc
```diff
@@ -27,22 +27,30 @@ FloatRandomGenerator: abstract class {
...
 	}
 	next: func -> Float {
-		this _state = 214013 * this _state + 2531011
-		value := ((((this _state >> 16) & 0x7fff) % 10000) as Float) / 10000.0f
-		Float linearInterpolation(this _min, this _max, value)
+		this _state ^= (this _state << 5)
+		this _state ^= (this _state >> 13)
+		this _state ^= (this _state << 6)
```
davidhesselbom: I have no idea what's going on here, but that doesn't mean it's wrong. @davidpiuva, @sebastianbaginski, does this look alright to you?

source/math/FloatRandomGenerator.ooc
```diff
@@ -50,14 +58,16 @@ FloatGaussianRandomGenerator: class extends FloatRandomGenerator {
 	_backend: FloatUniformRandomGenerator
 	_mu := 0.0f
 	_sigma := 1.0f
+	_secondValue : Float
+	_hasSecond := false
```
davidhesselbom: `_hasSecondValue`

davidhesselbom: Actually, what do you even mean by "second value"? With a single value, you will generate numbers "around" that value, and with a second value, you will generate numbers in the range between them, is that it?

source/math/FloatRandomGenerator.ooc
```diff
@@ -65,12 +75,24 @@ FloatGaussianRandomGenerator: class extends FloatRandomGenerator {
 		this _backend free()
 		super()
 	}
+	setRange: func (=_mu, =_sigma) {
+		_hasSecond = false
```
davidhesselbom: I don't understand why you're doing this... could you try to explain?

source/math/IntRandomGenerator.ooc
```diff
@@ -15,17 +15,41 @@
...
+		this setRange(min, max)
+		this _state = seed
+	}
+	_generate: func -> Int {
+		this _state = 214013 * this _state + 2531011
+		(this _state >> 16) & 0x7FFF
+	}
+	setRange: func (=_min, =_max) {
+		this _range = this _max - this _min + 1
+		this _shortRange = (this _range < 0x7FFF)
```
davidhesselbom: Once again, this should be `Short maximum`, if that's what this is.

source/math/IntRandomGenerator.ooc
```diff
@@ -15,17 +15,41 @@
...
+	init: func (seed := Time microtime()) {
+		this setRange(0, Int maximumValue)
+		this _state = seed
+	}
+	init: func ~withParameters (min, max: Int, seed := Time microtime()) {
+		this setRange(min, max)
+		this _state = seed
+	}
+	_generate: func -> Int {
+		this _state = 214013 * this _state + 2531011
```
davidhesselbom: Maybe a link to some website that explains the reason why these numbers are used here, if it can't be explained briefly. I have to admit I never worked out how PRNGs work, exactly, so for someone like me, it could help.

source/math/IntRandomGenerator.ooc
```diff
@@ -35,35 +59,31 @@ IntRandomGenerator: class {
...
-	init: func ~withBackend (backend: FloatGaussianRandomGenerator) {
-		super(backend)
+	init: func ~withUniformBackend (=_backend)
+	init: func ~withUniformBackendAndParameters (mu, sigma: Float, backend: FloatUniformRandomGenerator) {
+		this _backend = FloatGaussianRandomGenerator new~withBackendAndParameters(mu, sigma, backend)
 	}
-	init: func ~withUniformBackend (backend: FloatUniformRandomGenerator) {
-		super(FloatGaussianRandomGenerator new~withBackendAndParameters(0.0f, 10.0f, backend))
+	free: func {
+		this _backend free()
```
davidhesselbom: `free` is always an override for Objects. Also, you forgot to call `super()`.

source/math/FloatRandomGenerator.ooc
sebastianbaginski: Main idea here is to mess up the bit representation of the `_state`. I guess this is accomplished.
I didn't read the whole code, but it is not clear to me why having a `_rangeCoefficient` is better than `linearInterpolation` call, which was kind of more self-explanatory.

source/math/FloatRandomGenerator.ooc
marcusnaslund: The Box-Muller transform takes two uniformly distributed numbers and generates two normally distributed numbers: https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform
We only return one of them, so we save the second value for the next call.

source/math/FloatRandomGenerator.ooc
marcusnaslund: The stored value (if any) was calculated for other choices of mean and standard deviation. If we change them, with this call, it is no longer correct.

source/math/IntRandomGenerator.ooc
marcusnaslund: These are the same coefficients that were used for the FloatUniformRandomGenerator - but yes, I can include a link.

source/math/FloatRandomGenerator.ooc
davidhesselbom: Alright, so
```ooc
// Implements the Box-Muller transform
 FloatGaussianRandomGenerator: class extends FloatRandomGenerator {
```
would help clear this out for the next person who looks at this.

source/math/FloatRandomGenerator.ooc
davidhesselbom: And who calls this function? Is there any reason someone would create a `FloatGaussianRandomGenerator`, use it for a while, and then set its range?

source/math/IntRandomGenerator.ooc
marcusnaslund: Do I always call super() even if there is no inheritance? What about IntUniformRandomGenerator, do I need a free() which justs calls super()?

davidhesselbom: No, if the super class has no implementation, it should use the on in that one's super class, and so on. Eventually, you will reach `Object`'s virtual `free` function, which deallocates the object itself. Without calling `super` here, you will free `this _backend`, but not the `IntGaussianRandomGenerator` instance itself.

davidhesselbom: So free `free: override func { super() }` is completely useless, everywhere.

source/math/FloatRandomGenerator.ooc
marcusnaslund: I used setRange (noticed I've been inconsistent with naming here, will fix) for the uniform generators to reuse code, thought I'd include them in the gaussian versions for similarity. Possibly unneccessary.

davidhesselbom: @sebastianbaginski, is this what you need?

davidhesselbom: Wait, wrong class, sorry.

source/math/IntRandomGenerator.ooc
davidhesselbom: @sebastianbaginski, is *this* what you need?

source/math/FloatRandomGenerator.ooc
sebastianbaginski: I would welcome the possibility to change the range of generated numbers for existing generator object.

davidhesselbom:
#443 should be merged first.

source/sdk/lang/Numbers.ooc
```diff
@@ -64,6 +64,8 @@ UShort: cover from unsigned short extends ULLong
 
 INT_MAX := 2147483647
 INT_MIN := -INT_MAX - 1
+SHRT_MIN : extern Short
+SHRT_MAX : extern Short
```
davidhesselbom: #443 fixed this and a few other things.

davidhesselbom:
Needs rebasing, sorry...

marcusnaslund:
Sorry, what do I need to do?

davidhesselbom:
```
git fetch upstream
git checkout randomgenerator-performance
git merge upstream/master
meld .
```
and solve conflicts. After you're done,
```
git add [whatever files you needed to fix conflicts for]
git rebase --continue
```
.

davidhesselbom:
The problem is most likely that you and I have made conflicting changes in `Numbers.ooc`.

marcusnaslund:
Yep, the only conflicts in were in `Numbers.ooc`.

PR #431, opened 2015-09-25
Changed delimeter character

thomasfanell:
also some readability changes.


source/io/CsvReader.ooc
```diff
@@ -33,7 +35,7 @@ CsvReader: class extends Iterator<VectorList<Text>> {
 		readCharacter: Char
 		for (i in 0 .. rowDataLength) {
 			textBuilder := TextBuilder new()
-			while (i < rowDataLength && ((readCharacter = rowData[i]) != ';')) {
+			while (i < rowDataLength && ((readCharacter = rowData[i]) != ',')) {
```
davidhesselbom: Maybe it would be a good idea to use static variables for the delimiter chars?

thomasfanell:
Moved the delimiter to a static property in CsvReader. I don't know if this is the best approach, but I don't really like the idea of having two static variables, one for each class.

source/io/CsvWriter.ooc
```diff
@@ -15,9 +15,17 @@ CsvWriter: class {
 	}
 	write: func (row: VectorList<Text>) {
 		for (i in 0 .. row count) {
-			this _fileWriter file write(row[i] toString())
+			value := TextBuilder new (row[i])
```
davidhesselbom: `TextBuilder new(row[i])`. magic doesn't catch this... odd.

source/io/CsvWriter.ooc
```diff
@@ -15,9 +15,17 @@ CsvWriter: class {
...
+			for (k in 0 .. row[i] count)
+				if (this _isWhitespace(row[i][k])) {
+					value prepend('\"')
+					value append('\"')
+					break
+				}
 			if (i < row count - 1)
-				this _fileWriter write(";")
+				value append(CsvReader delimiter)
+			this _fileWriter file write(value toString())
```
davidhesselbom: `this _fileWriter write` should work.

thomasfanell: ```FileWriter write()``` takes either ```(Char*, SizeT)```, or ```(Char)```. I want to write the entire string without having to provide length. Is this preferable: ```this _fileWriter write(value raw, value count as SizeT)```? ```SizeT``` is as you know platform dependant.

source/io/CsvWriter.ooc
thomasfanell: Should it?

source/io/CsvWriter.ooc
davidhesselbom: Mm. Maybe we should add a function in FileWriter that takes a `String`?

thomasfanell: Do you mean that I should do that in this PR?

davidhesselbom: I suppose it can wait.

PR #432, opened 2015-09-25
ResultPromise renamed to Future

marcusnaslund:
Also added getResult() to Future since you might not want the result right away.


marcusnaslund:
Pull request #438 takes care of this as well.

PR #434, opened 2015-09-25
allow to create Text from string literals without giving explicit size

sebastianbaginski:
@simonmika wanted to have this

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -32,7 +32,10 @@ Text: cover {
 		this _count = 0
 		this _owner = Owner Unknown
 	}
-	init: func@ ~fromStringLiteral (string: CString, =_count) {
+	init: func@ ~fromStringLiteral (string: CString) {
+		this init(string, strlen(string))
+	}
+	init: func@ ~fromStringLiteralWithCount (string: CString, =_count) {
 		this _buffer = string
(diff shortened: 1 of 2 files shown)


source/base/Text.ooc
```diff
@@ -32,7 +32,10 @@ Text: cover {
 		this _count = 0
 		this _owner = Owner Unknown
 	}
-	init: func@ ~fromStringLiteral (string: CString, =_count) {
+	init: func@ ~fromStringLiteral (string: CString) {
+		this init(string, strlen(string))
+	}
+	init: func@ ~fromStringLiteralWithCount (string: CString, =_count) {
```
davidhesselbom: Why would you want to set `_count` to anything other than `strlen(string)` anyway?

sebastianbaginski: To initialize from string partially, I guess.
I included it as i wanted to avoid the use of `strlen()`. If the string is not null-terminated, this function will crash your program if you are lucky. Can even expose your system. I'm not a big fan of this function, that's why I didn't include it and here it is because Simon wanted it.

simonmika: This is only a stopgap solution until the day we have a compiler that can handle text literals: 
```
t"Text Literal"
```

PR #436, opened 2015-09-25
Add FloatSize2D minimum~Float() and maximum~Float().

bettinaselig:
Added these functions to write less code otherwise.


bettinaselig:
READY.

PR #438, opened 2015-09-25
New Integration of Promises with Threadpool

marcusnaslund:
Multiple bug fixes and general code cleanup. 
ResultPromise has been renamed to Future
wait() and waitAll() removed from ThreadPool which will break code in vidproc (another pull request there will fix this)


davidhesselbom:
Conflicts.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,162 @@
...
+			if (this _state != _PromiseState Unfinished) {
+				status = (this _state == _PromiseState Finished)
+				this _mutex unlock()
+				break
+			}
 			this _mutex unlock()
-			this _mutex destroy()
-			this _finishedCondition free()
-			super()
+			Time sleepMilli(This _mutexUpdateTime)
```
emilwestergren: Shouldn't we use a condition here?

davidhesselbom: A condition on what? Thanks for stopping by, btw.

emilwestergren: Waiting for a condition to be signaled, instead of busy waiting with sleep.

davidhesselbom: @simonmika , input plz

test/collections/SynchronizedQueueTest.ooc
```diff
@@ -35,27 +35,37 @@ SynchronizedQueueTest: class extends Fixture {
...
-			for (i in 0 .. 20)
-				pool add(func4)
+			for (i in 0 .. limitD)
+				promises[limitA + limitB + limitC + i] = pool getPromise(func4)
 
-			pool waitAll()
+			//TODO This is not pretty but replaces waitAll for now
+			for (i in 0 .. limitA + limitB + limitC + limitD) {
+				promises[i] wait()
+				promises[i] free()
```
davidhesselbom: `wait() . free()`, again

test/collections/BlockedQueueTest.ooc
```diff
@@ -39,38 +39,54 @@ BlockedQueueTest: class extends Fixture {
...
-				pool add(func1)
+				promises[limitA + limitB + limitC + limitD + limitE + i] = pool getPromise(func1)
 				Time sleepMilli(10)
 			}
-			pool waitAll()
+
+			//TODO This is not pretty but replaces waitAll for now
+			for (i in 0 .. limitA + limitB + limitC + limitD + limitE + limitF) {
+				promises[i] wait()
+				promises[i] free()
```
davidhesselbom: And here...

test/collections/SynchronizedQueueTest.ooc
marcusnaslund: PromiseCollector will fix this soon, but OK

source/base/ThreadPool.ooc
marcusnaslund: I will let him speak for himself, but we talked about this earlier and he seemed to agree with this approach for the things we're using it for (this was also written before WaitCondition was finished and platform independent - I believe it is now?).

source/base/Promise.ooc
```diff
@@ -103,6 +104,13 @@ _ThreadResultPromise: class <T> extends ResultPromise<T> {
 		status := this wait()
```
simonmika: Why assign to variable?

source/base/Promise.ooc
```diff
@@ -103,6 +104,13 @@ _ThreadResultPromise: class <T> extends ResultPromise<T> {
...
 		status ? this _result[T] : defaultValue
 	}
+	getResult: func (defaultValue: T) -> T {
+		status: Bool
+		this _mutex lock()
+		status = (this _state == _PromiseState Finished)
```
simonmika: `status` could be declared here.

source/base/ThreadPool.ooc
sebastianbaginski: > before WaitCondition was finished and platform independent - I believe it is now?

Yes, it works both on Linux and Windows.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,162 @@
...
-	_freeOnCompletion := false
-	init: func (body: Func) { super(body) }
-	free: override func {
-		this _mutex lock()
-		if (!this _finished) {
-			this _freeOnCompletion = true
-			this _mutex unlock()
-		} else {
+_Task: abstract class {
+	_mutexUpdateTime: static Int = 1
```
simonmika: If this is only used once place it there.

source/base/ThreadPool.ooc
simonmika: I suggest leaving it for now. The sleep call _should_ give up execution for at least the given time, so it is not entirely bad so let's see how this performs first. The old solution created a wait condition for every new synchronized task, which isn't ideal. Also we always want to keep the API towards the OS as small as possible.

If we are to use wait conditions they should be combined in a class together with a mutex as they are pretty useless without mutexes.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,162 @@
...
+		this _finishedTask()
+	}
+}
+
+_ResultTask: class <T> extends _Task {
+	_result: Cell<T>
+	_action: Func -> T
+	init: func (=_action, =_mutex)
+	run: func (=_mutex) {
+		temporary := Cell<T> new(this _action())
```
simonmika: Why create new variable?

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,162 @@
...
+	free: override func {
+		this _task free()
+		super()
+	}
+	wait: override func -> Bool {
+		this _task wait()
+	}
+	wait: func ~default (defaultValue: T) -> T {
+		status := this wait()
+		status ? this _task _result[T] : defaultValue
```
simonmika: Why create new variable?

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,162 @@
...
+	}
+	getResult: func (defaultValue: T) -> T {
+		status := (this _task _state == _PromiseState Finished)
+		status ? this _task _result[T] : defaultValue
```
simonmika: Again

source/base/Promise.ooc
marcusnaslund: It doesn't work otherwise for some reason.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,162 @@
...
 }
 
 ThreadPool: class {
+	_globalMutex := Mutex new()
 	_threads: Thread[]
-	_jobs := BlockedQueue<ThreadJob> new()
-	_mutex := Mutex new()
-	_allFinishedCondition := WaitCondition new()
-	_activeJobs: Int
+	_threadMutexes: Mutex[]
```
simonmika: This _must_ be combined into a worker. This creates a lot of messy code, which may contain errors.

source/base/ThreadPool.ooc
marcusnaslund: Same answer.

source/base/ThreadPool.ooc
```diff
@@ -1,111 +1,162 @@
...
+			this _threadMutexes[i] destroy()
+		}
+		this _threads free()
+		this _threadMutexes free()
+		this _tasks free()
+		this _globalMutex destroy()
 		super()
 	}
-	threadLoop: func {
+	_threadLoop: func (identifier: Int) {
```
simonmika: Don't hand over index. Hand over each workers context instead.

source/base/Promise.ooc
davidhesselbom: Would `(status := this wait()) ? this _result[T] : defaultValue` work? Just curious.

marcusnaslund: It does, actually. Would you prefer it?

source/base/ThreadPool.ooc
marcusnaslund: Ooc gets all confused about type if I assign it directly.

source/base/ThreadPool.ooc
emilwestergren: Could we at least profile this theory before we merge it? This looks like a pretty ugly solution to me, but if the condition allocations are even more expensive then I accept it.

marcusnaslund: We only need one WaitCondition per thread here, right?

source/base/Promise.ooc
davidhesselbom: No, not really. It saves a line, but it's also much harder to read. I suppose `(this wait()) ? this _result[T] : defaultValue` doesn't work, right?

marcusnaslund: Unfortunately it does not.

marcusnaslund:
Updated with a Worker class. Waiting for a decision on WaitCondition vs Mutex.

marcusnaslund:
We still have problems with the SynchronizedQueue test, right?

davidhesselbom:
And blockedQueueTest.

source/base/ThreadPool.ooc
simonmika: Create a small running example on upstream rock and file an issue.

davidhesselbom: You'd need to test it with upstream rock, though.

marcusnaslund:
BlockedQueue inherits from SynchronizedQueue, so that makes sense.

source/base/Promise.ooc
```diff
@@ -40,11 +40,7 @@ _ThreadPromise: class extends Promise {
 	}
```
davidhesselbom: I missed this earlier - `free: override func {`.

davidhesselbom:
Waiting for memory leak fixes. Meanwhile, I'm marking this as not ready, again.

marcusnaslund:
Done. There seems to be some problem with calling `wait()` (and/or `cancel()`) and then `free()` if there are too many concurrent threads running, but this a problem somewhere in the Thread implementation, not here. Also doesn't affect what we're doing with it at the moment.

davidhesselbom:
How many is "too many"?

marcusnaslund:
Sorry, must remember to check Magic.
I somtimes run into trouble when ther are more than 4 pthread threads.

sebastianbaginski:
> this a problem somewhere in the Thread implementation

If you know a reliable way to reproduce it, I'd like to try fixing it.

davidhesselbom:
4 sounds like a number we'll reach pretty quickly.

marcusnaslund:
Yep. There is something wrong with `cancel()` on Future, I will fix it before I flag this as ready.

davidhesselbom:
Thanks.

marcusnaslund:
`Worker` should not lock the mutex before destroying it. This should be OK now (if we can live with the same memory leaks we've had with the threadpool before?).
@sebastianbaginski - Would you please take a look at the `Worker` class? I run into trouble with `this _thread cancel()` all the time.

source/base/Promise.ooc
davidhesselbom: This is still a legit comment. Why not `status := (this _state == _PromiseState Finished)`?

marcusnaslund: Sorry, missed this.

marcusnaslund:
Updated.

sebastianbaginski:
> Would you please take a look at the Worker class whenever you have time? I run into trouble with this _thread cancel() all the time.

Sure :)
The thing is, cancelling a thread should not be a part of a normal workflow. It is like with programs you run every day, typically you don't kill them with task manager, but rather you give them a chance to exit from `main()` / event loop etc. Same with threads. Thread exits its main function, you `join` it and everything is nice and clean.
Now as I understand it, there is no way a `Worker` can exit his thread function. This should be corrected, but before that, it should be possible to cancel a waiting `BlockedQueue`.
Current workflow is like
- `Worker` calls `wait` on a WaitCondition
- someone hopefully pushes a job into queue, `Worker` wakes up and runs the job
- `Worker` waits again
- someone wants to free the `ThreadPool`, but can't really do that nicely as there are `Worker`s sleeping no a wait condition

what if (2) never happens ? You end up with a blocked WaitCondition. I may be wrong, but I think that destroying a wait condition, while threads are still waiting on it, causes your program to enter a state called `undefined behaviour`, which we don't really want.

What we want is something like
- `Worker` calls `wait` on a WaitCondition
- someone hopefully pushes a job into queue, `Worker` wakes up and runs the job
- `Worker` waits again
- someone wants to free the `ThreadPool`, so he calls `cancel` on a job queue
- `cancel` wakes all `Worker` threads waiting on a job queue
- `Worker` notices that job queue has been cancelled, so it exits the thread function
- `ThreadPool` joins all worker threads (or `wait`s on a Thread, as it is called in ooc)
- everyone lives happily ever after

One more thing about the `pthread_cancel`. Cancelling is not an immediate operation, it merely pushes a cancellation request and a thread should be given a chance to receive it. There is no such chance while thread is sleeping on a wait condition. Cancelling is possible only when thread enters so called *cancellation point*, at which OS checks all pending cancellation requests and terminates the thread if there is a pending request. There are plenty of possible cancellation points and it is very easy to find that in the linux manual, so you can read more about it if you are interested.

sebastianbaginski:
So I have a little patch that implements the idea described above, you can just apply it to your branch and it should work without the need to cancel a thread.

```
diff --git a/source/base/ThreadPool.ooc b/source/base/ThreadPool.ooc
index 8534a7a..de39094 100644
--- a/source/base/ThreadPool.ooc
+++ b/source/base/ThreadPool.ooc
@@ -114,7 +114,7 @@ Worker: class {
 		this _thread start()
 	}
 	free: override func {
-		//FIXME this _thread cancel() // Should be here to prevent memory leaks, but causes segfaults
+		this _thread wait()
 		this _thread free()
 		(this _threadClosure as Closure) dispose()
 		this _mutex destroy()
@@ -122,8 +122,12 @@ Worker: class {
 	}
 	_threadLoop: func {
 		while (true) {
-			job := this _tasks wait()
-			job run(this _mutex)
+			isOk := true
+			job := this _tasks wait(isOk&)
+			if (isOk)
+				job run(this _mutex)
+			else
+				break
 		}
 	}
 }
@@ -140,8 +144,8 @@ ThreadPool: class {
 		for (i in 0 .. threadCount)
 			this _workers[i] = Worker new(this _tasks)
 	}
-	//TODO: Implement free with timeout
 	free: override func {
+		this _tasks cancel()
 		for (i in 0 .. this _threadCount) {
 			this _workers[i] free()
 		}
diff --git a/source/collections/SynchronizedQueue.ooc b/source/collections/SynchronizedQueue.ooc
index 1839d39..7035fae 100644
--- a/source/collections/SynchronizedQueue.ooc
+++ b/source/collections/SynchronizedQueue.ooc
@@ -65,6 +65,7 @@ SynchronizedQueue: class <T> extends Queue<T> {
 
 	_populated := WaitCondition new()
+	_canceled := false
 	init: func { super() }
 	free: override func {
 		this _populated free()
@@ -74,12 +75,25 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 		super(item)
 		this _populated signal()
 	}
-	wait: func -> T {
+	cancel: func {
+		this _mutex lock()
+		this _canceled = true
+		this _mutex unlock()
+		this _populated broadcast()
+	}
+	wait: func (isOk := null as Bool*) -> T {
 		result: T
 		this _mutex lock()
-		while (this empty)
+		while (this empty) {
 			this _populated wait(this _mutex)
-		this _backend dequeue(result&) as Void // Ignore the return value to avoid build warning
+			if (this _canceled) {
+				if (isOk)
+					isOk@ = false
+				break
+			}
+		}
+		if (!this _canceled)
+			this _backend dequeue(result&) as Void // Ignore the return value to avoid build warning
 		this _mutex unlock()
 		result
 	}

```

I'm not really experienced with git, but if I can figure out how to make a pull request into your branch I will do that instead of sending the patch file.

marcusnaslund:
That makes a lot of sense @sebastianbaginski - many thanks.
SynchronizedQueue failed again, but I get no errors here.

davidhesselbom:
So... status? Are we all done here?

emilwestergren:
Do we have any profiling? It would suck to merge this if it makes the performance worse while I'm chasing milliseconds to optimize...

davidhesselbom:
> Do we have any profiling?

Crit is legit. Does anyone have time to profile this? Is there a written guide anywhere that describes how to profile it?

marcusnaslund:
@simonmika @emilwestergren I think I heard you discussing this during lunch? What's the verdict?

simonmika:
Merge now, fix later.

marcusnaslund:
I traced the "problems with >4 concurrent threads" to calling `wait()` or `cancel()` on the thread if it already has completed (and thus through `_thread wait()` called `pthread_join()`). I don't know why it worked if there were only a few threads.
Anyway, the new update takes care of this, so the bug seems to be gone.

PR #439, opened 2015-09-25
FloatMatrix with bracket operators

marcusnaslund:
FloatMatrix now supports [] and []= instead of get/set
Also, the exceptions are no longer used and thus removed.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -51,8 +51,8 @@ FloatMatrix : cover {
 	// <param name="x">Column number of a matrix.</param>
 	// <param name="y">Row number of a matrix.</param>
 	// <returns></returns>
-	get: func@ (x, y: Int) -> Float { this elements[x + this width * y] } //TODO Deprecated, remove when no longer used
-	operator [] (x, y: Int) -> Float { this elements[x + dimensions width * y] }
+	get: func@ (x, y: Int) -> Float { this[x, y] } //TODO Deprecated, remove when no longer used
+	operator [] (x, y: Int) -> Float { this elements[x + this width * y] }
 
 	// <summary>
 	// Set an element in a matrix at position(x,y).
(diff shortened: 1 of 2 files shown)


source/math/FloatMatrix.ooc
```diff
@@ -60,7 +61,8 @@ FloatMatrix : cover {
 	// <param name="y">Row number of a matrix.</param>
 	// <param name="value">The value set at (x,y).</param>
 	// <returns></returns>
-	set: func@ (x, y: Int, value: Float) { this elements[x + this width * y] = value }
+	set: func@ (x, y: Int, value: Float) { this elements[x + this width * y] = value } //TODO Deprecated, remove when no longer used
+	operator []= (x, y: Int, value: Float) -> Float { this elements[x + dimensions width * y] = value }
```
davidhesselbom: This shouldn't return anything.

source/math/FloatMatrix.ooc
```diff
@@ -51,7 +51,8 @@ FloatMatrix : cover {
 	// <param name="x">Column number of a matrix.</param>
 	// <param name="y">Row number of a matrix.</param>
 	// <returns></returns>
-	get: func@ (x, y: Int) -> Float { this elements[x + this width * y] }
+	get: func@ (x, y: Int) -> Float { this elements[x + this width * y] } //TODO Deprecated, remove when no longer used
+	operator [] (x, y: Int) -> Float { this elements[x + dimensions width * y] }
```
davidhesselbom: Why not use `this width`, as above, instead of `dimensions width` (which, by the way, should be preceded by `this `)?

Actually, there's no point in implementing the same function twice. Make the operators call the `get` and `set` functions or vice versa, and when we've updated all callers to use the operators, remove the `get` and `set` functions entirely.

source/math/FloatMatrix.ooc
```diff
@@ -260,10 +260,8 @@ FloatMatrix : cover {
 				value := upper elements[y + y * upper width]
 				if (value != 0)
 					result elements[x + y * result width] = accumulator / value
-				else {
-					// TODO: What do we do about this?
-					/*DivisionByZeroException new() throw()*/
-				}
+				else
+					raise("Division by zero")
```
davidhesselbom: `raise("Division by zero in FloatMatrix backwardSubstitution")` would be more helpful...

davidhesselbom:
Merge https://github.com/marcusnaslund/ooc-kean/pull/1 first.

PR #441, opened 2015-09-28
Removed unix specific wait condition from GLFence

sebastianbaginski:


davidhesselbom:
Everything *seems* to work.

sebastianbaginski:
As long as 
```ooc
create: func -> T {
    T new()
}
x := create()
```
gives you the same type of object as
```ooc
x := T new()
```
we should be fine :)

PR #442, opened 2015-09-28
Broke out OwnedBuffer of Text for reuse in other situations.

simonmika:


source/base/OwnedBuffer.ooc
```diff
@@ -0,0 +1,73 @@
...
+OwnedBuffer: cover {
+	_pointer: UInt8*
+	_size: Int
+	_owner: Owner
+	pointer ::= this _pointer
+	size ::= this _size
+	init: func@ {
+		this _pointer = null
+		this _size = 0
+		this _owner = Owner Unknown
```
davidhesselbom: Why not make these the default values?

source/base/OwnedBuffer.ooc
```diff
@@ -0,0 +1,73 @@
...
+		}
+		result
+	}
+	free: func@ ~withCriteria (criteria: Owner) -> Bool {
+		this _owner == criteria && this free()
+	}
+	copy: func -> This {
+		this _copyData()
+		this _owner = Owner Callee
+		this
```
davidhesselbom: This looks really scary, but that's because I'm thinking of it as a class. Interesting stuff!

davidhesselbom:
I don't think you meant to add `dataFile.txt` and `example.svg`. Amend.

davidpiuva:
I always get those files when getting the latest Kean.

simonmika:
I did indeed not. Can somebody, please fix the tests so that they always clean up their garbage.

source/base/OwnedBuffer.ooc
simonmika: We have been adding comments to help people out.

davidhesselbom:
Gimme a sec...

source/base/OwnedBuffer.ooc
simonmika: Why initialize and then overwrite immediately? Covers are not zero the way objects are. But we should be calling another constructor.

davidhesselbom:
Should be fixed in #449.

simonmika:
It turned out that `Text`did not follow the rules for freeing.

source/base/TextBuffer.ooc
```diff
@@ -0,0 +1,85 @@
...
+	operator []= (index: Int, value: Char) {
+		this raw[index] = value
+	}
+	operator []= (range: Range, data: This) {
+		data copyTo(this[range])
+	}
+	operator == (other: This) -> Bool {
+		this _backend == other _backend
+	}
+	empty: static This { get { This new() } }
```
davidhesselbom: There are many places in our codebase where `empty` is a property that returns a `Bool`. Should those be changed into `isEmpty`? Either way, `empty` is a an adjective and a verb. Should it be obvious that
- `This empty` returns an "empty" instance of something,
- `this empty` returns a bool stating whether an instance is "empty", and
- `this empty()` "empties" an instance?

simonmika: Is this relevant to this PR? Lets discuss the `empty` / `isEmpty` property outside the scope of this PR.

davidhesselbom: It is relevant if we conclude that name of this property should change, but alright.

davidhesselbom:
See #475.

source/base/OwnedBuffer.ooc
```diff
@@ -0,0 +1,106 @@
...
+import math
+
+OwnedBuffer: cover {
+	_pointer: UInt8*
+	_size: Int
+	_owner: Owner
+	pointer ::= this _pointer
+	size ::= this _size
+	owner ::= this _owner
+	isOwned ::= this _owner != Owner Unknown && this _owner != Owner Static && this _owner != Owner Stack && this _pointer != null
```
davidhesselbom: Would it make sense to put the part of this logic that compares `this _owner` to various `Owner` variants in its own function?

davidhesselbom:
Aside from the single-letter variable names here and there (`t`, `n`, `s`, `c`), I guess this is ready. I can't say I understand everything that goes on in here, but it looks very cool...

source/base/OwnedBuffer.ooc
simonmika: `owner` is exposed as a property as well.

simonmika:
@sebastianbaginski can you have a look at this code and ensure I got it right?

source/base/OwnedBuffer.ooc
```diff
@@ -0,0 +1,106 @@
...
+			this _owner = Owner Caller
+		} else
+			result = this
+		result
+	}
+	copyTo: func (destination: This) -> Int {
+		result := 0
+		if (this _pointer != null && this _size != 0 && destination _pointer != null && destination _size != 0) {
+			result = Int minimum(this _size, destination _size)
+			memcpy(destination _pointer, this _pointer, result)
```
sebastianbaginski: What if `destination` is a string literal ? I think the memory can't be written to in that case.

source/base/Text.ooc
```diff
@@ -302,5 +312,4 @@ Text: cover {
 
 operator == (left: String, right: Text) -> Bool { Text new(left) == right }
 operator != (left: String, right: Text) -> Bool { !(left == right) }
-operator != (left: Text, right: String) -> Bool { !(left == right) }
-operator != (left: Text, right: Text) -> Bool { !(left == right) }
+operator []= (left: TextBuffer, range: Range, right: Text) { right _buffer copyTo(left slice(range min, range max - range min)) }
```
sebastianbaginski: `range count`?

source/base/TextBuilder.ooc
```diff
@@ -46,71 +48,73 @@ TextBuilder: class {
 		This new(this)
 	}
 	append: func ~char (c: Char) {
-		this append(Text new(c&, 1, Owner Stack) give())
+		this append(Text new(c&, 1, Owner Stack) copy())
```
sebastianbaginski: `Text new(c)`?

source/base/TextBuilder.ooc
```diff
@@ -46,71 +48,73 @@ TextBuilder: class {
...
+		this _data add(value claim())
 	}
 	append: func (other: This) {
 		for (i in 0 .. other count)
-			this append(other[i] copy())
+			this append(other _data[i] copy())
 	}
 	prepend: func ~char (c: Char) {
-		this prepend(Text new(c&, 1, Owner Stack) give())
+		this prepend(Text new(c&, 1, Owner Stack) copy())
```
sebastianbaginski: `Text new(c)`?

PR #444, opened 2015-09-28
Fixed TimeSpan test on Windows

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/433 .


davidhesselbom:
Do you have any idea why this fixes it?

sebastianbaginski:
As I wrote in the issue, possibly because of different representations of `double`. Looks like `1.2` couldn't be represented accurately on windows, and `1.25` is easier to represent exactly (as 2^0 + 2^(-1/4)). I don't think we should spend more time on this.

PR #445, opened 2015-09-28
Added drawing of mesh

emilwestergren:
It is now possible to create a 3D mesh and draw a texture on it.

### DIFF
--- a/source/draw/gpu/GpuCanvas.ooc
+++ b/source/draw/gpu/GpuCanvas.ooc
@@ -18,7 +18,7 @@ use ooc-math
 use ooc-draw
 use ooc-base
 use ooc-collections
-import GpuImage, GpuMap, GpuSurface, GpuContext, GpuYuv420Semiplanar
+import GpuImage, GpuMap, GpuSurface, GpuContext, GpuYuv420Semiplanar, GpuMesh
 
 GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 	_target: GpuYuv420Semiplanar
@@ -54,5 +54,10 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 	clear: override func {
 		this _target y canvas clear()
 		this _target uv canvas clear()
+	}
+	draw: override func ~mesh (image: Image, mesh: GpuMesh) {
+		yuv := image as GpuYuv420Semiplanar
+		this _target y canvas draw(yuv y, mesh)
+		this _target uv canvas draw(yuv uv, mesh)
 	}
 }
--- a/source/draw/gpu/opengl/OpenGLContext.ooc
+++ b/source/draw/gpu/opengl/OpenGLContext.ooc
@@ -178,11 +178,10 @@ OpenGLContext: class extends GpuContext {
 		fence sync()
 		(result, fence)
 	}
-	createMesh: override func (positions: FloatPoint3D[], textureCoordinates: FloatPoint2D[]) {
+	createMesh: override func (vertices: FloatPoint3D[], textureCoordinates: FloatPoint2D[]) {
 		toGL := FloatTransform3D createScaling(1.0f, -1.0f, -1.0f)
-		for (i in 0 .. positions length)
-			positions[i] = toGL * positions[i]
-		OpenGLMesh new(positions, textureCoordinates, this)
+		for (i in 0 .. vertices length)
+			vertices[i] = toGL * vertices[i]
+		OpenGLMesh new(vertices, textureCoordinates, this)
 	}
-
 }
--- a/source/draw/gpu/opengl/OpenGLContext.ooc
+++ b/source/draw/gpu/opengl/OpenGLContext.ooc
@@ -19,7 +19,7 @@ use ooc-draw
 use ooc-draw-gpu
 use ooc-collections
 use ooc-ui
-import OpenGLPacked, OpenGLMonochrome, OpenGLBgr, OpenGLBgra, OpenGLUv, OpenGLFence, RecycleBin
+import OpenGLPacked, OpenGLMonochrome, OpenGLBgr, OpenGLBgra, OpenGLUv, OpenGLFence, OpenGLMesh, RecycleBin
 import Map/OpenGLMap, Map/OpenGLMapPack
 import backend/[GLContext, GLRenderer]
 
@@ -31,6 +31,8 @@ OpenGLContext: class extends GpuContext {
 	_packUv: OpenGLMapPackUv
 	_linesShader: OpenGLMapLines
 	_pointsShader: OpenGLMapPoints
+	_meshShader: OpenGLMapMesh
+	meshShader ::= this _meshShader
 	_renderer: GLRenderer
 	defaultMap: GpuMap { get { this _transformTextureMap } }
 	_recycleBin := RecycleBin new()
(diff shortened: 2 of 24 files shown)


source/draw/gpu/GpuCanvas.ooc
```diff
@@ -18,7 +18,7 @@ use ooc-math
 use ooc-draw
 use ooc-base
 use ooc-collections
-import GpuImage, GpuMap, GpuSurface, GpuContext, GpuYuv420Semiplanar
+import GpuImage, GpuMap, GpuSurface, GpuContext, GpuYuv420Semiplanar, GpuMesh
```
davidhesselbom: Why not just `use ooc-draw-gpu`?

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -55,4 +55,9 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target y canvas clear()
 		this _target uv canvas clear()
 	}
+	draw: override func ~mesh (image: Image, mesh: GpuMesh) {
```
davidhesselbom: Shouldn't `image` be a `GpuImage`?

source/draw/gpu/opengl/OpenGLContext.ooc
```diff
@@ -174,4 +178,10 @@ OpenGLContext: class extends GpuContext {
 		fence sync()
 		(result, fence)
 	}
+	createMesh: override func (positions: FloatPoint3D[], textureCoordinates: FloatPoint2D[]) {
```
davidhesselbom: Would it be wrong to call `positions` `points` instead?

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -180,26 +180,17 @@ Gles3Context: class extends GLContext {
 	}
 
 	createQuad: func -> Gles3Quad {
-		version(debugGL) { validateStart() }
```
davidhesselbom: We don't need these anymore, or what?

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -208,22 +199,12 @@ Gles3Context: class extends GLContext {
...
 		result
 	}
-	createFence: func -> Gles3Fence {
-		version(debugGL) { validateStart() }
-		result := Gles3Fence new()
-		version(debugGL) { validateEnd("Fence create") }
-		result
-	}
+	createFence: func -> Gles3Fence { Gles3Fence new() }
 	createVolumeTexture: func (width, height, depth: UInt, pixels: UInt8*) -> Gles3VolumeTexture {
```
davidhesselbom: Why not use an `IntSize3D`?

source/draw/gpu/opengl/backend/gles3/Gles3VertexArrayObject.ooc
```diff
@@ -37,36 +37,70 @@ Gles3VertexArrayObject: class extends GLVertexArrayObject {
...
+		for (i in 0 .. vertexCount) {
+			//Positions
+			packedArray[floatsPerVertex * i + 0] = positions[i] x
+			packedArray[floatsPerVertex * i + 1] = positions[i] y
+			packedArray[floatsPerVertex * i + 2] = positions[i] z
+			//Texture coordinates
+			packedArray[floatsPerVertex * i + 3] = textureCoordinates[i] x
+			packedArray[floatsPerVertex * i + 4] = textureCoordinates[i] y
+		}
+		this _generate(packedArray[0]&, 3, vertexCount)
```
davidhesselbom: Plenty of duplicate code, here... Maybe not much you can do about it though.

source/draw/gpu/opengl/backend/gles3/Gles3VertexArrayObject.ooc
```diff
@@ -37,36 +37,70 @@ Gles3VertexArrayObject: class extends GLVertexArrayObject {
...
+	}
+	_generate: func (packedArray: Float*, dimensions: Int, vertexCount: Int) {
```
davidhesselbom: `dimensions, vertexCount: Int`

source/draw/gpu/GpuCanvas.ooc
emilwestergren: Because use-files are for imports outside of the module.

source/draw/gpu/GpuCanvas.ooc
emilwestergren: We might want to support RasterImages. In this case it's just a matter of creating a temporary GpuImage if the input image is a RasterImage. Since this function is (and probably will always be) a special case only used in the rolling shutter distortion correction I've been lazy regarding these matters. It's up to you if you think we should raise exceptions etc.

davidhesselbom: Okay, if we want to possibly be able to call this function with a `RasterImage`, fine.

source/draw/gpu/opengl/OpenGLContext.ooc
emilwestergren: I think in the context of computer graphics yes, *points* would probably create more confusion than clarity. Maybe *vertices* instead?

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
emilwestergren: No, they should only be wrapped around true OpenGL calls (i.e. with the prefix *gl*).

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
emilwestergren: Makes sense, I can change it.

source/draw/gpu/GpuCanvas.ooc
davidhesselbom: I wasn't aware of that. Thanks.

source/draw/gpu/opengl/OpenGLContext.ooc
davidhesselbom: I think I'd prefer that, yes.

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
davidhesselbom: Alright.

source/draw/gpu/opengl/backend/gles3/Gles3VertexArrayObject.ooc
emilwestergren: I know... I will remove the first constructor (the one with the C pointers) but I need an index operator for the FloatPoint classes to replace all this with a single loop (i.e. point[0] == point x). What do you think?

davidhesselbom: You'd still need one function fror `FloatPoint3D[]` and one for `FloatPoint2D` though, wouldn't you?

emilwestergren: Yes, but the for-loop could be generalized.

emilwestergren:
Updated.

Issue #448, opened 2015-09-28
FloatMatrix index operators should use bounds checks


davidhesselbom:
In `Vector.ooc`

```ooc
	version (safe) {
		if (index >= this capacity || index < 0)
			raise("Accessing Vector index out of range in set operator")
	}
```

Do the corresponding thing for `FloatMatrix`.

marcusnaslund:
Since I'm already doing some stuff on FloatMatrix, I'll fix it.

PR #452, opened 2015-09-28
Floatmatrix API update

marcusnaslund:
Removes get, set, dispose
Replaces isNull() with isNull
Adds safety checks for [], []= and swapRows ( #448 )

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -51,8 +51,13 @@ FloatMatrix : cover {
 	// <param name="x">Column number of a matrix.</param>
 	// <param name="y">Row number of a matrix.</param>
 	// <returns></returns>
-	get: func@ (x, y: Int) -> Float { this[x, y] } //TODO Deprecated, remove when no longer used
-	operator [] (x, y: Int) -> Float { this elements[x + y * this width] }
+	operator [] (x, y: Int) -> Float {
+		version (safe) {
+			if (x + y * this width >= this width * this height || x + y * this width < 0)
+				raise("Accessing Vector index out of range in get operator")
+		}
+		this elements[x + y * this width]
+	}
 	// NOTE: Because rock doesn't understand the concept of inline functions,
 	// this function has been inlined manually in many places in this file for performance reasons.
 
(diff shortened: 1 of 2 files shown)


source/math/FloatMatrix.ooc
```diff
@@ -51,8 +51,13 @@ FloatMatrix : cover {
 	// <param name="x">Column number of a matrix.</param>
 	// <param name="y">Row number of a matrix.</param>
 	// <returns></returns>
-	get: func@ (x, y: Int) -> Float { this[x, y] } //TODO Deprecated, remove when no longer used
-	operator [] (x, y: Int) -> Float { this elements[x + y * this width] }
+	operator [] (x, y: Int) -> Float {
+		version (safe) {
+			if (x + y * this width >= this width * this height || x + y * this width < 0)
```
davidhesselbom: This doesn't cover the case where both `x` and `y` are negative or only one of `x` or `y` exceed their respective bounds.

marcusnaslund: Good catch and it makes the code nicer, so will fix.
So the **safe** flag is concerned also with incorrect mathematical operations, not just to prevent invalid memory read/writes?

davidhesselbom: Right, it's so you won't have to scratch your head wondering why things aren't working and Valgrind doesn't tell you about invalid reads/writes.

test/math/FloatMatrixTest.ooc
```diff
@@ -4,8 +4,9 @@ import math
 import lang/IO
 
 FloatMatrixTest: class extends Fixture {
-matrix := FloatMatrix new (3, 3)
-matrixNonSquare := FloatMatrix new (IntSize2D new(2, 3))
+	matrix := FloatMatrix new (3, 3)
+	matrixNonSquare := FloatMatrix new (IntSize2D new(2, 3))
+	matrixNull := FloatMatrix new(0, 0)
```
davidhesselbom: These names are weird. Try `nonSquareMatrix` and `nullMatrix`.

marcusnaslund: Fixed.

PR #454, opened 2015-09-29
Added get and set columns in FloatMatrix.

tomasmahlberg:
Will use these in #425.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -45,30 +45,23 @@ FloatMatrix : cover {
 		result
 	}
 
-	setColumn: func (x: Int, values: ...) {
-		if (x < 0 || x >= this width)
-			raise("Column index out of range in FloatMatrix setColumn")
-		if (values count != this height)
-			raise("Number of values must match matrix height in FloatMatrix setColumn")
-		y := 0
-		values each (|value|
-			match value {
-				case floatValue: Float => this[x, y] = floatValue
-				case => raise("Values must be of type Float in FloatMatrix setColumn")
-			}
-			y += 1
-		)
-	}
-
-	setColumn: func ~floatpoint3d (x: Int, vector: FloatPoint3D) {
-		if (this height != 3)
-			raise("Matrix must be of height 3 in FloatMatrix setColumn")
-		this setColumn(x, vector x, vector y, vector z)
+	setVertical: func (xOffset, yOffset: Int, vector: FloatPoint3D) {
+		version (safe) {
+			if (xOffset < 0 || xOffset >= this width)
+				raise("Column index out of range in FloatMatrix setVertical")
+			if (this height - yOffset < 3)
+				raise("Element positions exceed matrix dimensions in FloatMatrix setVertical")
+		}
+		this[xOffset, yOffset] = vector x
+		this[xOffset, yOffset + 1] = vector y
+		this[xOffset, yOffset + 2] = vector z
 	}
 
 	getColumn: func (x: Int) -> This {
-		if (x < 0 || x >= this width)
-			raise("Column index out of range in FloatMatrix getColumn")
+		version (safe) {
+			if (x < 0 || x >= this width)
+				raise("Column index out of range in FloatMatrix getColumn")
+		}
 		result := This new(1, this height)
 		for (y in 0 .. this height)
 			result[0, y] = this[x, y]
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -49,9 +49,9 @@ FloatMatrix : cover {
 		version (safe) {
 			if (xOffset < 0 || xOffset >= this width)
 				raise("Column index out of range in FloatMatrix setVertical")
-			if (this height - yOffset < 3)
-				raise("Element positions exceed matrix dimensions in FloatMatrix setVertical")
 		}
+		if (this height - yOffset < 3)
+			raise("Element positions exceed matrix dimensions in FloatMatrix setVertical")
 		this[xOffset, yOffset] = vector x
 		this[xOffset, yOffset + 1] = vector y
 		this[xOffset, yOffset + 2] = vector z
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -45,6 +45,36 @@ FloatMatrix : cover {
 		result
 	}
 
+	setColumn: func (x: Int, values: ...) {
+		if (x < 0 || x >= this width)
+			raise("Column index out of range in FloatMatrix setColumn")
+		if (values count != this height)
+			raise("Number of values must match matrix height in FloatMatrix setColumn")
+		y := 0
+		values each (|value|
+			match value {
+				case floatValue: Float => this[x, y] = floatValue
+				case => raise("Values must be of type Float in FloatMatrix setColumn")
+			}
+			y += 1
+		)
+	}
+
+	setColumn: func ~floatpoint3d (x: Int, vector: FloatPoint3D) {
+		if (this height != 3)
+			raise("Matrix must be of height 3 in FloatMatrix setColumn")
+		this setColumn(x, vector x, vector y, vector z)
+	}
+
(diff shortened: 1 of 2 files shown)


source/math/FloatMatrix.ooc
```diff
@@ -45,6 +45,36 @@ FloatMatrix : cover {
 		result
 	}
 
+	setColumn: func (x: Int, values: ...) {
+		if (x < 0 || x >= this width)
+			raise("Column index out of range in FloatMatrix setColumn")
+		if (values count != this height)
+			raise("Number of values must match matrix height in FloatMatrix setColumn")
+		y := 0
+		values each (|value|
```
davidhesselbom: This is a closure, which allocates memory which will never be freed, and so it will leak memory.

source/math/FloatMatrix.ooc
```diff
@@ -45,6 +45,36 @@ FloatMatrix : cover {
 		result
 	}
 
+	setColumn: func (x: Int, values: ...) {
```
davidhesselbom: There is no way to check at compile time that the values in `values` are `Float`, and any other type of values will cause the program to crash at run time. As nice as it would be to use varargs, don't - use a `FloatMatrix` or something instead.

source/math/FloatMatrix.ooc
```diff
@@ -45,6 +45,36 @@ FloatMatrix : cover {
...
+		if (x < 0 || x >= this width)
+			raise("Column index out of range in FloatMatrix setColumn")
+		if (values count != this height)
+			raise("Number of values must match matrix height in FloatMatrix setColumn")
```
davidhesselbom: Put these checks in the version block `safe`, as elsewhere in this file.

source/math/FloatMatrix.ooc
```diff
@@ -45,6 +45,36 @@ FloatMatrix : cover {
...
+
+	setColumn: func ~floatpoint3d (x: Int, vector: FloatPoint3D) {
+		if (this height != 3)
+			raise("Matrix must be of height 3 in FloatMatrix setColumn")
+		this setColumn(x, vector x, vector y, vector z)
+	}
+
+	getColumn: func (x: Int) -> This {
+		if (x < 0 || x >= this width)
+			raise("Column index out of range in FloatMatrix getColumn")
```
davidhesselbom: This should also be in a `safe` block.

source/math/FloatMatrix.ooc
marcusnaslund: How about a general setter which takes a FloatMatrix/FloatPoint2D/etc. and (x,y) for the top-left location (for example) in the matrix and replaces the values starting from there?

source/math/FloatMatrix.ooc
```diff
@@ -45,6 +45,36 @@ FloatMatrix : cover {
...
+				case floatValue: Float => this[x, y] = floatValue
+				case => raise("Values must be of type Float in FloatMatrix setColumn")
+			}
+			y += 1
+		)
+	}
+
+	setColumn: func ~floatpoint3d (x: Int, vector: FloatPoint3D) {
+		if (this height != 3)
+			raise("Matrix must be of height 3 in FloatMatrix setColumn")
```
davidhesselbom: That seems like an unnecessary limitation... Can't we add a `y` parameter, bounds check inside a `safe` block and allow the caller to write 3 values horizontally anywhere in a matrix?

source/math/FloatMatrix.ooc
davidhesselbom: How do you mean, "general"? Generics?

marcusnaslund: General as in not specifically a column or a row, but a FloatMatrix of any size.

davidhesselbom: So... more like `setVertical`, then?

marcusnaslund: Like this:
```ooc
set(data: FloatMatrix2D, x, y: Int) {
// check x,y >= 0
// check x+data width < this width, and the same for height
for (i in 0 .. data width)
  for (j in 0 .. data height)
    this[x+i, y+j] = data[i,j]
}
```
And similar for FloatPoint2D/FloatPoint3D and so on. Setting a column or a row would mean inputting a `1xN` or a `Nx1` FloatMatrix.

davidhesselbom: Sure, I don't see why not. It'd be kinda like drawing images on a canvas. I think the loop should be changed so that we iterate over `x` in the inner loop, though (better for memory locality). It's likely going to be slower for vertical vectors (the inner loop might lead to a performance penalty when the width is 1).

We can still keep the functions that use a `FloatPoint3D`, for performance reasons.

sebastianbaginski: It should be possible to use `data height` `memcpy` calls instead of a loop and multiple operator function calls. Another thing is that memory access will be better if you switch the order of the loops (`for each row` x `for each column` is more efficient for a row-major data, as it is more likely that the continuous row data is on the same memory page).
These are minor details. Good idea, though.

davidhesselbom: @sebastianbaginski What if the width is 1, wouldn't `memcpy` be pretty slow in that case?

marcusnaslund: You're both right, this was just to show my idea.

davidhesselbom: I don't know how large the matrix typically is, so it might be a good idea to get an idea about that first.

sebastianbaginski: @davidhesselbom of course, you're right, memcpy-ing one byte isn't the most efficient operation :) should be possible to set a size threshold somehow, but maybe I'm micro-optimizing now.

marcusnaslund: Currently the only potential use of this (the only current use FloatMatrix in general?) I can find involves 2x2 and 2x1 matrices.

davidhesselbom: > memcpy-ing one byte

It'd be 4 bytes, but yeah :) 

> Currently the only potential use of this (the only current use FloatMatrix in general?) I can find involves 2x2 and 2x1 matrices.

In that case, loops seem like a better idea than memcpy.

tomasmahlberg: Let's do the copy functions etc. in another pull request when/if they are needed (right now they aren't). The point with this PR  was basically to move `_copyFloatPoint3DToFloatMatrixColumn` in #425 from `Quaternion` to `FloatMatrix` and to add a `getColumn` function.

davidhesselbom: Sure, but I'm not merging the varargs thing.

tomasmahlberg:
Updated.

davidhesselbom:
Could you add a test, as well?

tomasmahlberg:
Oh, sorry, will do!

source/math/FloatMatrix.ooc
```diff
@@ -45,6 +45,29 @@ FloatMatrix : cover {
 		result
 	}
 
+	setVertical: func (xOffset, yOffset: Int, vector: FloatPoint3D) {
+		version (safe) {
+			if (xOffset < 0 || xOffset >= this width)
+				raise("Column index out of range in FloatMatrix setVertical")
+		}
+		if (this height - yOffset < 3)
+			raise("Element positions exceed matrix dimensions in FloatMatrix setVertical")
```
davidhesselbom: I see no reason not to put this inside the `safe` block as well.

tomasmahlberg: OK. A general question though, what dictates whether a check should or shouldn't be inside a `version (safe)`? It's a bit unclear to me.

davidhesselbom: Well, it's not entirely clear to me either, but, the rationale for the `safe` block is performance - an `if` clause that needs to be run a lot will make the program run slower, and in most cases we should be able to index things correctly without having to check indices while running the program. `safe` blocks should only be active during development. Does that answer your question?

tomasmahlberg: Yes.

Issue #447, opened 2015-09-28
AppVeyor


davidhesselbom:
Now that we have Circle CI in place for running our tests in Ubuntu, and now that we are building for Windows as well, we should look into using [AppVeyor](http://www.appveyor.com/) to run our tests in Windows. It works similarly to Circle CI and is free for open source projects. I'd be happy to help, given the time to do so.

davidhesselbom:
Any progress here, @jonathanudd ? If not, unassign yourself to indicate that no one is working on it.

marcusnaslund:
This is working now, so closing it.

PR #457, opened 2015-09-29
Add limits.h header on Windows for numeric constants

sebastianbaginski:
Current version of kean can't be compiled on Windows, as `INT_MIN` and `INT_MAX` constants are defined in `<limits.h>`.


davidhesselbom:
Ah. This is why we should try to work on #447.

PR #458, opened 2015-09-29
Fixes in ImageFileTest and RasterYuv420Semiplanar

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/450
Corrected conversion function in RasterYuvSemiplanar, if source height was even, this function could write outside the result buffer.
Simplified conversion function in RasterBgra.
Fixes some minor memory leaks related to not `free`d closures.
Cleaned up `ImageFileTest`, added `expect` calls to test if the image file was created.
Added `free` calls on images in `ImageFileTest`, as on Linux there was a crash in `RasterYuvSemiplanar` destructor when image height was not even (this thing crashed sooner on Windows).

### DIFF
--- a/source/draw/RasterBgra.ooc
+++ b/source/draw/RasterBgra.ooc
@@ -37,22 +37,23 @@ RasterBgra: class extends RasterPacked {
 		result
 	}
 	apply: func ~bgr (action: Func(ColorBgr)) {
-		end := this buffer pointer as Long + this buffer size
-		rowLength := this size width * this bytesPerPixel
-		for (row: Long in this buffer pointer as Long .. end) {
-			rowEnd := row + rowLength
-			for (source: Long in row .. rowEnd) {
-				action((source as ColorBgr*)@)
-				source += 3
+		for (row in 0 .. this size height) {
+			source := this buffer pointer + row * this stride
+			for (pixel in 0 .. this size width) {
+				pixelPointer := (source + pixel * this bytesPerPixel) as ColorBgr*
+				action(pixelPointer@)
 			}
-			row += this stride - 1
 		}
 	}
 	apply: func ~yuv (action: Func(ColorYuv)) {
-		this apply(ColorConvert fromBgr(action))
+		convert := ColorConvert fromBgr(action)
+		this apply(convert)
+		(convert as Closure) dispose()
 	}
 	apply: func ~monochrome (action: Func(ColorMonochrome)) {
-		this apply(ColorConvert fromBgr(action))
+		convert := ColorConvert fromBgr(action)
+		this apply(convert)
+		(convert as Closure) dispose()
 	}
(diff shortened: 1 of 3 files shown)


source/draw/RasterBgra.ooc
```diff
@@ -37,22 +37,23 @@ RasterBgra: class extends RasterPacked {
 		result
 	}
 	apply: func ~bgr (action: Func(ColorBgr)) {
-		end := this buffer pointer as Long + this buffer size
-		rowLength := this size width * this bytesPerPixel
-		for (row: Long in this buffer pointer as Long .. end) {
-			rowEnd := row + rowLength
-			for (source: Long in row .. rowEnd) {
-				action((source as ColorBgr*)@)
-				source += 3
```
sebastianbaginski: I believe this was not correct in case of bgra images, as each pixel has `4` components, not `3`. It uses `this bytesPerPixel` now.

davidhesselbom: I think the idea was that `source` was incremented automatically by the `for` loop, hence `+3`.

source/draw/RasterBgra.ooc
```diff
@@ -37,22 +37,23 @@ RasterBgra: class extends RasterPacked {
...
+		for (row in 0 .. this size height) {
+			source := this buffer pointer + row * this stride
```
davidhesselbom: `row * this stride` should be the same on each row, but whatever.

source/draw/RasterBgra.ooc
```diff
@@ -37,22 +37,23 @@ RasterBgra: class extends RasterPacked {
...
+				action(pixelPointer@)
 			}
-			row += this stride - 1
 		}
 	}
 	apply: func ~yuv (action: Func(ColorYuv)) {
-		this apply(ColorConvert fromBgr(action))
+		convert := ColorConvert fromBgr(action)
+		this apply(convert)
+		(convert as Closure) dispose()
```
davidhesselbom: Hm, why was this change necessary?

sebastianbaginski: *valgrind* reported a leak here. Very small, but still.

source/draw/RasterBgra.ooc
sebastianbaginski: Oh yeah, you're right :+1: 
Too many casts anyway, imho.

source/draw/RasterBgra.ooc
davidhesselbom: Weird. Did the old version a closure implicitly, I wonder?

source/draw/RasterBgra.ooc
davidhesselbom: Agreed, this looks much better now.

PR #462, opened 2015-09-29
Added VectorList iterator

thomasfanell:
also removed some commented out test-code...

### DIFF
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -158,4 +158,18 @@ VectorList: class <T> {
 	getSliceInto: func ~indices (start, end: Int, buffer: This<T>) {
 		this getSliceInto(start .. end, buffer)
 	}
+	iterator: func -> Iterator<T> { _VectorListIterator<T> new(this) }
+}
+
+_VectorListIterator: class <T> extends Iterator<T> {
+	_backend: VectorList<T>
+	_position: UInt
+	init: func (=_backend)
+	hasNext?: func -> Bool { this _position < this _backend count }
+	next: func -> T {
+		result := this _backend[this _position]
+		this _position += 1
+		result
+	}
+	remove: func -> Bool { false }
 }
--- a/test/collections/VectorListTest.ooc
+++ b/test/collections/VectorListTest.ooc
@@ -179,13 +179,6 @@ VectorTest: class extends Fixture {
 			expect(newList[2], is equal to("3"))
 			list free(); newList free()
 		})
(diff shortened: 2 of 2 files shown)


sebastianbaginski:
This is nice.
Are you planning to write other types of iterators (bidirectional, random access, output iterators, inserters etc.) in the future ?
Are we switching to the C++ way of writing algorithms for *abstract ranges* instead of *concrete containers* ? I'd be so happy :joy:

davidhesselbom:
How are we going to solve the problem with iterators not being freed?

davidhesselbom:
@sebastianbaginski Some of those already exist in `Iterators.ooc`. The problem has always been that the iterator is a class and the instance needs to be freed after use.

thomasfanell:
@davidhesselbom by first obtaining the iterator manually, like in test #2. I makes it less convenient, but I could live with that. I am not sure how hard it would be to have rock inject a call to ```free()```. I'm sure there are non-obvious cases to consider with that.

davidhesselbom:
And what do we gain by using an iterator instead of a `for` loop?

davidhesselbom:
I'm serious, I'm trying to keep up with what's going on here.

sebastianbaginski:
If every container exposes the same `Iterator` interface, then it is possible to write generalized algorithms that operate on iterators (ranges) instead of containers themselves. You can for example write a `sort` method that take a `start` and `end` iterators and sorts everything in between. So you don't have to rewrite the `sort` for every new container or type, you only implement the correct iterators for it and every algorithm you have ever written (for the abstract iterator interface) will just work. This is how the C++ standard library (the STL) is designed.
I'm not sure if this answers your question.
If I understand correctly, problem seems to be that in ooc iterators are heavy-weight classes that needs to be explicitly free'd, while in C++ they are typically very lightweight with almost no additional performance penalty at runtime.

davidhesselbom:
Right. There is a cover called `VarArgsIterator`, however.

davidhesselbom:
And thanks for the explanation.

thomasfanell:
@davidhesselbom we're going to have a look at the iterator API and see how to handle ownership. I'm not too familiar with how that would work yet though, I think @sebastianbaginski did something similar with ```Text```?

But for now, the only way is to manually obtain the iterator and then to manually free it.

sebastianbaginski:
@thomasfanell I think you should ask @simonmika about it, as he has done some major refactoring to the `Owner` code.

PR #466, opened 2015-09-29
Expect Count

simonmika:
Added output of number of expect to fail.


davidhesselbom:
@thomasfanell I seem to remember discussing this with you, and you said you were replacing the whole test framework and didn't want to spend time on this. That shouldn't stop me from merging this, though, right?

thomasfanell:
Nope, go right ahead.

PR #471, opened 2015-09-30
Fixed some memory leaks in File class

sebastianbaginski:
These things keeps popping up in *valgrind* when I'm testing, I just wanted to get rid of them.

### DIFF
--- a/source/sdk/io/File.ooc
+++ b/source/sdk/io/File.ooc
@@ -251,6 +251,7 @@ File: abstract class {
         p := parent
         if (p) {
             p mkdirs(mode)
+            p free()
         }
         mkdir()
     }
@@ -466,7 +467,9 @@ File: abstract class {
      * @param dstFile the file to copy to
      */
     copyTo: func(dstFile: This) {
-        dstFile parent mkdirs()
+        dstParent := dstFile parent
+        dstParent mkdirs()
+        dstParent free()
 
         src := FileReader new(this)
         dst := FileWriter new(dstFile)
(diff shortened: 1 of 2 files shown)


davidhesselbom:
I haven't spotted these before. When do they occur?

davidhesselbom:
I mean, I probably haven't used the functions that you fixed now.

sebastianbaginski:
I think it was when I ran `RasterMonochrome` or something related to drawing, I got a leak related to `mkdirs` function.

sebastianbaginski:
`copyTo` is not used anywhere, but this leak was so obvious, it just hurts my eyes.

davidhesselbom:
I'm very happy to see memory leaks fixed, but I'm a little scared to merge something without testing it... And I've been completely ignoring memory leaks originating from the test suite.

sebastianbaginski:
Sure, I can write a test case for `File` class.

sebastianbaginski:
> And I've been completely ignoring memory leaks originating from the test suite.

I know, me too. Actually I think it would be better to not allow leaks in tests, I will do that in my tests from now on.

davidhesselbom:
> Actually I think it would be better to not allow leaks in tests

`ooc-unit` will need a serious overhaul then, I'm afraid, but the leaks from there *could* be added to a valgrind suppression file.

sebastianbaginski:
I have added tests for methods changed in this pull request.
I guess `FileTest` should be extended, but this should do for now.
It is ready.

PR #477, opened 2015-09-30
Updated test script to generate a single binary

thomasfanell:
* The test script now generates a use-file with all specified tests, which rock compiles into a single binary
* Modified Fixture to set a global flag if one or more tests failed
* Tests that are not extending Fixture now prints a TODO about it. (this is not auto-generated)
* Modified .gitignore to ignore the generated use file

There are no changes to the actual usage of ```test.sh``` from a user point of view.


sebastianbaginski:
Can we still generate one binary for each test case ? I like it for debugging.

thomasfanell:
No, but maybe we could add a script for that? @simonmika @davidhesselbom?

test/base/ByteBufferTest.ooc
```diff
@@ -1,7 +1,3 @@
 use ooc-base
 
-b := ByteBuffer new(4000)
-b referenceCount decrease()
-b = ByteBuffer new(4000)
-b referenceCount decrease()
-ByteBuffer clean()
+"ByteBufferTest [TODO: Not implemented as a fixture!]" printfln()
```
davidhesselbom: `printfln`?

test/base/DebugPrintTest.ooc
```diff
@@ -24,3 +24,5 @@ version(debugTests) {
 
 	Debug print("TEST2", 3)
 }
+
+"DebugPrintTest [TODO: Not implemented as a fixture!]"
```
davidhesselbom: Is this supposed to print something?

thomasfanell:
@sebastianbaginski you can still select a single file though, same as before. ```./test.sh io/CsvFileWriter```

test/base/TimerTest.ooc
```diff
@@ -58,3 +58,5 @@ t stop()
 t start()
 for (i in 0 .. 8) { }
 t stop()
+
+"TimerTest [TODO: Not implemented as a fixture!]"
```
davidhesselbom: And this?

sebastianbaginski:
@thomasfanell That's great, thanks !

test/base/DebugPrintTest.ooc
thomasfanell: Oh shoot, yes. Forgot the printfln(). :(

thomasfanell:
Sorry for the flag toggling. There might be a quirk I have to sort out.

PR #480, opened 2015-10-01
Disabled upload optimization due to bad performance

emilwestergren:


source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -38,7 +38,8 @@ AndroidContext: class extends OpenGLContext {
 	}
 	createGpuImage: override func (rasterImage: RasterImage) -> GpuImage {
 		result: GpuImage
-		if (rasterImage instanceOf?(GraphicBufferYuv420Semiplanar)) {
+		//This optimization is currently disabled since it's only faster for 4k resolution
+		if (false && rasterImage instanceOf?(GraphicBufferYuv420Semiplanar)) {
```
davidhesselbom: Can we do something less confusing than `if (false` here? How about a version flag or a setting, somewhere?

emilwestergren: That's why I added the comment. I could add a version flag around the entire function instead I guess.

davidhesselbom: Would making a decision based on the size of the image, right here in the function, be a bad idea?

emilwestergren: Yes, I thought of that too, but the decision will be very dependent of the platform so it's probably a bad idea.

davidhesselbom: Alright, then make it a version flag. Not sure what to name the flag, though...

emilwestergren: It seems you can't have a whole function inside a version block though.

davidhesselbom: Correct... but you can version parts of it.

emilwestergren:
Updated.

source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -38,14 +38,19 @@ AndroidContext: class extends OpenGLContext {
 	}
 	createGpuImage: override func (rasterImage: RasterImage) -> GpuImage {
 		result: GpuImage
-		if (rasterImage instanceOf?(GraphicBufferYuv420Semiplanar)) {
-			graphicBufferImage := rasterImage as GraphicBufferYuv420Semiplanar
-			rgba := graphicBufferImage toRgba(this)
-			result = this unpackBgraToYuv420Semiplanar(rgba, rasterImage size)
-			rgba free()
+		version (graphicbufferupload) {
```
davidhesselbom: @jonathanudd , what do you think of the name of this flag?

jonathanudd: Debug flags should be named with the prefix `debug` and then what's it's debugging. In this case something like `debugGraphicbufferupload` or similar.

emilwestergren: But this is not really a debugflag, more of a settings/feature flag.

davidhesselbom: Right, but it's not really a feature flag, is it?

emilwestergren: No, it's probably best described as an optimization flag.

emilwestergren:
Updated. We decided to use `opti` as prefix to optimization flags.

PR #483, opened 2015-10-01
Improved test script.

simonmika:
Removes warning when trying to remove non-existing cache and make test running depend on the result of the build step.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -35,9 +35,8 @@ do
 	fi
 done
 echo "Main: ./test/Tests.ooc" >> "$TESTS_USE_FILE"
-rm .libs/tests-linux64.*
-rock -q -lpthread --gc=off $ARGS $FLAGS $TESTS_USE_FILE 2>&1
-./Tests
+rm -f .libs/tests-linux64.*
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -35,9 +35,8 @@ do
 	fi
 done
 echo "Main: ./test/Tests.ooc" >> "$TESTS_USE_FILE"
-rm .libs/tests-linux64.*
-rock -q -lpthread --gc=off $ARGS $FLAGS $TESTS_USE_FILE 2>&1
```
davidhesselbom: What was the `2>&1` thing used for anyway?

thomasfanell: redirect stderr to stdout

simonmika: It was needed in the old script to suppress errors but I believe we don't need it anymore.

test.sh
```diff
@@ -35,9 +35,8 @@ do
...
-./Tests
+rm -f .libs/tests-linux64.*
+rock --allerrors -q -lpthread --gc=off $ARGS $FLAGS $TESTS_USE_FILE && ./Tests
```
thomasfanell: Why not -r?

simonmika: You are probably right.

PR #484, opened 2015-10-01
Fixed test script to suppress error message.

thomasfanell:
Previously, if the test script tried to remove cache files that does not exist, we got an error message. I added the ```-f``` flag to the command to suppress it.


davidhesselbom:
Already done in #483 . Closing...

PR #485, opened 2015-10-01
Fixed tests for TextLiteral

sebastianbaginski:
Updated to match latest changes in Owner and Text code


davidhesselbom:
Thanks. We actually have `t""` support now though.

PR #486, opened 2015-10-01
Imroved Code integration.

simonmika:
Two minor improvements on VS Code integration.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -36,7 +36,7 @@ do
 done
 echo "Main: ./test/Tests.ooc" >> "$TESTS_USE_FILE"
 rm -f .libs/tests-linux64.*
-rock --allerrors -q -lpthread --gc=off $ARGS $FLAGS $TESTS_USE_FILE && ./Tests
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -36,7 +36,7 @@ do
 done
 echo "Main: ./test/Tests.ooc" >> "$TESTS_USE_FILE"
 rm -f .libs/tests-linux64.*
-rock --allerrors -q -lpthread --gc=off $ARGS $FLAGS $TESTS_USE_FILE && ./Tests
+rock -q -lpthread --gc=off $ARGS $FLAGS $TESTS_USE_FILE && ./Tests
```
davidhesselbom: What if I'm not using Code?

simonmika: That is actually why I went for not displaying all errors. It is not that help full then.

simonmika:
There is trouble with the `ThreadPool`, nothing with this PR.

test.sh
simonmika: You can always call `./test.sh all --allerrors`.

PR #487, opened 2015-10-01
Test for ReferenceCounter

sebastianbaginski:
ReferenceCounter test implemented as a Fixture.

### DIFF
--- a/test/base/ReferenceCounterTest.ooc
+++ b/test/base/ReferenceCounterTest.ooc
@@ -1,11 +1,69 @@
+use ooc-unit
 use ooc-base
+import threading/Thread
 
-Thing: class {
-	init: func
+TestObject: class {
+	_isAlive: Bool*
+	init: func (=_isAlive) {
+		this _isAlive@ = true
+	}
+	free: override func {
+		this _isAlive@ = false
+		super()
+	}
 }
 
-t := Thing new()
-r := ReferenceCounter new(t)
-r update(-1)
+ReferenceCounterTest: class extends Fixture {
+	init: func {
+		super("ReferenceCounter")
+		this add("single thread", func {
+			isAlive: Bool
+			object := TestObject new(isAlive&)
+			expect(isAlive)
+			counter := ReferenceCounter new(object)
+			for (i in 0 .. 10)
+				counter increase()
+			for (i in 0 .. 10)
+				counter decrease()
+			expect(isAlive, is false)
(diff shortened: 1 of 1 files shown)


test/base/ReferenceCounterTest.ooc
```diff
@@ -1,11 +1,69 @@
...
+		(job as Closure) dispose()
+		threads free()
+		counter free()
+	}
+}
 
-"ReferenceCounterTest [TODO: Not implemented as a fixture!]" printfln()
+test := ReferenceCounterTest new()
+test run()
+test free()
```
thomasfanell: Nice, are you planning to fix so all tests are free'd like this?

test/base/ReferenceCounterTest.ooc
```diff
@@ -1,11 +1,69 @@
...
+			for (i in 0 .. countPerThread) {
+				counter increase()
+				Thread yield()
+				counter decrease()
+				Thread yield()
+			}
+		}
+		counter increase()
+		for (i in 0 .. numberOfThreads) {
+			threads[i] = Thread new(|| job())
```
davidhesselbom: Cool, but this will cause a memory leak, I think. Would `Thread new(job)` work?

test/base/ReferenceCounterTest.ooc
sebastianbaginski: We (me and @davidhesselbom ) agreed that we should not create memory leaks if they can be avoided, it makes it a lot easier to use *valgrind* on the test binary. I will try to write test like that from now on. I don't know about fixing all existing tests, though.
And I know that current implementation of Fixture doesn't really care about memory leaks, but since you are planning to rewrite it I guess it's not a problem.

test/base/ReferenceCounterTest.ooc
sebastianbaginski: Really ? I didn't know that. 
And just a second ago I wrote a comment about avoiding leaks :)

davidhesselbom: Yes, that's why I pointed it out. `|| job()` will create a closure that calls `job()` and that's all it does. So why not just give the closure `job` to the thread instead of wrapping it inside another closure? Is it because you can't, because it doesn't work?

sebastianbaginski: It worked, thanks

PR #488, opened 2015-10-01
FloatMatrix with OwnedBuffer backend.

simonmika:
Added a OwnedBuffer backend to FloatMatrix so that it can use the notion of ownership to simplify freeing.


simonmika:
@tomasmahlberg: we should try using this together for a start and see how it works in real usage.

tomasmahlberg:
Alright. This looks nice.

tomasmahlberg:
I've been experimenting a bit with this. Let's say we want to do the following (assuming `x` and `y` are two square `FloatMatrix`'s):
```ooc
for (i in 0 .. x width)
	x += 3.0f * y getColumn(i) * y getColumn(x width - 1 - i) transpose() + y
```
This will generate six memory leaks per iteration and would, if I understood this correctly, have to be rewritten as:
```ooc
for (i in 0 .. x width)
	x += ((3.0f * (y getColumn(i) give() * y getColumn(x width - 1 - i) give() transpose() give()) give() + y) give()) give()
```
This still looks a bit messy and should probably be split up onto more than one line. Would it make sense to instead let `give()` be the default behaviour in `FloatMatrix`'s methods, i.e. to have them return `result give()` instead of `result`? What consequences would this have?

simonmika:
Thats what I want to look at.

davidhesselbom:
> Would it make sense to instead let give() be the default behaviour in FloatMatrix's methods, i.e. to have them return result give() instead of result?

That sounds a lot like doing things in-place instead of making `FloatMatrix` immutable, which I what I suggested earlier...

davidhesselbom:
Anyway, is this still ready to be merged, despite the messy syntax?

simonmika:
An update to the syntax is coming. @tomasmahlberg can you do a PR on my branch?

tomasmahlberg:
There's still the problem with free being called multiple times in operator methods when instantiating a `FloatMatrix` with owner `Owner Callee` instead of `Owner Caller` as I have not had time to work on this yet. Merge this PR as is, and fix this later?

simonmika:
Better not to merge before fixing as it, most likely, will break existing code.

source/math/FloatMatrix.ooc
```diff
@@ -219,44 +224,27 @@ FloatMatrix : cover {
...
-				temp1 free()
-				temp2 free()
-				lup[0] free()
-				lup[1] free()
-				lup[2] free()
-				lup free()
+				outerProduct := (this transpose() * this) take()
+				(l, u, p) := outerProduct lupDecomposition()
+				result = (p * this transpose() * y) forwardSubstitution(l) backwardSubstitution(u)
+				outerProduct free()
```
davidhesselbom: Wow, nice!

source/math/FloatMatrix.ooc
```diff
@@ -24,25 +25,26 @@ FloatMatrix : cover {
 	dimensions ::= this _dimensions
 	width ::= this _dimensions width
 	height ::= this _dimensions height
-	elements: Float[]
+	_elements: OwnedBuffer
+	elements ::= this _elements pointer as Float*
```
davidhesselbom: So... a downside to doing things this way, that I just discovered, is that the `get` function that this syntactic sugar creates is called *every time* you access `this elements[]`.

This class has been littered with `this elements[y + x * this height]` in place of `this[x, y]` for performance reasons (by inlining the operator `[]` function and thereby accessing the array directly) and this change just nullified that optimization. Now it looks bad for no good reason, and `OwnedBuffer__OwnedBuffer__getpointer__` just bumped to the 3rd most-called function in the other project. :(

source/math/FloatMatrix.ooc
```diff
@@ -123,12 +126,13 @@ FloatMatrix : cover {
...
 	// </summary>
 	// <returns>Return current matrix tranposed.</returns>
-	transpose: func@ -> This {
+	transpose: func -> This {
 		result := This new(this dimensions swap())
 		for (y in 0 .. this height)
 			for (x in 0 .. this width)
 				result elements[y + x * this height] = this elements[x + y * this width]
-		result
+		this free(Owner Callee)
```
davidhesselbom: Wait, what happens to the original as a result of this call?

tomasmahlberg: We're calling `free~withCriteria()`, so it will be freed if the ownership is set to `Owner Callee`, i.e. if the instance was not explicitly taken with `take()` or directly instantiated via a constructor which defaults owner to `Owner Caller`.
```ooc
free: func@ ~withCriteria (criteria: Owner) -> Bool {
	this _owner == criteria && this free()
}
```

davidhesselbom: So in the event that the owner is `Callee`, and the matrix is square, we don't need to allocate memory for a new one but can transpose in-place?

tomasmahlberg: I guess we could reuse the memory space in that case, yes.

davidhesselbom: Actually... it wouldn't have to be square at all. The transposed matrix will need exactly as much memory as the original either way. And, this applies to any of the function in this class (if we can reuse memory, we should). More than half of all `malloc` calls in the project originate from the `FloatMatrix` constructor...

davidhesselbom: We can put it in a "pull request debt" issue, though.

source/math/FloatMatrix.ooc
tomasmahlberg: Should we try to do anything about this? @simonmika What's your take on this?

davidhesselbom: So I tried building in release mode and now the function is called more times than the next 7 or so functions combined. It's easily fixed, though - whenever you need `this elements` more than once in a function (especially before a loop), store it in a variable
```ooc
buffer := this elements
for (i in 0 .. width)
    buffer[x * i]
```
etc.

source/math/FloatMatrix.ooc
simonmika: I would say it belongs into a future improvement milestone, it does not make anything worse, at the moment.

davidhesselbom: No, that's true.

source/math/FloatMatrix.ooc
```diff
@@ -24,25 +25,26 @@ FloatMatrix : cover {
...
-	identity: static func@ (order: Int) -> This {
+	identity: static func (order: Int) -> This {
 		result := This new(order, order)
 		for (i in 0 .. order)
 			result elements[i + result width * i] = 1.0f
-		result
+		result give()
 	}
 
 	setVertical: func (xOffset, yOffset: Int, vector: FloatPoint3D) {
```
davidhesselbom: These checks should be in `safe` blocks.

source/math/FloatMatrix.ooc
```diff
@@ -113,9 +116,9 @@ FloatMatrix : cover {
 	// Creates a copy of the current matrix.
 	// </summary>
 	// <returns>Return a copy of the current matrix.</returns>
-	copy: func@ -> This {
-		result := This new(this dimensions)
-		memcpy(result elements data, this elements data, this dimensions area * Float size)
+	copy: func -> This {
+		result := This new(this _elements copy(), this dimensions)
+		this free(Owner Callee)
```
davidhesselbom: Okay, can someone explain this? Copy and free?

source/math/FloatMatrix.ooc
```diff
@@ -141,30 +145,31 @@ FloatMatrix : cover {
 		result := 0.0f
 		for (i in 0 .. this height)
 			result += this[i, i]
+		this free(Owner Callee)
```
davidhesselbom: Why... free?

source/math/FloatMatrix.ooc
```diff
@@ -141,30 +145,31 @@ FloatMatrix : cover {
...
+				this elements[i + second * this width] = buffer
 			}
 	}
-
-	toString: func@ -> String {
+	// TODO: DEPRECATED
+	swaprows: func@ (first, second: Int) {
+		c"FloatMatrix swaprows deprecated" println()
+		this swapRows(first, second)
+	}
```
davidhesselbom: Time to die, `swaprows`. This can safely be removed.

source/math/FloatMatrix.ooc
```diff
@@ -219,44 +224,27 @@ FloatMatrix : cover {
 	// </summary>
 	// <param name="y">The right hand column y vector of the equation system.</param>
 	// <returns>Return the least square solution to the system.</returns>
-	solve: func@ (y: This) -> This {
+	solve: func (y: This) -> This {
 		result: This
 		if (this width > this height)
 			raise("Invalid dimensions in FloatMatrix solve")
 		// TODO: This can probably be cleaned up...
```
davidhesselbom: I guess this can be removed, now.

source/math/FloatMatrix.ooc
simonmika: Which checks are you referring to?

source/math/FloatMatrix.ooc
simonmika: It probably is safe to change it to an either or in a later PR that focuses on in-place updates when owner is set to callee.

source/math/FloatMatrix.ooc
davidhesselbom: The ones right below this line. I guess it can wait, just noticed it now.

source/math/FloatMatrix.ooc
simonmika: Because thats the convention. All method calls and operators have to free all arguments including this that have the owner set to callee.

source/math/FloatMatrix.ooc
simonmika: Belongs in a different PR.

source/math/FloatMatrix.ooc
simonmika: What I can see there are more that can be cleaned up. Let us take it later.

source/math/FloatMatrix.ooc
davidhesselbom: It can be fixed later, I'm just wondering what's going on here, if there's something I'm missing. Why would you *ever* want to return a copy and free the original instead of just returning `this`?

source/math/FloatMatrix.ooc
davidhesselbom: No, it was added in this PR and it shouldn't have been.

source/math/FloatMatrix.ooc
davidhesselbom: Alright, I can see why that would be useful for this function.

source/math/FloatMatrix.ooc
davidhesselbom: Why fix something and leave the TODO? It here will only cause confusion, indicating there is something to be fixed here when there no longer is.

simonmika: There _are_ more things to fix.

davidhesselbom: Oh, that's what you mean. Okay then.

source/math/FloatMatrix.ooc
simonmika: `swapRows` is added in this PR not `swaprows` and since it is public I have to assume that somebody uses it.

source/math/FloatMatrix.ooc
simonmika: Because the normal thing is not to free.

source/math/FloatMatrix.ooc
davidhesselbom: By that logic, you also can't change the signature of `lupDecomposition` or `free`.

tomasmahlberg: Sorry, I  am to blame for modifying `lupDecomposition` here, but albeit being a public function the only use of it I could find was inside the class itself, in `solve`. So it _should_ be safe.

Also, what is it that need to be sorted out before this PR can be merged? It is not entirely clear from the conversation here.

davidhesselbom: I want `swaprows` removed.

PR #490, opened 2015-10-02
Updated BlockedQueue and SynchronizedQueue tests

sebastianbaginski:
`SynchronizedQueueTest` will now test the queue with cover and class, both with single and multiple threads.
`BlockedQueueTest` will now test multiple writers with single reader, and single writer with multiple readers, both with cover and a class.
It does not use a `ThreadPool` now, it has its own unit test.

### DIFF
--- a/test/collections/BlockedQueueTest.ooc
+++ b/test/collections/BlockedQueueTest.ooc
@@ -1,94 +1,92 @@
 use ooc-base
 use ooc-collections
 use ooc-unit
-import os/Time
+import threading/Thread
 
 BlockedQueueTest: class extends Fixture {
 	init: func {
-		queue := BlockedQueue<Int> new()
-
-		func1 := func { queue enqueue(123) }
-		func2 := func {
-			result: Int
-			success := queue peek(result&)
-			expect(success)
-			expect(result == 123)
-		}
-		func3 := func {
-			result: Int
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Does this seem to fix the occasional errors with these tests?

sebastianbaginski:
Seems like it, or they are just less frequent now :)

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/489

marcusnaslund:
There is something still wrong with `BlockedQueue` which shows up in stress tests, so at least this fixes something that was also wrong with the tests.

marcusnaslund:
This will conflict with #438 (but this is of course the tests we should use).

sebastianbaginski:
> There is something still wrong with BlockedQueue which shows up in stress tests

If you know how to reproduce a bug in one of the classes we use, then create an issue about it, ideally with a sample code showing how to reproduce the error. This way others will know too, and maybe someone will fix it, or provide other kind of useful input ;)

marcusnaslund:
@thomasfanell has been working on this for some time.

thomasfanell:
Actually I haven't worked on it for quite some time, I was under the impression that someone else took over a few weeks ago.

PR #491, opened 2015-10-02
Improved math test coverage + bug fixes in Math

marcusnaslund:
Adds a looooot of tests for properties and public functions in nearly all math classes.
Cleans up a few functions whose existence isn't warranted (like `static create` instead of `new`) and fixes some bugs with computations that were discovered with the new tests, for example in `Quaternion inverse`.
Removes dependency on `ArrayList` and instead uses `VectorList` for `IntBox2D` and `FloatBox2D`.


source/math/FloatBox2D.ooc
```diff
@@ -126,19 +126,24 @@ FloatBox2D: cover {
...
-		this createAround(newCenter, newSize)
+		This createAround(newCenter, newSize)
 	}
 	toString: func -> String { "#{this leftTop toString()}, #{this size toString()}" }
 	parse: static func (input: String) -> This {
 		array := input split(',')
 		This new(array[0] toFloat(), array[1] toFloat(), array[2] toFloat(), array[3] toFloat())
 	}
-	create: static func (leftTop: FloatPoint2D, size: FloatSize2D) -> This { This new(leftTop, size) }
-	create: static func ~fromFloats (left, top, width, height: Float) -> This { This new(left, top, width, height) }
```
davidhesselbom: I take it no one was using this, but it's still useful to have around, I think.

davidhesselbom: Wait, there's a constructor... nevermind.

marcusnaslund: Yep, and `create` just calls the corresponding constructor, so I have no idea why anyone would want these.

source/math/FloatSize2D.ooc
```diff
@@ -25,7 +25,7 @@ FloatSize2D: cover {
 	length ::= this norm
 	empty ::= !(this width > 0 && this height > 0)
 	norm ::= (this width squared() + this height squared()) sqrt()
-	azimuth ::= this height atan2(this width)
+	azimuth ::= atan2(this height, this width) as Float
```
davidhesselbom: Why?

source/math/FloatSize3D.ooc
```diff
@@ -34,8 +34,8 @@ FloatSize3D: cover {
 	init: func@ ~default { this init(0.0f, 0.0f, 0.0f) }
 	pNorm: func (p: Float) -> Float {
 		p == 1 ?
-		this width abs() + this height abs() :
-		(this width abs() pow(p) + this height abs() pow(p)) pow(1 / p)
+		this width abs() + this height abs() + this depth abs() :
+		(this width abs() pow(p) + this height abs() pow(p) + this depth abs() pow(p)) pow(1 / p)
```
davidhesselbom: I'm a little worried `1 / p` will return an `Int`...  can you make it `1.0f / p` instead?

source/math/FloatSize3D.ooc
```diff
@@ -46,12 +46,12 @@ FloatSize3D: cover {
 		)
 	}
 	angle: func (other: This) -> Float {
-		(this scalarProduct(other) / (this norm * other norm)) clamp(-1.0f, 1.0f) acos() * (this width * other height - this height * other width < 0.0f ? -1.0f : 1.0f)
+		(this scalarProduct(other) / (this norm * other norm)) acos() * (this width * other height - this height * other width < 0.0f ? -1.0f : 1.0f)
```
davidhesselbom: Do you know for a fact that the `clamp` isn't needed, here?

source/math/FloatSize2D.ooc
marcusnaslund: I thought this was a lot easier to read.. no?

source/math/FloatSize3D.ooc
marcusnaslund: Even if the return type is `Float`? Sure.

davidhesselbom: Yes, I'm worried that you're putting an `Int` as the argument to `pow`.

source/math/FloatSize3D.ooc
marcusnaslund: Yes, the value will always fall between those values anyway, this is a fact of linear algebra.

source/math/FloatSize3D.ooc
marcusnaslund: Done. Fixed the corresponding thing in `FloatSize2D` as well.

source/math/FloatSize3D.ooc
sebastianbaginski: And here we are in the beautiful world of computer programs, where `0.1 + 0.1 + 0.1  != 0.3` :)

source/math/IntSize3D.ooc
```diff
@@ -32,6 +32,7 @@ IntSize3D: cover {
 	minimum: func (ceiling: This) -> This { This new(Int minimum~two(this width, ceiling width), Int minimum~two(this height, ceiling height), Int minimum~two(this depth, ceiling depth)) }
 	maximum: func (floor: This) -> This { This new(Int maximum~two(this width, floor width), Int maximum~two(this height, floor height), Int maximum~two(this depth, floor depth)) }
 	clamp: func (floor, ceiling: This) -> This { This new(this width clamp(floor width, ceiling width), this height clamp(floor height, ceiling height), this depth clamp(floor depth, ceiling depth)) }
+	//FIXME: Modulo operator does not work for negative numbers
```
davidhesselbom: You mean as in `-5 % 3 == -2`, but you want it to be `== 1`?

marcusnaslund: That was probably it, yes. I see that `modulo` in `Math.ooc` does this, but the `%` operator does not.

marcusnaslund: Ah, this is standard behaviour in C, so it is to be expected.

davidhesselbom: Actually,

```ooc
import math

(-5 % 3) toString() println() //prints -2
(-5 modulo(3)) toString() println() // also prints -2
```

davidpiuva: You can write a modulo that works for negative numbers.

davidhesselbom: Yes, but `Int modulo` in `math.ooc` should do that. For some reason it doesn't seem to work for me...

marcusnaslund: Taking another look it seems that `modulo` only takes care of cases like `5 modulo(-3)`, not when the minus sign is reversed.

davidpiuva: The divisor should never be less than 1 so obviously it does not work.

marcusnaslund: Anyway, I believe the `Size` classes are only expected to have positive values? In that case this isn't "really" an issue.

davidhesselbom: > Taking another look it seems that modulo only takes care of cases like 5 modulo(-3), not when the location of the minus sign is reversed.

But it doesn't. `(-5 modulo(3))` gets me `-2`. `Int modulo` should always return a positive number... and looking at the first `if` clause in it, I think it's supposed to handle the case where the divisor is negative.

davidpiuva: I have a working modulo in my own language at home.

marcusnaslund:
Since reviewing wasn't finished, I went ahead and included three more tests.

source/math/FloatSize2D.ooc
davidhesselbom: That's true, but it makes `atan2` a global function of sorts, and we're trying to avoid that, judging from the rest of the code (`this width abs()` and so on).

source/math/FloatSize3D.ooc
davidhesselbom: Alright. If possible, I'd like to see a test added that verifies that this function will never do anything stupid, like, throw domain errors.

source/math/FloatSize3D.ooc
```diff
@@ -46,12 +46,12 @@ FloatSize3D: cover {
...
 	}
 	distance: func (other: This) -> Float { (this - other) norm }
 	round: func -> This { This new(this width round(), this height round(), this depth round()) }
 	ceiling: func -> This { This new(this width ceil(), this height ceil(), this depth ceil()) }
-	floor: func -> This { This new(this width floor(), this height floor(), this height floor()) }
+	floor: func -> This { This new(this width floor(), this height floor(), this depth floor()) }
```
davidhesselbom: Haha, oops.

source/math/FloatSize2D.ooc
marcusnaslund: Alright, will fix.

source/math/IntShell2D.ooc
```diff
@@ -21,13 +21,17 @@ import text/StringTokenizer
 import structs/ArrayList
 
 IntShell2D: cover {
-	left, right, top, bottom: Int
+	_left, _right, _top, _bottom: Int
+	left ::= this _left
+	right ::= this _right
+	top ::= this _top
+	bottom ::= this _bottom
```
davidhesselbom: Okay... why were these made into properties?

source/math/Quaternion.ooc
```diff
@@ -33,7 +33,7 @@ Quaternion: cover {
 	// z = this y
 	// w = this z
 
-	inverse ::= This new(this w, -this x, -this y, -this z)
+	inverse ::= This new(this w / (this norm * this norm), -this x / (this norm * this norm), -this y / (this norm * this norm), - this z / (this norm * this norm))
```
davidhesselbom: This... looks like it'll break things elsewhere. Are you *sure* this is the correct implementation?

marcusnaslund: Yes, see http://se.mathworks.com/help/aeroblks/quaternioninverse.html
Previously, `inverse` returned the same result as `conjugate`, which is definitely not true.

test/math/FloatBox2DTest.ooc
```diff
@@ -23,19 +24,38 @@ FloatBox2DTest: class extends Fixture {
...
+			result := box0 + box1
+			expect(result left, is equal to(1.0f) within (this precision))
+			expect(result top, is equal to(2.0f) within (this precision))
+			expect(result width, is equal to(5.0f) within (this precision))
+			expect(result height, is equal to(4.0f) within (this precision))
+			expect(result == box0 union(box1))
+		})
+		this add("subtraction, intersection", func {
+			result := box0 - box2
+			resultb := box2 - box0
```
davidhesselbom: `resultb` isn't a very good name...

source/math/IntShell2D.ooc
marcusnaslund: That was.. hm.. I.. don't know what I was thinking here. Will fix.

test/math/FloatBox2DTest.ooc
```diff
@@ -23,19 +24,38 @@ FloatBox2DTest: class extends Fixture {
...
+			expect(result width, is equal to(2.0f) within(this precision))
+			expect(result height, is equal to(2.0f) within(this precision))
+			expect(result == box0 intersection(box2))
 		})
-		this add("addition", func)
-		this add("subtraction", func)
-		this add("scalar multiplication", func)
 		this add("casts", func {
-//			FIXME: We have no integer versions of anything yet
+			intVersion := this box0 toIntBox2D()
```
davidhesselbom: `intBox` would be a better name.

test/math/FloatBox2DTest.ooc
```diff
@@ -56,6 +76,73 @@ FloatBox2DTest: class extends Fixture {
 			shrunkBox := paddedBox shrink(1.0f / 11.0f)
 			expect(box == shrunkBox, is true)
 		})
+		this add("rounding", func {
+			_round := this box3 round()
+			_ceiling := this box3 ceiling()
+			_floor := this box3 floor()
```
davidhesselbom: ... underscore? Why?

test/math/FloatBox2DTest.ooc
```diff
@@ -56,6 +76,73 @@ FloatBox2DTest: class extends Fixture {
...
+		this add("swap", func {
+			swapped := this box0 swap()
+			expect(swapped top, is equal to(this box0 left))
+			expect(swapped left, is equal to(this box0 top))
+			expect(swapped width, is equal to(this box0 height))
+			expect(swapped height, is equal to(this box0 width))
+		})
+		this add("adaptTo", func {
+			small := FloatBox2D new(1.0f, 1.0f, 1.0f, 1.0f)
+			large := FloatBox2D new(3.0f, 3.0f, 2.0f, 2.0f)
```
davidhesselbom: `smallBox`, `largeBox`

test/math/FloatBox2DTest.ooc
```diff
@@ -56,6 +76,73 @@ FloatBox2DTest: class extends Fixture {
...
+			points add(FloatPoint2D new(-1.0f, -2.0f))
+			points add(FloatPoint2D new(1.0f, 3.0f))
+			points add(FloatPoint2D new(-2.0f, 4.0f))
+			points add(FloatPoint2D new(0.0f, 0.0f))
+			points add(FloatPoint2D new(-1.0f, 3.0f))
+			box := FloatBox2D bounds(points)
+			expect(box left, is equal to(-2.0f) within(this precision))
+			expect(box top, is equal to(-2.0f) within(this precision))
+			expect(box right, is equal to(1.0f) within(this precision))
+			expect(box bottom, is equal to(4.0f) within(this precision))
```
davidhesselbom: Doesn't matter, but you *could* free `points` here if you want.

test/math/FloatMatrixTest.ooc
```diff
@@ -99,11 +100,28 @@ FloatMatrixTest: class extends Fixture {
...
+			A := createMatrix(3, 3, [1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f, 9.0f])
+			column := A getColumn(1)
+			expect(column toString() == "4.00; 5.00; 6.00; ")
+			expect(A toString() == "1.00, 4.00, 7.00; 2.00, 5.00, 8.00; 3.00, 6.00, 9.00; ")
+			A free()
+			column free()
+		})
 	}
 
 	createMatrix: func (width: Int, height: Int, values: Float[]) -> FloatMatrix {
```
davidhesselbom: `width, height: Int`, while you're at it.

source/math/Quaternion.ooc
davidhesselbom: Alright, well, any depending code will have to be updated, or things will go bad. If what you say is true, then it should be easy. Without committing anything further in `ooc-kean`,

1. Rename `inverse` here to `inverse2` or something (but don't commit)
1. Try to build the project
1. Any `inverse` call on a `Quaternion` will cause an error, so switch for `conjugate`, commit, and make another PR in the other project
1. I will merge the other PR first, and then this one.

test/math/FloatMatrixTest.ooc
```diff
@@ -123,7 +141,7 @@ FloatMatrixTest: class extends Fixture {
 		// 2 5
 		for (x in 0 .. matrix dimensions width) {
 			for (y in 0 .. matrix dimensions height) {
-				expect(matrix[x, y], is equal to(values[x * matrix dimensions height + y]))
+				expect(matrix[x, y], is equal to(values[x * matrix dimensions height + y]) within(this tolerance))
 			}
 		}
```
davidhesselbom: You need neither of these brackets, btw.

test/math/FloatPoint2DTest.ooc
```diff
@@ -102,6 +102,59 @@ FloatPoint2DTest: class extends Fixture {
...
+			_min := this point0 minimum(this point1)
+			expect(_max x, is equal to(22.221f) within(this precision))
+			expect(_max y, is equal to(13.1f) within(this precision))
+			expect(_min x, is equal to(12.221f) within(this precision))
+			expect(_min y, is equal to(-3.1f) within(this precision))
+		})
+		this add("rounding", func {
+			_round := this point1 round()
+			_ceiling := this point1 ceiling()
+			_floor := this point1 floor()
```
davidhesselbom: Again... underscore. Why?

test/math/FloatPoint2DTest.ooc
```diff
@@ -102,6 +102,59 @@ FloatPoint2DTest: class extends Fixture {
...
+			expect(_round x, is equal to(12.0f) within(this precision))
+			expect(_round y, is equal to(13.0f) within(this precision))
+			expect(_ceiling x, is equal to(13.0f) within(this precision))
+			expect(_ceiling y, is equal to(14.0f) within(this precision))
+			expect(_floor x, is equal to(12.0f) within(this precision))
+			expect(_floor y, is equal to(13.0f) within(this precision))
+		})
+		this add("p norm", func {
+			onenorm := this point0 pNorm(1)
```
davidhesselbom: `oneNorm`

test/math/FloatPoint3DTest.ooc
```diff
@@ -57,12 +58,81 @@ FloatPoint3DTest: class extends Fixture {
 //			FIXME: Equals interface
 //			expect(FloatSize2D parse(value), is equal to(this vector3))
 		})
```
davidhesselbom: Well, since you're probably right about interfaces, this commented out test could be removed.

source/math/Quaternion.ooc
marcusnaslund: Right, I can change all occurances of inverse on `Quaternion` with conjugate and we'd get code that works the same as before. But, this begs the question, is this **really** what we want? Conjugate and inverse are very different things.

test/math/FloatPoint3DTest.ooc
marcusnaslund: There are several other mentions of an equals interface, I will remove them as well.

source/math/Quaternion.ooc
tomasmahlberg: They are not _very_ different though, are they? In most cases we are working with unit quaternions in which case they are the same. I agree with the naming confusion though. Right now `conjugate` and `inverse` does the exact same thing.

test/math/FloatBox2DTest.ooc
marcusnaslund: Hmm if I recall correctly I had some trouble somewhere were I was using a variable together with a function of the same name.. but here it seems to work fine. Rock works in mysterious ways. Will fix.

source/math/Quaternion.ooc
marcusnaslund: Well as a mathematician I would say they are. :) But not if they are unit quaternions, you're right, that's a very good explanation.

source/math/IntSize3D.ooc
marcusnaslund: Should I just leave this as is, for now?

marcusnaslund:
Updated according to comments, but haven't fixed anything related to `Quaternion inverse` or `modulo` yet.

marcusnaslund:
Test timeout again. :/

thomasfanell:
Just restart the test

source/math/Quaternion.ooc
davidhesselbom: I don't know whether `inverse` or `conjugate` is truly what we want in any given situation, but for now, let's assume that everything relies on the current implementation of `inverse` being what it is and will break if we change its implementation. It's entirely possible that the correct implementation is what we want in all of those cases, but let's not guess.

source/math/IntSize3D.ooc
davidhesselbom: Well... you only added this comment and I'm not sure why. Were you planning on modifying the `fillEven` function somehow, but got stuck because of how `%` works? Anyway, `Int odd` should do the trick there (and it uses the `modulo` function, which apparently works for negative *variables*, just not for negative non-parenthesized literals).

test/math/FloatPoint2DTest.ooc
```diff
@@ -102,6 +101,59 @@ FloatPoint2DTest: class extends Fixture {
...
+			expect(this point0 scalarProduct(this point1), is equal to (230.95f) within(0.01f))
+		})
+		this add("distance", func {
+			distance := point0 distance(point1)
+			expect(distance, is equal to(19.04f) within(0.01f))
+		})
+		this add("clamp", func {
+			clamped := this point1 clamp(this point0, this point2)
+			expect(clamped x, is equal to(22.221f) within(this precision))
+			expect(clamped y, is equal to(10.0f) within(this precision))
```
davidhesselbom: Why not compare to `point0 x` and `point2 y` instead of the literal values?

test/math/FloatPoint3DTest.ooc
```diff
@@ -63,6 +64,69 @@ FloatPoint3DTest: class extends Fixture {
...
+		this add("minimum maximum", func {
+			_max := this point0 maximum(this point1)
+			_min := this point0 minimum(this point1)
+			expect(_max x, is equal to(22.0f) within(this precision))
+			expect(_max y, is equal to(13.0f) within(this precision))
+			expect(_max z, is equal to(20.0f) within(this precision))
+			expect(_min x, is equal to(12.0f) within(this precision))
+			expect(_min y, is equal to(-3.0f) within(this precision))
+			expect(_min z, is equal to(10.0f) within(this precision))
```
davidhesselbom: Again, `_`, why?

source/math/Quaternion.ooc
niklasborwell: I agree that we should fix this but I agree even more with David that we have to make sure nothing breaks when we do this. It's much better to keep code is mathematically suspicious and fix it when we have time to double check the result in a way that guarantees that nothing breaks. Otherwise it could create some bugs that are seriously difficult to find. 
The fact that we have the wrong implementation shows that we had some things mixed up when we wrote this so we don't know what we should have used in the situations where we used inverse, and in the end getting the correct result is more important than having the code mathematically correct.

test/math/FloatShell2DTest.ooc
```diff
@@ -0,0 +1,96 @@
...
+			expect(added left, is equal to(5.0f) within(tolerance))
+			expect(added right, is equal to(5.5f) within(tolerance))
+			expect(added top, is equal to(6.0f) within(tolerance))
+			expect(added bottom, is equal to(6.5f) within(tolerance))
+		})
+		this add("increase, decrease (box)", func {
+			shell := FloatShell2D new(1.0f)
+			box := FloatBox2D new(1.0f, 2.0f, 3.0f, 4.0f)
+			inc := shell increase~byBox(box)
+			dec := shell decrease~byBox(box)
```
davidhesselbom: `increased`, `decreased`

test/math/FloatSize2DTest.ooc
```diff
@@ -90,10 +90,62 @@ FloatSize2DTest: class extends Fixture {
 			expect(FloatSize2D basisX angle(-FloatSize2D basisY), is equal to(-PI as Float / 2.0f) within(this precision))
```
davidhesselbom: Not your fault, but this should be `Float pi`.

test/math/FloatSize2DTest.ooc
```diff
@@ -90,10 +90,62 @@ FloatSize2DTest: class extends Fixture {
...
+			_min := this vector0 minimum(this vector1)
+			expect(_max width, is equal to(22.221f) within(this precision))
+			expect(_max height, is equal to(13.1f) within(this precision))
+			expect(_min width, is equal to(12.221f) within(this precision))
+			expect(_min height, is equal to(-3.1f) within(this precision))
+		})
+		this add("rounding", func {
+			_round := this vector1 round()
+			_ceiling := this vector1 ceiling()
+			_floor := this vector1 floor()
```
davidhesselbom: More `_`.

test/math/FloatSize2DTest.ooc
```diff
@@ -90,10 +90,62 @@ FloatSize2DTest: class extends Fixture {
...
+			expect(interpolate2 height, is equal to(5.0f) within(0.01f))
+			expect(interpolate3 width, is equal to(this vector1 width) within(this precision))
+			expect(interpolate3 height, is equal to(this vector1 height) within(this precision))
+		})
+		this add("area, length, empty", func {
+			_empty := FloatSize2D new()
+			expect(this vector0 area, is equal to(-68.89f) within(0.01f))
+			expect(this vector0 length, is equal to(22.44f) within(0.01f))
+			expect(_empty empty, is equal to(true))
+			expect(_empty area, is equal to(0.0f) within(this precision))
```
davidhesselbom: Again.

test/math/FloatSize2DTest.ooc
```diff
@@ -90,10 +90,62 @@ FloatSize2DTest: class extends Fixture {
...
+			//expect(this vector0 empty, is equal to(false)) //FIXME This crashes Fixture
```
davidhesselbom: Eh? Any idea why?

thomasfanell: ```expect(this vector0 empty, is false)``` // this is the correct way of writing this.

test/math/FloatSize3DTest.ooc
```diff
@@ -59,15 +55,89 @@ FloatSize3DTest: class extends Fixture {
...
+			_floor := this vector4 floor()
+			expect(_round width, is equal to(10.0f) within(this precision))
+			expect(_round height, is equal to(21.0f) within(this precision))
+			expect(_round depth, is equal to(30.0f) within(this precision))
+			expect(_ceiling width, is equal to(11.0f) within(this precision))
+			expect(_ceiling height, is equal to(21.0f) within(this precision))
+			expect(_ceiling depth, is equal to(31.0f) within(this precision))
+			expect(_floor width, is equal to(10.0f) within(this precision))
+			expect(_floor height, is equal to(20.0f) within(this precision))
+			expect(_floor depth, is equal to(30.0f) within(this precision))
```
davidhesselbom: More underscores.

test/math/FloatSize3DTest.ooc
```diff
@@ -59,15 +55,89 @@ FloatSize3DTest: class extends Fixture {
...
+			first := FloatSize3D new(2.0f, -3.0f, 5.0f)
+			second := FloatSize3D new(5.0f, 3.0f, -7.0f)
+			expect(first angle(second), is equal to(2.221f) within(0.01f))
+			expect(first distance(second), is equal to(13.74f) within(0.01f))
+		})
+		this add("length, empty", func {
+			_empty := FloatSize3D new()
+			expect(this vector0 length, is equal to(24.35f) within(0.01f))
+			expect(_empty empty, is equal to(true))
+			expect(_empty volume, is equal to(0.0f) within(this precision))
```
davidhesselbom: And here.

test/math/IntBox2DTest.ooc
```diff
@@ -29,11 +30,84 @@ IntBox2DTest: class extends Fixture {
...
+			result := box0 + box1
+			expect(result left, is equal to(1))
+			expect(result top, is equal to(2))
+			expect(result width, is equal to(5))
+			expect(result height, is equal to(4))
+			expect(result == box0 union(box1))
+		})
+		this add("subtraction, intersection", func {
+			result := box0 - box2
+			resultb := box2 - box0
```
davidhesselbom: `resultb` again...

test/math/IntBox2DTest.ooc
```diff
@@ -29,11 +30,84 @@ IntBox2DTest: class extends Fixture {
...
+			expect(result == resultb)
+			expect(result top, is equal to(2))
+			expect(result left, is equal to(2))
+			expect(result width, is equal to(2))
+			expect(result height, is equal to(2))
+			expect(result == box0 intersection(box2))
+		})
 		this add("casts", func {
-//			FIXME: We have no integer versions of anything yet
+			floatVersion := this box0 toFloatBox2D()
```
davidhesselbom: `floatBox` plz

test/math/IntPoint2DTest.ooc
```diff
@@ -41,14 +39,33 @@ IntPoint2DTest: class extends Fixture {
...
+			expect(min y, is equal to(-3))
+		})
+		this add("scalar product", func {
+			product := this point0 scalarProduct(this point1)
+			expect(product, is equal to(225))
+		})
+		this add("clamp", func {
+			result := point1 clamp(point0, point2)
+			expect(result x, is equal to(22))
+			expect(result y, is equal to(10))
```
davidhesselbom: Again, you could use `point0 x` and `point2 y`.

test/math/IntRandomGeneratorTest.ooc
```diff
@@ -51,6 +52,23 @@ IntRandomGeneratorTest: class extends Fixture {
 			expect(Int absolute(mean - expectedMean) < 2)
 			expect(Int absolute(deviation - expectedDeviation) < 2)
 		})
+		this add("uniform range", func {
+			uniformGenerator := IntUniformRandomGenerator new()
+			uniformGenerator setRange(-250000, 250000)
+			uniformLowest := 250000
+			uniformHighest := -250000
```
davidhesselbom: Use `_` to group numbers by 3, like this: `-250_000`

test/math/IntShell2DTest.ooc
```diff
@@ -0,0 +1,95 @@
...
+			expect(added left, is equal to(5))
+			expect(added right, is equal to(5))
+			expect(added top, is equal to(6))
+			expect(added bottom, is equal to(6))
+		})
+		this add("increase, decrease (box)", func {
+			shell := IntShell2D new(1)
+			box := IntBox2D new(1, 2, 3, 4)
+			inc := shell increase~byBox(box)
+			dec := shell decrease~byBox(box)
```
davidhesselbom: `increased`, `decreased`.

test/math/IntSize2DTest.ooc
```diff
@@ -50,14 +47,52 @@ IntSize2DTest: class extends Fixture {
...
+		})
+		this add("polar", func {
+			sqrttwo := IntSize2D polar(1.415f, 0.785f)
+			expect(sqrttwo width, is equal to(1))
+			expect(sqrttwo height, is equal to(1))
+		})
+		this add("area, square, empty", func {
+			_rectangle := IntSize2D new(10, 20)
+			_square := IntSize2D new(5, 5)
+			_empty := IntSize2D new()
```
davidhesselbom: More underscores.

test/math/IntSize3DTest.ooc
```diff
@@ -59,6 +54,38 @@ IntSize3DTest: class extends Fixture {
...
+			expect(even width, is equal to(22))
+			expect(even height, is equal to(-3)) //FIXME Incorrect due to mod operator not correct about negative numbers
+			expect(even depth, is equal to(8))
+		})
+		this add("volume, empty", func {
+			_empty := IntSize3D new()
+			expect(_empty empty, is equal to(true))
+			expect(this vector1 empty, is equal to(false))
+			expect(this vector1 volume, is equal to(-1248))
+			expect(_empty volume, is equal to(0))
```
davidhesselbom: Underscores, again.

test/math/FloatSize2DTest.ooc
davidhesselbom: Ah. Thanks.

thomasfanell: And FYI, it crashes because Cell toString has no overload for Bool, which we probably should fix.

davidhesselbom: Oh. Yeah, we should.

marcusnaslund:
All underscores removed.

test/math/FloatSize2DTest.ooc
marcusnaslund: `PI as Float` is used in several places, I will replace all of them with `Float pi`

source/math/Quaternion.ooc
marcusnaslund: For now I agree - I could perhaps take the change out and bring it up in a separate PR/issue (?) - but in the long run it makes little sense to have, and keep using, a function we know is generally incorrect.

davidhesselbom: I'm not suggesting we keep the incorrect implementation at all &mdash; I'm asking that we *first* make sure that any code that is dependent on the current implementation of `inverse` calls `conjugate` and *then* correct the implementation of `inverse`.

source/math/FloatSize3D.ooc
marcusnaslund: This really "should" be fine, but to be absolutely sure we don't get something weird due to floating point arithmetic as @sebastianbaginski points (no pun intended) out, I will keep `clamp` here.

source/math/IntSize3D.ooc
marcusnaslund: Yes, using `Int odd` gives correct results.

marcusnaslund:
Updated. Changed the use of `Quaternion inverse` to `Quaternion conjugate`.

source/math/Quaternion.ooc
```diff
@@ -207,7 +207,7 @@ Quaternion: cover {
 		This new(realResult, imaginaryResult)
 	}
 	operator * (value: FloatPoint3D) -> FloatPoint3D {
-		(this * This new(0.0f, value) * this inverse) imaginary
+		(this * This new(0.0f, value) * this conjugate) imaginary
```
davidhesselbom: Does this need a `FIXME`, maybe?

marcusnaslund: Oh, I missed one. Will fix.

source/math/Quaternion.ooc
davidhesselbom: There's a space here: `- this z` and there isn't for `y` or `x`.

marcusnaslund: Shouldn't magic catch this? @thomasfanell

test/math/FloatPoint3DTest.ooc
```diff
@@ -63,6 +64,69 @@ FloatPoint3DTest: class extends Fixture {
...
+		})
+		this add("spherical", func {
+			point := FloatPoint3D spherical(5.0f, 1.23f, 0.57f)
+			expect(point x, is equal to(0.90f) within(0.01f))
+			expect(point y, is equal to(2.54f) within(0.01f))
+			expect(point z, is equal to(4.21f) within(0.01f))
+		})
+		this add("azimuth", func {
+			mypoint := FloatPoint3D new(1.0, 5.5, 0.1)
+			expect(mypoint azimuth, is equal to(atan2(5.5, 1.0) as Float) within(this precision))
```
davidhesselbom: `5.5 atan2(1.0)`

test/math/FloatSize3DTest.ooc
```diff
@@ -59,15 +55,89 @@ FloatSize3DTest: class extends Fixture {
...
+		this add("length, empty", func {
+			empty := FloatSize3D new()
+			expect(this vector0 length, is equal to(24.35f) within(0.01f))
+			expect(empty empty, is equal to(true))
+			expect(empty volume, is equal to(0.0f) within(this precision))
+			expect(this vector0 empty, is equal to(false))
+		})
+		this add("azimuth", func {
+			myvector := FloatSize3D new(1.0, 5.5, 0.1)
+			expect(myvector azimuth, is equal to(atan2(5.5, 1.0) as Float) within(this precision))
```
davidhesselbom: `5.5 atan2(1.0)`

source/math/Quaternion.ooc
thomasfanell: It should yes, however, ```-``` along with a few other operators were "temporarily" disabled early on due to being both unary and binary, so we were going to catch those kind of rules in the parsing phase instead. However, we decided to rewrite magic from the ground up, so it was never fixed in this version of magic.

davidhesselbom: I think unary looks a little strange with a space, but maybe that's just me. Either way, be consistent.

marcusnaslund:
I had these changes to `FloatVectorList` and `FloatComplexVectorList` in another working branch but I figured they belonged to this PR. Now I'm done, won't add anything more. :)

test/math/FloatVectorListTest.ooc
```diff
@@ -294,6 +290,36 @@ FloatVectorListTest: class extends Fixture {
...
 			filtered free()
 		})
+		this add("absolute", func {
+			list := FloatVectorList new()
+			list add(2.0f)
+			list add(-1.0f)
+			list add(-6.0f)
+			list add(4.0f)
+			list add(-7.0f)
+			_absolute := list absolute()
```
davidhesselbom: `_`

test/math/FloatComplexVectorListTest.ooc
```diff
@@ -63,6 +63,50 @@ FloatComplexVectorListTest: class extends Fixture {
...
+			list add(FloatComplex new(1, 1))
+			list add(FloatComplex new(2, -3))
+			list add(FloatComplex new(4, -2))
+			list add(FloatComplex new(-1, 8))
+			_sum := list sum
+			_mean := list mean
+			expect(_sum real, is equal to(6.0f) within(tolerance))
+			expect(_sum imaginary, is equal to(4.0f) within(tolerance))
+			expect(_mean real, is equal to(1.5f) within(tolerance))
+			expect(_mean imaginary, is equal to(1.0f) within(tolerance))
```
davidhesselbom: `_`

source/math/FloatVectorList.ooc
```diff
@@ -24,12 +24,18 @@ FloatVectorList: class extends VectorList<Float> {
 		this super()
 	}
 	init: func ~heap (capacity: Int) {
-		super(capacity)
+		this super(capacity)
```
davidhesselbom: `this super`? I'm not sure about that...

marcusnaslund: Now that you mention it, neither am I. `init~default` had it, but none of the others. I should have removed them all instead. Will fix.

marcusnaslund:
Updated: `this super` and underscores (again) removed.

marcusnaslund:
Updated - found an error in `FloatComplexTest`.

source/math/FloatComplexVectorList.ooc
```diff
@@ -30,21 +30,27 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 		super(other _vector)
 		this _count = other count
 	}
+	init: func ~withValue (capacity: Int, value: FloatComplex) -> This {
+		super(capacity)
+		for (i in 0 .. capacity)
+			this add(value)
+		this
```
davidhesselbom: I don't think this needs to return anything.

marcusnaslund: Fixed.

test/math/QuaternionTest.ooc
```diff
@@ -54,12 +58,26 @@ QuaternionTest: class extends Fixture {
...
 			expect((-1.0f) * this quaternion0 == -this quaternion0)
 		})
 		this add("norm", func {
 			expect(this quaternion0 norm, is equal to(65.5991592f))
 		})
+		this add("inverse", func {
+			inverse := this quaternion2 inverse
+			product := this quaternion2 * inverse
+			expect(product real == 1.0f)
+			expect(product x > -tolerance) //FIXME The standard "is equal to(0.0) within(tolerance)" does not work here?
```
davidhesselbom: Is this still true, after merging with upstream?

marcusnaslund: Yep. This doesn't work either (the line above): 
```ooc
productReal := product real
expect(productReal, is equal to(1.0f) within(tolerance))
```

davidhesselbom: What about `expect(productReal as Float, is equal to(1.0f) within(tolerance))`? Not that I like the look of it, but I'm curious.

marcusnaslund: Huh. This works too: `expect(product real as Float, is equal to(1.0f) within(tolerance))`

davidhesselbom: Stupid, stupid compiler...

source/math/FloatBox2D.ooc
```diff
@@ -125,19 +125,24 @@ FloatBox2D: cover {
...
+		for (i in 0 .. points length)
+			pointsAsVectorList add(points[i])
+		result := This bounds(pointsAsVectorList)
+		pointsAsVectorList free()
+		result
+	}
+	bounds: static func ~fromList (points: VectorList<FloatPoint2D>) -> This {
 		xMinimum := 0.0f
 		xMaximum := xMinimum
 		yMinimum := xMinimum
```
davidhesselbom: I know this wasn't introduced here, but `for (point in points)` leaks memory. You need to use
```ooc
for (i in 0 .. points count) {
    point := points[i]
```.

marcusnaslund: The same is true in `IntBox2D`. Will fix.

marcusnaslund:
Hmm, AppVeyor fails. I guess I haven't merged in windows-related stuff in a while.

davidhesselbom:
I just merged some changes in `FloatMatrix`, and now there are conflicts. Sorry...

davidhesselbom:
> Hmm, AppVeyor fails.

AppVeyor integration was added just yesterday, so no surprise. Probably need #580 to fix it, but I'm not going to let AppVeyor stop me from merging things just yet.

marcusnaslund:
No problem, the only things I did with `FloatMatrix` were cleaning up amongst the comments and fixing the `toString()` function. Ready soon.

davidhesselbom:
New conflicts. Can you please rebase again?

davidhesselbom:
Also, if you could split this up into several smaller pull requests, that would really help...

marcusnaslund:
Conflicts resolved.

marcusnaslund:
Yes, this PR got a bit out of hand. Smaller versions on the way. :)

marcusnaslund:
Split up into seven smaller PRs, so closing this.

PR #493, opened 2015-10-02
AbstractContext as parent for GpuContext

sebastianbaginski:
Problem with this one is that I want to keep as much compatibility with existing code base in the other project as possible. There are `context createGpuImage` calls spread out in there everywhere. I know it looks weird to have `createGPUIImage` in `AbstractContext`, but I think it is wise to do that in small steps, and not try to change everything in one pull request. So for now we could just keep the API compatible.

### DIFF
--- a/source/draw/AbstractContext.ooc
+++ b/source/draw/AbstractContext.ooc
@@ -1,0 +1,39 @@
+//
+// Copyright (c) 2011-2014 Simon Mika <simon@mika.se>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+use ooc-base
+use ooc-draw
+use ooc-math
+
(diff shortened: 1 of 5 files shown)


source/draw/AbstractContext.ooc
```diff
@@ -0,0 +1,39 @@
...
+	Ceiling
+}
+
+AbstractContext: abstract class {
+	init: func
+	createMonochrome: abstract func (size: IntSize2D) -> Image
+	createBgr: abstract func (size: IntSize2D) -> Image
+	createBgra: abstract func (size: IntSize2D) -> Image
+	createUv: abstract func (size: IntSize2D) -> Image
+	createGpuImage: abstract func (rasterImage: RasterImage) -> Image
```
davidhesselbom: Would it make sense to return a `GpuImage` in this case?

sebastianbaginski: `AbstractContext` will have another subclass called `CpuContext`, in which this method will return a `RasterImage`. I know this interface will require to be updated and I will do that, but for now I just wanted to keep the code in the *other project* unchanged.
Right now you can just merge this pull request and everything will work as usual.
This method is meant to return an image class *native* to given context implementation, so `GpuContext` will create `GpuImage`, while `CpuContext` will create `RasterImage`. So if you run with `CpuContext` the frame content will be a raster image, and that's what every client class will expect, because they will be replaced (via version blocks) by the cpu-only versions.
I guess we should simply call this method `createImage` or something. I ~~can~~ will do that in another pull request later on.

davidhesselbom: Wait, a function called `createGpuImage` will return a `RasterImage`?

davidhesselbom: Oh. Okay. Yeah, okay, do it in the next PR, then...

sebastianbaginski: Yes, for now. Just to keep the API unchanged.

Issue #460, opened 2015-09-29
PointerVector, PointerVectorList and HashDictionary


thomasfanell:
Currently, [PointerVector](https://github.com/cogneco/ooc-kean/blob/master/source/collections/PointerVector.ooc),  [PointerVectorList](https://github.com/cogneco/ooc-kean/blob/master/source/collections/PointerVectorList.ooc) and  [HashDictionary](https://github.com/cogneco/ooc-kean/blob/master/source/collections/HashDictionary.ooc) does not appear to be used in any of our projects. Do we want to use these for something? Can anyone think of a case where one of these would be better to use?

davidhesselbom:
`PointerVectorList` could be useful as a replacement for `VectorList` where `T` is known to be an object, for performance reasons. I can't think of anywhere this would actually help, though. It would be better to create dedicated, non-generic versions of e.g. `FloatVectorList`.

marcusnaslund:
`HashDictionary` seems like a reasonable thing to have in any standard library. The first two I'm not so sure about, though. Why were they written in the first place?

davidhesselbom:
The first two I just explained, but `HashDictionary` was meant to be used for something before we realized how difficult it was to handle memory whenever `String` was involved.

marcusnaslund:
It should be pretty simple to rewrite `HashDictionary` to use `Text` instead of `String` as key, right?

davidhesselbom:
Probably, but I don't think we'll be switching back now anyway, so don't spend time on it right now.

PR #494, opened 2015-10-05
Reworked HashDictionary

marcusnaslund:
Adds missing tests for `HashDictionary` and `VectorList`.
Some cleanup of `HashDictionary`. More can be done (it uses `ArrayList` and `HashBag` from the sdk for example) but we can look at that if/when we will use this class. 
Right now it doesn't seem to be used (see #460 )


test/collections/VectorListTest.ooc
```diff
@@ -195,8 +243,12 @@ VectorListTest: class extends Fixture {
 			list add(16)
 			list add(32)
 			iterator := list iterator()
+			expect(iterator hasNext?(), is equal to(true))
 			for ((index, item) in iterator)
 				expect(item, is equal to(list[index]))
+			expect(iterator hasNext?(), is equal to(false))
+			secondIterator := list iterator()
```
thomasfanell: You forgot to free secondIterator

marcusnaslund: @thomasfanell I've seen many tests where nothing at all is freed. I thought I was just being pedantic sometimes. But you're right, will fix.

source/collections/HashDictionary.ooc
```diff
@@ -3,15 +3,30 @@ import structs/[HashBag, HashMap, ArrayList]
 HashDictionary: class {
 	_myHashBag: HashBag
 	_capacity: Int
+	size ::= this _myHashBag size()
```
davidhesselbom: Would it be a good idea to call this `count` instead, as in our other collections?

source/collections/HashDictionary.ooc
```diff
@@ -20,17 +35,16 @@ HashDictionary: class {
...
 	get: func <T> (key: String, defaultValue: T) -> T {
 		result := defaultValue
 		if (_myHashBag contains?(key)) {
-			storedType := _myHashBag getClass(key)
-			// is `T` a derived type or the same type as the stored type?
+			storedType := this _myHashBag getClass(key)
 			if (T inheritsFrom?(storedType)) {
-				result = _myHashBag getEntry(key, T) value as T
+				result = this _myHashBag getEntry(key, T) value as T
 			}
```
davidhesselbom: Since you're cleaning this up, you might as well get rid of these unnecessary brackets...

marcusnaslund:
Fixed according to comments.

test/collections/HashDictionaryTest.ooc
```diff
@@ -104,7 +104,7 @@ HashDictionaryTest: class extends Fixture {
 			dictionary add("First", 1337)
 			expect(dictionary getAsType("First", Int) == 1337)
 			expect(dictionary getAsType("First", String) == null)
-			expect(dictionary getAsType("Second", Int) == null)
+			//expect(dictionary getAsType("Second", Int) == null) //FIXME Can we do this without obnoxious Pointer-To-Int warnings?
```
davidhesselbom: `getAsType: func <T>(key: String, T: Class) -> T {`, which means that `dictionary getAsType("Second", Int)` will return an `Int`. If you compare an `Int` to `null`, you will get the warning. If the test would follow the warning `The Class parameter for Covers must be Cell<Cover>` by calling `dictionary getAsType("Second", Cell<Int>)` the warning should go away, but then you also need to use `Cell` when adding things to the dictionary.

marcusnaslund: Ah, this should be solved as I did with `Future`. Will fix.

marcusnaslund:
The updated `HashDictionary` now handles wrapping of covers into cells automatically (like `Promise`) - no need to do it manually and no bothersome warnings in the test.

test/collections/HashDictionaryTest.ooc
```diff
@@ -40,41 +43,37 @@ HashDictionaryTest: class extends Fixture {
...
-			expect(dictionary get("TestClassValue", defaultClass) stringVal == "String", is true)
-			expect(dictionary get("Nonexistent", defaultClass) stringVal == "Default", is true)
-			expect(dictionary get("TestClassValue", defaultClass) intVal == 1, is true)
-			expect(dictionary get("Nonexistent", defaultClass) intVal == 0, is true)
+			expect(dictionary get("TestClassValue", this defaultClass) stringVal == "String", is true)
+			expect(dictionary get("Nonexistent", this defaultClass) stringVal == "Default", is true)
+			expect(dictionary get("TestClassValue", this defaultClass) intVal == 1, is true)
+			expect(dictionary get("Nonexistent", this defaultClass) intVal == 0, is true)
 		})
 		this add("ArrayList", func {
```
davidhesselbom: It's not an `ArrayList` anymore, so you should change this thing as well.

source/collections/HashDictionary.ooc
```diff
@@ -3,15 +3,30 @@ import structs/[HashBag, HashMap, ArrayList]
 HashDictionary: class {
 	_myHashBag: HashBag
```
davidhesselbom: `my` isn't used anywhere else afaik. Should probably just be `_hashBag`, right?

test/collections/HashDictionaryTest.ooc
```diff
@@ -98,43 +97,41 @@ HashDictionaryTest: class extends Fixture {
...
-
 		this add("Get from primitive", func {
 			dictionary := HashDictionary new()
 			dictionary add("First", 1337)
-			expect(dictionary getAsType("First", Int) == 1337)
-			expect(dictionary getAsType("First", String) == null)
-			expect(dictionary getAsType("Second", Int) == null)
-			expect(dictionary getAsType("First", Cell<TestCover>) == null)
-			expect(dictionary getAsType("First", TestClass) == null)
+			dictionary add("First2", TestClass new(1337, "String"))
```
davidhesselbom: `First2` doesn't make much sense. How about `Third`, or replace `Second` below with `Third` and `First2` with `Second`?

marcusnaslund:
Updated: `_myHashBag` renamed to `_hashBag`, `first2` renamed to `second` and *ArrayList* test renamed to *VectorList*.

source/sdk/lang/types.ooc
```diff
@@ -122,8 +122,10 @@ Closure: cover {
 /** An object storing a value and its class. */
 Cell: class <T> {
     val: __onheap__ T
+	type: Class = null
 
     init: func(=val)
+	init: func ~withtype(=val, =type)
```
davidhesselbom: Indentation

davidhesselbom: What do you need this for?

marcusnaslund: Oh, the sdk uses spaces, I see.
This is to please the `get` functions in `HashDictionary` - I need to store the classtype `T` in order to check whether the requested type actually exists for a particular key.

source/sdk/lang/types.ooc
```diff
@@ -122,8 +122,10 @@ Closure: cover {
 /** An object storing a value and its class. */
 Cell: class <T> {
     val: __onheap__ T
+    type: Class = null
 
     init: func(=val)
+    init: func ~withtype(=val, =type)
```
davidhesselbom: I don't think we should add this just to get `HashDictionary` (which we're not even using) working. Besides, `Cell` should already know what `T` is, it shouldn't need to be stored like this. I'm going to let this sit for a while until someone comes up with a better idea.

marcusnaslund: We could store the classtype in `HashEntry` instead perhaps?

source/collections/HashDictionary.ooc
```diff
@@ -1,91 +1,100 @@
...
 	}
-	/* WARNING: The Class parameter for Covers must be Cell<Cover> */
-	getAsType: func <T>(key: String, T: Class) -> T {
-		result := null
-		if (_myHashBag contains?(key)) {
-			storedType := _myHashBag getClass(key)
-			// is `T` a derived type or the same type as the stored type?
-			if (T inheritsFrom?(storedType))
-				result = _myHashBag getEntry(key, T) value as T
+	getAsType: func <T>(key: String, T: Class, defaultValue: T) -> T {
```
davidhesselbom: Is there any reason you would want `T: Class` to be a different `T` than that of  `defaultValue`?

davidhesselbom: This looks plain dangerous. In `T inheritsFrom?`, which T are you referring to &mdash; the variable, or the type?

marcusnaslund: For a class, you may pass `null`, and rock will go crazy trying to understand the expected type.
This is also the way I did it in for example `ResultPromise`: 
```ooc
start: static func<S> (S: Class, action: Func -> S) -> This<S>
```

marcusnaslund:
Updated with cleanup of `VectorListTest`.
Possibly, the `HashDictionary` update should be a separate PR - since we don't seem to be clear on what to do with it (and it will require more work anyhow). Then the rest of the stuff here has a chance of getting passed faster. What do you think?

davidhesselbom:
Yes, split it up.

source/collections/HashDictionary.ooc
marcusnaslund: Wait a minute. Looking at the `Cell` implementation, the `[] operator` looks like this:
```ooc
operator [] <T> (c: Cell<T>, T: Class) -> T {
  if(!c T inheritsFrom?(T)) {
 //...
```
Does that mean I can read the stored type as `c T` without storing the classtype as a separate member? In that case this problem is easily solved.

davidhesselbom: I'm pretty sure you could, yes. An instance of something `foo: Bar<T>` should always know its own T, anything else would be crazy.

marcusnaslund: Yes, nothing crazy in this language. :)

davidhesselbom: Related (and not fixed in our version of rock): https://github.com/fasterthanlime/rock/issues/415

davidhesselbom: Still waiting for this to be fixed.

marcusnaslund: What should I fix? It doesn't store the type in `Cell` anymore. The reason we specify `T` explicitly here is so that rock can understand the expected type. As mentioned previously, this is what makes `Future start()` work.

marcusnaslund: @davidhesselbom I never got an answer to this. Want to know if I can remove the `not ready` label.

davidhesselbom: `T: Class` still looks unnecessary, but I need to read this through again before I know for sure. Hold on.

marcusnaslund: In any other language it would be, but if you don't have it you get something like `could not resolve return type T` from rock in most cases.

davidhesselbom: I'm not so sure about that, so give me some time to try it. I'll let you know as soon as I do.

marcusnaslund:
Couldn't find a need for `getAsType` - it's been removed.

source/collections/HashDictionary.ooc
```diff
@@ -1,97 +1,90 @@
 import structs/[HashBag, HashMap, ArrayList]
 
 HashDictionary: class {
-	_myHashBag: HashBag
+	_hashBag: HashBag
 	_capacity: Int
```
davidhesselbom: Doesn't look like this is actually used for anything at all.

marcusnaslund: Now.

marcusnaslund:
Updated and ready.

PR #504, opened 2015-10-07
Fixed bug in RasterMonochrome copy()

marcusnaslund:
The `ByteBuffer` was never initialized, so copying to the new buffer resulted in segfaults.

### DIFF
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -33,6 +33,7 @@ RasterMonochrome: class extends RasterPacked {
 	create: func (size: IntSize2D) -> Image { This new(size) }
 	copy: func -> This {
 		result := This new(this)
+		result _buffer = ByteBuffer new(this _buffer size)
 		this buffer copyTo(result buffer)
 		result
 	}
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -29,14 +29,9 @@ RasterMonochrome: class extends RasterPacked {
 	init: func ~allocateStride (size: IntSize2D, stride: UInt) { super(size, stride) }
 	init: func ~fromByteBufferStride (buffer: ByteBuffer, size: IntSize2D, stride: UInt) { super(buffer, size, stride) }
 	init: func ~fromByteBuffer (buffer: ByteBuffer, size: IntSize2D) { this init(buffer, size, this bytesPerPixel * size width) }
-	init: func ~fromRasterImage (original: RasterImage) { super(original) }
+	init: func ~fromRasterImage (original: This) { super(original) }
 	create: func (size: IntSize2D) -> Image { This new(size) }
-	copy: func -> This {
-		result := This new(this)
-		result _buffer = ByteBuffer new(this _buffer size)
-		this buffer copyTo(result buffer)
-		result
-	}
+	copy: func -> This { This new(this) }
 	apply: func ~bgr (action: Func(ColorBgr)) {
 		this apply(ColorConvert fromMonochrome(action))
 	}
--- a/test/draw/RasterMonochromeTest.ooc
+++ b/test/draw/RasterMonochromeTest.ooc
@@ -68,6 +68,11 @@ RasterMonochromeTest: class extends Fixture {
 			image2 = image2 resizeTo(size)
 			expect(image2 size == size)
 		})
+		this add("copy", func {
+			image := RasterMonochrome open(this sourceFlower)
+			image2 := image copy()
+			expect(image size == image2 size)
+		})
 		/*this add("distance, convertFrom RasterBgra", func {
 			source := this sourceFlower
 			output := "test/draw/output/RasterBgrToMonochrome.png"
--- a/test/draw/RasterMonochromeTest.ooc
+++ b/test/draw/RasterMonochromeTest.ooc
@@ -72,6 +72,12 @@ RasterMonochromeTest: class extends Fixture {
 			image := RasterMonochrome open(this sourceFlower)
 			image2 := image copy()
 			expect(image size == image2 size)
+			expect(image stride, is equal to(image2 stride))
+			expect(image transform == image2 transform)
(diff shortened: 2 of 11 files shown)


davidpiuva:
Maybe we need another laber to separate "has bug" from "fixed bug".

davidhesselbom:
> Maybe we need another laber to separate "has bug" from "fixed bug".

I think it's clear enough since the label can be used both for issues and for pull requests. However, it should say "Fixed bug", not "fix bug".

marcusnaslund:
Oops. Title **fixed** :)

davidhesselbom:
The title is like that because your commit message was like that. That, too, should probably have read "Fixed"...

test/draw/RasterMonochromeTest.ooc
```diff
@@ -68,6 +68,11 @@ RasterMonochromeTest: class extends Fixture {
 			image2 = image2 resizeTo(size)
 			expect(image2 size == size)
 		})
+		this add("copy", func {
+			image := RasterMonochrome open(this sourceFlower)
+			image2 := image copy()
+			expect(image size == image2 size)
```
davidhesselbom: Add
```ooc
expect(image stride == image2 stride)'
expect(image stride == image2 stride)
```
I'm pretty sure this check will fail. The `copy` function needs to call `result := This new~allocateStride(this

sebastianbaginski: In addition to what @davidhesselbom pointed out, you can also use `equals` method from `RasterPacked`.

davidhesselbom:
```ooc
expect(image referenceCount != image2 referenceCount)
```
would probably also be a good idea...

davidhesselbom: > you can also use equals method from RasterPacked.

Nice catch.

thomasfanell: If I could chime in here, do prefer:
```expect(image stride, is equal to(image2 stride))```
over
```expect(image stride == image2 stride)```

Comment (marcusnaslund) [inline test/draw/RasterMonochromeTest.ooc]: Thanks everyone. `RasterMonochrome` was supposed to look for the super constructor
```ooc
init: func ~fromOriginal (original: This)
```
in `RasterPacked`, but with an argument of type `RasterImage`, it instead went looking further up in `RasterImage` where it found an appropriate constructor, which led to `this _stride` never being set.
Changing the constructor in `RasterPacked` to
```ooc
init: func ~fromOriginal (original: RasterImage)
```
passes all your tests.
Will push shortly.

davidhesselbom: Great, I was hoping it would be something that simple.

source/draw/RasterMonochrome.ooc
```diff
@@ -33,6 +33,7 @@ RasterMonochrome: class extends RasterPacked {
 	create: func (size: IntSize2D) -> Image { This new(size) }
 	copy: func -> This {
 		result := This new(this)
+		result _buffer = ByteBuffer new(this _buffer size)
```
davidhesselbom: Problem is, this will have to be added to the other `RasterPacked` classes as well (Bgr, Bgra, Uv`. Maybe what this function does should be moved to a function in `RasterPacked` instead? `copy` looks identical in all of them anyway.

marcusnaslund: Since `_buffer` even is a member of `RasterPacked`, that makes sense.

marcusnaslund: On second note, it doesn't. The arguments are of type `RasterImage`, and they don't have a buffer (the subclass `RasterPacked` does). It would work if we changed the argument to be of type `RasterPacked`. Or, we could add an `init~fromRasterPacked` in `RasterPacked` which would be called instead. Or, all subclasses to `RasterPacked` create the buffer in their respective `Copy` method (like `RasterMonochrome` does here).
I don't feel too comfortable messing around here, so I leave it for someone else to decide.

davidhesselbom: > Or, all subclasses to RasterPacked create the buffer in their respective Copy method (like RasterMonochrome does here).

That's exactly what I want to avoid - the code in `copy` will look 100% identical in all classes that inherit from `RasterPacked` (it already does, except for the line you just added to `RasterMonochrome` here).

marcusnaslund: Sorry, I was refering to the constructor in `RasterPacked`:
```ooc
init: func ~fromOriginal (original: RasterImage)
```

We cannot do anything with the buffer in `original` here unless it explicitly is set to be of type `RasterPacked`. But for it to be, we also have to update the appropriate constructor in each subclass since they expect the same type:
```ooc
RasterMonochrome: class extends RasterPacked {
// ...
init: func ~fromRasterImage (original: RasterImage) { super(original) }
// ...
```
Or, add a new `init ~fromRasterPacked`.

davidhesselbom: Ah, I see. I can look into fixing this.

marcusnaslund: Alright.

davidhesselbom: Hehe, in `RasterImage`,
```ooc
	copy: func ~fromMoreParams (size: FloatSize2D, source: FloatBox2D, upperLeft, upperRight, lowerLeft: FloatPoint2D) -> This {
		result := RasterBgra new(size ceiling() toIntSize2D())
//		TODO: The stuff
		result
	}
```

Maybe I should just leave this as is, open an issue about `copy` being broken and get back to it later.

davidhesselbom:
Setting to high since #518 needs it.

sebastianbaginski:
It's ok, no rush, we can use `swapRedBlue()` instead of `redBlueSwapped()` for now.

marcusnaslund:
Your code is merged @davidhesselbom - looks good and I assume you test this before you merge yourself so I did not.

davidhesselbom:
Sure. I'll test this again before I merge with upstream.

PR #505, opened 2015-10-07
Morphological tests for BoolVectorList added

marcusnaslund:
Including tests for logical operators.

### DIFF
--- a/test/math/BoolVectorListTest.ooc
+++ b/test/math/BoolVectorListTest.ooc
@@ -15,6 +15,7 @@
  * along with this software. If not, see <http://www.gnu.org/licenses/>.
  */
 
+use ooc-base
 use ooc-math
 use ooc-unit
 import math
@@ -43,26 +44,10 @@ BoolVectorListTest: class extends Fixture {
 			expect(!reversed[4])
 			expect(reversed[5])
 			list free()
+			reversed free()
 		})
-		
 		this add("dilation", func {
-			list := BoolVectorList new() //1100001110001110
-			list add(true)
-			list add(true)
-			list add(false)
-			list add(false)
-			list add(false)
-			list add(false)
-			list add(true)
-			list add(true)
-			list add(true)
-			list add(false)
-			list add(false)
-			list add(false)
-			list add(true)
-			list add(true)
-			list add(true)
-			list add(false)
+			list := this _createFromText(t"1100 0011 1000 1110")
 			dilated := list dilation(3)
 			expect(dilated[2])
 			expect(!dilated[3])
@@ -74,25 +59,8 @@ BoolVectorListTest: class extends Fixture {
 			list free()
 			dilated free()
 		})
-		
 		this add("erosion", func {
-			list := BoolVectorList new() //1100001110001110
-			list add(true)
-			list add(true)
-			list add(false)
-			list add(false)
-			list add(false)
-			list add(false)
-			list add(true)
-			list add(true)
-			list add(true)
-			list add(false)
-			list add(false)
-			list add(false)
-			list add(true)
-			list add(true)
-			list add(true)
-			list add(false)
+			list := this _createFromText(t"1100 0011 1000 1110")
 			eroded := list erosion(3)
 			expect(!eroded[1])
 			expect(!eroded[2])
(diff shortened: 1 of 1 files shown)


test/math/BoolVectorListTest.ooc
```diff
@@ -0,0 +1,183 @@
...
+			list add(true)
+			list add(true)
+			list add(true)
+			list add(false)
+			list add(false)
+			list add(false)
+			list add(true)
+			list add(true)
+			list add(true)
+			list add(false)
```
davidhesselbom: If you want to make this a *little* bit easier to read, you could group the `add` calls by 4 or so:
```ooc
			list add(true) . add(true) . add(false) . add(false)
			list add(false) . add(false) . add(true) . add(true)
			list add(true) . add(false) . add(false) . add(false)
			list add(true) . add(true) . add(true) . add(false)

File: test/math/BoolVectorListTest.ooc [hunk 2]
```diff
@@ -0,0 +1,183 @@
...
+			expect(reversed[1])
+			expect(reversed[2])
+			expect(reversed[3])
+			expect(!reversed[4])
+			expect(reversed[5])
+			list free()
+		})
+		
+		this add("dilation", func {
+			list := BoolVectorList new() //1100001110001110
```
davidhesselbom: Similarly, you could group this by 4: `//1100 0011 1000 1110`

sebastianbaginski: Would it make sense to add a constructor which takes a cstring of `0` and `1` and creates a vector from it ? It would make this tests so much more readable.

marcusnaslund: Add a constructor just for this test? I could write a helper function in this test that does that, though.

sebastianbaginski: I don't know if just for this test, maybe it could be useful elsewhere.
Another thing, `BoolVectorList` could be greatly optimized for storage space by changing its backend to a vector of integers. One 32 bit integer can keep the information for 32 entries in the boolean vector. But that's a separate issue I guess.

marcusnaslund: Isn't it that exact same idea which caused so much debate over the `std::vector<bool>` specialization in c++? :)

davidhesselbom: > Would it make sense to add a constructor which takes a cstring of 0 and 1 and creates a vector from it?

Not a bad idea.

> I could write a helper function in this test that does that, though.

You could also extend the class right here in the test file and give it an extra constructor that way.

> Another thing, BoolVectorList could be greatly optimized for storage space by changing its backend to a vector of integers. 

Sure, but then it's not a `VectorList` anymore.

davidhesselbom: Oh wait, I see what you mean now!

davidhesselbom: Let's open an issue for that idea, shall we?

sebastianbaginski: Yes :) But vector in ooc can't return a reference to an element, so the problem of having 
```C++
bool& operator[] (size_t pos);
```
does not exist here :)

marcusnaslund: So I add a new constructor and the memory optimization becomes a separate issue?

davidhesselbom: Yes, I think so.

marcusnaslund: How about also overloading `==` to compare with `Text` in the same way? Is that useful anywhere else?

davidhesselbom: I don't think we'll be comparing to literals anywhere, so I vote against it.

davidhesselbom: Same with the constructor, it'll only be used in this test, so `extend BoolVectorList` instead of adding the constructor in the class itself.

marcusnaslund: Hmm, extending the constructor works but gives an ugly warning for `implicit function declaration`.

davidhesselbom: How did you implement it?

marcusnaslund:
```ooc
extend BoolVectorList {
	init: func ~fromText (content: Text) {
		//lalala
	}
}
```

marcusnaslund: Doing the same thing in a private function of `BoolVectorListTest` gives no warnings.

davidhesselbom: Huh. Okay. Maybe `init` is a function that can't readly be added using an extension. Too bad.

marcusnaslund:
Updated. Clean enough?

davidhesselbom:
Sure, I guess the first list can be left as is.

marcusnaslund:
That was intentional, otherwise we never test `add`.

davidhesselbom:
Right.

PR #507, opened 2015-10-07
Painter and PaintEngine implementation

sebastianbaginski:
This pull request introduces the `Painter` class and `PaintEngine` hierarchy of classes.
`Painter` uses the `PaintEngine` as a backend to render primitives on images.
Each image subclass is responsible for creating a `PaintEngine` implementation suitable for the image type. It is enough to provide implementation for `drawPoint` for new `PaintEngine` type, drawing lines and boxes is based on that method.
As this is the first version, only basic primitives are supported, and only basic anti-aliasing is implemented. This can be extended later on by introducing a `RenderHint` flags for the paint engines and providing better painting algorithms if needed. Basic cover for `Pen` type is created and meant to be extended later (with line width, line styles, maybe a brush to fill entire shapes with colour etc.).
This version will be enough for our current needs in the *other project*, I will open a pull request there soon.

### DIFF
--- a/source/draw/PaintEngine.ooc
+++ b/source/draw/PaintEngine.ooc
@@ -1,0 +1,100 @@
+//
+// Copyright (c) 2011-2014 Simon Mika <simon@mika.se>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use ooc-base
+use ooc-collections
+use ooc-math
+import math
+import Color
+
+Pen: cover {
+	color: ColorBgra { get set }
+	alpha ::= this color alpha
+	alphaAsFloat ::= this alpha as Float / 255.0f
+	init: func@ (=color)
+	init: func@ ~withBgr (colorBgr: ColorBgr) {
+		this color = colorBgr toBgra()
+	}
+	setApha: func@ (alpha: UInt8) {
+		this color = ColorBgra new(this color toBgr(), alpha)
+	}
+	setAlpha: func@ ~withFloat (value: Float) {
+		this color = ColorBgra new(this color toBgr(), value * 255)
+	}
+}
+
+PaintEngine: abstract class {
+	pen: Pen { get set }
+	init: func {
+		this pen = Pen new(ColorBgra new(255, 255, 255, 255))
+	}
+	drawPoint: abstract func (x, y: Int)
+	drawPoint: virtual func ~withIntPoint2D (position: IntPoint2D) {
+		this drawPoint(position x, position y)
+	}
+	drawPoint: virtual func ~withFloat (x, y: Float) {
+		this drawPoint(x as Int, y as Int)
+	}
+	drawPoint: virtual func ~withFloatPoint2D (position: FloatPoint2D) {
+		this drawPoint(position toIntPoint2D())
+	}
+	drawLine: virtual func (start, end: IntPoint2D) {
+		if (start y == end y) {
+			startX := Int minimum~two(start x, end x)
+			endX := Int maximum~two(start x, end x)
+			for (x in startX .. endX + 1)
+				this drawPoint(x, start y)
+		} else if (start x == end x) {
+			startY := Int minimum~two(start y, end y)
+			endY := Int maximum~two(start y, end y)
+			for (y in startY .. endY + 1)
+				this drawPoint(start x, y)
+		} else {
+			originalPen := this pen
+			originalAlpha := originalPen alphaAsFloat
+			slope := (end y - start y) as Float / (end x - start x) as Float
+			startX := Int minimum~two(start x, end x)
+			endX := Int maximum~two(start x, end x)
+			for (x in startX .. endX + 1) {
+				idealY := slope * (x - start x) + start y
+				floor := idealY floor()
+				weight := (idealY - floor) abs()
+				this pen setAlpha(originalAlpha * (1.0f - weight))
+				this drawPoint(x, floor)
+				this pen setAlpha(originalAlpha * weight)
+				this drawPoint(x, floor + 1)
+			}
+			this pen = originalPen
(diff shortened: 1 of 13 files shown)


sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/501

source/draw/PaintEngine.ooc
```diff
@@ -0,0 +1,99 @@
...
+Pen: cover {
+	color: ColorBgra { get set }
+	alpha ::= this color alpha
+	alphaAsFloat ::= this alpha as Float / 255.0f
+	init: func@ (=color)
+	init: func@ ~withBgr (colorBgr: ColorBgr) {
+		this color = colorBgr toBgra()
+	}
+	setApha: func@ (alpha: UInt8) {
+		this color = ColorBgra new(this color toBgr(), alpha)
```
davidhesselbom: `this color alpha = alpha` doesn't do it?

source/draw/PaintEngine.ooc
```diff
@@ -0,0 +1,99 @@
...
+		} else {
+			originalPen := this pen
+			slope := (end y - start y) as Float / (end x - start x) as Float
+			startX := Int minimum~two(start x, end x)
+			endX := Int maximum~two(start x, end x)
+			for (x in startX .. endX + 1) {
+				idealY := slope * (x - start x) + start y
+				floor := idealY floor()
+				weight := (idealY - floor) abs()
+				this pen setAlpha(1.0f - weight)
```
davidhesselbom: Don't you somehow have to take into consideration the alpha channel of `originalPen`?

source/draw/Painter.ooc
```diff
@@ -0,0 +1,83 @@
...
+		this _backend drawLine(start, end)
+	}
+	drawLine: func ~withCoordinates (x1, y1, x2, y2: Int) {
+		this _backend drawLine(IntPoint2D new(x1, y1), IntPoint2D new(x2, y2))
+	}
+	drawBox: func (box: IntBox2D) {
+		this _backend drawBox(box)
+	}
+	drawPoint: func ~withFloat (x, y: Float) {
+		this _backend drawPoint(x, y)
```
davidhesselbom: This doesn't look necessary. If `PaintEngine drawPoint`, which takes 2 `Int`s, accepts 2 `Float`s, then so should `Painter drawPoint`.

source/draw/RasterPaintEngine.ooc
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+BgraPaintEngine: class extends RasterPaintEngine {
+	_image: RasterBgra
+	init: func (=_image) { super(this _image size) }
+	drawPoint: func (x, y: Int) {
+		position := this _map(IntPoint2D new(x, y))
+		if (this _image isValidIn(position x, position y))
+			this _image[position x, position y] = this _image[position x, position y] blend(this pen alphaAsFloat, this pen color toBgra())
```
davidhesselbom: `this pen color` is already a `ColorBgra`, isn't it?

source/draw/RasterPaintEngine.ooc
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+Yuv420PaintEngine: class extends RasterPaintEngine {
+	_image: RasterYuv420Semiplanar
+	init: func (=_image) { super(this _image size) }
+	drawPoint: func (x, y: Int) {
+		position := this _map(IntPoint2D new(x, y))
+		if (this _image isValidIn(position x, position y))
```
davidhesselbom: Should this check be put in a `safe` block, perhaps? Or should what `isValidIn` does be put in a `safe` block instead?

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -249,4 +251,5 @@ RasterYuv420Semiplanar: class extends RasterYuvSemiplanar {
 		fileWriter write(this uv buffer pointer as Char*, this uv buffer size)
 		fileWriter close()
 	}
+	createPaintEngine: override func -> PaintEngine { Yuv420PaintEngine new(this) }
```
davidhesselbom: What about `RasterYuv420Planar`?

source/draw/PaintEngine.ooc
sebastianbaginski: Unfortunately, you will get an error
> lvalue required as left operand of assignment

source/draw/PaintEngine.ooc
sebastianbaginski: Good catch, fixed.

source/draw/Painter.ooc
sebastianbaginski: `GpuPaintEngine` is based on `GpuCanvas`, which accepts floats for rendering primitives. Here I just want to forward arguments to the paint engine, without any conversions. Most of the `PaintEngine`s will use integers, but gpu can work on floats.

source/draw/RasterPaintEngine.ooc
sebastianbaginski: Here I wanted to have explicit check on the bounds during runtime, even without the `safe` mode. Sometimes the tracker boundaries were off the image, and it crashed. Since I wanted to minimize the changes made in the client code, I decided to have this check. 
It seems like the original rendering code (`GpuCanvas`) can handle incorrect coordinates without the `safe` block, so I think every `PaintEngine` should behave like this.

source/draw/RasterYuv420Semiplanar.ooc
sebastianbaginski: Same as `RasterUv` - I didn't need it.
I created implementation only for classes possibly needed for the windows port. Also, it keeps the pull request smaller :)

source/draw/RasterPaintEngine.ooc
sebastianbaginski: I'm not sure about this. When I started with the Painter, `this pen color` was `ColorBgr`. If you keep this method call here, then you won't have to change this code when `this pen color` changes to something else than `ColorBgra`.
I don't know, I can remove it if you want.

source/draw/PaintEngine.ooc
davidhesselbom: Hm, probably because `this color` is a property and not a variable... Oh well, I guess it doesn't matter.

source/draw/RasterPaintEngine.ooc
davidhesselbom: I think we should be able to avoid drawing outside the bounds, but alright, let's add the check later if need be.

davidhesselbom:
Just wanted to add, because I forgot to earlier, that this is superb OOP. Nice work!

sebastianbaginski:
Thank you, that's nice:) I guess I should give a credit to Qt folks (http://doc.qt.io/qt-4.8/qpainter.html) :)

PR #509, opened 2015-10-07
Updates to Promise and Future

marcusnaslund:
Contains several related changes:
* A new `PromiseCollector` class is added to facilitate the use of many `Promises`. Test included.
* Added `wait` with timeout for `Promise` and `Future`.
* Fixed bug related to waiting on a `Promise` more than once. This should solve related memory issues.
* `_threadPromise` and `_threadFuture` no longer have mutexes as there are no possible race conditions. This makes them lighter and faster.
* `Future` now only wraps covers in a `Cell` (instead of everything), since only they need it.
* Reduced the loop lengths in `PromiseTest` and `ThreadPoolTest` to make them faster to run.


marcusnaslund:
Resolved conflicts.

test/base/PromiseCollectorTest.ooc
```diff
@@ -0,0 +1,64 @@
...
+			promises := promise + promise2 + promise3 + promise4
+			expect(promises wait())
+			promises free()
+		})
+		this add("advanced (+= operator)", func {
+			promises := PromiseCollector new()
+
+			for (j in 0 .. 5) {
+				promises += Promise start(func { for (i in 0 .. 10_000_000) { } } )
+			}
```
davidhesselbom: You don't need these brackets

test/base/PromiseCollectorTest.ooc
```diff
@@ -0,0 +1,64 @@
...
+
+			extra := Promise start(func { for (i in 0 .. 10_000_000) { } } )
+			extra cancel()
+			promises += extra
+
+			others := PromiseCollector new()
+			for (j in 0 .. 5) {
+				others += Promise start(func { for (i in 0 .. 10_000_000) { } } )
+			}
```
davidhesselbom: or these

test/base/PromiseCollectorTest.ooc
```diff
@@ -0,0 +1,64 @@
...
+			promises += Promise start(func { for (i in 0 .. 10_000_000) { } } )
+			promises += Promise start(func { for (i in 0 .. 10_000_000) { } } )
+			expect(promises wait())
+			promises clear()
+		})
+		this add("wait with timeout", func {
+			promises := PromiseCollector new()
+			for (j in 0 .. 5) {
+				promises += Promise start(func { for (i in 0 .. 50_000_000) { } } )
+			}
```
davidhesselbom: or these

davidhesselbom:
@emilwestergren Could you help me review this?

source/base/Promise.ooc
```diff
@@ -11,53 +12,59 @@ Promise: abstract class {
...
 				this _state = _PromiseState Finished
-			this _mutex unlock()
 		}
 		this _thread = Thread new(this _action)
 		this _thread start()
 	}
 	free: override func {
-		if (this _state == _PromiseState Unfinished)
+		if (this _threadAlive)
 			this _thread wait()
```
emilwestergren: What if I don't want to wait on the promise, then there is no way to free it.

source/base/Promise.ooc
```diff
@@ -65,7 +72,15 @@ Future: abstract class <T> {
 	_state: _PromiseState
 	init: func
 	wait: abstract func -> Bool
-	wait: abstract func ~default (defaultValue: T) -> T
+	wait: abstract func ~timeout (seconds: Double) -> Bool
+	wait: virtual func ~default (defaultValue: T) -> T {
+		status := this wait()
+		status ? this getResult(defaultValue) : defaultValue
```
emilwestergren: Or on a single line:  `this wait() ? this getResult(defaultValue) : defaultValue`

source/base/Promise.ooc
```diff
@@ -65,7 +72,15 @@ Future: abstract class <T> {
...
+	}
+	wait: virtual func ~defaulttimeout (seconds: Double, defaultValue: T) -> T {
+		status := this wait(seconds)
+		status ? this getResult(defaultValue) : defaultValue
```
emilwestergren: Or on a single line:  `this wait() ? this getResult(defaultValue) : defaultValue`

source/base/Promise.ooc
```diff
@@ -74,57 +89,62 @@ Future: abstract class <T> {
...
 				this _state = _PromiseState Finished
-			this _mutex unlock()
 		}
 		this _thread = Thread new(this _action)
 		this _thread start()
 	}
 	free: override func {
-		if (this _state == _PromiseState Unfinished)
+		if (this _threadAlive)
 			this _thread wait()
```
emilwestergren: What if I don't want to wait on the future, then there is no way to free it.

source/base/ThreadPool.ooc
```diff
@@ -46,19 +46,25 @@ _Task: abstract class {
 _ActionTask: class extends _Task {
 	_action: Func
 	init: func (=_action, mutex: Mutex) { super(mutex) }
-	run: override func {
+	run: func {
```
emilwestergren: Why remove override?

source/base/ThreadPool.ooc
```diff
@@ -46,19 +46,25 @@ _Task: abstract class {
...
 _ResultTask: class <T> extends _Task {
-	_result: Cell<T>
+	_result: Object
 	_action: Func -> T
+	_hasCover := false
 	init: func (=_action, mutex: Mutex) { super(mutex) }
-	run: override func {
-		temporary := Cell<T> new(this _action())
-		this _result = temporary
+	run: func {
```
emilwestergren: Why remove override?

source/base/Promise.ooc
marcusnaslund: Yes, that would be nice, but rock doesn't like it.

source/base/Promise.ooc
marcusnaslund: I have to join the thread before I can free its reference, and that means waiting for it. Do you have another suggestion?

source/base/ThreadPool.ooc
marcusnaslund: Bad merge on my part, I'm afraid.

source/base/ThreadPool.ooc
```diff
@@ -70,7 +76,15 @@ _TaskPromise: class extends Promise {
 		this _task free()
 		super()
 	}
-	wait: override func -> Bool { this _task wait() }
+	wait: func -> Bool { this _task wait() }
+	wait: func ~timeout (seconds: Double) -> Bool {
+		timer := ClockTimer new() . start()
+		status := false
+		while (timer stop() / 1000.0 < seconds && !status)
```
emilwestergren: Don't you need to call `timer start()` again after you call `timer stop()`? Or does it simply return the current time difference since the latest call to `timer start()`?

Nonetheless, this is a very ugly solution to wait for something (as I think you already know by now). At least sleep for `timeout / 10` or so between every check.

source/base/Promise.ooc
emilwestergren: If the promise is unfinished when free is called, a `_freeOnCompletion` flag is set to true. This flag is checked after the task has completed (in the thread) and the object freed if the flag is set to true.

source/base/ThreadPool.ooc
marcusnaslund: No, this works. But yes, I agree with you, it's ugly. I'll add some `sleep` at least.

source/base/Promise.ooc
marcusnaslund: I understand, but in this case the object to free is the instance of the thread itself, which first has to be joined. Can I do that from inside the thread itself?

emilwestergren: The thread doesn't have to be joined, but if you don't join it before the process exits then Valgrind will see it as a memory leak. You could simply free the object data for ThreadPromise and Thread which can be done from inside the thread itself, but I'm not sure if this is compatible with every other platform.

Just leave it for now then, maybe add a TODO?

marcusnaslund: Sure, I will add a `TODO` and create a relevant issue. Is that OK @davidhesselbom ?

marcusnaslund:
Updated according to comments.

source/base/Promise.ooc
```diff
@@ -8,56 +9,63 @@ _PromiseState: enum {
...
 		this _thread free()
 		(this _action as Closure) dispose()
-		this _mutex free()
 		super()
 	}
 	wait: func -> Bool {
-		if (this _state == _PromiseState Unfinished)
-			this _thread wait()
+		if (this _threadAlive)
+			if (this _thread wait())
```
davidhesselbom: Briefly, what does it mean when `wait()` returns false?

source/base/Promise.ooc
```diff
@@ -8,56 +9,63 @@ _PromiseState: enum {
...
-		status := false
-		this _mutex lock()
 		if (this _state == _PromiseState Unfinished) {
+			this _thread cancel()
 			this _state = _PromiseState Cancelled
-			status = true
 		}
-		this _mutex unlock()
-		status
+		this _state == _PromiseState Cancelled
```
davidhesselbom: So this will return `false` if it was already finished when cancelled?

source/base/Promise.ooc
emilwestergren: The promise has not been fulfilled, for example if someone cancelled it before it completed.

source/base/Promise.ooc
marcusnaslund: Exactly

PR #510, opened 2015-10-08
Added a test coverage tester

marcusnaslund:
Far from perfect, but has been a helpful tool for me when updating/writing new tests. I'm not sure if it's worth the extra effort of perfecting it.
I also don't know what the policy is on Python, or whether this belongs in kean, so feel free to cancel this PR if it doesn't belong here.
In short: Checks all files in a specified source folder (base, collections, math, etc.) and outputs easy-to-read information on which functions that aren't tested.

### DIFF
--- a/coverage-test.py
+++ b/coverage-test.py
@@ -1,0 +1,92 @@
+#!/usr/bin/python
+import re, sys, os
+from os import listdir
+from os.path import isfile, join
+
+TCOLOR_OKGREEN = '\033[92m'
+TCOLOR_FAIL = '\033[91m'
+TCOLOR_UNDERLINE = '\033[4m'
+TCOLOR_ENDC = '\033[0m'
+
+def get_unused_items(sourcetext, testtext):
+	# Extract everything that can be used in source file
+	# TODO: Only checks overloaded functions once
+	functions = re.findall(r'\s.+: func ', sourcetext)
+	functions += re.findall(r'\s.+: static func ', sourcetext)
+	#functions += re.findall(r'\s.+: func ~.+?\s', sourcetext)
+	#functions += re.findall(r'\s.+: static func ~.+?\s', sourcetext)	
+	properties = re.findall(r'\s.+ ::=', sourcetext)
+	operators = re.findall(r'\s+operator \S+', sourcetext)
+
+	# Clean up results
+	functions = [w.replace(' static func ','').replace(' func ','').replace(':','').strip() for w in functions]
+	properties = [w.replace('::=','').strip() for w in properties]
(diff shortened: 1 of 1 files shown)


thomasfanell:
Nice, I think maybe we should do this in Typescript/Node though, as we are already depending on those two for magic. I think an even better solution would be the integrate this into magic (which is already planned by the way).

marcusnaslund:
That's probably a good idea, I just don't have any experience in (or knowledge of) Typescript/Node.

thomasfanell:
How about writing it in ooc, and stick it ```ooc-kean/tools```, then make a script that compiles and runs it?

marcusnaslund:
To clarify: The script has to be in the ooc-kean root, and is run for example with:
`python coverage-test.py base`

marcusnaslund:
I don't dare to rely on the `ooc` regex stuff, so I vote for the first option.

davidhesselbom:
I think it should be converted into (or rewritten in) TypeScript. I have nothing against Python per se, but we don't need *another* language being used in our project, and I don't want anything we do to be dependent on any one person, if we can avoid it.

marcusnaslund:
I think @thomasfanell is taking a look at it.

thomasfanell:
Indeed I am, writing on it as we speak.

davidhesselbom:
Nice. Glhf!

thomasfanell:
Just a side-note: It'll be tricky to get this 100% accurate. Things like function overload coverage, unless called explicitly (with suffix), will be very difficult to detect. I very much doubt I have the time required to cover all cases. I can get us in the ball park no problem, but are we happy with that?

davidhesselbom:
How accurate is the Python version?

marcusnaslund:
It has the same limitations.

davidhesselbom:
Well then. I see no harm in replacing it with a TypeScript version that has the same functionality and flaws.

thomasfanell:
Well it'll be much improved, but my point was it's not going to be perfect. :-)

davidhesselbom:
Sure, but it's better than nothing, and if it's as good as what's in this here pull request, I have no complaints for now.

thomasfanell:
Ok, i'll work on it as quickly as I can then. :+1:

davidhesselbom:
I think @marcusnaslund has already covered as much as he can using the Python version, so I don't think there's any hurry, really...

marcusnaslund:
That's true, the version here is enough for me at the moment.

marcusnaslund:
So, close this and wait for the TypeScript version?

davidhesselbom:
Yes.

PR #511, opened 2015-10-08
CreateAlphaByteString now produces magic-compliant output

marcusnaslund:
Since there is a "malformed" ooc-file there.

### DIFF
--- a/.magicignore
+++ b/.magicignore
@@ -2,3 +2,4 @@
 .libs
 rock_tmp
 source/sdk
+test/draw/output
--- a/.magicignore
+++ b/.magicignore
@@ -2,4 +2,3 @@
 .libs
 rock_tmp
 source/sdk
-test/draw/output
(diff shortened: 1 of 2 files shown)


davidhesselbom:
Quick fix, but wrong. We're actually using the class that generates that "malformed" ooc code to create arrays containing textures. The solution is to fix `CreateAlphaByteString` so that it produces its output correctly, not to stick our heads in the sand and ignore the incorrect output. And that's an easy fix, too: 
```ooc
		for (i in 0 .. ipbuffer size / 4) {
			imageArray = imageArray + ip[i] toString() + ", "
			counter += 1
			if (i % 32 == 0) {
				imageArray = imageArray + "\n"
			}
		}
```
should be
```
		for (i in 0 .. ipbuffer size / 4) {
			imageArray = imageArray + ip[i] toString() + ","
			counter += 1
			if (i % 32 == 0)
				imageArray = imageArray + "\n"
			else
				imageArray = imageArray + " "
		}
```
Not elegant, but performance and memory leaks are non-issues in this class.

marcusnaslund:
Good point. Here we go - had to change one more line as well.

PR #512, opened 2015-10-08
CpuContext

sebastianbaginski:
Subclass of `AbstractContext` used when running without GPU.

### DIFF
--- a/ooc-draw.use
+++ b/ooc-draw.use
@@ -5,6 +5,7 @@ Requires: ooc-unit
 Requires: ooc-math
 Requires: ooc-base
 Imports: AbstractContext
+Imports: CpuContext
 Imports: Color
 Imports: Image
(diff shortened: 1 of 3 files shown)


marcusnaslund:
The CircleCI "timeout" problem I've seen once before from `SynchronizedQueue`, so I will venture to say it is not the fault of `ThreadPool` but rather the infamous queues. Haven't been able to reproduce it on my machine, though.

sebastianbaginski:
> sudo apt-get install  --> 4:16

Looks like I got a really slow machine ...

marcusnaslund:
Ah, maybe that's the problem.

davidhesselbom:
Maybe we should split up the `apt-get install` calls into one for each package. It won't speed things up, but it'll reduce the risk of timeouts.

PR #516, opened 2015-10-08
Added setSeed and global seeds for random generators

marcusnaslund:
Adds `permanentSeed` for Int and Float random generators. If they are non-zero will override the seed automatically generated from `Time microtime()`. Useful for debug purposes.
Also adds the somehow missing `setSeed(Int)` which was missing - if you only want to control a single generator instance.

### DIFF
--- a/source/math/FloatRandomGenerator.ooc
+++ b/source/math/FloatRandomGenerator.ooc
@@ -17,6 +17,7 @@ import math
 import os/Time
 
 FloatRandomGenerator: abstract class {
+	permanentSeed: static UInt = 0
 	next: abstract func -> Float
 	next: func ~withCount (count: Int) -> Float[] {
 		result := Float[count] new()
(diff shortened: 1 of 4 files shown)


source/math/FloatRandomGenerator.ooc
```diff
@@ -29,24 +30,21 @@ FloatRandomGenerator: abstract class {
 FloatUniformRandomGenerator: class extends FloatRandomGenerator {
 	_state: UInt
 	_min, _max, _rangeCoefficient: Float
+	_unsignedIntMaxAsFloat : static Float = 4294967295U as Float
```
davidhesselbom: No, use `UINT_MAX` and cast that to `Float` instead.

davidhesselbom: Actually, I don't see this being used more than once. And it has nothing to do with the class, really, so just put type `UINT_MAX as Float` instead of `This _unsignedIntMaxAsFloat` below.

marcusnaslund:
Fixed, changed to use `UINT_MAX`.

source/math/FloatRandomGenerator.ooc
```diff
@@ -31,22 +32,18 @@ FloatUniformRandomGenerator: class extends FloatRandomGenerator {
...
 	minimum ::= this _min
 	maximum ::= this _max
-	init: func {
-		this _state = Time microtime()
-		this setRange(0.0f, 1.0f)
-	}
-	init: func ~withSeed (seed: Int) {
-		this _state = seed
+	init: func (seed := Time microtime()) {
+		this _state = FloatRandomGenerator permanentSeed != 0 ? This permanentSeed : seed
```
davidhesselbom: `this _state = This permanentSeed != 0 ? This permanentSeed : seed` doesn't work?

marcusnaslund: Oh, it does, will fix. I thought magic caught these sorts of things? @thomasfanell ?

PR #517, opened 2015-10-08
Removed interface files and tests

marcusnaslund:
Because they do not work, never have and never will, this removes all interfaces.
Also removed the `OnOff` enum which serves no purpose and is never used.


davidhesselbom:
I'm conflicted about this. It's true that we haven't been able to get interfaces to work the way we want them to, but these interfaces actually do work, at least in their tests, don't they? How do we know we can't fix interfaces at some point in the near future and make these interfaces useful? But on the other hand, why keep code around that isn't used for anything...

marcusnaslund:
The plan is to drop support for interfaces completely for now, as far as I understand. Right @thomasfanell ?

thomasfanell:
If you mean dropping them from Kean, you're right. I don't see the point in having them around at all, since they're apparently so unreliable. They are on my hit-list of things to kill while doing the SDK-sweep. I suppose the final decision is on David or Simon.

davidhesselbom:
You're probably right, but I'd like to hear from @simonmika about it first.

simonmika:
Go ahead and remove it, they are not used at the moment anyhow. I will take another look at interfaces and try to estimate what works and what is broken. When I wrote this we intended to use them with covers and that is not possible. They might work better with classes. OnOff was used a while for some settings.

PR #518, opened 2015-10-09
Method to swap red and blue channels in rgb and rgba images

sebastianbaginski:
We will need it on windows for debugging, @jonathanudd discovered that while working on the window class.
This pull request also introduces the `width` and `height` properties for `Image` class. I think it is better to write `for (row in 0 .. image height)` than `for (row in 0 .. image size height)`.

### DIFF
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -142,5 +142,13 @@ RasterBgr: class extends RasterPacked {
 	}
 	operator [] (x, y: Int) -> ColorBgr { this isValidIn(x, y) ? ((this buffer pointer + y * this stride) as ColorBgr* + x)@ : ColorBgr new(0, 0, 0) }
 	operator []= (x, y: Int, value: ColorBgr) { ((this buffer pointer + y * this stride) as ColorBgr* + x)@ = value }
+	swapRedBlue: func {
+		this swapChannels(0, 2)
+	}
+	redBlueSwapped: func -> This {
+		result := this copy()
+		result swapRedBlue()
+		result
+	}
 	createPaintEngine: override func -> PaintEngine { BgrPaintEngine new(this) }
 }
--- a/source/draw/RasterBgra.ooc
+++ b/source/draw/RasterBgra.ooc
@@ -171,5 +171,13 @@ RasterBgra: class extends RasterPacked {
 			(top * (left * topLeft alpha + (1 - left) * topRight alpha) + (1 - top) * (left * bottomLeft alpha + (1 - left) * bottomRight alpha))
 		)
 	}
+	swapRedBlue: func {
+		this swapChannels(0, 2)
+	}
+	redBlueSwapped: func -> This {
+		result := this copy()
+		result swapRedBlue()
+		result
+	}
 	createPaintEngine: override func -> PaintEngine { BgraPaintEngine new(this) }
 }
--- a/source/draw/RasterPacked.ooc
+++ b/source/draw/RasterPacked.ooc
@@ -89,4 +89,18 @@ RasterPacked: abstract class extends RasterImage {
 		file free()
 		StbImage writePng(filename, this size width, this size height, this bytesPerPixel, this buffer pointer, this size width * this bytesPerPixel)
 	}
+	swapChannels: func (first, second: Int) {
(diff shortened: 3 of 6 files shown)


source/draw/RasterBgr.ooc
```diff
@@ -140,4 +140,12 @@ RasterBgr: class extends RasterPacked {
...
 	operator [] (x, y: Int) -> ColorBgr { this isValidIn(x, y) ? ((this buffer pointer + y * this stride) as ColorBgr* + x)@ : ColorBgr new(0, 0, 0) }
 	operator []= (x, y: Int, value: ColorBgr) { ((this buffer pointer + y * this stride) as ColorBgr* + x)@ = value }
+	swapRedBlue: func {
+		this swapChannels(0, 2)
+	}
+	redBlueSwapped: func -> This {
+		result := this copy()
+		result swapRedBlue()
+		result
+	}
```
sebastianbaginski: This won't work until https://github.com/cogneco/ooc-kean/issues/500 is solved.

source/draw/RasterPacked.ooc
```diff
@@ -89,4 +89,18 @@ RasterPacked: abstract class extends RasterImage {
...
+			if (first > this bytesPerPixel || second > this bytesPerPixel)
+				raise("Channel number too large")
+		}
+		pointer := this buffer pointer
+		index := 0
+		while (index < this buffer size) {
+			value := pointer[index + first]
+			pointer[index + first] = pointer[index + second]
+			pointer[index + second] = value
+			index += this bytesPerPixel
```
davidhesselbom: I wonder how good gcc is at optimizing this for a `RasterMonochrome`...

sebastianbaginski: Should we have an empty `swapChannels: override func (first, second: Int)` in `RasterMonochrome` ? Or do we trust that no one will ever want to swap channels on an image with single channel ? :)

davidhesselbom: No no, I don't expect it to be a problem, I'm only curious.

davidhesselbom:
That's odd, the two output images look very different.

sebastianbaginski:
If you just save them without swapping channels they look different too, the BGRA version looks more saturated.

davidhesselbom:
Weird. Must be a bug in `stb_image`.

davidpiuva:
Swapping blue and red between RGB and BGR can look similar to swapping blue and orange between YUV and YVU. PNG also has built in Gamma correction that often fails.

source/draw/RasterBgra.ooc
```diff
@@ -171,5 +171,13 @@ RasterBgra: class extends RasterPacked {
 			(top * (left * topLeft alpha + (1 - left) * topRight alpha) + (1 - top) * (left * bottomLeft alpha + (1 - left) * bottomRight alpha))
 		)
 	}
+	swapRedBlue: func {
```
davidhesselbom: Personally, I would prefer a name like `RgbToBgr`, but then it would do the exact same thing as `BgrToRgb` and things would get confusing...

test/draw/RasterBgrTest.ooc
```diff
@@ -45,7 +45,27 @@ RasterBgrTest: class extends Fixture {
...
+			output free()
+		})
 	}
 }
 
-RasterBgrTest new() run()
\ No newline at end of file
+test := RasterBgrTest new()
+test run()
+test free()
```
davidhesselbom: Any of the following would also work:
```ooc
test := RasterBgrTest new()
test run() . free()
```
```ooc
test := RasterBgraTest new() . run() . free()
```
```ooc
RasterBgraTest new() run() . free()
```
What do you think?

sebastianbaginski: I don't really have an opinion. Which one do you prefer ?

davidhesselbom: The last one. No variables for something we're not keeping anyway, fits on one line.

sebastianbaginski: Updated.

davidhesselbom: Nice, thanks.

PR #520, opened 2015-10-09
added Bool to Cell toString()

thomasfanell:
This fixes any Fixture crashes where these tests fail:
```expect(someCondition, is equal to(false))```
```expect(someCondition, is equal to(true))```

Hint: For booleans, use these instead:
```expect(someCondition, is false)```
```expect(someCondition, is true)```

### DIFF
--- a/source/sdk/lang/types.ooc
+++ b/source/sdk/lang/types.ooc
@@ -135,6 +135,8 @@ Cell: class <T> {
 	// TODO: Move this to an extension in ooc-kean later
 	toString: func -> String {
 		match T {
+			case Bool =>
+				(this val as Bool) toString()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by thomasfanell** [comment]
Fixes #519

File: source/sdk/lang/types.ooc [hunk 1]
```diff
@@ -135,6 +135,8 @@ Cell: class <T> {
 	// TODO: Move this to an extension in ooc-kean later
 	toString: func -> String {
 		match T {
+            case Bool =>
+                (this val as Bool) toString()
```
davidhesselbom: Indentation...

thomasfanell:
Fixed

PR #526, opened 2015-10-09
Improved DateTime toStringFormat and adjusted and improved test accor…

### LINKED ISSUE #525: DateTime toString() bugs
Instead of padding with a leading `0`, the seconds part of the time is printed as a single digit if less than 10.

Also, a year like `2003` or `1903` will be printed as `3` instead of `03`, and there is no way to print the hours or minutes parts as a single digit, and `%h`, `%m`, `%s` (if in the format string) will not be replaced at all. EDIT: Apparently, the single-digit second format is on purpose, judging from the test. However, I will consider this a mistake, as I have never before seen a clock display the first 10 seconds in the minute using single digits and no leading `0`. Maybe for a `TimeSpan`, it would be acceptable..

Finally, `%zzzz` is used for the milisecond part, even though will it will never hold more than 3 digits. In the current implementation, `%z` or `%zzz` will not be replaced at all, and the milliseconds part will never be padded by leading `0`s.

davidhesselbom:
…dingly. Should fix #525 unless I missed something. @sebastianbaginski, I just want to run this by you before I merge, since you wrote the original version (I think).


sebastianbaginski:
That's true, I created the first version, but I'm also sure that I corrected the date formatting (I clearly remember fixing the "%zzzz" bug), since I was using this http://doc.qt.io/qt-5/qdatetime.html#toString as a reference.
But I also recall that my pull request was a mess and you cleaned it up. I have probably committed the changes to my branch after you merged the clean version into master. Committed them, without making a pull request.
I was truly a *git noob* back then.

> I just want to run this by you 

I tested it, works good.

davidhesselbom:
Haha, okay. This implementation still leaks loads of memory, but I wanted to get it fixed before we migrate to `Text`, if that's something we're planning on doing soon.

PR #529, opened 2015-10-12
Added Int64, UInt64 and Char to Fixture and relevant constraints.

### LINKED ISSUE #528: Fixture does not handle Int64
Because the `TimeSpan elapsedDays()` (among others') return type is `Int64` it is not possible to do the following in a test:
```ooc
expect(span elapsedDays(), is equal to(1))
```
This gives an error like `expected equal to '24', was '24.00000000'`. This should be adressed. For now, the following workaround can be used:
```ooc
expect(span elapsedDays() as Int, is equal to(1))
```
The same may be true for other basic numeric types as well.

thomasfanell:
* Added Int64, UInt64 and Char to Fixture and relevant constraints.
* Added Int64, UInt64 and Char to ```Cell toString()```

### DIFF
--- a/source/sdk/lang/types.ooc
+++ b/source/sdk/lang/types.ooc
@@ -135,6 +135,8 @@ Cell: class <T> {
 	// TODO: Move this to an extension in ooc-kean later
 	toString: func -> String {
 		match T {
+			case Char =>
+				(this val as Char) toString()
 			case Bool =>
 				(this val as Bool) toString()
 			case Int =>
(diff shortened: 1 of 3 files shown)


thomasfanell:
Is there a reason why this hasn't been merged yet @davidhesselbom? Need me to fix anything here?

davidhesselbom:
Dunno, have been prioritizing other things. I'll let you know!

PR #530, opened 2015-10-12
Bilinear interpolation for `resize` in RasterMonochrome

sebastianbaginski:
Introduces bilinear interpolation in `RasterMonochrome::resizeTo` method.
I want to generalize this method and introduce bilinear rescaling for each image class, hopefully without duplicating much of the code (the only difference is in the pixel type, so it should be possible to create a generic method for this).

### DIFF
--- a/source/draw/Image.ooc
+++ b/source/draw/Image.ooc
@@ -27,6 +27,11 @@ CoordinateSystem: enum {
 	YUpward = 0x02
 }
 
+TransformMethod: enum {
+	Fast, // nearest neighbour
+	Smooth // bilinear
+}
+
 Image: abstract class {
 	_size: IntSize2D
 	size ::= this _size
@@ -78,6 +83,9 @@ Image: abstract class {
 		this resizeTo(((this size toFloatSize2D()) * Float minimum(restriction width as Float / this size width as Float, restriction height as Float / this size height as Float)) toIntSize2D())
 	}
 	resizeTo: abstract func (size: IntSize2D) -> This
+	resizeTo: virtual func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
+		this resizeTo(size)
+	}
 	create: virtual func (size: IntSize2D) -> This { raise("Image type not implemented."); null }
 	copy: abstract func -> This
 	copy: abstract func ~fromParams (size: IntSize2D, transform: FloatTransform2D) -> This
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -58,20 +58,61 @@ RasterMonochrome: class extends RasterPacked {
 			}
 	}
 	resizeTo: override func (size: IntSize2D) -> This {
+		this resizeTo(size, TransformMethod Smooth) as This
+	}
+	resizeTo: override func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
 		result: This
 		if (this size == size)
 			result = this copy()
 		else {
 			result = This new(size)
-			for (row in 0 .. size height) {
-				srcRow := (this size height * row) / size height
-				for (column in 0 .. size width) {
-					srcColumn := (this size width * column) / size width
-					result[column, row] = this[srcColumn, srcRow]
-				}
+			match (method) {
+				case TransformMethod Smooth => This _resizeBilinear(this, result)
+				case => This _resizeNearestNeighbour(this, result)
 			}
 		}
 		result
+	}
+	_resizeNearestNeighbour: static func (source, result: This) {
+		resultBuffer := result buffer pointer
+		sourceBuffer := source buffer pointer
+		for (row in 0 .. result size height) {
+			sourceRow := (source size height * row) / result size height
+			for (column in 0 .. result size width) {
+				sourceColumn := (source size width * column) / result size width
+				resultBuffer[column + result stride * row] = sourceBuffer[sourceColumn + source stride * sourceRow]
+			}
+		}
+	}
+	_resizeBilinear: static func (source, result: This) {
+		resultBuffer := result buffer pointer
+		sourceBuffer := source buffer pointer
+		for (row in 0 .. result size height) {
+			sourceRow := ((source size height as Float) * row) / result size height
+			sourceRowUp := sourceRow floor() as Int
+			weightDown := sourceRow - sourceRowUp as Float
+			sourceRowDown := (sourceRow - weightDown) as Int + 1
+			rowDownValid := sourceRowDown < source size height
+			if (!rowDownValid)
+				weightDown = 0.0f
+			for (column in 0 .. result size width) {
+				sourceColumn := ((source size width as Float) * column) / result size width
+				sourceColumnLeft := sourceColumn floor() as Int
+				weightRight := sourceColumn - sourceColumnLeft as Float
+				sourceColumnRight := (sourceColumn - weightRight) as Int + 1
+				columnRightValid := sourceColumnRight < source size width
+				if (!columnRightValid)
+					weightRight = 0.0f
+				valueRowUp := (1.0f - weightRight) * sourceBuffer[sourceColumnLeft + sourceRowUp * source stride]
+				if (columnRightValid)
+					valueRowUp += weightRight * sourceBuffer[sourceColumnRight + sourceRowUp * source stride]
+				valueRowDown := 0.0f
+				if (rowDownValid)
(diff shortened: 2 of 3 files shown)


source/draw/Image.ooc
```diff
@@ -26,6 +26,11 @@ CoordinateSystem: enum {
 	YUpward = 0x02
 }
 
+TransformMethod: enum {
+	Fast,
+	Smooth
+}
```
sebastianbaginski: Do we need anything more here ? I thought it's better to have it like that than
```ooc
TransformMethod: enum {
    NearestNeighbour,
    Bilinear
}
```
As a user of this class, I wouldn't really care about what method is used under the hood. I would simply want to have a `fast` or `smooth` method of rescaling.
What do you think ?

davidhesselbom: True, but you could also leave a comment for each entry
```ooc
TransformMethod: enum {
    Fast, // nearest neighbor,
    Smooth // bilinear
}
```

sebastianbaginski: OK, done.

source/draw/RasterMonochrome.ooc
```diff
@@ -49,21 +49,63 @@ RasterMonochrome: class extends RasterPacked {
...
-			for (row in 0 .. size height) {
-				srcRow := (this size height * row) / size height
-				for (column in 0 .. size width) {
-					srcColumn := (this size width * column) / size width
-					result[column, row] = this[srcColumn, srcRow]
-				}
+			match (method) {
+				case TransformMethod Fast => This _resizeNearestNeighbour(this, result)
+				case TransformMethod Smooth => This _resizeBilinear(this, result)
+				case =>
```
davidhesselbom: So in the default case, return an empty image? Do you think we should default to `NearestNeighbor`, maybe?

sebastianbaginski: But `method` enum has only two values. Is it even possible to hit the default case ?

davidhesselbom: What do you mean? `Fast` could be the default:
```ooc
match (method) {
	case TransformMethod Smooth => This _resizeBilinear(this, result)
	case => This _resizeNearestNeighbour(this, result)
```

davidhesselbom: I mean, if someone were to call this function with an invalid `method`.

sebastianbaginski: Yes your code is correct, but I mean, when enum has two values and you explicitly create cases for all of them in a match clause, is it even possible to fall into the last, default case ?

sebastianbaginski: Oh so it is possible in ooc.

davidhesselbom: Someone could cast `1337 as TransformMethod`, I guess. We *could* do it this way and return an empty image when someone uses an invalid `TransformMethod`, of course. We could also throw an exception. I'm just asking what you think.

sebastianbaginski: > cast 1337 as TransformMethod

I wasn't aware that such abomination is allowed in ooc :) Let's do that your way, I think it's the safest.

source/draw/Image.ooc
```diff
@@ -66,6 +71,9 @@ Image: abstract class {
 		this resizeTo(((this size toFloatSize2D()) * Float minimum(restriction width as Float / this size width as Float, restriction height as Float / this size height as Float)) toIntSize2D())
 	}
 	resizeTo: abstract func (size: IntSize2D) -> This
+	resizeTo: virtual func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
+		this resizeTo(size)
```
davidhesselbom: `method` isn't used in this function.

source/draw/RasterMonochrome.ooc
```diff
@@ -49,21 +49,62 @@ RasterMonochrome: class extends RasterPacked {
 			}
 	}
 	resizeTo: override func (size: IntSize2D) -> This {
+		this resizeTo(size, TransformMethod Smooth) as This
+	}
+	resizeTo: override func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
```
davidhesselbom: Why not one function, but with a default value for `method`?

source/draw/Image.ooc
```diff
@@ -27,6 +27,11 @@ CoordinateSystem: enum {
 	YUpward = 0x02
 }
 
+TransformMethod: enum {
```
davidhesselbom: This won't affect `GpuImage` at all, will it?

source/draw/Image.ooc
sebastianbaginski: Yes, it is meant to be reimplemented in derived classes. If a derived class *knows* how to resize itself according to given method, then it should reimplemented this function. Otherwise it just defaults to standard `resizeTo`.

source/draw/RasterMonochrome.ooc
```diff
@@ -49,21 +49,62 @@ RasterMonochrome: class extends RasterPacked {
...
+				case => This _resizeNearestNeighbour(this, result)
 			}
 		}
 		result
 	}
+	_resizeNearestNeighbour: static func (source, result: This) {
+		resultBuffer := result buffer pointer
+		sourceBuffer := source buffer pointer
+		for (row in 0 .. result size height) {
+			srcRow := (source size height * row) / result size height
```
davidhesselbom: `sourceRow`

source/draw/RasterMonochrome.ooc
```diff
@@ -49,21 +49,62 @@ RasterMonochrome: class extends RasterPacked {
...
+				srcColumn := (source size width * column) / result size width
+				resultBuffer[column + result stride * row] = sourceBuffer[srcColumn + source stride * srcRow]
+			}
+		}
+	}
+	_resizeBilinear: static func (source, result: This) {
+		resultBuffer := result buffer pointer
+		sourceBuffer := source buffer pointer
+		for (row in 0 .. result size height) {
+			srcRow := ((source size height as Float) * row) / result size height
```
davidhesselbom: `sourceRow`

test/draw/RasterMonochromeTest.ooc
```diff
@@ -56,17 +56,34 @@ RasterMonochromeTest: class extends Fixture {
...
 			expect(image2 size == image size / 2)
-			image2 = image2 resizeTo(image size)
-			expect(image2 size == image size)
-			expect(image distance(image2) < image size width / 10)
-			image2 = image2 resizeTo(size)
-			expect(image2 size == size)
+			image3 := image2 resizeTo(image size)
+			expect(image3 size == image size)
+			expect(image distance(image3) < image size width / 10)
+			image3 free()
```
davidhesselbom: `image3 referenceCount decrease()`

source/draw/RasterMonochrome.ooc
sebastianbaginski: I think I had some problems with virtual functions and default values for arguments. I will try again.

davidhesselbom: Well, that's possible. Should be doable, though...

source/draw/Image.ooc
sebastianbaginski: I don't think so.
Ok now that you mentioned it I remember why I don't have a single `resizeTo` with two arguments. It is because all derived classes currently have the single argument version of `resizeTo`. If you remove it and add the two argument version with default second argument, then compiler will complain that all derived image classes should reimplement it. I wanted to keep the changes at minimum.

davidhesselbom: Alright.

sebastianbaginski:
Updated.

source/draw/Image.ooc
davidhesselbom: So... do we plan to let this enum be used by `GpuImage` as well? Otherwise, it should be moved to `RasterImage`.

sebastianbaginski: I think it could be nice to use it with any class that implements the `Image` *interface*.

davidhesselbom: @emilwestergren , do you think there's a way we could use this in the GpuImages? I'm not asking for it to happen right now, obviously.

emilwestergren: 1. I don't like the name `TransformMethod`, it's more a `Filter` or `Interpolation` setting, or if you have a better suggestion. Other than that we should have it in `Image` and adapt the code in `GpuImage`.
2. We need to decide if this should be a state for the Image or a parameter. Currently for GpuImages, filtering is a state for the Image that is being drawn (sampled). The good thing about this is that you don't have to supply it as a parameter and it also handles the case with draw calls where you have multiple images as input (for example in a shader) and only some of them should use bilinear filtering. Also, in this specific case there would be no need for an overloaded `resizeTo`.

davidhesselbom: Looks like I shouldn't have merged this until I got your answer, but I'll open an issue with those concerns instead.

PR #532, opened 2015-10-12
Pbo debug fix.

ziquanyu:
bug fix in Pbo.

### DIFF
--- a/source/draw/gpu/opengl/backend/gles3/Gles3FramebufferObject.ooc
+++ b/source/draw/gpu/opengl/backend/gles3/Gles3FramebufferObject.ooc
@@ -81,12 +81,14 @@ Gles3FramebufferObject: class extends GLFramebufferObject {
 		this bind()
 		glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, texture _backend, 0)
 		status: UInt = glCheckFramebufferStatus(GL_FRAMEBUFFER)
-		if (status != GL_FRAMEBUFFER_COMPLETE) {
-			statusMessage := getErrorMessage(status)
-			errorMessage := "Framebuffer Object creation failed with status: " + statusMessage + " for texture of size " +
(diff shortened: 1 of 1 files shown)


emilwestergren:
This shouldn't be merged, if glCheckFramebufferStatus returns GL_NO_ERROR then something is wrong and we want to catch it.

sebastianbaginski:
> function returns NO_ERROR -> something is wrong

LOL :P Feels like win32api

davidhesselbom:
... What does `glCheckFramebufferStatus` return when there is, in fact, no error, I wonder?

emilwestergren:
"If the framebuffer is complete, then `GL_FRAMEBUFFER_COMPLETE` is returned."
"Additionally, if an error occurs, zero is returned."
`#define GL_NO_ERROR                       0`
https://www.khronos.org/opengles/sdk/docs/man/xhtml/glCheckFramebufferStatus.xml

davidhesselbom:
Confusing. Thanks for the link, though!

ziquanyu:
If the error queue is empty, it will return GL_NO_ERROR. 

Maybe we can have a better way to check if there's error and if frameBuffer is complete. It's wired to have a raise saying error code : no error!


I've got one situation: glCheckFramebufferStatus resturn 0 instread for GL_FRAMEBUFFER_COMPLETE when it tried to generate Fbo.

ziquanyu:
@davidhesselbom 
https://www.opengl.org/wiki/OpenGL_Error

emilwestergren:
This error is not from the error queue, it's a status value from the `glCheckFramebufferStatus` function. The documentation clearly states that if it doesn't return `GL_FRAMEBUFFER_COMPLETE` then there is an error. So for this function, 0 is not equal to GL_NO_ERROR, it is simply an unknown error.

ziquanyu:
This is true.

thomasfanell:
@sebastianbaginski like ERROR_SUCCESS :+1:

PR #535, opened 2015-10-13
GpuOff flag

jonathanudd:
Introduced a compile flag to version out all gpu related parts.


source/draw/gpu/opengl/backend/egl/egl.ooc
```diff
@@ -14,7 +14,7 @@
 * You should have received a copy of the GNU Lesser General Public License
 * along with this software. If not, see <http://www.gnu.org/licenses/>.
 */
-
```
davidhesselbom: No, keep the empty line here.

PR #536, opened 2015-10-13
Run kean tests on appveyor

jonathanudd:
- The yml file for appveyor installs rock and then run the tests for kean.
It runs the tests without gpu and also with the compile flag `-DgpuOff` and for the tests to not fail PR #535 is needed because it introduced `-DgpuOff`.
- A step for installing a windows version of rock in the install_rock script is also added.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,14 @@
+version: 1.0.{build}
+os: Windows Server 2012 R2
+environment:
+  PATH: C:\MinGW\bin;C:\msys64\usr\bin;%PATH%
+install:
+- cmd: >-
+    bash install_rock.sh
+
+    mv rock.exe C:\msys64\usr\bin\
+
+    mingw-get install pthreads
+build: off
+test_script:
+- cmd: bash test.sh nogpu
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 1.0.{build}
+version: 0.1.{build}
 os: Windows Server 2012 R2
 environment:
   PATH: C:\MinGW\bin;C:\msys64\usr\bin;%PATH%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,4 +11,4 @@ install:
     mingw-get install pthreads
 build: off
 test_script:
-- cmd: bash test.sh nogpu
+- cmd: bash test.sh nogpu -DgpuOff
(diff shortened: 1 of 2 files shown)


jonathanudd:
Someone also needs to create a project for cogneco kean on appveyor when this is merged.

appveyor.yml
```diff
@@ -0,0 +1,14 @@
+version: 1.0.{build}
+os: Windows Server 2012 R2
+environment:
+  PATH: C:\MinGW\bin;C:\msys64\usr\bin;%PATH%
+install:
+- cmd: >-
```
davidhesselbom: What does `>_` do?

install_rock.sh
```diff
@@ -1,6 +1,14 @@
...
+	wget https://github.com/cogneco/rock/releases/$tmp
+	echo ${tmp##*/}
+	7z e ${tmp##*/}
+else
+	tmp="$(wget -O - https://github.com/cogneco/rock/releases/latest | grep -o '\<download/rock_.*deb\>')"
+	wget https://github.com/cogneco/rock/releases/$tmp
+	echo ${tmp##*/}
+	# dpkg-deb -x ${tmp##*/} .
+	# mv usr/bin/rock .
+	# mv usr/lib/fancy_backtrace.so .
```
davidhesselbom: Commenting this part out causes Circle CI to fail. Change it back.

appveyor.yml
jonathanudd: I don't know actually, I made all the settings on the web GUI for appveyor and then downloaded the yml file based on the settings.

install_rock.sh
jonathanudd: Fixed it as soon as I saw the Circle CI test fail. It was not intentional but a mistake to leave the comment there.

appveyor.yml
davidhesselbom: Ah, that's why there's so much stuff in here.

appveyor.yml
```diff
@@ -0,0 +1,14 @@
+version: 1.0.{build}
```
davidhesselbom: Should it really be 1.0.anything though?

jonathanudd: Maybe just `version: {build}`?

davidhesselbom: Is the `version` part required, btw? Until we start making releases, I don't see the point.

jonathanudd: It is the unique identifier for a build so it is required.

davidhesselbom: Alright...

jonathanudd: Do you want me to changed it to `version: {build}` or just leave it as it is?

davidhesselbom: Well... what happens if we realize that we actually want to use a real version number and we've already made several builds named 1.0? Feels like `0.1.{build]` would be a better idea. Is `{build}` even a valid version number?

jonathanudd: `{build}` is a valid version number as far as appveyor knows. But I guess its not a good version number to use.

PR #537, opened 2015-10-13
Fixed invalid memory reads related to Raster* conversion methods

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/522 .
Detailed description of the problem is in the issue.

### DIFF
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -35,19 +35,17 @@ RasterBgr: class extends RasterPacked {
 	create: func (size: IntSize2D) -> Image { This new(size) }
 	copy: func -> This { This new(this) }
 	apply: func ~bgr (action: Func(ColorBgr)) {
-		end := this buffer pointer as Long + this buffer size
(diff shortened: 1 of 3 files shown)


source/draw/RasterBgr.ooc
```diff
@@ -39,19 +39,17 @@ RasterBgr: class extends RasterPacked {
...
+			for (pixel in 0 .. this size width) {
+				pointer := this buffer pointer + pixel * this bytesPerPixel + row * this stride
+				color := (pointer as ColorBgr*)@
+				action(color)
 			}
-			row += this stride - 1
-		}
 	}
 	apply: func ~yuv (action: Func(ColorYuv)) {
-		this apply(ColorConvert fromBgr(action))
```
sebastianbaginski: This leaked memory.

sebastianbaginski:
Looks like these *invalid reads* were not really related to YUV images, as you can find them also in `ImageFileTest` when running for example `RasterMonochrome -> RasterBgra` conversion. 
These errors are also fixed by this pull request.

PR #538, opened 2015-10-13
Collections tests cleanup

marcusnaslund:
Added missing coverage and cleaned up tests for `VectorList` `VectorQueue` and `SynchronizedQueue`.
With the exception of `HashDictionary` (separate PR) the test coverage is complete in `collections` now.


test/collections/VectorListTest.ooc
```diff
@@ -21,6 +21,7 @@ use ooc-collections
 VectorListTest: class extends Fixture {
 	init: func {
 		super("VectorList")
+		tolerance := 0.00001f
```
davidhesselbom: Next time, consider using `1.0e-5f` to make it easier to count the zeroes.

PR #542, opened 2015-10-13
Valgrind suppression file to filter out Fixture from logs

sebastianbaginski:
I use this file to suppress memory leaks from `Fixture`. I know that @thomasfanell wants to rewrite it, but maybe someone will find this file useful until then.

### DIFF
--- a/kean.supp
+++ b/kean.supp
@@ -1,0 +1,18 @@
+{
+   fixture_getis
+   Memcheck:Leak
+   ...
+   fun:Fixture__Fixture___getis__
+}
+{
+   fixture_expect
+   Memcheck:Leak
+   ...
+   fun:Fixture__Fixture_expect_*
+}
+{
+   fixture_getequal
+   Memcheck:Leak
+   ...
+   fun:Constraints__IsConstraints___getequal__
+}


kean.supp
```diff
@@ -0,0 +1,66 @@
...
+   fixture
+   Memcheck:Leak
+   ...
+   fun:Fixture__Fixture_expect_double
+}
+{
+   fixture
+   Memcheck:Leak
+   ...
+   fun:Fixture__Fixture_expect_ldouble
```
davidhesselbom: `fun:Fixture__Fixture_expect_*`

should cover most of these cases.

kean.supp
```diff
@@ -0,0 +1,66 @@
...
+   fun:Fixture__Fixture_expect_float
+}
+{
+   fixture
+   Memcheck:Leak
+   ...
+   fun:Fixture__Fixture_expect_double
+}
+{
+   fixture
```
davidhesselbom: I think these "names" are supposed to be unique to each suppression in the file.

kean.supp
sebastianbaginski: You're right, updated.

PR #545, opened 2015-10-13
Plot

davidpiuva:
Minimum and maximum funcitons in PlotData2D have been made abstract to allow different plotters to store different types of information. Then I could move dataSeries from PlotData2D to LinePlotData2D and ScatterPlotData2D.

A new plotter called ShapePlotData2D has a list of free shapes to draw. Each shape can override the default color using an optional argument.

This change may break backward compability in compile time if someone tries to access dataSeries from the abstract class instead of the specific type of plotter since they do not have the same data list any more.

### DIFF
--- a/ooc-plot.use
+++ b/ooc-plot.use
@@ -3,6 +3,7 @@ Description: Plot functionality library.
 SourcePath: source/plot
 Imports: LinePlotData2D
 Imports: ScatterPlotData2D
+Imports: ShapePlotData2D
 Imports: PlotData2D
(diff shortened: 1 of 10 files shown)


davidhesselbom:
This probably at least borderline "too complex", but I need to give it an honest attempt before I can decide. If you can, please split this up into smaller pieces.

davidpiuva:
Divided into shape and round PR.

davidhesselbom:
Thank you.

Issue #508, opened 2015-10-07
Int modulo


davidhesselbom:
`Int modulo(Int)`, defined in `math.ooc`, behaves oddly.

```ooc
import math

(5 % 3) toString() println() // outputs 2.
(5 % -3) toString() println() // also outputs 2.
(-5 % 3) toString() println() // outputs -2, but should output 1. 
// Too bad, but that's how the operator works in C.
((-5) % 3) toString() println() // also outputs -2.
(-5 % -3) toString() println() // also outputs -2.

(5 modulo(3)) toString() println() // outputs 2, as expected.
(5 modulo(-3)) toString() println() // outputs 1. Is this correct?
(-5 modulo(3)) toString() println() // outputs -2, but it should output 1. 
// Unlike the % operator, modulo() was written explicitly to handle this situation,
// but it still doesn't work correctly.
((-5) modulo(3)) toString() println() // outputs 1, which is what we want.
// So... it's not a bug in the `modulo` function itself.
(-5 modulo(-3)) toString() println() // outputs -1. Is this correct?
((-5) modulo(-3)) toString() println() // outputs 2. Is this correct?
```

We're not going to change the behavior of the `%` operator, but the `modulo` function that we added should work the way we want it to. Currently, it doesn't for `-5 modulo(3)`, but what about the cases where the divisor is negative? Should `modulo` always return a positive number?

davidpiuva:
A % B equals the largest X satisfying the equation A + B * I = X < B.

sebastianbaginski:
```ooc
(-5 % 3) toString() println() // outputs -2, but should output 1. 
// Too bad, but that's how the operator works in C.
(-5 % -3) toString() println() // also outputs -2.
```

The result is not specified by the C standard for negative operands. It depends entirely on the compiler, so I guess it is not safe to assume anything about the result in C. It can be `-2` now, but you can switch from `gcc` to `clang` and the same C code suddenly returns `1`.

davidpiuva:
There seems to be many dialects of modulo being defined for negative divisors that different compilers might use.
https://en.wikipedia.org/wiki/Modulo_operation

davidhesselbom:
Okay, but let's forget about the `%` operator for now, and focus on our `modulo` function. What should its behavior be for negative divisors? Should we always return a non-negative value?

> A % B equals the largest X satisfying the equation A + B * I = X < B.

Do you mean `X < |B|` ?

davidpiuva:
[...]

PR #547, opened 2015-10-13
Added TimeSpan toString()

### LINKED ISSUE #524: TimeSpan has no toString() function

marcusnaslund:
Also had to add `modulo(Int)` for `Int64` types. We could inline this directly if we want to avoid this dependancy. You choose.
Fixes #524


source/sdk/math.ooc
```diff
@@ -59,6 +59,10 @@ floor: extern(floorl) func ~Long (LDouble) -> LDouble
 	 maximumValue ::= static SHRT_MAX
 }
 
+extend Int64 {
+	modulo: func(divisor: Int) -> This { this - divisor * (this / divisor) }
+}
```
davidhesselbom: 1. This implementation differs from the one for `Int`. That means at least one of them is wrong. Perhaps now would be a good time to decide which one is correct?
2. Shouldn't `divisor` also be an `Int64`, i.e. `divisor: This`?

marcusnaslund: 1. It differs in how it handles negative numbers, if that's what you mean. Yes, we need to decide that. Guess that discussion belongs to #508 
2. Yep, you're right, will fix.

davidhesselbom:
Did we decide on anything final in #508? I'd like to do that before I merge this.

marcusnaslund:
No we did not, but fixing that (if anything needs fixing) would be a different PR, which could also include this.
For all our purposes, the modulo here works. /Mr. Pragmatist

source/sdk/math.ooc
```diff
@@ -59,6 +59,13 @@ floor: extern(floorl) func ~Long (LDouble) -> LDouble
 	 maximumValue ::= static SHRT_MAX
 }
 
+extend Int64 {
+	modulo: func(divisor: This) -> This { 
+		result := this - (this / divisor) * divisor
+		result < 0 ? result + divisor : result
+	}
+}
+
```
davidhesselbom: Okay, so exactly how does the behavior of this function differ from the one for `Int`?

marcusnaslund: The `Int` version cares if `divisor < 0`. From the decision in #508, we should not.

davidhesselbom: Okay. I'd *really* like to avoid having these two functions behave differently, but at the same time, modifying the other one here seems risky. Can you make an issue anda top priority to update the one for `Int` as soon as we're done with this merge sprint?

marcusnaslund: Sure.

marcusnaslund: Done, updated #508

source/base/TimeSpan.ooc
```diff
@@ -43,6 +44,38 @@ TimeSpan: cover {
...
+	//  %m - minutes (<60)
+	//  %s - seconds (<60)
+	//  %z - milliseconds (<1000)	
+	//  %z - milliseconds (<1000)
+	//  %D - days (based on total ticks)
+	//  %H - hours (based on total ticks)
+	//  %M - minutes (based on total ticks)
+	//  %S - seconds (based on total ticks)
+	//  %Z - milliseconds (based on total ticks)	
+	toStringFormat: func (format: String) -> String {
```
simonmika: This should be:
```ooc
toString: func ~format (format: String) -> String {
```

marcusnaslund: Wouldn't have been consistent with `DateTime`.

davidhesselbom: Alright - it'll need fixing in `DateTime` as well.

simonmika: The rule is to _always_ use overloads when possible.

marcusnaslund: I will create an issue.

PR #550, opened 2015-10-14
Fixed bug that broke gaussian kernel calculations.

niklasborwell:


source/math/FloatVectorList.ooc
```diff
@@ -224,7 +224,7 @@ FloatVectorList: class extends VectorList<Float> {
 		result := This new(size)
 		factor := 1.0f / (sqrt(2.0f * Float pi) * sigma)
 		for (i in 0 .. size)
-			result add(factor * pow(Float e, -0.5f * ((i - (size - 1.0f) / 2.0f) squared()) / (sigma squared())))
+			result add((factor * pow(Float e, -0.5f * ((i - (size - 1.0f) / 2.0f) squared()) / (sigma squared()))) as Float)
```
davidhesselbom: What part of this formula turned everything into something other than a `Float`? Does `Double` need casting into `Float` before it can be added to a `FloatVectorList`?

niklasborwell: I don't know, I didn't investigate it. It felt like an "ooc moment" to me so I just fixed it and moved on. According to my understanding this should already be a Float, first when `factor` is defined and again when used as an argument.

davidhesselbom: > "ooc moment"

That would be a "rock moment", but yeah :D

PR #553, opened 2015-10-14
Fixed race condition causing deadlock.

emilwestergren:


fredrikbryntesson:
Builds and run on both pc and android, memcheck ok, magic ok.

source/base/ThreadPool.ooc
```diff
@@ -54,8 +55,8 @@ _ActionTask: class extends _Task {
 _ResultTask: class <T> extends _Task {
 	_result: Cell<T>
 	_action: Func -> T
-	init: func (=_action, =_mutex)
-	run: func (=_mutex) {
```
davidhesselbom: Wow. Did I merge this? ...

marcusnaslund:
My bad. 
There are other problems as well (like trying to join a thread more than once), which are still waiting in #509.

source/base/ThreadPool.ooc
davidhesselbom: Wait, I remember now. It wasn't as big a problem as I thought it was. Still, this new solution is much better.

PR #555, opened 2015-10-15
First step in dissolving sdk: math/Random

marcusnaslund:
Completely removes `sdk/math/Random`.
Randomness in `sdk/structs/List` is replaced with what goes on in `IntRandomGenerator` (I did not use the generator to avoid potential/future dependancy issues).


source/sdk/structs/List.ooc
```diff
@@ -194,10 +194,12 @@ List: abstract class <T> extends BackIterable<T> {
...
         indexes := ArrayList<Int> new()
         for(i in 0..getSize()) indexes add(i)
+        currentIndex := Time microtime()
 
         while(!indexes empty?()) {
-            i := Random randRange(0, indexes getSize())
-            shuffled add(this[indexes removeAt(i)])
+            currentIndex = 214013 * currentIndex + 2531011 // Copied directly from source/math/IntRandomGenerator
+            currentIndex = ((currentIndex >> 16) & Short maximumValue) % indexes getSize()
+            shuffled add(this[indexes removeAt(currentIndex)])
```
davidhesselbom: You could have simply removed the whole `shuffle` function and no one would have noticed, but alright, why not.

PR #559, opened 2015-10-16
Cleaned and improved ByteBuffer

emilwestergren:


source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -77,9 +77,10 @@ AndroidContext: class extends OpenGLContext {
 		sourcePointer := eglImage buffer lock(false)
 		length := channels * eglImage size width * eglImage size height
 		buffer := ByteBuffer new(sourcePointer, length,
-			func (buffer: ByteBuffer) {
+			func (b: ByteBuffer) {
 				eglImage buffer unlock()
 				this recyclePacker(gpuRgba)
+				b forceFree()
```
davidhesselbom: Why does the buffer need to be freed so forcefully in this case?

emilwestergren: Because this closure defines what should happen when the ByteBuffer gets freed. I only want to recycle the backend memory and free the object data for the ByteBuffer. A call to `free` would result in an infinite loop.

source/base/ByteBuffer.ooc
```diff
@@ -28,26 +28,29 @@ ByteBuffer: class {
...
-		memset(_pointer + offset, 0, length)
-	}
-	slice: func (offset, size: Int) -> This {
-		_SlicedByteBuffer new(this, offset, size)
+	forceFree: func {
+		this _forceFree = true
+		this free()
 	}
+	zero: func ~whole { memset(_pointer, 0, _size) }
+	zero: func ~range (offset, length: Int) { memset(_pointer + offset, 0, length) }
```
marcusnaslund: `this _pointer`

source/base/ByteBuffer.ooc
```diff
@@ -61,68 +64,54 @@ ByteBuffer: class {
...
-	}
+	clean: static func { _RecyclableByteBuffer _clean() }
 }
 _SlicedByteBuffer: class extends ByteBuffer {
 	_parent: ByteBuffer
 	_offset: Int
 	init: func (=_parent, =_offset, size: Int) {
 		_parent referenceCount increase()
-		super(_parent pointer + _offset, size)
+		super(_parent pointer + _offset, size, false)
```
marcusnaslund: `false` is already the default value

emilwestergren:
Updated.

source/draw/gpu/opengl/AndroidContext.ooc
marcusnaslund: I agree `buffer` wasn't a smart name here, @davidhesselbom isn't usually happy with names like `b` either but I'll let him decide for this case. Maybe `bytebuffer`.
Also - rock understands that you actually want a `_RecoverableByteBuffer`? That's nice.

marcusnaslund:
Besides the comment above it looks good @davidhesselbom

source/draw/gpu/opengl/AndroidContext.ooc
emilwestergren: Yes, due to this:
`new: static func ~recover (pointer: UInt8*, size: Int, recover: Func (This)) -> This {
	}`

marcusnaslund: Ah, I see.

source/base/ByteBuffer.ooc
```diff
@@ -28,26 +28,29 @@ ByteBuffer: class {
 	size ::= this _size
 	_referenceCount: ReferenceCounter
 	referenceCount ::= this _referenceCount
-	init: func (=_pointer, =_size) {
+	_owner: Bool
```
davidhesselbom: This is a strange name for a `Bool`. I can't think of a better one that doesn't also sound silly, though. `_ownsItself`, `_hasOwner`, `_isHisOwnMaster`, etc. are all obviously bad, too.

source/base/ByteBuffer.ooc
```diff
@@ -28,26 +28,29 @@ ByteBuffer: class {
...
+	_forceFree := false
```
davidhesselbom: I would prefer `_isRecyclable` or something. `forceFree` sounds like a function (and it is).

source/base/ByteBuffer.ooc
```diff
@@ -28,26 +28,29 @@ ByteBuffer: class {
...
 	}
-	zero: func ~whole {
-		memset(_pointer, 0, _size)
-	}
-	zero: func ~range (offset, length: Int) {
-		memset(_pointer + offset, 0, length)
-	}
-	slice: func (offset, size: Int) -> This {
-		_SlicedByteBuffer new(this, offset, size)
+	forceFree: func {
```
davidhesselbom: I still don't like the fact that we apparently need this. No one who uses a `ByteBuffer` should have to keep track of how it frees itself, it should be abstracted and kept inside the class. Can we find another way?

source/base/ByteBuffer.ooc
emilwestergren: `_ownsMemory` ` _isOwner` ?

emilwestergren: It's used to know if the ByteBuffer is responsible of freeing the allocated memory on `free`

davidhesselbom: Yeah, I understand, I'm just having a hard time coming up with a good name.

source/base/ByteBuffer.ooc
emilwestergren: It's not only for Recyclables though, it is used for Recoverables and any other ByteBuffer we might implement.

source/base/ByteBuffer.ooc
sebastianbaginski: In vectorList it is called `_freeContent`

source/base/ByteBuffer.ooc
emilwestergren: We do have Recoverable ByteBuffers where you specifically define how it frees itself. If I want to make something and just free it normally after, how am I supposed to achieve this without a state or function in the public ByteBuffer API?
` buffer := ByteBuffer new(pointer, length, func { /* First make something... but how do I free the ByteBuffer after? */  })`

source/base/ByteBuffer.ooc
emilwestergren: That works for me.

source/base/ByteBuffer.ooc
emilwestergren: Hmm, maybe the recover closure can return a Bool which decides if it should free it after?

source/base/ByteBuffer.ooc
davidhesselbom: In `VectorList` it refers to each element in the list though, not the backend itself, as is the case here. Still, I guess it would work. Thanks for the suggestion.

emilwestergren: `_freeBackend` ?

davidhesselbom: Really, `_pointer` is a bad name, too, so it should probably be `_buffer` and then you could call it `_freeBuffer`. On the other hand,

```ooc
if (this _ownsTheBuffer)
    // do stuff
```
reads a lot better than
```ooc
if (this _freeBuffer)
    // do stuff
```

source/base/ByteBuffer.ooc
davidhesselbom: Yeah... maybe.

I think I should probably just merge this and leave the details in a PR debt issue instead.

emilwestergren: Sure, it is still an improvement from before. The Bool-returning closure is a simple fix though if it is important to you.

davidhesselbom: I think maybe a better solution is to create a ByteBuffer type that cannot actually be recycled, i.e. one that uses the same `free` function as the base class.

source/base/ByteBuffer.ooc
emilwestergren: So what's the verdict here? `_freeBuffer` or `ownsBuffer` ?

source/base/ByteBuffer.ooc
davidhesselbom: But yes, let's leave it until later.

emilwestergren: Although then `forceFree` wouldn't work for Recyclable buffers... let's just leave it then.

source/base/ByteBuffer.ooc
davidhesselbom: Leave it for now, I created #698 to remind us.

Issue #561, opened 2015-10-16
Plenty of TODO and FIXME in Raster* classes


marcusnaslund:
Most of the `TODO`s and `FIXME`s in kean are in the `Raster*` classes such as `RasterMonochrome`, `RasterYuv420SemiPlanar` etc.

These are mostly related to commented-out sections of code or whole functions.

- Do we (still) need/want them?
- Are there ways to improve what's there?
- Do we now know how to solve these problems?

I'm making this issue so that this doesn't go unnoticed forever.

davidhesselbom:
Can't give you an answer that covers all the cases, but many of them can probably be removed, and other can be fixed. For instance, in `RasterMonochrome` (and other related classes) `open`, memory is copied unnecessarily:

```ooc
		buffer := ByteBuffer new(x * y * requiredComponents)
		// FIXME: Find a better way to do this using Dispose() or something
		memcpy(buffer pointer, data, x * y * requiredComponents)
		StbImage free(data)
```

This isn't a high priority, but it would be nice to get it fixed, and it deserves a `FIXME` until it is indeed fixed.

davidhesselbom:
```ooc
//	FIXME
//	openResource(assembly: ???, name: String) {
//		Image openResource
//	}
```
can be removed, everywhere.

marcusnaslund:
All `openResource` FIXMEs removed with #563

davidhesselbom:
Great. Btw, all `StbImage free` does under the hood is `free(data)`, so fixing that needless copy should be a piece of cake:
```ooc
buffer := ByteBuffer new(data, x * y * requiredComponents)
```
instead of 
```ooc
		buffer := ByteBuffer new(x * y * requiredComponents)
		// FIXME: Find a better way to do this using Dispose() or something
		memcpy(buffer pointer, data, x * y * requiredComponents)
		StbImage free(data)
```

davidhesselbom:
While you're at it, `requiredComponents` can be replaced by `this bytesPerPixel`, and boom, the function can be made identical for nearly all `RasterPacked` classes...

marcusnaslund:
Awesome. How about the recurring
```ooc
//		else if (!other instanceOf?(This))
//			FIXME
//		else if (this size != other size)
//			FIXME
```
in `distance()`?

sebastianbaginski:
I think in the first case we should do the comparison anyway (convert the other to `This` and call `distance()`).
[...]

PR #567, opened 2015-10-19
More todo fixme cleanup

marcusnaslund:
Mostly changes to `open()` in `Raster*` classes. "Fixes" #561 
Removed commented-out code in `Synchronized` (we never use try-catch anywhere else, so...)
And fixed `TODO`s in `Vector`.

### DIFF
--- a/source/collections/Vector.ooc
+++ b/source/collections/Vector.ooc
@@ -112,15 +112,15 @@ HeapVector: class <T> extends Vector<T> {
 
 	operator [] (index: Int) -> T {
 		version (safe) {
-			if (index >= this capacity || index < 0 || index >= this capacity)
-				raise("Accessing Vector index out of range in set operator")
+			if (index >= this capacity || index < 0)
+				raise("Accessing Vector index out of range in get operator")
 		}
 		this _backend[index]
 	}
 
 	operator []= (index: Int, item: T) {
 		version (safe) {
-			if (index >= this capacity || index < 0 || index >= this capacity)
+			if (index >= this capacity || index < 0)
 				raise("Accessing Vector index out of range in set operator")
 		}
 		if (this _freeContent && T inheritsFrom?(Object)) {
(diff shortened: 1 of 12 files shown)


source/collections/Vector.ooc
```diff
@@ -121,9 +120,7 @@ HeapVector: class <T> extends Vector<T> {
 
 	operator []= (index: Int, item: T) {
 		version (safe) {
-			// TODO: Should we also check that index < count?
-			// TODO: Should this affect count, if index > count?
-			if (index >= this capacity || index < 0)
+			if (index >= this capacity || index < 0 || index >= this capacity)
```
davidhesselbom: Derp, you're checking `index >= this capacity` twice...

source/draw/RasterBgr.ooc
```diff
@@ -105,14 +105,10 @@ RasterBgr: class extends RasterPacked {
 		}
 	}
 	open: static func (filename: String) -> This {
-		x, y, n: Int
+		x, y, totalComponents: Int
```
davidhesselbom: I think `imageComponents` would be a better name... don't you?

source/collections/Vector.ooc
marcusnaslund: If the computer miscomputes the first time, it's good to check again. :)

marcusnaslund: Also: oops. The `TODO`s refer to `count` which doesn't even exist here. I didn't see the whole if-statement once I corrected for that.

davidhesselbom: Right, `count` needs to be checked by `VectorList`.

source/draw/RasterBgr.ooc
marcusnaslund: Hmm... yes.

marcusnaslund:
Updated. 
`VectorList` now checks `count` (we don't need to check if it's `< 0` here).
Renamed `totalComponents` to `imageComponents`.

PR #571, opened 2015-10-19
Cleaned up GraphicBuffer alignWidth()

marcusnaslund:
The `TODO` said to remove the check for empty array, I put it in a `safe` block instead.

### DIFF
--- a/source/draw/gpu/opengl/GraphicBuffer.ooc
+++ b/source/draw/gpu/opengl/GraphicBuffer.ooc
@@ -93,13 +93,8 @@ GraphicBuffer: class {
 		memcpy(This _alignedWidth data, alignedWidth, count * Int size)
 	}
 	alignWidth: static func (width: Int, align := AlignWidth Nearest) -> Int {
-		result: Int
-		version (safe) {
-			if (This _alignedWidth length == 0)
-				result = width
-			else
-				result = align == AlignWidth Ceiling ? This _alignedWidth[This _alignedWidth length-1] : This _alignedWidth[0]
-		} else
+		result := width
+		if (This _alignedWidth length > 0)
 			result = align == AlignWidth Ceiling ? This _alignedWidth[This _alignedWidth length-1] : This _alignedWidth[0]
 		for (i in 0 .. This _alignedWidth length) {
 			currentWidth := This _alignedWidth[i]
--- a/source/draw/gpu/opengl/GraphicBuffer.ooc
+++ b/source/draw/gpu/opengl/GraphicBuffer.ooc
@@ -93,37 +93,17 @@ GraphicBuffer: class {
 		memcpy(This _alignedWidth data, alignedWidth, count * Int size)
 	}
 	alignWidth: static func (width: Int, align := AlignWidth Nearest) -> Int {
-		//TODO: Clean this messy function and remove check for empty array
-		result: Int
-		if (This _alignedWidth length == 0) {
-			result = width
+		version (safe) {
+			if (This _alignedWidth length == 0)
+				return width
 		}
-		else {
-			result = This _alignedWidth[0]
-			match (align) {
-				case AlignWidth Nearest => {
-					for (i in 0 .. This _alignedWidth length) {
-						currentWidth := This _alignedWidth[i]
-						if (abs(result - width) > abs(currentWidth - width))
-							result = currentWidth
-					}
-				}
-				case AlignWidth Floor => {
-					for (i in 0 .. This _alignedWidth length) {
-						currentWidth := This _alignedWidth[i]
-						if (abs(result - width) > abs(currentWidth - width) && currentWidth <= width)
-							result = currentWidth
-					}
-				}
-				case AlignWidth Ceiling => {
-					result = This _alignedWidth[This _alignedWidth length-1]
-					for (i in 0 .. This _alignedWidth length) {
-						currentWidth := This _alignedWidth[i]
-						if (abs(result - width) > abs(currentWidth - width) && currentWidth >= width)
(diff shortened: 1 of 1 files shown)


source/draw/gpu/opengl/GraphicBuffer.ooc
```diff
@@ -93,37 +93,17 @@ GraphicBuffer: class {
...
-						if (abs(result - width) > abs(currentWidth - width) && currentWidth >= width)
-							result = currentWidth
-					}
-				}
-			}
+		for (i in 0 .. This _alignedWidth length) {
+			currentWidth := This _alignedWidth[i]
+			if ((abs(result - width) > abs(currentWidth - width) && align == AlignWidth Nearest) ||
+				(abs(result - width) > abs(currentWidth - width) && currentWidth <= width && align == AlignWidth Floor) ||
+				(abs(result - width) > abs(currentWidth - width) && currentWidth >= width && align == AlignWidth Ceiling))
```
davidhesselbom: You're checking `(abs(result - width) > abs(currentWidth - width)` 3 times.

davidhesselbom: So...
```ooc
if (abs(result - width) > abs(currentWidth - width) && ( 
				align == AlignWidth Nearest ||
				(align == AlignWidth Floor && currentWidth <= width) ||
				(align == AlignWidth Ceiling && currentWidth >= width))
```

davidhesselbom: Why are we assigning `result` so many times (first once before the loop, then possibly `This _alignedWidth length` times in the loop)? Should this loop ever break, or should it keep going through the whole array? If the last hit is the only one that counts, why not iterate backwards instead?

marcusnaslund: Right.
We're using the current value of `result` inside the if-statement, so not sure what you mean.

davidhesselbom: Oooh... right. Nevermind :)

marcusnaslund:
Updated.

source/draw/gpu/opengl/GraphicBuffer.ooc
```diff
@@ -93,37 +93,18 @@ GraphicBuffer: class {
...
-							result = currentWidth
-					}
-				}
-			}
+		for (i in 0 .. This _alignedWidth length) {
+			currentWidth := This _alignedWidth[i]
+			if (abs(result - width) > abs(currentWidth - width) && 
+				(align == AlignWidth Nearest ||
+				(currentWidth <= width && align == AlignWidth Floor) ||
+				(currentWidth >= width && align == AlignWidth Ceiling)))
```
davidhesselbom: You don't need the extra parentheses here, `&&` already has higher precedence than `||`.

marcusnaslund:
Hang on, the `safe`-block happens *after* I've already tried to read `This _alignedWidth[0]`, which doesn't exist then.
Is it OK to start the function with
```ooc
version (safe) {
			if (This _alignedWidth length == 0)
				return width
		}
```
?

davidhesselbom:
Start with `result: Int`, always return `result`, and then assign it in the version block.

marcusnaslund:
Sure, but if we're running the safe version and `This _alignedWidth length == 0`, whatever value `result` is set to inside the `safe` block will be changed right afterwards by
```ooc
result = align == AlignWidth Ceiling ? This _alignedWidth[This _alignedWidth length-1] : This _alignedWidth[0]
```
which will crash, even. Adding an `if` statement there negates the whole purpose of the `safe` block.

davidhesselbom:
Not if you put the rest of the function in an `else` block immediately after the `version` block.

marcusnaslund:
We get two sets of `result = align == AlignWidth Ceiling ? This _alignedWidth[This _alignedWidth length-1] : This _alignedWidth[0]`, but I guess that's better than before.

source/draw/gpu/opengl/GraphicBuffer.ooc
```diff
@@ -93,37 +93,21 @@ GraphicBuffer: class {
...
-				}
-			}
+		version (safe) {
+			if (This _alignedWidth length == 0)
+				result = width
+			else
+				result = align == AlignWidth Ceiling ? This _alignedWidth[This _alignedWidth length-1] : This _alignedWidth[0]
+		} else
+			result = align == AlignWidth Ceiling ? This _alignedWidth[This _alignedWidth length-1] : This _alignedWidth[0]
+		for (i in 0 .. This _alignedWidth length) {
```
davidhesselbom: Giving it some more thought, you are already checking the value of `This _alignedWidth length`, a cheap lookup that will be performed a limited amount of times (unlike, say, indexing a list or an image). So is the `safe` block really necessary?

marcusnaslund: The original code wanted specifically `result = width` if the length was zero.

davidhesselbom: Sure, but why not
```ooc
result = width // cheap, easy to read
if (This _alignedWidth length > 0)
    result = align == AlignWidth Ceiling ? This _alignedWidth[This _alignedWidth length-1] : This _alignedWidth[0]
for ...
```

marcusnaslund: I wanted to avoid that `if` statement but you're right, we do more of them in the `for` loop anyway. Your will is my command.

marcusnaslund:
Updated.

PR #573, opened 2015-10-19
Version flags for libraries in ui module

sebastianbaginski:


sebastianbaginski:
Updated enums in x11 to workaround https://github.com/cogneco/ooc-kean/issues/572

PR #575, opened 2015-10-19
Parse in math classes now use Text instead of String

### LINKED ISSUE #569: Parse methods in math use String and ArrayList
A lot of `math` classes (like `FloatPoint`, `FloatSize`, `IntPoint`, etc.) have a `static parse` method which takes a `String` argument. These methods call `split` on the input which returns an `ArrayList`.

Since we want to move away from both `String` and `ArrayList`, shouldn't `parse()` move on to take arguments of `Text` instead? Then `split` would return a `VectorList` which is more easily cleaned up. Right now, most of the `parse` methods leak memory.

marcusnaslund:
Fixes #569 
As noted before, `toString` and `parse` generally don't match. I will create a new issue for that.


source/math/FloatBox2D.ooc
```diff
@@ -128,9 +129,11 @@ FloatBox2D: cover {
 		this createAround(newCenter, newSize)
 	}
 	toString: func -> String { "#{this leftTop toString()}, #{this size toString()}" }
-	parse: static func (input: String) -> This {
+	parse: static func (input: Text) -> This {
 		array := input split(',')
-		This new(array[0] toFloat(), array[1] toFloat(), array[2] toFloat(), array[3] toFloat())
+		result := This new(array[0] toFloat(), array[1] toFloat(), array[2] toFloat(), array[3] toFloat())
+		array free()
```
davidhesselbom: So, does this also free all of the `Text`s we create by splitting `input`?

marcusnaslund: Yes.

marcusnaslund:
Might conflict with #491 but I will fix that once you've merged either of them.

test/math/FloatPoint2DTest.ooc
```diff
@@ -49,9 +50,8 @@ FloatPoint2DTest: class extends Fixture {
 		this add("casting", func {
 			value := "10.00, 20.00"
 			expect(this point3 toString(), is equal to(value))
-//			FIXME: Equals interface
-			expect((FloatPoint2D parse(value)) x, is equal to((this point3) x))
-			expect((FloatPoint2D parse(value)) y, is equal to((this point3) y))
+			expect((FloatPoint2D parse(t"10.00,20.00")) x, is equal to((this point3) x))
```
davidhesselbom: Why not parse `value` instead?

marcusnaslund: Because of the space after the comma.

davidhesselbom: Okay, but why parse a format without the space, but print a value with the space?

marcusnaslund: Well, my point exactly in the PR description. Not all of them expect comma-separated data either, so I want to discuss before I change anything. For now it works as it did before.

davidhesselbom: Maybe `toFloat` is smart enough that it can discard the space by itself... But alright. Assign `t"10.00,20.00"` to a variable at least, so you don't have to define it twice.

marcusnaslund: @davidhesselbom No, I tried, it returns 0. Alright.

source/math/FloatBox2D.ooc
```diff
@@ -20,6 +20,7 @@ import IntBox2D
 import FloatPoint2DVectorList
 import text/StringTokenizer
 import structs/ArrayList
+use ooc-base
```
davidhesselbom: Why? The other covers in the PR apparently do fine without it.

test/math/FloatComplexTest.ooc
```diff
@@ -71,15 +72,15 @@ FloatComplexTest: class extends Fixture {
...
-			expect((FloatComplex parse("2.00 +1.00i")) == this complexNumber0, is true)
+			expect((FloatComplex parse(t"2.00 1.00")) == this complexNumber0, is true)
 			expect(this complexNumber3 toString(), is equal to("-2.00 -1.00i"))
-			expect((FloatComplex parse("-2.00 -1.00i")) == this complexNumber3, is true)
+			expect((FloatComplex parse(t"-2.00 -1.00")) == this complexNumber3, is true)
 			expect(FloatComplex new (2, -1) toString(), is equal to("2.00 -1.00i"))
-			expect((FloatComplex parse("2.00 -1.00i")) == FloatComplex new (2, -1), is true)
+			expect((FloatComplex parse(t"2.00 -1.00")) == FloatComplex new (2, -1), is true)
 			expect(FloatComplex new (-2, 1) toString(), is equal to("-2.00 +1.00i"))
-			expect((FloatComplex parse("-2.00 +1.00i")) == FloatComplex new (-2, 1), is true)
```
davidhesselbom: The trailing `i` doesn't break anything, does it?

test/math/FloatPoint2DTest.ooc
davidhesselbom: > For now it works as it did before.

Wait, no, it doesn't. It used to be able to parse `10.00, 20.00`, now it no longer is, it can only parse `10.00,20.00`. #643 needs to be fixed before I can merge this, or we will break lots of things.

davidhesselbom:
#643 needs fixing first.

Also, (and it can wait) you don't need to import `StringTokenizer` or `ArrayList` anymore.

davidhesselbom:
#719 adds `trim()`, so rebase and try again.

davidhesselbom:
All of these classes should be able to parse their own `toString()` output, as before.

marcusnaslund:
Finally :) I'm on it

marcusnaslund:
Works fine now. No side-effects since we're not using `parse` anywhere atm. @davidhesselbom

source/math/FloatBox2D.ooc
```diff
@@ -152,9 +153,11 @@ FloatBox2D: cover {
 		this createAround(newCenter, newSize)
 	}
 	toString: func -> String { "#{this leftTop toString()}, #{this size toString()}" }
-	parse: static func (input: String) -> This {
+	parse: static func (input: Text) -> This {
 		array := input split(',')
```
davidpiuva: `array` could be named with something less generic telling what type of elements that are given.

marcusnaslund: @davidhesselbom ?

davidhesselbom: Agreed, especially since it's not even an array anymore (it used to be an `ArrayList`). It can be fixed later, I guess.

marcusnaslund: Putting it on my list

davidpiuva:
Syntax looks okay.

davidhesselbom:
> No side-effects since we're not using parse anywhere atm

Yes, we are. Should be fine though.

source/math/FloatComplex.ooc
```diff
@@ -41,12 +42,10 @@ FloatComplex: cover {
...
 		this real toString() >> (this imaginary > 0 ? " +" : " ") & this imaginary toString() >> "i"
 	}
-	parse: static func (input: String) -> This {
-		parts := input split(' ')
-		result := This new(parts[0] toFloat(), parts[1] toFloat())
-		parts[0] free()
-		parts[1] free()
-		parts free()
+	parse: static func (input: Text) -> This {
+		array := input find('+') >= 0 ? input split('+') : input split(' ')
```
davidhesselbom: I really don't know why you changed `parts` to `array` here.

test/math/FloatPoint2DTest.ooc
```diff
@@ -49,8 +50,8 @@ FloatPoint2DTest: class extends Fixture {
 		this add("casting", func {
 			value := "10.00, 20.00"
 			expect(this point3 toString(), is equal to(value))
-			expect((FloatPoint2D parse(value)) x, is equal to((this point3) x))
-			expect((FloatPoint2D parse(value)) y, is equal to((this point3) y))
+			expect((FloatPoint2D parse(Text new(value))) x, is equal to((this point3) x))
+			expect((FloatPoint2D parse(Text new(value))) y, is equal to((this point3) y))
```
davidhesselbom: Why not
```ooc
value := t"10.00, 20.00"
expect(this point3 toString(), is equal to(value toString()))
expect((FloatPoint2D parse(value)) x, is equal to((this point3) x))
expect((FloatPoint2D parse(value)) y, is equal to((this point3) y))
```
?

source/math/FloatComplex.ooc
marcusnaslund: For consistency since all the other places had `array`.

test/math/FloatPoint2DTest.ooc
marcusnaslund: Sure, that works too, but we'll be revisiting this soon enough when `toText()` is introduced and `toString()` finally deprecated and removed. So, do I need to fix it now?

source/math/FloatComplex.ooc
davidhesselbom: Yeah, but that's because they were...

davidhesselbom: Otoh, you still access them like arrays, so okay, fine.

test/math/FloatPoint2DTest.ooc
davidhesselbom: This "fix" of mine is a smaller change than the one you already implemented, which is why I suggest it now.

marcusnaslund: Fair enough :) Done.

marcusnaslund:
Hmm, `Promise` and `PromiseCollector` now fail every time on Appveyor...

PR #580, opened 2015-10-21
Added version blocks around EGL and OpenGL lib links in usefiles

jonathanudd:
It would be more correct to use `version (!gpuOff)` but at the moment we are not able to use custom flags in version blocks in use files but we need this now when running on windows.
This should also solve the failing appveyor tests.


jonathanudd:
This solved the link problem on appveyor but it seems we have a problem with the blockedqueue test on windows.

PR #584, opened 2015-10-21
Added resize functions for Box that don't affect the center

### LINKED ISSUE #551: Box needs resize functions
Changing the size of an existing `IntBox` or `FloatBox` without moving its center is currently too complicated - they need functions for making this easier.

marcusnaslund:
Fixes #551 
Adds the following four functions to change the size without affecting the center of a `FloatBox2D` or `IntBox2D`:
- [x] `setSize`: a specified size
- [x] `scale`: change by a scale factor
- [x] `enlarge`: change to a size that is >= the Size passed to the function (if the box is already >= the Size argument, use this size)
- [x] `reduce`: a size that is <= the Size passed to the function (if the box is already <= the Size argument, use this size).

If this isn't what you meant of if you prefer other names, let me know.

### DIFF
--- a/source/math/FloatBox2D.ooc
+++ b/source/math/FloatBox2D.ooc
@@ -70,10 +70,10 @@ FloatBox2D: cover {
 		This createAround(this center, value * this size)
 	}
 	enlargeTo: func (size: FloatSize2D) -> This {
-		This createAround(this center, this size + ((size - this size) maximum~Float(0.0f)))
+		This createAround(this center, FloatSize2D maximum(this size, size))
 	}
 	shrinkTo: func (size: FloatSize2D) -> This {
-		This createAround(this center, this size + ((size - this size) minimum~Float(0.0f)))
+		This createAround(this center, FloatSize2D minimum(this size, size))
 	}
 	intersection: func (other: This) -> This {
 		left := Float maximum(this left, other left)
--- a/source/math/FloatBox2D.ooc
+++ b/source/math/FloatBox2D.ooc
@@ -66,7 +66,15 @@ FloatBox2D: cover {
 	scale: func (value: Float) -> This {
 		sizeChange := this size - (this size * value)
 		This new(this leftTop + (sizeChange / 2.0f), this size - sizeChange)
-	}	
+	}
+	enlarge: func (size: FloatSize2D) -> This {
+		sizeChange := (size - this size) maximum~Float(0.0f)
+		This new(this leftTop - (sizeChange / 2.0f), this size + sizeChange)
+	}
+	reduce: func (size: FloatSize2D) -> This {
+		sizeChange := (size - this size) minimum~Float(0.0f)
+		This new(this leftTop - (sizeChange / 2.0f), this size + sizeChange)
+	}
 	intersection: func (other: This) -> This {
 		left := Float maximum(this left, other left)
 		top := Float maximum(this top, other top)
--- a/source/math/FloatBox2D.ooc
+++ b/source/math/FloatBox2D.ooc
@@ -72,7 +72,7 @@ FloatBox2D: cover {
 	enlargeTo: func (size: FloatSize2D) -> This {
 		This createAround(this center, this size + ((size - this size) maximum~Float(0.0f)))
 	}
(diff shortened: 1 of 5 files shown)


source/math/FloatBox2D.ooc
```diff
@@ -63,6 +63,18 @@ FloatBox2D: cover {
...
+		This new(this leftTop + (sizeChange / 2.0f), this size - sizeChange)
+	}
+	enlarge: func (size: FloatSize2D) -> This {
+		sizeChange := (size - this size) maximum~Float(0.0f)
+		This new(this leftTop - (sizeChange / 2.0f), this size + sizeChange)
+	}
+	reduce: func (size: FloatSize2D) -> This {
+		sizeChange := (size - this size) minimum~Float(0.0f)
+		This new(this leftTop - (sizeChange / 2.0f), this size + sizeChange)
+	}
```
davidhesselbom: `reduce` could mean either `reduceTo` or `reduceBy`. How about `enlargeTo` and `shrinkTo`? Or are there even better names?

marcusnaslund: Right. I prefer the `To` version.

tomasmahlberg: I had to scale a `FloatBox2D` the other day. I did it like this:
```ooc
This createAround(this center, value * this size)
```

marcusnaslund: Oh, there is a `this center`, didn't notice. Thanks!

marcusnaslund:
There we go.

source/math/FloatBox2D.ooc
```diff
@@ -63,6 +63,18 @@ FloatBox2D: cover {
 	shrink: func ~fraction (margin: Float) -> This {
 		this pad(-margin * this height / 2.0f)
 	}
+	setSize: func (size: FloatSize2D) -> This {
```
davidhesselbom: I don't like this name, because it makes it sound like you are only modifying the size (and knowing a box consists of a top-left position and a size, that's misleading). `resizeTo` would make things clearer.

marcusnaslund: Yep. `scale` or `scaleTo`?

source/math/FloatBox2D.ooc
```diff
@@ -63,6 +63,18 @@ FloatBox2D: cover {
...
+	resizeTo: func (size: FloatSize2D) -> This {
+		This createAround(this center, size)
+	}
+	scale: func (value: Float) -> This {
+		This createAround(this center, value * this size)
+	}
+	enlargeTo: func (size: FloatSize2D) -> This {
+		This createAround(this center, this size + ((size - this size) maximum~Float(0.0f)))
+	}
+	reduceTo: func (size: FloatSize2D) -> This {
```
davidhesselbom: I maintain that "reduce" is the opposite of "increase", not the opposite of "enlarge". What's wrong with `shrinkTo`?

marcusnaslund: Sorry, missed it.

source/math/FloatBox2D.ooc
```diff
@@ -63,6 +63,18 @@ FloatBox2D: cover {
...
+	}
+	scale: func (value: Float) -> This {
+		This createAround(this center, value * this size)
+	}
+	enlargeTo: func (size: FloatSize2D) -> This {
+		This createAround(this center, this size + ((size - this size) maximum~Float(0.0f)))
+	}
+	shrinkTo: func (size: FloatSize2D) -> This {
+		This createAround(this center, this size + ((size - this size) minimum~Float(0.0f)))
+	}
```
tomasmahlberg: I think this would become more readable if we changed `enlargeTo` and `shrinkTo` to
```ooc
This createAround(this center, FloatSize2D maximum(this size, size))
```
and
```ooc
This createAround(this center, FloatSize2D minimum(this size, size))
```
respectively.

davidhesselbom: Agreed.

marcusnaslund: Done.

PR #585, opened 2015-10-21
Changed Caller and Callee to Sender and Receiver.

simonmika:


simonmika:
This breaks the tests but @tomasmahlberg wanted to start looking at it.

davidhesselbom:
Breaks test in the other project.

PR #586, opened 2015-10-21
Exported functions from date/time, raster and float transform classes to public API

sebastianbaginski:
Needed in the other project.

### DIFF
--- a/source/base/DateTime.ooc
+++ b/source/base/DateTime.ooc
@@ -279,12 +279,12 @@ DateTime: cover {
 	kean_base_dateTime_fromDateTime: unmangled static func (year, month, day, hour, minute, second, millisecond: Int) -> This {
 		This new(year, month, day, hour, minute, second, millisecond)
 	}
-	kean_base_dateTime_now: unmangled static func -> This { This now }
-	kean_base_dateTime_millisecond: unmangled func -> Int { this millisecond() }
-	kean_base_dateTime_second: unmangled func -> Int { this second() }
-	kean_base_dateTime_minute: unmangled func -> Int { this minute() }
-	kean_base_dateTime_hour: unmangled func -> Int { this hour() }
-	kean_base_dateTime_day: unmangled func -> Int { this day() }
-	kean_base_dateTime_month: unmangled func -> Int { this month() }
-	kean_base_dateTime_year: unmangled func -> Int { this year() }
+	kean_base_dateTime_getNow: unmangled static func -> This { This now }
+	kean_base_dateTime_getMillisecond: unmangled func -> Int { this millisecond() }
+	kean_base_dateTime_getSecond: unmangled func -> Int { this second() }
+	kean_base_dateTime_getMinute: unmangled func -> Int { this minute() }
+	kean_base_dateTime_getHour: unmangled func -> Int { this hour() }
+	kean_base_dateTime_getDay: unmangled func -> Int { this day() }
+	kean_base_dateTime_getMonth: unmangled func -> Int { this month() }
+	kean_base_dateTime_getYear: unmangled func -> Int { this year() }
 }
--- a/source/base/DateTime.ooc
+++ b/source/base/DateTime.ooc
@@ -71,8 +71,6 @@ DateTime: cover {
 			raise ("invalid input specified for constructor(year,month,day,hour,minute,second,ms)")
 	}
 	isLeapYear: static func (year: Int) -> Bool { (year % 100 == 0) ? (year % 400 == 0) : (year % 4 == 0) }
-	kean_base_dateTime_getTicks: unmangled func -> UInt64 { this _ticks }
-	kean_base_dateTime_new: unmangled static func (ticks: UInt64) -> This { This new(ticks) }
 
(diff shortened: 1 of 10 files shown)


source/math/FloatEuclidTransform.ooc
```diff
@@ -46,4 +46,5 @@ FloatEuclidTransform: cover {
 	toString: func -> String {
 		"Translation: " << this translation toString() >> " Rotation: " & this rotation toString() >> " Scaling: " & this scaling toString()
 	}
+	kean_math_floatEuclidTransform_toFloatTransform3D: unmangled func -> FloatTransform3D { this transform }
```
davidhesselbom: Isn't it `getFloatTransform3D` rather than `toFloatTransform3D`?

source/base/DateTime.ooc
```diff
@@ -274,4 +272,19 @@ DateTime: cover {
...
+		This new(year, month, day, hour, minute, second, millisecond)
+	}
+	kean_base_dateTime_now: unmangled static func -> This { This now }
+	kean_base_dateTime_millisecond: unmangled func -> Int { this millisecond() }
+	kean_base_dateTime_second: unmangled func -> Int { this second() }
+	kean_base_dateTime_minute: unmangled func -> Int { this minute() }
+	kean_base_dateTime_hour: unmangled func -> Int { this hour() }
+	kean_base_dateTime_day: unmangled func -> Int { this day() }
+	kean_base_dateTime_month: unmangled func -> Int { this month() }
+	kean_base_dateTime_year: unmangled func -> Int { this year() }
```
davidhesselbom: I want to ask before we release the API: Why are these functions and not properties?

sebastianbaginski: Probably because this was my first task and my background is C++. There are no properties there (like in ooc), so I just used functions like I always did (with the difference that you can't mark a function as `const` in ooc) :)

davidhesselbom: Ah. @simonmika , what do you think? The ooc function look like static properties, or constructors, to me, just like `FloatTransform2D identity`. Should we fix this before publishing the API?

davidhesselbom: Wait... I guess it doesn't affect the C API, anyway, does it?

sebastianbaginski: No, it doesn't.

davidhesselbom: But why do these functions return `Int` and not `This`?

sebastianbaginski: Good catch. This is an error.

davidhesselbom: Ohh, I get it now. In `DateTime`, `minute()` gives you the `minute` part of the DateTime. In `TimeSpan`, it creates a `TimeSpan` that's 1 minute long. This is really confusing... :)

davidhesselbom: I guess we can fix that later, but I'd really like not to have to change the C API later, so can we decide what these functions should actually do before I merge?

sebastianbaginski: No wait, it is correct, it is like that in the ooc api, I was confused with the TimeSpan class.
So what do we do about it ?

davidhesselbom: Lemme talk to Simon...

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -247,4 +247,12 @@ RasterYuv420Semiplanar: class extends RasterYuvSemiplanar {
 		fileWriter close()
 	}
 	createPaintEngine: override func -> PaintEngine { Yuv420PaintEngine new(this) }
+	kean_draw_rasterYuv420Semiplanar_new: static unmangled func (width, height, stride: Int, monochromeData: Void*, uvData: Void*) -> This {
```
davidhesselbom: `monochromeData, uvData: Void*`

sebastianbaginski: Fixed, but I created another commit as I don't know if I can amend the last one (PR merged)

davidhesselbom: No problem, but yes, you could (you would have to create a new branch at your last one and rebase my commit on top of your new one, but this works too).

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -260,4 +260,12 @@ RasterYuv420Semiplanar: class extends RasterYuvSemiplanar {
 		fileWriter close()
 	}
 	_createCanvas: override func -> Canvas { Yuv420RasterCanvas new(this) }
+	kean_draw_rasterYuv420Semiplanar_new: static unmangled func (width, height, stride: Int, monochromeData: Void*, uvData: Void*) -> This {
```
davidhesselbom: `monochromeData, uvData: Void*`

Looks like this got lost somewhere along the way?

sebastianbaginski: Yeah, maybe during the last rebase. Fixed now.

Issue #589, opened 2015-10-21
Test of BlockedQueue and SynchronizedQueue fails


jonathanudd:
The tests fails alot on Windows and sometimes on Unix.

jonathanudd:
Testoutput on Windows
```
raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/BlockedQueue -DgpuOff
Cogneco Rock Compiler
BlockedQueue .. done

raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/BlockedQueue -DgpuOff
Cogneco Rock Compiler
BlockedQueue [lang/Backtrace] No backtrace extension nor execinfo - use a debugger!
[TestFailedException]:

raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/BlockedQueue -DgpuOff
Cogneco Rock Compiler
BlockedQueue f. failed
  ->  cover : expected  equal to  '1', was '5'
[lang/Backtrace] No backtrace extension nor execinfo - use a debugger!

raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/BlockedQueue -DgpuOff
Cogneco Rock Compiler
BlockedQueue f. failed
  ->  cover : expected  equal to  '1', was '2'
[lang/Backtrace] No backtrace extension nor execinfo - use a debugger!

raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/BlockedQueue -DgpuOff
Cogneco Rock Compiler
BlockedQueue f. failed
  ->  cover : expected  equal to  '1', was '2'
[lang/Backtrace] No backtrace extension nor execinfo - use a debugger!

raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/BlockedQueue -DgpuOff
Cogneco Rock Compiler
BlockedQueue [lang/Backtrace] No backtrace extension nor execinfo - use a debugger!
[TestFailedException]:

raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/SynchronizedQueue -DgpuOff
Cogneco Rock Compiler
BlockedQueue [lang/Backtrace] No backtrace extension nor execinfo - use a debugger!
[TestFailedException]:

raven+judd@raven MSYS ~/ooc/ooc-kean
$ ./test.sh collections/SynchronizedQueue -DgpuOff
Cogneco Rock Compiler
[...]

PR #590, opened 2015-10-21
Cleanup of Queue API

### LINKED ISSUE #568: Clean up Queue API
Right now there are three different `dequeue` and `peek`, only because of issues with rock. The latest ones (marked `~default`) are the only ones that always work. 

Since the older ones won't be fixed anytime soon, and in order to avoid confusion, we should strive to:
1) Eliminate all uses of the obsoleted versions (possibly already done?)
2) Remove the obsoleted versions

marcusnaslund:
Fixes #568 
Removes the deprecated `dequeue` and `peek` versions from `Queue` and derived classes and updates the corresponding tests to match.
Works well in linux but doesn't help with #589 - it actually makes it worse. Looking into it.


source/collections/Queue.ooc
```diff
@@ -6,16 +6,7 @@ Queue: abstract class <T> {
...
-	peek: abstract func -> (T, Bool)
-	// Obsoleted from using template pointers that are not reliable in the compiler's type safety
-	dequeue: abstract func ~out (result: T*) -> Bool
-	peek: abstract func ~out (result: T*) -> Bool
-	// Yet another interface to go around compiler bugs
-	// T must be nullable or have a reserved value to know when it failed
+	enqueue: virtual func (item: T) //FIXME: This should be abstract but it messes up for sub-subclasses that overrides
 	dequeue: abstract func ~default (fallback: T) -> T
 	peek: abstract func ~default (fallback: T) -> T
 	//new: static func -> Queue<T> { VectorQueue<T> new() }
```
marcusnaslund: This commented-out `new` function has been there since April. Remove it?

davidpiuva: I don't see a reason not to since VectorQueue should not be mentioned in Queue.

marcusnaslund: Agreed.

PR #593, opened 2015-10-23
Cleanup in RasterMonocrhome apply

sebastianbaginski:


source/draw/RasterMonochrome.ooc
```diff
@@ -41,14 +41,12 @@ RasterMonochrome: class extends RasterPacked {
...
-		rowLength := this size width * this bytesPerPixel
-		for (row in this buffer pointer as Long .. end) {
-			rowEnd := row + rowLength
-			for (source: Long in row .. rowEnd)
-				action((source as ColorMonochrome*)@)
-			row += this stride - 1
-		}
+		pointer := this buffer pointer as ColorMonochrome*
+		for (row in 0 .. this height)
+			for (col in 0 .. this width) {
```
davidhesselbom: I might merge this before you have a chance to fix it, but do try to keep in mind to type out the full name of things, at least unless the result is ridiculously long. `col` should be `column`.

sebastianbaginski: Fixed.

davidhesselbom: Thanks!

PR #594, opened 2015-10-23
Optimized some raster conversion methods

sebastianbaginski:
Raster conversion methods use direct copy if converting from the same image type. It is faster than creating a closure and applying it pixel by pixel.
For example:
> RasterBgr from RasterBgr (with closure)
> 21.408001 msec.
> RasterBgr from RasterBgr (direct copy)
> 0.0507032 msec.

There is still room for optimization left (`RasterMonochrome` from `RasterYuvPlanar` - direct copy of `y` channel etc.), but I guess it is a good start.

### DIFF
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -111,22 +111,27 @@ RasterBgr: class extends RasterPacked {
 		This new(ByteBuffer new(data as UInt8*, x * y * requiredComponents), IntSize2D new(x, y))
 	}
 	convertFrom: static func (original: RasterImage) -> This {
-		result := This new(original size)
-		row := result buffer pointer as Long
-		rowLength := result size width
-		rowEnd := row as ColorBgr* + rowLength
-		destination := row as ColorBgr*
-		f := func (color: ColorBgr) {
-			(destination as ColorBgr*)@ = color
-			destination += 1
-			if (destination >= rowEnd) {
-				row += result stride
-				destination = row as ColorBgr*
-				rowEnd = row as ColorBgr* + rowLength
+		result: This
+		if (original instanceOf?(This))
+			result = (original as This) copy()
+		else {
+			result = This new(original size)
+			row := result buffer pointer as Long
+			rowLength := result size width
+			rowEnd := row as ColorBgr* + rowLength
+			destination := row as ColorBgr*
+			f := func (color: ColorBgr) {
+				(destination as ColorBgr*)@ = color
+				destination += 1
+				if (destination >= rowEnd) {
+					row += result stride
+					destination = row as ColorBgr*
+					rowEnd = row as ColorBgr* + rowLength
+				}
 			}
+			original apply(f)
+			(f as Closure) dispose()
 		}
-		original apply(f)
-		(f as Closure) dispose()
 		result
 	}
 	operator [] (x, y: Int) -> ColorBgr { this isValidIn(x, y) ? ((this buffer pointer + y * this stride) as ColorBgr* + x)@ : ColorBgr new(0, 0, 0) }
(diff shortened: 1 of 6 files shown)


source/draw/RasterBgr.ooc
```diff
@@ -111,18 +111,23 @@ RasterBgr: class extends RasterPacked {
...
+				(destination as ColorBgr*)@ = color
+				destination += 1
+				if (destination >= rowEnd) {
+					row += result stride
+					destination = row as ColorBgr*
+					rowEnd = row as ColorBgr* + rowLength
+				}
 			}
 		}
 		original apply(f)
```
davidhesselbom: This needs to be moved inside the `else` clause or `f` will be out of scope when you call `original appl(f)`.

source/draw/RasterMonochrome.ooc
```diff
@@ -107,21 +107,26 @@ RasterMonochrome: class extends RasterPacked {
...
+			f := func (color: ColorMonochrome) {
+				(destination as ColorMonochrome*)@ = color
+				destination += 1
+				if (destination >= rowEnd) {
+					row += result stride
+					destination = row
+					rowEnd = row + rowLength
+				}
 			}
+			original apply(f)
```
davidhesselbom: Also, `(f as Closure) dispose()`, right?

source/draw/RasterUv.ooc
```diff
@@ -118,21 +118,26 @@ RasterUv: class extends RasterPacked {
...
+			f := func (color: ColorYuv) {
+				(destination as ColorUv*)@ = ColorUv new(color u, color v)
+				destination += 1
+				if (destination >= rowEnd) {
+					row += result stride
+					destination = row as ColorUv*
+					rowEnd = row as ColorUv* + rowLength
+				}
 			}
+			original apply(f)
```
davidhesselbom: Also, `(f as Closure) dispose()`, right?

source/draw/RasterYuv420Planar.ooc
```diff
@@ -96,42 +96,47 @@ RasterYuv420Planar: class extends RasterYuvPlanar {
...
+					yDestination = yRow
+					if (y % 2 == 0) {
+						uRow += result u stride
+						uDestination = uRow
+						vRow += result v stride
+						vDestination = vRow
+					}
 				}
 			}
+			original apply(f)
```
davidhesselbom: Also, `(f as Closure) dispose()`, right?

davidhesselbom:
>21.408001 msec.
0.0507032 msec.

Lol :D

source/draw/RasterBgr.ooc
sebastianbaginski: Yes, before I merged with master. This line was added automatically.

sebastianbaginski:
I know, the performance penalty when using closures is really frightening...
Code updated.

thomasfanell:
Why is the test hanging on ThreadPool? I thought we sorted out the deadlocks?

sebastianbaginski:
No idea. I worked on the `Thread`, `Mutex` and `WaitCondition` classes some time ago (so you can blame me when something is wrong with them, I guess :P), but I think ThreadPool and the test itself is @marcusnaslund 's recent creation.

marcusnaslund:
#509 fixes some remaining issues but hasn't been merged yet.

davidhesselbom:
> I know, the performance penalty when using closures is really frightening...

Well, in the one case you do a single `memcpy`, and in the other you run a loop that does lots and lots of arrary lookups that end up replacing each byte with its original value. Can't really blame closures for that...

sebastianbaginski:
Yes of course, but I mean now you have that nice conversion code dealing with all the differences between image formats by closures, imagine how much faster would things be if we used hard coded conversions for each format.
It's the similar thing we did with the vector `operator[]` vs. `image buffer pointer[]`, avoiding function calls whenever you can.

davidhesselbom:
I guess that's doable, assuming we will never want to add new formats. You will still need the color conversion arrays though, right? I don't see how you could feasibly hard-code those (and I think array lookup is faster than computing the values using a formula in this case).

Did you benchmark it in release or debug mode, btw?

sebastianbaginski:
Release.
Yeah arrays will still stay there, as it is indeed faster to lookup things than compute.

PR #595, opened 2015-10-23
Add FloatBox2D pad~fractionAverage().

bettinaselig:
Example:
```
FloatBox2D box = FloatBox2D new(FloatPoint new(0, 0), FloatSize(100, 80))
FloatBox2D box2 = box pad~fractionAverage(0.2f) 
```
Then `box2`'s boundaries are`0.2f * (100 + 80) / 2.0 = 18` pixels shifted outwards compared to `box`.

### DIFF
--- a/source/math/FloatBox2D.ooc
+++ b/source/math/FloatBox2D.ooc
@@ -60,6 +60,9 @@ FloatBox2D: cover {
 	pad: func ~fraction (pad: Float) -> This {
 		this pad(pad * this size / 2.0f)
 	}
+	padFractionAverage: func (pad: Float) -> This {
+		this pad(pad * (this size width + this size height) / 2.0f)
+	}
 	shrink: func ~fraction (margin: Float) -> This {
 		this pad(-margin * this height / 2.0f)
 	}


davidhesselbom:
So... if you call this function enough times, the box will asymptotically grow into a very large square shape?

source/math/FloatBox2D.ooc
```diff
@@ -60,6 +60,9 @@ FloatBox2D: cover {
 	pad: func ~fraction (pad: Float) -> This {
 		this pad(pad * this size / 2.0f)
 	}
+	pad: func ~fractionAverage (pad: Float) -> This {
```
davidhesselbom: I don't like the name. Suffixes are meant to help the C compiler disambiguate between function with the same name, but with a different signature. Now you have 2 `pad` functions with the same signature, but with different suffixes. Come up with a real name for this function instead.

bettinaselig:
We can discuss name changes tomorrow. I will tell Sebastian to rebase on my branch instead of upstream, so we can deliver today.

bettinaselig:
Sebastian suggested to make an issue in the pull request dept milestone and merge it now. What do you think about that?

davidhesselbom:
No. If someone is calling `pad(0.2f)` somewhere, we don't know which of these two functions will be used; `~fractionAverage` or `~fraction`. Give it a name, call it `padFractionAverage` if you can't think of anything, but never have two functions with the same signature and only a suffix to tell them apart.

sebastianbaginski:
> we don't know which of these two functions will be used

How does compiler know that ? Shouldn't it refuse to compile such function call because of ambiguity ?

davidhesselbom:
Maybe it will pick one for us, I don't know.

bettinaselig:
I can change this, but there exist already `pad: func ~fromFloat (pad: Float)` and `pad: func ~fraction (pad: Float)` on upstream.

davidhesselbom:
Yes, I see that now. This is really, really bad, but let's not make it even worse. Give your new function a name, and we'll solve the other ambiguity later.

bettinaselig:
Ok.

bettinaselig:
Done.

PR #597, opened 2015-10-25
Removed exceptions in Transform classes

marcusnaslund:
Instead uses `raise()` inside `safe` blocks.

### DIFF
--- a/source/math/FloatTransform2D.ooc
+++ b/source/math/FloatTransform2D.ooc
@@ -35,7 +35,7 @@ FloatTransform2D: cover {
 		result := 0.0f
 		version (safe) {
 			if (x < 0 || x > 2 || y < 0 || y > 2)
-				raise("Out of bounds in FloatTransform2D get operator")
+				raise("Out of bounds in FloatTransform2D get operator (#{x}, #{y})")
 		}
 		match (x) {
 			case 0 =>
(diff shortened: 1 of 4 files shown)


source/math/FloatTransform2D.ooc
```diff
@@ -33,29 +33,29 @@ FloatTransform2D: cover {
 	a, b, c, d, e, f, g, h, i: Float
 	operator [] (x, y: Int) -> Float {
 		result := 0.0f
+		version (safe) {
+			if (x < 0 || x > 2 || y < 0 || y > 2)
+				raise("Out of bounds in FloatTransform2D get operator")
```
davidhesselbom: The version has the advantage that it tells you the index you incorrectly tried to access. This should say `"Out of bounds in FloatTransform2D get operator (#{x}, #{y})"` or something

source/math/FloatTransform2D.ooc
```diff
@@ -68,6 +68,10 @@ FloatTransform2D: cover {
 	rotationZ ::= this b atan2(this a)
 	inverse: This { get {
 		determinant := this determinant
+		version (safe) {
+			if (determinant == 0)
+				raise("Determinant is zero in FloatTransform2D inverse()")
```
davidhesselbom: This check needed adding here, yes, but I'm not sure it belongs in a `safe` block. `safe` blocks are used for one reason: performance. Bounds checking every single indexing of an array or collection means lots of unnecessary work, because we should never access out of bounds, and when we do, it's because we made a mistake somewhere, and then the bounds check can help us figure out where, so we can fix it. But when we want to compute the inverse of a transform, and the determinant is 0, something is wrong with the transform, and there is no way we can know this unless we compute the determinant (as far as I know). When this happens, it should not be hidden from us inside a `safe` block. Does this make sense?

marcusnaslund: The question is `can the matrix be inverted` and you can answer this without computing the determinant, but it's not computationally more efficient in this case. Your explanation makes sense, so I will fix.

marcusnaslund:
Updated.

PR #601, opened 2015-10-26
Removed try-catch block in Synchronized

marcusnaslund:
To avoid memory problems.


davidhesselbom:
Just be sure, this function is never used anywhere anyway, right?

marcusnaslund:
That's right. I doubt it even works, given how it's written and the trouble we've had with generic functions as parameters in other cases (like `Future`).

davidhesselbom:
Maybe we should consider removing it instead?

marcusnaslund:
Well, that's your call, but at least this cleans it up.

PR #607, opened 2015-10-27
x11 window without opengl

sebastianbaginski:
After merging this it is possible to use `gpuOff` version of the other project on linux.
It uses OpenGL as before when running with GPU.

It fails to clear the window with background color, I think this is because we do not process any events from the window system event queue (normally you should draw on the window only on `paint / refresh` event, our approach is kind of brute force). The same code clears the window if you have an event loop running, maybe we should fix this (it would allow us to close the instance of the program when pressing the `X` button on window status bar).

### DIFF
--- a/source/ui/x11/X11Window.ooc
+++ b/source/ui/x11/X11Window.ooc
@@ -17,10 +17,17 @@
 use ooc-math
 use ooc-ui
 import include/x11
+use ooc-draw
 
 version(unix || apple) {
 X11Window: class extends NativeWindow {
+	_x11GraphicsContext: GraphicsContextX11
+	_defaultScreen := 0
+	_displayDepth := 0
+	_cacheSize: IntSize2D
+	_xImage: XImageOoc*
 	init: func (size: IntSize2D, title: String) {
+		this _xImage = null
 		/* Note: ":0" is the usual identifier for the default display but this should be read from the DISPLAY variable in the system by passing null as parameter,
 		i.e. this _display = XOpenDisplay(null) */
 		display := XOpenDisplay(null)
@@ -30,10 +37,15 @@ X11Window: class extends NativeWindow {
 		}
 		if (display == null)
 			raise("Unable to open X Display")
+		this _defaultScreen = XDefaultScreen(display)
+		this _displayDepth = DefaultDepth(display, this _defaultScreen)
 		root: Long = DefaultRootWindow(display)
 		swa: XSetWindowAttributesOoc
 		swa eventMask = ExposureMask | PointerMotionMask | KeyPressMask | ButtonPressMask
 		backend := XCreateWindow(display, root, 0, 0, size width, size height, 0u, CopyFromParent as Int, InputOutput as UInt, null, CWEventMask, swa&)
+
+		this _x11GraphicsContext = XCreateGC(display, backend, 0, 0)
+		XSetBackground(display, this _x11GraphicsContext, 0)
 		//Disable fit to screen
 		sh: XSizeHintsOoc
 		sh width = sh min_width = size width
@@ -52,8 +64,25 @@ X11Window: class extends NativeWindow {
 	}
 	resize: func (size: IntSize2D) { XResizeWindow(this display, this backend, size width, size height) }
 	free: override func {
+		if (this _xImage)
+			XDestroyImage(this _xImage)
+		XFreeGC(this display, this _x11GraphicsContext)
+		XDestroyWindow(this display, this backend)
 		XCloseDisplay(this display)
 		super()
+	}
+	draw: func (image: RasterBgra) {
+		if (this _cacheSize != image size) {
+			if (this _xImage)
+				XDestroyImage(this _xImage)
+			imageData := gc_malloc(image height * image stride * 4)
+			this _xImage = XCreateImage(this display, CopyFromParent, this _displayDepth, ZPixmap, 0, imageData, image width, image height, 8, 0)
+			this _cacheSize = image size
+		}
+		image swapRedBlue()
+		memcpy(this _xImage@ data, image buffer pointer, image buffer size)
+		XClearWindow(this display, this backend)
+		XPutImage(this display, this backend, this _x11GraphicsContext, this _xImage, 0, 0, 0, 0, image width, image height)
 	}
 }
 }
--- a/source/widgets/UnixWindow.ooc
+++ b/source/widgets/UnixWindow.ooc
@@ -22,19 +22,45 @@ use ooc-draw
 use ooc-opengl
 use ooc-x11
 
-version((unix || apple) && !gpuOff) {
-UnixWindow: class extends DisplayWindow {
+version(unix || apple) {
+UnixWindowBase: class extends DisplayWindow {
 	_xWindow: X11Window
-	_openGLWindow: OpenGLWindow
-	context ::= this _openGLWindow context
 	init: func (size: IntSize2D, title: String) {
 		super()
 		this _xWindow = X11Window new(size, title)
+	}
+	free: override func {
+		this _xWindow free()
+		super()
+	}
+	draw: override func (image: Image) {
+		if (image instanceOf?(RasterBgra))
+			this _xWindow draw(image as RasterBgra)
+		else {
+			raster := RasterBgra convertFrom(image as RasterImage)
+			this _xWindow draw(raster)
+			raster referenceCount decrease()
+		}
+	}
+	create: static func (size: IntSize2D, title: String) -> This {
+		result: This
+		version(gpuOff)
+			result = This new(size, title)
+		else
+			result = UnixWindow new(size, title)
+		result
+	}
+}
+version(!gpuOff) {
+UnixWindow: class extends UnixWindowBase {
+	_openGLWindow: OpenGLWindow
+	context ::= this _openGLWindow context
+	init: func (size: IntSize2D, title: String) {
+		super(size, title)
 		this _openGLWindow = OpenGLWindow new(this _xWindow)
 	}
 	free: override func {
 		this _openGLWindow free()
-		this _xWindow free()
 		super()
 	}
 	draw: override func (image: Image) {
@@ -43,5 +69,6 @@ UnixWindow: class extends DisplayWindow {
 	refresh: override func {
 		this _openGLWindow refresh()
 	}
+}
 }
 }
--- a/source/ui/x11/include/x11.ooc
+++ b/source/ui/x11/include/x11.ooc
@@ -72,6 +72,8 @@ XEventOoc: cover from XEvent {
 	xkey: extern XKeyEventOoc
 }
 
+GraphicsContextX11: cover from GC
+
 XOpenDisplay: extern func (displayName: Char*) -> Pointer
 XCloseDisplay: extern func (display: Pointer) -> Int
 XCreateWindow: extern func (display: Pointer, window: Long, x, y: Int,
@@ -86,6 +88,23 @@ XFlush: extern func (display: Pointer)
 XSendEvent: extern func (display: Pointer, window: Long, propagate: Bool, event_mask: Long, event_send: XEventOoc*)
 XClearWindow: extern func (display: Pointer, window: Long)
 XMoveWindow: extern func (display: Pointer, window: Long, x, y: Int)
+XDefaultScreen: extern func (display: Pointer) -> Int
+DefaultDepth: extern func (display: Pointer, screen: Int) -> Int
+XSetBackground: extern func (...)
+XCreateGC: extern func (...) -> GraphicsContextX11
+XFreeGC: extern func (...)
+XDestroyWindow: extern func (...)
+
+XImageOoc: cover from XImage {
+	width, height, xoffset, format: Int
+	data: Char*
+	byteOrder, bitmapUnit, bitmapBitOrder, bitmapPad, depth, bytesPerLine, bitsPerPixel: Int
+	redMask, greenMask, blueMask: ULong
+}
+XCreateImage: extern func (...) -> XImageOoc*
+XPutImage: extern func (...)
+XDestroyImage: extern func (image: XImageOoc*)
+XPutPixel: extern func (...)
 
 XKeyEventOoc: cover from XKeyEvent {
 	type: extern Int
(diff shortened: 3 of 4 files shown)


source/ui/x11/X11Window.ooc
```diff
@@ -52,8 +64,34 @@ X11Window: class extends NativeWindow {
...
 	resize: func (size: IntSize2D) { XResizeWindow(this display, this backend, size width, size height) }
 	free: override func {
+		if (this _xImage)
+			XDestroyImage(this _xImage)
+		XFreeGC(this display, this _x11GraphicsContext)
+		XDestroyWindow(this display, this backend)
 		XCloseDisplay(this display)
 		super()
 	}
+	draw: func (image: RasterYuv420Semiplanar) {
```
sebastianbaginski: This format is chosen only because this is the default one when running without gpu. Could be more like *any format that can return color at (x,y)*, but I'm too afraid of generics in ooc.

source/ui/x11/X11Window.ooc
```diff
@@ -52,8 +64,34 @@ X11Window: class extends NativeWindow {
...
+				pixel := 0 as ULong
+				pixelData := pixel& as UChar*
+				color := image[column, row] toBgr()
+				pixelData@ = color blue as UChar
+				(pixelData + 1)@ = color green as UChar
+				(pixelData + 2)@ = color red as UChar
+				XPutPixel(this _xImage, column, row, pixel)
+			}
+		}
+		XClearWindow(this display, this backend)
```
sebastianbaginski: This one does not work properly (I think because of missing event processing)

sebastianbaginski:
So there are two separate classes for `UnixWindow` now. I don't think we should remove the one with OpenGL backend, as it works better than the one I just implemented. It would allow us to get rid of the `widgets` module, though...

source/ui/x11/X11Window.ooc
marcusnaslund: Generics don't work well, but would probably work OK for what you want to do here.

sebastianbaginski: You can do that if you want after this PR is merged, I just have more interesting things to do than trying to get generics in ooc to work properly :speak_no_evil:

marcusnaslund: Haha, don't know how to respond to that.

sebastianbaginski: I just don't touch generics in ooc if I don't have to. If you enjoy working with them in ooc then you have my admiration and respect :D

marcusnaslund: I have been burned many times, but I keep trying. :)

davidhesselbom: I think you're confusing generics with inheritance...

sebastianbaginski: Can `operator[]` be declared as virtual in `RasterImage` or `RasterPacked` ? Will it work if we reimplement it in derived classes ?
Another option is to have a `getColor: abstract func (x,y: Int) -> ColorBgr` somewhere in the image hierarchy.
Anyway let's fix that in another PR, this one is good enough to run noGpu on linux.

sebastianbaginski: I'm asking because now `operator[]` is just declared in some image classes, without any base class interface for it, so now it's not possible to accept a *generic* image type and use `operator[]` on it.

davidhesselbom: I don't think operators can be declared as virtual, period, and I have no idea why.

marcusnaslund: Operators are not inherited.

davidhesselbom: Yes, but why?

sebastianbaginski:
Updated with base class for `UnixWindow` used in both `gpu` and `nogpu` modes.

source/ui/x11/X11Window.ooc
marcusnaslund: Add a `TODO` explaining this here. If this is something we want within the forseeable future, also create an issue.

sebastianbaginski: I will fix this as soon as https://github.com/cogneco/ooc-kean/pull/665 is merged. This is on top of my TODO list in my head because it looks horribly :)

source/widgets/UnixWindow.ooc
```diff
@@ -22,26 +22,51 @@ use ooc-draw
...
-UnixWindow: class extends DisplayWindow {
-	_xWindow: X11Window
-	_openGLWindow: OpenGLWindow
-	context ::= this _openGLWindow context
-	init: func (size: IntSize2D, title: String) {
-		super()
-		this _xWindow = X11Window new(size, title)
-		this _openGLWindow = OpenGLWindow new(this _xWindow)
+version(unix || apple) {
+	UnixWindowBase: class extends DisplayWindow {
```
marcusnaslund: Shouldn't this be `abstract`?

source/widgets/UnixWindow.ooc
```diff
@@ -22,26 +22,51 @@ use ooc-draw
...
+				this _xWindow draw(raster)
+				raster referenceCount decrease()
+			}
+		}
 	}
-	free: override func {
-		this _openGLWindow free()
-		this _xWindow free()
-		super()
+version(gpuOff) {
```
marcusnaslund: I think I would rather see the constructor for `UnixWindowBase` create and return one of two subclasses depending on the flag rather than having two separate classes inside version blocks. For example like you(?) did with the `Thread` class. I leave that to @davidhesselbom to decide, though.

source/widgets/UnixWindow.ooc
sebastianbaginski: It has no abstract methods. In fact `UnixWindow` in `nogpu` mode should be a typedef for this one. But since ooc does not support typedefs...

marcusnaslund:
Besides the comment above, this looks good @davidhesselbom

source/widgets/UnixWindow.ooc
sebastianbaginski: I agree, this can be done better. Leave it for now, I will mark it as ready when I fix this.

sebastianbaginski:
Updated.

source/ui/x11/X11Window.ooc
```diff
@@ -52,8 +64,33 @@ X11Window: class extends NativeWindow {
...
 	}
+	draw: func (image: RasterYuv420Semiplanar) {
+		if (this _cacheSize != image size) {
+			if (this _xImage)
+				XDestroyImage(this _xImage)
+			imageData := gc_malloc(image width * image height * 4)
+			this _xImage = XCreateImage(this display, CopyFromParent, this _displayDepth, ZPixmap, 0, imageData, image width, image height, 8, 0)
+			this _cacheSize = image size
+		}
+		for (row in 0 .. image height )
```
davidhesselbom: No space before the `)`, please

source/ui/x11/X11Window.ooc
```diff
@@ -52,8 +64,33 @@ X11Window: class extends NativeWindow {
...
+			for (column in 0 .. image width) {
+				pixel := 0 as ULong
+				pixelData := pixel& as UInt8*
+				color := image[column, row] toBgr()
+				pixelData@ = color blue
+				(pixelData + 1)@ = color green
+				(pixelData + 2)@ = color red
+				XPutPixel(this _xImage, column, row, pixel)
```
davidhesselbom: This looks incredibly expensive. Why does this need doing this way?

sebastianbaginski: It is expensive and the display is lagging with big images. I don't know what is the internal representation of the `XImage`, I tried bgr, rgb, bgra, rgba, yuv and monochrome on my computer at home and none of it worked, result was a complete mess displayed instead of the image. This way it looks like it copies the pixel into the internal buffer with correct data internal representation. This is only for debugging anyway, we can (try ?) improve it later.

source/ui/x11/X11Window.ooc
```diff
@@ -52,8 +64,33 @@ X11Window: class extends NativeWindow {
...
+		XDestroyWindow(this display, this backend)
 		XCloseDisplay(this display)
 		super()
 	}
+	draw: func (image: RasterYuv420Semiplanar) {
+		if (this _cacheSize != image size) {
+			if (this _xImage)
+				XDestroyImage(this _xImage)
+			imageData := gc_malloc(image width * image height * 4)
+			this _xImage = XCreateImage(this display, CopyFromParent, this _displayDepth, ZPixmap, 0, imageData, image width, image height, 8, 0)
```
marcusnaslund: Question: That last argument is `bytes_per_line`. Why is it zero?

sebastianbaginski: > If you pass a zero value in bytes_per_line, Xlib assumes that the scanlines are contiguous in memory and calculates the value of bytes_per_line itself.

marcusnaslund: Thanks. :)

davidhesselbom: So what we should pass here is actually `image stride` or something?

source/ui/x11/X11Window.ooc
davidhesselbom: Couldn't you at least do `bgr := RasterBgr convertFrom(image)` and read each pixel from that image? The representation in `pixelData`  here is evidently `b g r` with one byte for each color component, just as in `RasterBgr`.

source/ui/x11/X11Window.ooc
sebastianbaginski: It should be the value of `bytesPerLine` for the `imageData` array up there, not the `image` object we are trying to display. It is allocated with size `image width * image height * 4` because we can't assume that every X11 system is using 3 bytes per pixel. If you alloc it as `image width * image height * 3` then it will crash on some systems (for example my linux at home). So I don't know it it actually needs to be `image width * 3` or `image width * 4`. The way it is done now you let the Xlib handle this.

source/ui/x11/X11Window.ooc
sebastianbaginski: So you mean allocate another memory for bgr image object and data just to read the pixel values from it ?

source/ui/x11/X11Window.ooc
davidhesselbom: Alright, thanks for explaining. I'm not sure, but I think maybe the `0` deserves a comment right there in the file.

source/ui/x11/X11Window.ooc
davidhesselbom: It would clean up the code here a lot. You wouldn't actually allocate the memory each time, since the underlying `ByteBuffer` is recycled when the temporary bgr image is freed.

sebastianbaginski: It works even if you convert to Bgra and memcpy the data, but I'm not sure if it's gonna work on every machine, I will commit and let you test it.

sebastianbaginski: Done. I'm pretty sure I tried to do it like that at home and it crashed. Then I started to search for more general alternatives like the put pixel loop

sebastianbaginski: Well not exactly working as it requires to swap r<->b channels now. Even if the first byte in `pixel` value was `blue` it looks like the internal layout is not `bgra`, it looks like `rgba` to me.

sebastianbaginski: So it is 
```ooc
	bgra := RasterBgra convertFrom(image)
	bgra swapRedBlue()
	memcpy(this _xImage@ data, bgra buffer pointer, bgra buffer size)
```
instead of the pixel loop now. Do you want to keep it this way ?

davidhesselbom: Wait, you did this
```ooc
pixelData@ = color blue
(pixelData + 1)@ = color green
(pixelData + 2)@ = color red
```
and it worked, but if you use a `RasterBgr´, red and blue needs to be swapped?

sebastianbaginski: Yes. Looks like the fact that first byte in `pixel` is blue does not imply the `bgra` internal layout. I think XPutPixel does the inversion on the provided pixel data.

sebastianbaginski: So I think now you understand why I did it that way, after experimenting with it I just thought it is not safe to assume anything about the internal representation.

davidhesselbom: Okay. What if you don't swap red and blue, keep the loop, but don't convert pixels yourself, and just give `XPutPixel` a pointer to a pixel in the `RasterBgr`?

davidhesselbom: Also, you still need to call `bgra referenceCount decrease()` before the function returns, or you'll have a massive memory leak :)

davidhesselbom: And it should be a `RasterBgr`, not a `RasterBgra`. The incoming YUV image has no alpha channel anyway...

davidhesselbom: Or does the target image need to have 4 bytes per pixel?

sebastianbaginski: `XPutPixel` expects `Long` so in order to give it a `ColorBgra` you would need to do
```ooc
for (row in 0 .. image height)
	for (column in 0 .. image width)
		XPutPixel(this _xImage, column, row, ((bgra[column, row]&) as Long*)@)
```

sebastianbaginski: If the XImage does not have enough memory to keep 4 bytes per pixel it will crash on some systems.

sebastianbaginski: This is hard, I'm not very experienced X11 programmer, in fact this is the first thing I've ever done using XLib. I'd say lets just keep the memcpy as the code is clean and performance is better than with the pixel loop. IF something ever breaks for someone around here then we can revisit this code.

davidhesselbom: > ((bgra[column, row]&) as Long*)@

Wow, that's not pretty at all, haha... perhaps `RasterImage` should have a function for returning the pointer to a pixel, instead. I still prefer this over the 7 lines of code you have before, though.

davidhesselbom: I hear you about XLib, I thought it was hard, too, and I've heard many people criticize it for various reasons. We can keep the memcpy, that works for me.

sebastianbaginski: OK, commit updated.

PR #610, opened 2015-10-27
Math: Updates to Box, Size2D, and Size3D

marcusnaslund:
Part 2 of #491


marcusnaslund:
Conflicts resolved.

source/math/IntBox2D.ooc
```diff
@@ -119,18 +120,24 @@ IntBox2D: cover {
 		array := input split(',')
 		This new(array[0] toInt(), array[1] toInt(), array[2] toInt(), array[3] toInt())
 	}
-	create: static func (leftTop: IntPoint2D, size: IntSize2D) -> This { This new(leftTop, size) }
-	create: static func ~fromFloats (left, top, width, height: Int) -> This { This new(left, top, width, height) }
 	createAround: static func (center: IntPoint2D, size: IntSize2D) -> This { This new(center + (-size) / 2, size) }
```
davidhesselbom: You wouldn't mind fixing this one, would you?

source/math/IntSize2D.ooc
```diff
@@ -36,7 +36,7 @@ IntSize2D: cover {
 	minimum: func ~Int (ceiling: Int) -> This { this minimum(This new(ceiling)) }
 	maximum: func ~Int (floor: Int) -> This { this maximum(This new(floor)) }
 	clamp: func (floor, ceiling: This) -> This { This new(this width clamp(floor width, ceiling width), this height clamp(floor height, ceiling height)) }
-	fillEven: static func (other: This) -> This { This new(other width + (other width % 2 == 1 ? 1 : 0), other height + (other height % 2 == 1 ? 1 : 0)) }
+	fillEven: static func (other: This) -> This { This new(other width + (Int odd(other width) ? 1 : 0), other height + (Int odd(other height) ? 1 : 0)) }
```
davidhesselbom: ... why didn't we just use `other width % 2`? Oh well...

davidhesselbom: Wait, maybe it's because it didn't work well for negative values... but in that case, I think `Int modulo(other width, 2)` would still read more clearly than this.

source/math/FloatBox2D.ooc
```diff
@@ -140,25 +140,31 @@ FloatBox2D: cover {
...
+	createAround: static func (center: FloatPoint2D, size: FloatSize2D) -> This { This new(center - (size / 2.0f), size) }
 	bounds: static func (left, right, top, bottom: Float) -> This { This new(left, top, right - left, bottom - top) }
-	bounds: static func ~fromArray (points: FloatPoint2D[]) -> This { This bounds(points as ArrayList<FloatPoint2D>) }
-	bounds: static func ~fromList (points: ArrayList<FloatPoint2D>) -> This {
+	bounds: static func ~fromArray (points: FloatPoint2D[]) -> This {
+		pointsAsVectorList := FloatPoint2DVectorList new()
+		for (i in 0 .. points length)
+			pointsAsVectorList add(points[i])
+		result := This bounds(pointsAsVectorList)
+		pointsAsVectorList free()
```
davidhesselbom: This isn't very good... maybe we can create a new function that takes a `FloatPoint2D*` and a `count: Int` that runs the loop. Then the array version can call that version as `This bounds(points data, points length)` and the `VectorList` version can call `This bounds(points pointer as This*, points count)`.

source/math/FloatBox2D.ooc
```diff
@@ -140,25 +140,31 @@ FloatBox2D: cover {
...
 	}
-	create: static func (leftTop: FloatPoint2D, size: FloatSize2D) -> This { This new(leftTop, size) }
-	create: static func ~fromFloats (left, top, width, height: Float) -> This { This new(left, top, width, height) }
-	createAround: static func (center: FloatPoint2D, size: FloatSize2D) -> This { This new(center + (-size) / 2, size) }
+	createAround: static func (center: FloatPoint2D, size: FloatSize2D) -> This { This new(center - (size / 2.0f), size) }
 	bounds: static func (left, right, top, bottom: Float) -> This { This new(left, top, right - left, bottom - top) }
-	bounds: static func ~fromArray (points: FloatPoint2D[]) -> This { This bounds(points as ArrayList<FloatPoint2D>) }
-	bounds: static func ~fromList (points: ArrayList<FloatPoint2D>) -> This {
+	bounds: static func ~fromArray (points: FloatPoint2D[]) -> This {
+		pointsAsVectorList := FloatPoint2DVectorList new()
```
davidhesselbom: At the very least, this should be `FloatPoint2DVectorList new(points length)`.

source/math/IntSize2D.ooc
marcusnaslund: After I wrote this I've realized the `size` covers are assumed to have nonnegative values - they're not general vectors. Right? I don't know why though. In that case using `%` is just fine.

source/math/IntBox2D.ooc
marcusnaslund: Fix what? The `+ (-size)` thing?

source/math/IntSize2D.ooc
davidhesselbom: You know, this function is never actually used... anywhere. We could just, you know, get rid of it.

marcusnaslund: Just because it isn't used at the moment doesn't mean to remove it, but sure, that solves a lot of problems.

marcusnaslund:
Updated.

marcusnaslund:
Validated @davidhesselbom

source/math/IntBox2D.ooc
```diff
@@ -118,18 +119,18 @@ IntBox2D: cover {
...
-	create: static func ~fromFloats (left, top, width, height: Int) -> This { This new(left, top, width, height) }
-	createAround: static func (center: IntPoint2D, size: IntSize2D) -> This { This new(center + (-size) / 2, size) }
-	bounds: func (left, right, top, bottom: Int) -> This { This new(left, top, right - left, bottom - top) }
-	bounds: func ~fromArray (points: IntPoint2D[]) -> This { this bounds(points as ArrayList<IntPoint2D>) }
-	bounds: func ~fromList (points: ArrayList<IntPoint2D>) -> This {
+	createAround: static func (center: IntPoint2D, size: IntSize2D) -> This { This new(center - (size / 2), size) }
+	bounds: static func (left, right, top, bottom: Int) -> This { This new(left, top, right - left, bottom - top) }
+	bounds: static func ~fromArray (points: IntPoint2D[]) -> This { This bounds(points data, points length) }
+	bounds: static func ~fromList (points: VectorList<IntPoint2D>) -> This { This bounds(points pointer as IntPoint2D*, points count) }
+	bounds: static func ~fromPointer (data: IntPoint2D*, count: Int) -> This {
```
davidhesselbom: The irony is, this function will have no bounds checking at all... ^^

source/math/IntBox2D.ooc
```diff
@@ -118,18 +119,18 @@ IntBox2D: cover {
 		array := input split(',')
```
davidpiuva: I remember that this one leaked memory in Cutter.

davidhesselbom: It's fixed in #575.

marcusnaslund: This isn't even in this PR...

source/math/FloatSize3D.ooc
```diff
@@ -79,6 +79,9 @@ FloatSize3D: cover {
 		array := input split(',')
 		This new (array[0] toFloat(), array[1] toFloat(), array[2] toFloat())
 	}
+	linearInterpolation: static func (a, b: This, ratio: Float) -> This {
+		This new(Float linearInterpolation(a width, b width, ratio), Float linearInterpolation(a height, b height, ratio), Float linearInterpolation(a depth, b depth, ratio))
+	}
```
davidpiuva: Once every PR is in, we could try to crate naming and argument consistency with color's Blend function. Both could be called `mix` as GLSL does for both colors and vectors.

davidhesselbom: `mix` is shorter than `linearInterpolation`, that's for sure. I wouldn't mind `lerp` myself, but it's not up to me, so :)

marcusnaslund:
Thank you `Promise`... @davidhesselbom please restart the Appveyor test.

source/math/FloatBox2D.ooc
```diff
@@ -149,25 +149,25 @@ FloatBox2D: cover {
...
-	create: static func (leftTop: FloatPoint2D, size: FloatSize2D) -> This { This new(leftTop, size) }
-	create: static func ~fromFloats (left, top, width, height: Float) -> This { This new(left, top, width, height) }
-	createAround: static func (center: FloatPoint2D, size: FloatSize2D) -> This { This new(center + (-size) / 2, size) }
+	createAround: static func (center: FloatPoint2D, size: FloatSize2D) -> This { This new(center - (size / 2.0f), size) }
 	bounds: static func (left, right, top, bottom: Float) -> This { This new(left, top, right - left, bottom - top) }
-	bounds: static func ~fromArray (points: FloatPoint2D[]) -> This { This bounds(points as ArrayList<FloatPoint2D>) }
-	bounds: static func ~fromList (points: ArrayList<FloatPoint2D>) -> This {
+	bounds: static func ~fromArray (points: FloatPoint2D[]) -> This { This bounds(points data, points length) }
+	bounds: static func ~fromList (points: VectorList<FloatPoint2D>) -> This { This bounds(points pointer as FloatPoint2D*, points count) }
+	bounds: static func ~fromPointer (data: FloatPoint2D*, count: Int) -> This {
```
davidpiuva: Maybe ~fromCArray?

davidhesselbom:
Can't you do it?

marcusnaslund:
No.

source/math/FloatBox2D.ooc
davidhesselbom: Isn't this clearer, though?

davidpiuva: A C array is a pointer too but pointers can also point to scalars.

marcusnaslund: Yes, and `count` could be 1. I don't think we should include mentions to C in ooc.

davidpiuva: That makes sense.

davidhesselbom: Also, the parameter is, in fact, a pointer, and the `[]` argument is an ooc array.

davidpiuva:
The syntax looks good to me.

source/math/FloatSize2D.ooc
```diff
@@ -34,7 +34,7 @@ FloatSize2D: cover {
 	pNorm: func (p: Float) -> Float {
 		p == 1 ?
 		this width abs() + this height abs() :
-		(this width abs() pow(p) + this height abs() pow(p)) pow(1 / p)
+		(this width abs() pow(p) + this height abs() pow(p)) pow(1.0f / p)
```
davidhesselbom: These two lines should be indented one step.

source/math/FloatSize3D.ooc
```diff
@@ -33,8 +33,8 @@ FloatSize3D: cover {
 	init: func@ ~default { this init(0.0f, 0.0f, 0.0f) }
 	pNorm: func (p: Float) -> Float {
 		p == 1 ?
-		this width abs() + this height abs() :
-		(this width abs() pow(p) + this height abs() pow(p)) pow(1 / p)
+		this width abs() + this height abs() + this depth abs() :
+		(this width abs() pow(p) + this height abs() pow(p) + this depth abs() pow(p)) pow(1.0f / p)
```
davidhesselbom: Same here.

source/math/FloatSize3D.ooc
```diff
@@ -33,8 +33,8 @@ FloatSize3D: cover {
 	init: func@ ~default { this init(0.0f, 0.0f, 0.0f) }
 	pNorm: func (p: Float) -> Float {
 		p == 1 ?
```
davidhesselbom: `p == 1.0f`

source/math/FloatSize3D.ooc
davidhesselbom: On the other hand, how likely is `p` to be `1.0f`? I think maybe this whole function should just be reduced to
```ooc
(this width abs() pow(p) + this height abs() pow(p) + this depth abs() pow(p)) pow(1.0f / p)
```
but if `1.0f` is a very common value for `p`, the performance boost might make up for the decreased readability.

source/math/IntBox2D.ooc
```diff
@@ -18,6 +18,7 @@ import IntPoint2D
 import IntSize2D
 import FloatPoint2D
 import FloatBox2D
+import VectorList
```
davidhesselbom: No, `use ooc-collections`.

source/math/FloatSize3D.ooc
marcusnaslund: I can't really say. I'll leave this optimization here for now.

davidhesselbom:
@marcusnaslund Conflicts.

marcusnaslund:
Resolved.

PR #612, opened 2015-10-27
Math: Updates to FloatComplex and Float(Complex)VectorList

marcusnaslund:
Part 4 of #491


marcusnaslund:
Wait what? How did the Pen color commit get in here when I rebased?

davidhesselbom:
Beats me...

marcusnaslund:
Why is this not ready?

davidhesselbom:
Because of those other commits that don't belong there. Rebase again and you should be good to go.

marcusnaslund:
Validated @davidhesselbom

source/math/FloatComplexVectorList.ooc
```diff
@@ -30,21 +30,26 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
...
-		sum() / this _count
+	sum: FloatComplex {
+		get {
+			result := FloatComplex new()
+			for (i in 0 .. this _count)
+				result = result + this[i]
+			result
+		}
 	}
+	mean ::= this sum / this _count
```
sebastianbaginski: I still don't understand it. From this guide (thanks @davidhesselbom ) https://msdn.microsoft.com/en-us/library/vstudio/ms229054%28v=vs.100%29.aspx : "*properties should not be computationally complex or produce side effects.*"
This `mean` property definitely does not follow this guideline - it is computationally complex. Later when this property is used somewhere it won't be clear that it calls a complex function behind the scenes, like
```ooc
for (i in 0 .. bignumber)
    dosomething(vector mean)
```
compared to 
```ooc
for (i in 0 .. bignumber)
    dosomething(vector mean()) // now I see you, you sneaky little funtction...
```
it looks like `vector mean` is a variable, while it is just a function lurking behind obfuscated syntax. In order to find out, one needs to open the definition of this class. So right now it looks to me like our guidelines about properties are : "everything that is read-only should be a property". It should be more like "everything that is read-only and cheap to compute should be a property".
That's why I don't get it. Something that looks like a variable access but actually calls several functions every time it is used is just unnecessary confusion.

davidhesselbom: I know - we currently disregard the complexity thing. One way to make this a proper property (well, almost) would be to store the mean somewhere and invalidate it whenever the list is updated instead of computing it from scratch every time. But, that would add a few more lines of code...

source/math/FloatComplexVectorList.ooc
```diff
@@ -30,21 +30,26 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
...
-	}
-	mean: func -> FloatComplex {
-		sum() / this _count
+	sum: FloatComplex {
+		get {
+			result := FloatComplex new()
+			for (i in 0 .. this _count)
+				result = result + this[i]
+			result
+		}
```
davidhesselbom: Do you want fix this?
```ooc
sum: FloatComplex { get {
   ...
}}
```

source/math/FloatVectorList.ooc
```diff
@@ -21,15 +21,20 @@ import FloatComplexVectorList
...
 	}
 	init: func ~fromVectorList (other: VectorList<Float>) {
-		this super(other _vector)
+		super(other _vector)
 		this _count = other count
 	}
+	init: func ~withValue (capacity: Int, value: Float) {
+		super(capacity)
+		for (i in 0 .. capacity)
+			this add(value)
```
davidhesselbom: This doesn't need doing if `value` is `0.0f`...

source/math/FloatVectorList.ooc
```diff
@@ -181,16 +186,10 @@ FloatVectorList: class extends VectorList<Float> {
...
-			result add(1.0f)
-		result
+		This new(count, 1.0f)
 	}
 	getZeros: static func (count: Float) -> This {
-		result := This new(count)
-		for (i in 0 .. count)
-			result add(0.0f)
-		result
+		This new(count, 0.0f)
```
davidhesselbom: ... so I suggest this instead does this
```ooc
result := This new(count)
result _count = count
result
```

source/math/FloatVectorList.ooc
```diff
@@ -181,16 +186,10 @@ FloatVectorList: class extends VectorList<Float> {
 		max != 0 ? (this / max) : this copy()
 	}
 	getOnes: static func (count: Float) -> This {
-		result := This new(count)
-		for (i in 0 .. count)
-			result add(1.0f)
-		result
+		This new(count, 1.0f)
 	}
 	getZeros: static func (count: Float) -> This {
```
davidhesselbom: `count` definitely shouldn't be a `Float`, but an `Int`.

source/math/FloatVectorList.ooc
```diff
@@ -181,16 +186,10 @@ FloatVectorList: class extends VectorList<Float> {
 		max != 0 ? (this / max) : this copy()
 	}
 	getOnes: static func (count: Float) -> This {
```
davidhesselbom: I think we discussed calling these `ones` and `zeros`, respectively, but I guess it can wait.

test/math/FloatComplexTest.ooc
```diff
@@ -34,7 +34,7 @@ FloatComplexTest: class extends Fixture {
 	complexNumberArray[2] = complexNumber2
 	complexNumberArray[3] = complexNumber3
 
-	tolerance := 0.00001
+	tolerance := 0.00001f
```
davidhesselbom: `tolerance := 1.0e-5f`

source/math/FloatVectorList.ooc
davidhesselbom: Since this PR is about cleanup, I mean...

source/math/FloatComplexVectorList.ooc
```diff
@@ -155,7 +163,9 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 		This _fastFourierTransformHelper(input, 0, input count, buffer, 0, result, 0)
 	}
 	createFFTBuffer: static func (inputSize: Int) -> This {
-		This createDefault(This _fastFourierTransformBufferSize(inputSize))
+		result := This new(This _fastFourierTransformBufferSize(inputSize))
+		result _count = inputSize
```
davidhesselbom: `_fastFourierTransformBufferSize` returns an `Int` which is used in the constructor to set the capacity of `result`. Then you assign `_count` to a different value. Doesn't seem like a good idea...

davidhesselbom: `createDefault`, which was removed here, sets `_count` to the same value as `capacity`. So you probably want this function to do this:
```ooc
count := This _fastFourierTransformBufferSize(inputSize)
result := This new(count)
result _count = count
result
```
```

File: source/math/FloatVectorList.ooc (hunk 4 above)
Comment (davidhesselbom) [inline source/math/FloatVectorList.ooc]: And then this function will do exactly what the `createDefault` function in `FloatComplexVectorList`, which you are currently removing, does. So maybe you shouldn't remove that function there, but rename it to `getZeroes` and save a few lines of code by calling 
```ooc
This getZeros(count)
```
instead of 
```
result := This new(input count)
result _count = input count
```
in, what, 4 places.

source/math/FloatComplexVectorList.ooc
davidhesselbom: Or, aligning with my suggestion below,
```ooc
This getZeros(This _fastFourierTransformBufferSize(inputSize))
```
after renaming `createDefault` to `getZeros` instead of removing it.

source/math/FloatComplexVectorList.ooc
```diff
@@ -167,17 +177,6 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 			result[i] = (result[i] conjugate) / (input count)
 		result
 	}
-	createDefault: static func (capacity: Int) -> This {
-		result := This new(capacity)
-		result _count = capacity
-		result
-	}
-	createDefault: static func ~withValue (capacity: Int, value: FloatComplex) -> This {
```
davidhesselbom: This function can still be removed though.

source/math/FloatComplexVectorList.ooc
marcusnaslund: No. :)

marcusnaslund: My plan is to go through and clean up all the math classes & covers soon enough.

source/math/FloatVectorList.ooc
marcusnaslund: Should I add an extra check for this? Then I have to set `_count` manually instead.

source/math/FloatVectorList.ooc
marcusnaslund: Ah, yes, good catch.

source/math/FloatVectorList.ooc
marcusnaslund: Ah yes. I'm not sure there's a net save in terms of lines of code, but sure. Done.

source/math/FloatVectorList.ooc
davidhesselbom: No, ignore this, the other thing you did fixed this.

source/math/FloatComplexVectorList.ooc
```diff
@@ -155,7 +165,9 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 		This _fastFourierTransformHelper(input, 0, input count, buffer, 0, result, 0)
 	}
 	createFFTBuffer: static func (inputSize: Int) -> This {
-		This createDefault(This _fastFourierTransformBufferSize(inputSize))
+		bufferSize := This _fastFourierTransformBufferSize(inputSize)
+		result := This getZeros(bufferSize)
+		result
```
davidhesselbom: This wouldn't compile if you put it all on a single line?

marcusnaslund: It does, thanks.

source/math/FloatComplexVectorList.ooc
```diff
@@ -155,7 +165,8 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
 		This _fastFourierTransformHelper(input, 0, input count, buffer, 0, result, 0)
 	}
 	createFFTBuffer: static func (inputSize: Int) -> This {
-		This createDefault(This _fastFourierTransformBufferSize(inputSize))
+		bufferSize := This _fastFourierTransformBufferSize(inputSize)
+		This getZeros(bufferSize)
```
davidhesselbom: Again, I ask: This wouldn't compile if you put it on a single line? :)

marcusnaslund: It does. :)

PR #614, opened 2015-10-27
Math: Updates to Shell and ConvexHull

marcusnaslund:
Part 6 of #491

### DIFF
--- a/source/math/FloatShell2D.ooc
+++ b/source/math/FloatShell2D.ooc
@@ -28,9 +28,9 @@ FloatShell2D: cover {
 	isZero ::= this left == 0 && this right == 0 && this top == 0 && this bottom == 0
 	notZero ::= this left != 0 && this right != 0 && this top != 0 && this bottom != 0
 	init: func@ (=left, =right, =top, =bottom)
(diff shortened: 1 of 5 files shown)


source/math/FloatShell2D.ooc
```diff
@@ -28,9 +28,9 @@ FloatShell2D: cover {
 	isZero ::= this left == 0 && this right == 0 && this top == 0 && this bottom == 0
 	notZero ::= this left != 0 && this right != 0 && this top != 0 && this bottom != 0
 	init: func@ (=left, =right, =top, =bottom)
-	init: func@ ~fromFloat (value: Float) { this init(value, value) }
+	init: func@ ~fromFloat (value: Float) { this init(value, value, value, value) }
 	init: func@ ~fromFloats (x, y: Float) { this init(x, x, y, y) }
-	init: func@ ~default { this init(0.0f) }
+	init: func@ ~default { this init(0.0f, 0.0f, 0.0f, 0.0f) }
```
davidhesselbom: This was actually on purpose (chained constructors), but I guess it doesn't matter either way.

marcusnaslund: Eliminating one call in the chain seems good to me.

PR #622, opened 2015-10-28
Added focalLength check in FloatTransform3D.

ziquanyu:
bug fix for FloatTransform3D.

### DIFF
--- a/source/math/FloatTransform3D.ooc
+++ b/source/math/FloatTransform3D.ooc
@@ -208,7 +208,7 @@ FloatTransform3D: cover {
 	}
 	transformAndProject: func ~FloatPoint2D (point: FloatPoint2D, focalLength: Float) -> FloatPoint2D {
 		transformedWorldPoint := this * FloatPoint3D new(point x, point y, focalLength)
-		focalLength == 0.0f ? FloatPoint2D new(transformedWorldPoint x, transformedWorldPoint y) : this project(transformedWorldPoint, focalLength)
+		focalLength < 0.000000001f ? FloatPoint2D new(transformedWorldPoint x, transformedWorldPoint y) : this project(transformedWorldPoint, focalLength)
 	}
 	transformAndProject: func ~FloatBox2D (box: FloatBox2D, focalLength: Float) -> FloatBox2D {
 		FloatBox2D new(this transformAndProject(box leftTop, focalLength), this transformAndProject(box rightBottom, focalLength))
--- a/source/math/FloatTransform3D.ooc
+++ b/source/math/FloatTransform3D.ooc
@@ -208,7 +208,7 @@ FloatTransform3D: cover {
 	}
 	transformAndProject: func ~FloatPoint2D (point: FloatPoint2D, focalLength: Float) -> FloatPoint2D {
 		transformedWorldPoint := this * FloatPoint3D new(point x, point y, focalLength)
-		focalLength < 0.000000001f ? FloatPoint2D new(transformedWorldPoint x, transformedWorldPoint y) : this project(transformedWorldPoint, focalLength)
+		focalLength < Float epsilon ? FloatPoint2D new(transformedWorldPoint x, transformedWorldPoint y) : this project(transformedWorldPoint, focalLength)
 	}
 	transformAndProject: func ~FloatBox2D (box: FloatBox2D, focalLength: Float) -> FloatBox2D {
 		FloatBox2D new(this transformAndProject(box leftTop, focalLength), this transformAndProject(box rightBottom, focalLength))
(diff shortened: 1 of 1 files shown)


source/math/FloatTransform3D.ooc
```diff
@@ -208,7 +208,7 @@ FloatTransform3D: cover {
 	}
 	transformAndProject: func ~FloatPoint2D (point: FloatPoint2D, focalLength: Float) -> FloatPoint2D {
 		transformedWorldPoint := this * FloatPoint3D new(point x, point y, focalLength)
-		this project(transformedWorldPoint, focalLength)
+		focalLength == 0.0f ? FloatPoint2D new(transformedWorldPoint x, transformedWorldPoint y) : this project(transformedWorldPoint, focalLength)
```
davidhesselbom: Looks good, but before I merge, what does it mean when the focal lenght is 0? Can it be negative?

ziquanyu: https://en.wikipedia.org/wiki/Focal_length
As I understand about working camera lens, focal length should be a positive number (greater than 0).

You can read link below to know when focal length is negative.
https://www.rp-photonics.com/focal_length.html

davidhesselbom: Yes, ok. Thanks for the links! But, this PR adds special behavior for the case when focal length is 0. How can this happen?

source/math/FloatTransform3D.ooc
```diff
@@ -208,7 +208,7 @@ FloatTransform3D: cover {
 	}
 	transformAndProject: func ~FloatPoint2D (point: FloatPoint2D, focalLength: Float) -> FloatPoint2D {
 		transformedWorldPoint := this * FloatPoint3D new(point x, point y, focalLength)
-		this project(transformedWorldPoint, focalLength)
+		focalLength < 0.000000001f ? FloatPoint2D new(transformedWorldPoint x, transformedWorldPoint y) : this project(transformedWorldPoint, focalLength)
```
davidhesselbom: How was this number chosen? A comment to explain why we should use this number would be nice. Also, please write it as `1.0e-9f`

source/math/FloatTransform3D.ooc
ziquanyu: I tried to integrate on the test phone, there was no data about focal length in the phone's metaData, which would lead to NaN issue because there was no focal length check. 

Gyro stabilization needs it very much. But another one does not need it at all.  If we do not check focal length here, which will affect every transform hereafter.

davidhesselbom: > NaN issue

Oh, I see. Thanks!

source/math/FloatTransform3D.ooc
ziquanyu: L269: epsilon ::= static FLT_EPSILON in math.ooc
This should be a better choice.

source/math/FloatTransform3D.ooc
ziquanyu: When focal length is 0, it will cause 0.0 / 0.0 = NaN. We don't have "isNaN" either. 
It means when there comes an "NaN", it will be like an virus. All related computation will have same result --NaN. :)

ziquanyu:
One question about "integration/appveyor". What does this do? It seems every latest PR failed.

source/math/FloatTransform3D.ooc
davidhesselbom: > We don't have "isNaN" either.

What do you mean? You can check if a number is `NaN` by comparing it to itself. If `value` is a `Float` and `value != value`, then `value` is `NaN`. Are you suggesting we should have a static function for this?

davidhesselbom:
AppVeyor is like Circle CI, but on Windows. It always fails because something is wrong with threads, and we still haven't fixed it. Ignore it for now.

source/math/FloatTransform3D.ooc
ziquanyu: I mean kean does not have NaN check. 
You are right, if we know somewhere may cause NaN, this simple check can be placed there.
It's not a problem now. But maybe it's better to have one "isNumber()" in the future as Simon suggested.

ziquanyu:
Got it.

source/math/FloatTransform3D.ooc
davidhesselbom: You mean
```ooc
//  math.ooc

extend Float {
    ....
    isNumber: func -> Bool {
        this == this
    }
    ....
```
?

ziquanyu: This is what I mean, yet simon wants more, I guess. We can maybe include that later.  It's not necessary to have "isNumber" everywhere either.

davidhesselbom: No, we should avoid having to check for `NaN`, of course, but when we do, it would be nice to have a property for it. See https://github.com/cogneco/ooc-kean/issues/626.

ziquanyu: agree.

PR #623, opened 2015-10-28
Fixed the mess with pad(Float) in FloatBox2D

### LINKED ISSUE #599: FloatBox2D pad(Float)
There are two functions with the same signature in `FloatBox2D` (and possibly elsewhere, too):
```ooc
	pad: func ~fromFloat (pad: Float) -> This { this pad(pad, pad, pad, pad) }

		this pad(pad * this size / 2.0f)
	}
```
Evidently, these two functions don't do that same thing, but both can be called like `pad(myFloat)`, which means we don't know which one of them is actually called. This is not good.

Rename one of these functions to resolve the ambiguity, but first, find every place `pad` is called and figure out which one of them is actually being called in the C code.

@thomasfanell, is there any way we can make magic detect this (functions that have a "twin" with the same name and the same signature in the same namespace)? It should be rock's job, really, but since it doesn't do its job, someone else will have to.

marcusnaslund:
Fixes #599 
Also added missing test for the new `padFractionAverage`.


source/math/FloatBox2D.ooc
```diff
@@ -57,14 +57,14 @@ FloatBox2D: cover {
...
 	pad: func ~fromSize (pad: FloatSize2D) -> This { this pad(pad width, pad width, pad height, pad height) }
-	pad: func ~fraction (pad: Float) -> This {
-		this pad(pad * this size / 2.0f)
+	enlargeEvenly: func (fraction: Float) -> This {
+		this pad(fraction * (this size width + this size height) / 2.0f)
 	}
-	padFractionAverage: func (pad: Float) -> This {
-		this pad(pad * (this size width + this size height) / 2.0f)
+	enlarge: func (fraction: Float) -> This {
+		this pad(fraction * this size / 2.0f)
```
marcusnaslund: This looks confusing, but I switched places because this one was so alike `shrink`.

marcusnaslund:
Conflicts resolved.

source/math/FloatBox2D.ooc
davidhesselbom: Can you just call `this scale(fraction + 1.0f)`?

marcusnaslund: Oh, we have that. Yes. But.. do we need `extend`? Seems like the use of `scale` is pretty straight-forward in similar circumstances.

davidhesselbom: `extend`?

marcusnaslund: Sorry, `enlarge`.

davidhesselbom: I think it could be useful, yes. Not strictly necessary, no, but the whole point is to make code elsewhere more readable (without making it harder to read here). I'm not sure what to do with `enlargeEvenly` though.

davidhesselbom: `enlargeEvenly` could, of course, call `resizeTo` instead of `pad`.

marcusnaslund: Sure, but that just makes the call in `enlargeEvenly` longer without any clear benefit.

davidhesselbom: The benefit is that it doesn't call `pad`... :)

marcusnaslund: Why is that a benefit?

marcusnaslund:
Validated @davidhesselbom

source/math/FloatBox2D.ooc
```diff
@@ -58,14 +58,14 @@ FloatBox2D: cover {
 	}
 	pad: func ~fromFloat (pad: Float) -> This { this pad(pad, pad, pad, pad) }
 	pad: func ~fromSize (pad: FloatSize2D) -> This { this pad(pad width, pad width, pad height, pad height) }
-	pad: func ~fraction (pad: Float) -> This {
-		this pad(pad * this size / 2.0f)
+	enlargeEvenly: func (fraction: Float) -> This {
```
davidhesselbom: My only problem here is the name `enlargeEvenly`. It's not terrible, but it's not good either. A fix can be left for later, but then it will need a corresponding PR in depending projects, again.

source/math/FloatBox2D.ooc
```diff
@@ -58,14 +58,14 @@ FloatBox2D: cover {
...
 	}
-	padFractionAverage: func (pad: Float) -> This {
-		this pad(pad * (this size width + this size height) / 2.0f)
+	enlarge: func (fraction: Float) -> This {
+		this scale(1.0f + fraction)
 	}
-	shrink: func ~fraction (margin: Float) -> This {
-		this pad(-margin * this height / 2.0f)
+	shrink: func (fraction: Float) -> This {
+		this scale(1.0f - fraction)
```
tomasmahlberg: Maybe it would make sense to rename `enlarge` and `shrink` to `scaleUp` and `scaleDown` respectively since they are basically wrappers for `scale`?

test/math/FloatBox2DTest.ooc
```diff
@@ -50,12 +50,17 @@ FloatBox2DTest: class extends Fixture {
 			expect(inBox[1], is equal to(1))
 			expect(inBox[2], is equal to(3))
 		})
-		this add("pad and shrink ~fraction", func {
+		this add("enlarge and shrink ~fraction", func {
```
davidhesselbom: You can get rid of `~fraction`.

source/math/FloatBox2D.ooc
davidhesselbom: I don't think an API should contain `scale` and `scaleUp`.

test/math/FloatBox2DTest.ooc
marcusnaslund: Sure

source/math/FloatBox2D.ooc
marcusnaslund: If you have a better name, we can change it right away

davidhesselbom: Well, that's the problem, I don't, so I guess it'll have to do. A comment explaining what you mean by "evenly" would be nice, though.

marcusnaslund: I don't know what my comment will say that the line afterwards won't.

davidhesselbom: Well... okay. Fine.

PR #624, opened 2015-10-28
Added test for IntTransform2D

### LINKED ISSUE #602: IntTransform2D is feeling lonely
Right now, there is no test for `IntTransform2D`. Before I go and create one, I would like to make sure we need it. Right now, we are not using this class (cover) at all, and we never have.

marcusnaslund:
Fixes #602 
Basically copied and adapted the test from `FloatTransform2D`.

### DIFF
--- a/test/math/IntTransform2DTest.ooc
+++ b/test/math/IntTransform2DTest.ooc
@@ -1,0 +1,149 @@
+use ooc-unit
+use ooc-math
+import math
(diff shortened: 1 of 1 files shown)


marcusnaslund:
The test coverage isn't complete, but I think we can get rid of a lot of stuff in `IntTransform2D`, so let's leave that for a future PR.

PR #627, opened 2015-10-28
Fixed memory leaks in Exception class

sebastianbaginski:
Keep track of stack frames for correct cleanup.
Fixed memory leak in `Stack` class.

### DIFF
--- a/source/sdk/lang/Exception.ooc
+++ b/source/sdk/lang/Exception.ooc
@@ -34,7 +34,10 @@ exceptionStack := ThreadLocal<Stack<StackFrame>> new()
 _exception := ThreadLocal<Exception> new()
 _EXCEPTION: Int = 1
 
+stackFramesToDelete := Stack<StackFrame> new()
+
 _pushStackFrame: inline func -> StackFrame {
+  _cleanupStackFrames()
     stack: Stack<StackFrame>
     if(!exceptionStack hasValue?()) {
         stack = Stack<StackFrame> new()
@@ -44,7 +47,20 @@ _pushStackFrame: inline func -> StackFrame {
     }
     buf := StackFrame new()
     stack push(buf)
+    exceptionStack set(stack)
     buf
+}
+
+_markStackFrame: inline func (frame: StackFrame) {
+  stackFramesToDelete push(frame)
+}
+
+_cleanupStackFrames: inline func {
+  while (stackFramesToDelete empty?() == false) {
+    frame := stackFramesToDelete pop() as StackFrame
+    if (frame)
+      gc_free(frame as Void*)
+  }
 }
 
 _setException: inline func (e: Exception) {
@@ -55,8 +71,14 @@ _getException: inline func -> Exception {
     _exception get()
 }
 
-_popStackFrame: inline func -> StackFrame {
-    exceptionStack get() as Stack<StackFrame> pop() as StackFrame
+_popStackFrame: inline func {
+    frame := exceptionStack get() as Stack<StackFrame> pop() as StackFrame
+    if (frame)
+      gc_free(frame as Void*)
+}
+
+_takeStackFrame: func -> StackFrame {
+  exceptionStack get() as Stack<StackFrame> pop() as StackFrame
 }
 
 _hasStackFrame: inline func -> Bool {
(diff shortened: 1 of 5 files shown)


source/sdk/lang/Exception.ooc
```diff
@@ -47,6 +49,18 @@ _pushStackFrame: inline func -> StackFrame {
     buf
 }
 
+_markStackFrame: inline func (frame: StackFrame) {
```
davidhesselbom: Just so you know, `inline` doesn't actually work.

source/sdk/lang/Exception.ooc
```diff
@@ -34,6 +34,8 @@ exceptionStack := ThreadLocal<Stack<StackFrame>> new()
 _exception := ThreadLocal<Exception> new()
 _EXCEPTION: Int = 1
 
+stackFramesToDelete := Stack<StackFrame> new()
```
davidhesselbom: I know this file was already a mess, but, does this really need to be a global thing? Could it be put inside a class and made static, perhaps?

source/sdk/lang/Exception.ooc
```diff
@@ -47,6 +49,18 @@ _pushStackFrame: inline func -> StackFrame {
...
 
+_markStackFrame: inline func (frame: StackFrame) {
+  stackFramesToDelete push(frame)
+}
+
+_cleanupStackFrames: inline func {
+  while (stackFramesToDelete empty?() == false) {
+    frame := stackFramesToDelete pop() as StackFrame
+    if (frame)
+      gc_free(frame as Void*)
```
davidhesselbom: I wonder if it's possible to define a `free` function on the `StackFrame` cover so you could call `frame free()` instead...

source/sdk/lang/Exception.ooc
```diff
@@ -142,6 +155,15 @@ Exception: class {
         init(null, message)
     }
 
+    free: override func {
+      while (this backtraces empty?() == false) {
+        bt := this backtraces pop() as Backtrace
```
davidhesselbom: `backtrace`, please?

source/sdk/lang/Exception.ooc
sebastianbaginski: Good to know.
I was just following the conventions :)

source/sdk/lang/Exception.ooc
sebastianbaginski: Let's clear this up in another PR / debt issue, along with all other global stuff in Exception / Backtrace. What do you think ?

davidhesselbom: Fine.

source/sdk/lang/Exception.ooc
sebastianbaginski: Fixed.

sebastianbaginski:
Updated. Now it releases the stack frame in `_popStackFrame`, which is automatically called by the compiler.
`Exception` class uses another function to take frame from the frame stack, so now everything is released even when no exception is thrown.

source/sdk/lang/Exception.ooc
sebastianbaginski: This actually works. It will be included in the cleanup PR. Thanks !

davidhesselbom:
The exception still needs freeing manually in the `catch` clause, though, right?

sebastianbaginski:
Yes. Do you want me to make it automatic ?
Maybe in the upcoming *cleanup* PR as the code there is easier to work with.

davidhesselbom:
Sure, it can wait, I was just asking.

sebastianbaginski:
Sure, I think it is a good idea.

davidhesselbom:
... but what happens if the `catch` throws the exception again?

sebastianbaginski:
`raise` creates new `Exception` instance each time. Is there another way to throw exceptions, other than `raise` function ?

davidhesselbom:
Well, you can create an exception manually and throw it, but the result should be the same.

davidhesselbom:
Okay, just tested it. Leak is gone, finally, but I still get invalid reads.

sebastianbaginski:
OK, thanks, I will try to address this issue in the next PR.

marcusnaslund:
Woah, on Windows, `ExceptionTest` generates hundreds of lines of `[lang/Backtrace] No backtrace extension nor execinfo - use a debugger!` when running it.

davidhesselbom:
Yeah, the library for backtraces is most likely missing on Windows.

marcusnaslund:
It's fine - AppVeyor is still happy, it's just not a pretty output.

davidhesselbom:
Well, either we put that printout inside a some kind of version block in order to ignore it, or we get the backtrace library working on Windows. The latter option will probably not be very easy, but I'm sure it's doable.

PR #629, opened 2015-10-28
Implemented CpuCanvas

emilwestergren:
Made Canvas an abstract class and replaced Painter/PaintEngine with a CpuCanvas.


emilwestergren:
@sebastianbaginski have a look again if you want

sebastianbaginski:
I don't think I need to look into the details too much, if it works then it's fine by me.
But, few days ago Simon said something like "we should look into the Painter thing again". I'm not sure if he was referring to this PR or something else ?

davidhesselbom:
The diff view really leaves a lot to ask for in these cases... Looks like all you did was move some stuff around, and boom, page after page of deletions and additions. Anything I should know about that involved more than moving code from one file to another?

emilwestergren:
Not really. These are the major changes:
* Created abstract Canvas class which is a property for Image
* Implemented RasterCanvas for RasterImages
* Removed Painter/PainterEngine
* Moved lots of code from GpuCanvas to abstract Canvas class
* Moved the StbImage c-files to a folder

source/draw/Canvas.ooc
```diff
@@ -0,0 +1,68 @@
...
+		this viewport = IntBox2D new(size)
+		this focalLength = 0.0f
+		this blend = false
+		this opacity = 1.0f
+	}
+	drawPoint: virtual func (position: FloatPoint2D) {
+		list := VectorList<FloatPoint2D> new()
+		list add(position)
+		this drawPoints(list)
+		list free()
```
davidhesselbom: This seems like a lot of work for drawing a single point, but I guess it can be optimized later.

source/draw/Canvas.ooc
```diff
@@ -0,0 +1,68 @@
...
+		positions add(box leftTop)
+		positions add(box rightTop)
+		positions add(box rightBottom)
+		positions add(box leftBottom)
+		positions add(box leftTop)
+		this drawLines(positions)
+		positions free()
+	}
+	fill: abstract func
+	draw: abstract func ~ImageSourceDestination (image: Image, source: IntBox2D, destination: IntBox2D)
```
davidhesselbom: `source, destination: IntBox2D`

source/draw/RasterCanvas.ooc
```diff
@@ -0,0 +1,74 @@
...
+use ooc-math
+use ooc-collections
+import math
+import Image, Canvas, Pen, RasterImage
+
+RasterCanvas: abstract class extends Canvas {
+	_target: RasterImage
+	init: func (=_target) { super(this _target size) }
+	fill: override func { raise("RasterCanvas fill unimplemented!") }
+	draw: override func ~ImageSourceDestination (image: Image, source: IntBox2D, destination: IntBox2D) { Debug raise("RasterCanvas draw~ImageSourceDestination unimplemented!") }
```
davidhesselbom: `source, destination: IntBox2D`

source/draw/gpu/GpuSurface.ooc
```diff
@@ -97,18 +82,11 @@ GpuSurface: abstract class {
 		else
 			Debug raise("Trying to draw unsupported image format to OpenGLCanvas!")
 	}
-	draw: func ~ImageSourceDestination (image: Image, source: IntBox2D, destination: IntBox2D) { this draw(image, source, destination, this _getDefaultMap(image)) }
-	draw: func ~ImageDestination (image: Image, destination: IntBox2D) { this draw(image, IntBox2D new(image size), destination) }
-	draw: func ~Image (image: Image) { this draw(image, IntBox2D new(image size)) }
-	draw: func ~ImageTargetSize (image: Image, targetSize: IntSize2D) { this draw(image, IntBox2D new(targetSize)) }
+	draw: override func ~ImageSourceDestination (image: Image, source: IntBox2D, destination: IntBox2D) { this draw(image, source, destination, this _getDefaultMap(image)) }
```
davidhesselbom: `source, destination: IntBox2D`

source/draw/Canvas.ooc
emilwestergren: The thing is, on a GPU the only implementation needed is `drawPoints` since they are always batched. Therefore `drawPoint` is a special case of `drawPoints`. In the CPU implementation, the opposite is true. Therefore it is not obvious which of `drawPoint` and `drawPoints` can be generalized in the abstract class. I just picked one as default and then it's up to the implementation to override if it can be optimized.

emilwestergren:
Updated.

source/draw/Canvas.ooc
davidhesselbom: Fair enough.

PR #633, opened 2015-10-29
Modified tests to make them Windows compliant

marcusnaslund:
Minor changes to tests to avoid roundoff errors and different thread scheduling that occurs in Windows.

### DIFF
--- a/test/math/FloatMatrixTest.ooc
+++ b/test/math/FloatMatrixTest.ooc
@@ -7,7 +7,7 @@ FloatMatrixTest: class extends Fixture {
 	matrix := FloatMatrix new (3, 3)
 	nonSquareMatrix := FloatMatrix new (IntSize2D new(2, 3))
 	nullMatrix := FloatMatrix new(0, 0)
-	tolerance := 1.0e-5
+	tolerance := 0.00001f
 
 	init: func {
 		super ("FloatMatrix")
--- a/test/math/FloatMatrixTest.ooc
+++ b/test/math/FloatMatrixTest.ooc
@@ -7,7 +7,7 @@ FloatMatrixTest: class extends Fixture {
 	matrix := FloatMatrix new (3, 3)
 	nonSquareMatrix := FloatMatrix new (IntSize2D new(2, 3))
 	nullMatrix := FloatMatrix new(0, 0)
-	tolerance := 0.00001f
+	precision := 1.0e-5f
 
 	init: func {
 		super ("FloatMatrix")
@@ -144,7 +144,7 @@ FloatMatrixTest: class extends Fixture {
 		// 2 5
 		for (x in 0 .. matrix dimensions width)
 			for (y in 0 .. matrix dimensions height)
-				expect(matrix[x, y], is equal to(values[x * matrix dimensions height + y]) within(this tolerance))
+				expect(matrix[x, y], is equal to(values[x * matrix dimensions height + y]) within(this precision))
 	}
 }
 FloatMatrixTest new() run()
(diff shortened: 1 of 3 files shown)


marcusnaslund:
`expected  equal to  '10.00000000', was '10.00000000'  [tolerance: 0.00001000]`...

test/math/FloatMatrixTest.ooc
```diff
@@ -74,7 +75,13 @@ FloatMatrixTest: class extends Fixture {
 
 		this add("multiplication (scalar)", func {
 			matrix = createMatrix(3, 3, [1.0f, 0, 0, 0, 2.0f, 0, 0, 0, 3.0f])
-			checkAllElements(2.0f * matrix, [2.0f, 0, 0, 0, 4.0f, 0, 0, 0, 6.0f])
+			product := 2.0f * matrix
+			expect(product[0, 0] > 2.0f - this tolerance, is true) //FIXME checkAllElements fails here for some reason
+			expect(product[0, 0] < 2.0f + this tolerance, is true)
```
davidhesselbom: What about `expect(product[0, 0] abs()` something, at the very least?

test/math/FloatMatrixTest.ooc
```diff
@@ -145,7 +152,7 @@ FloatMatrixTest: class extends Fixture {
 		// 2 5
 		for (x in 0 .. matrix dimensions width) {
 			for (y in 0 .. matrix dimensions height) {
-				expect(matrix[x, y], is equal to(values[x * matrix dimensions height + y]))
+				expect(matrix[x, y], is equal to(values[x * matrix dimensions height + y]) within(this tolerance))
```
davidhesselbom: Try storing the result of `matrix[x, y]` in a variable and using that as the first argument to `expect` and see if that works better:
```ooc
(element, value) := matrix[x, y]
expect(element, is equal to(value) within(this precision)
```

marcusnaslund: Gah, now the original thing works on both platforms. I think the original problem was `this tolerance` was accidentally a `Double` which is problematic sometimes, but not always. Hang on.

marcusnaslund:
Updated. As an extra bonus, I got rid of four curly brackets.

test/math/FloatMatrixTest.ooc
```diff
@@ -7,6 +7,7 @@ FloatMatrixTest: class extends Fixture {
 	matrix := FloatMatrix new (3, 3)
 	nonSquareMatrix := FloatMatrix new (IntSize2D new(2, 3))
 	nullMatrix := FloatMatrix new(0, 0)
+	tolerance := 0.00001f
```
davidhesselbom: - In just about every other test, this variable is called `precision`.
- Use scientific notation: `1.0e-5f´ (makes it easier to count the zeroes).

marcusnaslund:
Done.

davidhesselbom:
This should be the last thing we need to get things to work, right? I'll let the tests run before I merge, then.

marcusnaslund:
I hope so. It might a be a good idea to cancel all AppVeyor builds before this one, otherwise we will wait a long time.

davidhesselbom:
Let me guess - I'm the only one who can do that, right?

marcusnaslund:
Yep, I see no cancel button like I do for my own branch.

davidhesselbom:
Done. I must have cancelled at least 20 builds...

davidhesselbom:
Unbelievable. AppVeyor succeeds, but Circle fails. Is is Monday already?

marcusnaslund:
Hmmmm.

davidhesselbom:
`ThreadPool` now consistently fails on Circle CI for some reason. It has nothing to do with this PR, so what's causing it?

marcusnaslund:
It was green a second ago, now it says failed again. I also have trouble seeing updates from certain runs, and some seem to fail between tests. Is CircleCI having problems?

davidhesselbom:
Isn't it great when a system you rely on to find your bugs has bugs? ;)

PR #636, opened 2015-10-29
Removed cancel() from win32 threads

marcusnaslund:
Another step forward in making threaded things run smoothly on Windows and AppVeyor. The use of cancel is behind most of the problems.

Microsoft says `TerminateThread is a dangerous function that should only be used in the most extreme cases.`

Random people say:
```TerminateThread is able to kill threads that are currently holding critical sections deep within the Windows API, such as in the sockets layer. If a thread is killed while holding a critical section, that section stays locked forever, and your program can grind mysteriously to a halt. ```

```The warnings above are a gross understatement. TerminateThread is NOT a solution for shutting down worker threads. You will break your app, and it will fail randomly.```

### DIFF
--- a/source/sdk/threading/native/ThreadWin32.ooc
+++ b/source/sdk/threading/native/ThreadWin32.ooc
@@ -47,7 +47,9 @@ version(windows) {
         }
 
         cancel: func -> Bool {
-            this alive?() && TerminateThread(this handle, 0)
+            false
+            //this alive?() && TerminateThread(this handle, 0)
+            //TODO Find a better way to terminate Win32 threads, if any 
         }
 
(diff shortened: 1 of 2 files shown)


davidhesselbom:
@sebastianbaginski, do you have any objections?

sebastianbaginski:
No, it's fine. I think we should implement the `cancel` logic on the upper layer anyway via `cancelation flags`. Canceling a thread should mean to set the cancel flag to true and let the thread to exit its main function. Then `join` the thread. It will work on every platform.

davidhesselbom:
Great, I'll go ahead and merge this, then.

PR #637, opened 2015-10-29
Fixed a memory leak in CsvReader open(), added String overload.

thomasfanell:
* Fixed a memory leak in ```open()``` where a String was not properly freed.
* Added a String overload to ```open()```.

### DIFF
--- a/source/io/CsvReader.ooc
+++ b/source/io/CsvReader.ooc
@@ -72,9 +72,15 @@ CsvReader: class extends Iterator<VectorList<Text>> {
 		result
 	}
 	delimiter ::= static ','
-	open: static func (filename: Text) -> This {
+	open: static func ~text (filename: Text) -> This {
+		filenameString := filename toString()
+		result := This open(filenameString)
+		filenameString free()
+		result
+	}
+	open: static func ~string (filename: String) -> This {
 		result: This = null
-		file := File new(filename toString())
+		file := File new(filename)
 		if (file exists?())
 			result = This new(FileReader new(file))
 		file free()


davidhesselbom:
Why was this marked as `high`?

source/io/CsvReader.ooc
```diff
@@ -72,12 +72,16 @@ CsvReader: class extends Iterator<VectorList<Text>> {
...
+		This open(filename toString())
+	}
+	open: static func ~string (filename: String) -> This {
 		result: This = null
-		file := File new(filename toString())
+		file := File new(filename)
 		if (file exists?())
 			result = This new(FileReader new(file))
 		file free()
+		filename free()
```
davidhesselbom: Wouldn't it make more sense to let whoever calls this function free the string? We don't know whether the person who called the function wants to keep the string or not, do we?

thomasfanell:
It's a memory leak and @davidpiuva needs the overload.

source/io/CsvReader.ooc
davidpiuva: The text overload would then be responsible for freeing as well.

davidhesselbom: Yes, that's true, so it's would then be
```ooc
open: static func ~text (filename: Text) -> This {
    string := filename toString()
    result := This open(filename toString())
    string free()
    result
}
```

thomasfanell: I considered that and that would be the normal thing to do, yes. But what about cases like this:
```CsvReader open("path") // String literal ```?

davidhesselbom: Or, alternatively, the main code is in the `Text` overload, and all the `String` one does is call the `Text` version.

davidhesselbom: > ```CsvReader open("path") // String literal ```?

That's the caller's problem.

thomasfanell: Sure, that would work. I'll try that out.

thomasfanell: Would that mean that we're not freeing ```filename: Text``` either?

davidpiuva: You can free the text with a condition so that the sender can take or give ownership.

thomasfanell: I've changed it back to where the bulk code is in the Text overload, where I'm creating a String from the specified Text and then I free the String. Is this the way we want it?

davidhesselbom: No, I scratched that solution. Put the bulk in the `String` overload as you did, but let the `Text` overload create and free the `String` it passes to the `String` overload. The `String` overload does not free the `String`.

thomasfanell: Okidoki, amended.

davidpiuva:
I think there is a problem with freeing the result from CsvReader because of the nested Iterator<VectorList<Text>>. Freeing the tokens before freeing the rows removed a memory leak but also caused invalid free which is worse than leaking.

davidhesselbom:
> I think there is a problem with freeing the result from CsvReader

Not related to this Pull Request, is it?

davidpiuva:
It can be a follow up PR if someone wants to do it.

PR #638, opened 2015-10-29
Removed failing BQ test from Windows

marcusnaslund:
Yes, this is a cheap "solution", but I've identified the problem as the same one as on linux (invalid read/writes for some reason) - it just manifests itself a lot more on Windows.

The correct approach is to rebuild `BlockedQueue` and `SynchronizedQueue`, which will happen in the near future (I will create an issue about it). Until then, this will let us use AppVeyor.


davidhesselbom:
Fine.

PR #640, opened 2015-10-29
Removed the pthreads install from AppVeyor

marcusnaslund:
We haven't had to use `pthreads` to compile rock for some time, and it's not the threading backend for Windows, so this only consumes time.
Right @jonathanudd ?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,8 +7,6 @@ install:
     bash install_rock.sh
 
     mv rock.exe C:\msys64\usr\bin\
-
-    mingw-get install pthreads
 build: off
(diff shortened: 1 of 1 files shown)


marcusnaslund:
AppVeyor failed because it didn't include stuff from other AppVeyor-related PRs, but let's hold off on this for a while to be sure.

jonathanudd:
I think `pthreads` is not needed any more when running on Windows. And the test didn't fail because it was missing `pthreads` right? So it should be ok.

marcusnaslund:
Yes, this works.

PR #642, opened 2015-10-29
Fixed memory leak in CsvWriter

thomasfanell:
I forgot about the writer in the previous PR, so in the interest of symmetry, here is a fix for ```CsvWriter``` as well. @simonmika raised a good point regarding allowing ```String``` overloads in that the very purpose of ```Text``` was to force away the use of ```String``` (which is why I didn't make a ```String``` overload to begin with).

Anyway, I guess we could consider this a temporary fix until we rid ourselves of ```String```.

### DIFF
--- a/source/io/CsvWriter.ooc
+++ b/source/io/CsvWriter.ooc
@@ -36,10 +36,8 @@ CsvWriter: class {
 		result
 	}
 	open: static func ~string (filename: String) -> This {
-		result: This = null
 		file := File new(filename)
-		if (file exists?())
-			result = This new(FileWriter new(file))
+		result := This new(FileWriter new(file))
 		file free()
 		result
 	}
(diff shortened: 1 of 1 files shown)


davidhesselbom:
> until we rid ourselves of String

I don't expect that to happen very soon, so this may not be very temporary.

Note: `Fixed memory` isn't a very descriptive commit message. The PR title would have been a lot better.

davidhesselbom:
@simonmika, what should we do here, merge this, or close it and wait for another one that fixes the leak, but doesn't add an overload?

thomasfanell:
Yeah I edited the commit message and forgot to add 'leak'

davidhesselbom:
... and 'in CsvWriter'. ;)

thomasfanell:
Noted, i'll be more careful next time

simonmika:
Merge the PR at the moment. We need `Text` to mature just a little bit more.

thomasfanell:
Hang on, I spotted an error (which is probably causing the segfault)

PR #644, opened 2015-10-30
Refactored Stack class from sdk

sebastianbaginski:
`Stack` class from sdk refactored to our coding standards.
The reason why it's marked with `memory` is that now using exceptions in other project does not generate `invalid memory reads` in valgrind. They were related to the stack internal backend and `memmove` calls.

### DIFF
--- a/source/sdk/structs/Stack.ooc
+++ b/source/sdk/structs/Stack.ooc
@@ -1,73 +1,48 @@
-import structs/ArrayList
-
-Stack: class <T> extends BackIterable<T> {
-    data: ArrayList<T>
-    
-    size: SSizeT {
-        get {
-            data != null ? data size : 0		
-        }
-    }
-    
-    init: func {
-        data = ArrayList<T> new()
-    }
-
-    free: override func {
-      this data free()
-      super()
-    }
-
-    push: func(element: T) {
-        data add(element)
-    }
-    
-    pop: func -> T {
-        if (empty?())
-            Exception new(This, "Trying to pop an empty stack.") throw()
-            
-        return data removeAt(lastIndex())
-    }
-    
-    peek: func -> T {
-        if (empty?())
-            Exception new(This, "Trying to peek an empty stack.") throw()
-                
-        return data last()
-    }
-
-    peek: func ~index (index: SSizeT) -> T {
-        mysize := data size
-        if (index < 1)
-            Exception new(This, "Trying to peek(%d)! index must be >= 1 <= size" format(index)) throw()
-
-        if (index > mysize)
-            Exception new(This, "Trying to peek(%d) a stack of size %d" format(index, mysize)) throw()
-
-        return data get(mysize - index)
-    }
-
-    indexOf: func(element: T) -> SSizeT {
-        return data indexOf(element)
-    }
-	
-    getSize: func -> SSizeT {	
-        return data size
-    }
-    
-    empty?: func -> Bool {
-        return data empty?()
-    }
-    
-    lastIndex: func -> Int {
-        return size - 1
-    }
-
-    clear: func {
-        data clear()
-    }
-
-    iterator: func -> BackIterator<T> { data iterator() }
-
-    backIterator: func -> BackIterator<T> { data backIterator() }
+Stack: class <T> {
+	_data: T*
+	_size: SizeT = 0
+	_capacity: SizeT = 8
+	size ::= this _size
+	isEmpty ::= this size == 0
+	init: func {
+		this _data = gc_malloc(this _capacity * T size)
+	}
+	free: override func {
+		gc_free(this _data)
+		super()
+	}
+	push: func (element: T) {
+		if (this size >= this _capacity)
+			this _grow()
+		this _data[this size] = element
+		++this _size
+	}
+	pop: func -> T {
+		version(safe)
+			if (this isEmpty)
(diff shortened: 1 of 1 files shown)


marcusnaslund:
AppVeyor failed because
`08:41:17 Vector2D ...f.. failed in 0.00s, total: 6.05s
  ->  resize decrease size : expected  equal to  '10', was '10'`
I don't...what...How can this happen some of the time, but not always?

marcusnaslund:
We don't want iterators because of memory problems right?

sebastianbaginski:
Yes, I was told that they leak memory.
As they are not used now , I just removed them from the `Stack` api. They would have to be rewritten from scratch for this class anyway. I don't think it's worth the effort now, since we don't use them.

source/sdk/structs/Stack.ooc
```diff
@@ -1,73 +1,54 @@
...
-    backIterator: func -> BackIterator<T> { data backIterator() }
+Stack: class <T> {
+	_data: T*
+	_size: SizeT = 0
+	_capacity: SizeT = 8
+	size: SizeT {
+		get {
+			this _size
+		}
+	}
```
davidhesselbom: Since this PR is all about cleaning up:
```ooc
size ::= this _size
```

source/sdk/structs/Stack.ooc
```diff
@@ -1,73 +1,54 @@
...
+		version(safe)
+			if (index < 0)
+				Exception new(This, "Trying to peek(%d)! index must be >= 1 <= size" format(index)) throw()
+			else if (index >= this size)
+				Exception new(This, "Trying to peek(%d) a stack of size %d" format(index, this size)) throw()
+		this _data[index]
+	}
+	isEmpty: func -> Bool {
+		this size == 0
+	}
```
davidhesselbom: This, too, should be `isEmpty ::= this size == 0`, but since it changes the function from a function to a property, it may have side effects.

davidhesselbom: It *should* be a property, though.

source/sdk/structs/Stack.ooc
marcusnaslund: Beware, I think `expect()` doesn't handle `SizeT` too well.

davidhesselbom: Unrelated, but thanks for the warning.

sebastianbaginski: So what do we use here ? `UInt` ?

source/sdk/structs/Stack.ooc
```diff
@@ -1,73 +1,54 @@
-import structs/ArrayList
-
-Stack: class <T> extends BackIterable<T> {
-    data: ArrayList<T>
```
davidhesselbom: Excellent, thanks.

source/sdk/structs/Stack.ooc
sebastianbaginski: OK.

source/sdk/structs/Stack.ooc
davidhesselbom: Isn't `SizeT` platform dependent?

source/sdk/structs/Stack.ooc
```diff
@@ -1,73 +1,54 @@
...
+		if (this size >= this _capacity)
+			this _grow()
+		this _data[this size] = element
+		++this _size
+	}
+	pop: func -> T {
+		version(safe)
+			if (this isEmpty())
+				Exception new(This, "Trying to pop an empty stack.") throw()
+		this _data[--this _size]
```
davidhesselbom: Ohh, fancy!

source/sdk/structs/Stack.ooc
sebastianbaginski: > Alias of one of the fundamental unsigned integer types.
>size_t is the type returned by the sizeof operator

Also it is heavily use in C++ STL. So I guess it is platform dependent the same way as `sizeof` is

source/sdk/structs/Stack.ooc
sebastianbaginski: hehe:D

source/sdk/structs/Stack.ooc
sebastianbaginski: Sure:) I just thought that the logic in this class is easy enough to implement it directly on a plain C array.

davidhesselbom: Let's see whether it actually fixes the memory problem... Gimme a minute.

source/sdk/structs/Stack.ooc
```diff
@@ -1,73 +1,48 @@
...
+	}
+	peek: func (index := 0) -> T {
+		version(safe)
+			if (index < 0)
+				Exception new(This, "Trying to peek(%d)! index must be >= 1 <= size" format(index)) throw()
+			else if (index >= this size)
+				Exception new(This, "Trying to peek(%d) a stack of size %d" format(index, this size)) throw()
+		this _data[index]
```
sebastianbaginski: I made a mistake, this should be `this _data[this size - index - 1]`.

davidhesselbom: Right... maybe you should add add `last` property to compute exactly that value.

davidhesselbom: Or `top` or whatever the term is for a stack.

PR #648, opened 2015-10-30
Removed Abstractions from the SDK

marcusnaslund:
We're not going to miss this.


davidhesselbom:
I actually kinda like it, now that I look at it, but no, I guess we won't.

PR #652, opened 2015-10-30
Removed siphash.use from the SDK

marcusnaslund:
The c file it refers to doesn't even exist.


davidhesselbom:
It's probably used when building rock (which uses its own SDK to build itself (yes, I know)).

PR #653, opened 2015-10-30
Removed Opts from the SDK

marcusnaslund:
We already have `CLIParser` in `base`.


davidhesselbom:
... and if this thing was useful, we wouldn't have made our own, I guess.

PR #655, opened 2015-10-30
Removed Regexp from SDK

marcusnaslund:
Removed this. We don't even have `pcre.h`, and the whole implementation relies on memory-leaking iterators. If we ever want this we would want to rewrite the whole thing anyway.


davidhesselbom:
Nah, we could get `pcre.h` by installing `libpcre3-dev`, I think, and replacing the iterators with loops wouldn't be as hard as rewriting what you're removing now. A stronger case against keeping it is the fact that, well, we're not using it anyway.

davidhesselbom:
Probably also used to build rock.

PR #661, opened 2015-11-02
Removed flush and finish from Image interface

sebastianbaginski:


davidhesselbom:
Does anyone know why these functions were added in the first place? Is this just something that was left lying around after draw-gpu was refactored?

emilwestergren:
They were part of the Image class in the C# code as well.

sebastianbaginski:
Works on Android, according to @emilwestergren . Thanks for testing

davidhesselbom:
Oh, okay. Well then, I guess there's nothing to stop me from merging this.

PR #662, opened 2015-11-02
Changed FloatEuclidTransform translation from Point to Size.

davidhesselbom:
Will require corresponding updates in other projects.

### DIFF
--- a/source/math/FloatEuclidTransform.ooc
+++ b/source/math/FloatEuclidTransform.ooc
@@ -15,26 +15,26 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 import math
-import FloatPoint3D
+import FloatSize3D
 import FloatRotation3D
 import FloatTransform2D
 import FloatTransform3D
 
 FloatEuclidTransform: cover {
 	rotation: FloatRotation3D
-	translation: FloatPoint3D
+	translation: FloatSize3D
 	scaling: Float
 
 	inverse ::= This new(-this translation, this rotation inverse, 1.0f / this scaling)
 	transform ::= FloatTransform3D createScaling(this scaling, this scaling, 1.0f) * FloatTransform3D createTranslation(this translation) * this rotation transform
 
-	init: func@ ~default { this init(FloatPoint3D new(), FloatRotation3D identity, 1.0f) }
-	init: func@ ~translationAndRotation (translation: FloatPoint3D, rotation: FloatRotation3D) { this init(translation, rotation, 1.0f) }
+	init: func@ ~default { this init(FloatSize3D new(), FloatRotation3D identity, 1.0f) }
+	init: func@ ~translationAndRotation (translation: FloatSize3D, rotation: FloatRotation3D) { this init(translation, rotation, 1.0f) }
 	init: func@ ~full (=translation, =rotation, =scaling)
 	init: func@ ~fromTransform (transform: FloatTransform2D) {
 		rotationZ := atan(- transform d / transform a)
 		scaling := ((transform a * transform a + transform b * transform b) sqrt() + (transform d * transform d + transform e * transform e) sqrt()) / 2.0f
-		this init(FloatPoint3D new(transform g, transform h, 0.0f), FloatRotation3D createRotationZ(rotationZ), scaling)
+		this init(FloatSize3D new(transform g, transform h, 0.0f), FloatRotation3D createRotationZ(rotationZ), scaling)
 	}
 	operator + (other: This) -> This { This new(this translation + other translation, this rotation * other rotation, this scaling * other scaling) }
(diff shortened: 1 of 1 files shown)


niklasborwell:
This should probably be discussed. In my view translation is a vector and should be represented by a point, and some properties in FloatSize3D such as volume doesn't make sense when we are talking about a translation.

davidhesselbom:
Sure, I'm open to discussion, hence "not ready". It still bothers me that we can't have a base cover `Vector2D` that can be interpreted as a `Size` or a `Point` depending on context, but here we are.

sebastianbaginski:
If I may add something. From *coding* point of view it is better to keep translation as `point`, because then you can define all the math operators just for the `point` class and use it like
`point1 = point2 + translation`.
If you keep translation as `size` instead, you will have to define operators for adding / subtracting `size` from `point` (and other way around). I know that all of this is probably done, but still ...
Other than that, I simply think that negative translation expresses better with `point` than `size` (I'm used to keep `size` as unsigned).

bettinaselig:
By definition the coordinates of a points are the point's *translation* from the origin. So in my eyes a point is always a vector, hence a translation. When I realized that `FloatSize2D` is allowed to be negative I was very surprised. (This was when I wanted to add some lines to make sure that `FloatSize2D` cannot be negative.) 
Are there any arguments that translation should be a `FloatSize2D` other than "we used to do it like this"?

davidhesselbom:
I don't care either way, I just want to the codebase to be consistent. @simonmika, the ball's in your court. Perhaps we can reach a final decision here so it doesn't keep coming back for discussion?

simonmika:
A size is not affected by offset changes when transformed but a point is. So distances are expressed using size.

davidhesselbom:
@niklasborwell @bettinaselig @sebastianbaginski Comments?

sebastianbaginski:
I understand Simon's argument, but I still think translation is expressed better as `point` (or `vector`) in programming.

simonmika:
Both `point` and `size` are vectors, they both have direction, the difference is how they are affected by transforms. Read up on homogenous coordinate systems and get used to it.

bettinaselig:
So maybe we need another vector type... for distances.

sebastianbaginski:
Haha I know that I'm not using it every day but I do have a degree in mathematics :) I'm talking about programming and not mathematical theories. All the SDKs I worked with expressed translation as `point`, it is just more convenient and natural thing to do.

simonmika:
It is an implementation of mathematics.

bettinaselig:
To be a bit constructive here: 
- Several people think that expressing a *translation as* a `size` is not intuitive.
- The advantage of expressing it as a `size` instead of a `point` is that it is not affected by translations.

@niklasborwell Do we actually apply translations to *translations*? Or how are things handled internally? 

The question is: Do we need a new data type or is it ok to rename `size` to something else?

simonmika:
You would need to split up the use of `size` into two types: 
* `translation` for directional distances
* `distance` for non-directional distances
In the end we would need to convert between them in a lot of places.

simonmika:
Although I'm not that happy with those names.

davidhesselbom:
Isn't a non-directional `distance` just a scalar, i.e. `Float` or `Int`?

niklasborwell:
I can't see a situation where someone would transform a translation, since it doesn't make sense anyway.
You could say the same about a size. What if someone would use the `volume` of a translation? 
It doesn't make sense either, so we wouldn't get a correct representation anyway.
It seems to me we would just spend time replacing one flawed representation for another, that also feels less intuitive.

simonmika:
@davidhesselbom not exactly, thats why the names aren't perfect. It would still contain the angle information, just not the direction along that angle.

simonmika:
I would like to set a stop to this discussion. Let's stick with the way we have been doing things for the last 5+ years. It works even if it takes some getting used to.

davidhesselbom:
> I would like to set a stop to this discussion.

Fine, just be aware that if we stop now, we will be doing things your way because you said so, not because anyone actually understands why or agrees it's how we should do things. The discussion will keep coming back again and again.

simonmika:
The discussion will always pop up again for like every new person who starts working with it. I just don't think it belongs to this PR and I also doubt we will take the time and effort to change concepts soon. However to apply the current concept as stringent as possible will make it somewhat easier to grasp. So, in the end, let's not take this PR hostage, anybody who thinks we should change the concept, let us continue the discussion IRL.

bettinaselig:
The thing is that now everywhere `FloatPoint3D` is used to save translations and we do not have any problems with it. So why changing it "back" to `FloatSize3D`? 
Most of us think it is better to save a translation as a `point`, even though it is not mathematical correct.

davidhesselbom:
Like I said earlier, I don't care which way we go, as long as we keep it consistent, so please do continue this discussion IRL. I'm locking the conversation now.

PR #663, opened 2015-11-02
Plot shape

davidpiuva:
Allowing drawing of free shapes in the SVG plotter.

### DIFF
--- a/source/plot/Axis.ooc
+++ b/source/plot/Axis.ooc
@@ -7,7 +7,6 @@ Orientation: enum {
 	Horizontal
 	Vertical
 }
-
 Axis: class {
 	visible: Bool { get set }
 	label: String { get set }
@@ -28,7 +27,6 @@ Axis: class {
 	gridOn: Bool { get set }
 	precision: Int { get set }
 	roundAxisEndpoints: Bool { get set }
-
 	init: func (=orientation, label := "") {
 		this visible = true
 		this label = label
@@ -36,12 +34,9 @@ Axis: class {
 		this precision = 3
 		this roundAxisEndpoints = true
 	}
-
 	length: func -> Float {
-		result := this max - this min
-		result
+		this max - this min
 	}
-
 	getSvg: func (plotAreaSize, margin: FloatSize2D, transform: FloatTransform2D, fontSize: Int) -> String {
 		result := ""
 		if (this visible) {
@@ -68,7 +63,6 @@ Axis: class {
 		topTickMarkerStartOffset := FloatPoint2D new(0.0f, - plotAreaSize height)
 		result = result & Shapes text(position + labelOffset, this label, this fontSize, "middle")
 		result = result & this getRadixSvg(position + radixOffset, radix, "middle")
-
 		tickValue := this getFirstTickValue()
 		position x += transform translation width + transform scalingX * tickValue
 		while (tickValue <= this max) {
@@ -76,8 +70,7 @@ Axis: class {
 			tickValue += this tick
 			position x += transform scalingX * tick
 		}
-		result = result >> "</g>\n"
-		result
+		result >> "</g>\n"
 	}
 
 	getVerticalSvg: func (plotAreaSize, margin: FloatSize2D, position: FloatPoint2D, transform: FloatTransform2D, radix: Float) -> String {
@@ -91,7 +84,6 @@ Axis: class {
 		result = result & Shapes text(position + labelOffset, this label, this fontSize, "middle")
 		result = result >> "</g>\n"
 		result = result & this getRadixSvg(position + radixOffset, radix, "end")
-
 		tickValue := this getFirstTickValue()
 		position y += transform translation height - plotAreaSize height - transform scalingY * tickValue
 		while (tickValue <= this max) {
@@ -99,10 +91,8 @@ Axis: class {
 			tickValue += this tick
 			position y -= transform scalingY * tick
 		}
-		result = result >> "</g>\n"
-		result
+		result >> "</g>\n"
 	}
-
 	getRadixSvg: func (position: FloatPoint2D, radix: Float, textAnchor: String) -> String {
 		result := ""
 		if (radix >= pow(10, this precision - 1) || radix <= pow(10, - this precision)) {
@@ -112,7 +102,6 @@ Axis: class {
 		}
 		result
 	}
-
 	getTickSvg: func (tickValue, radix: Float, position, numberOffset, tickMarkerOnOtherSideOffset, tickMarkerEndOffset: FloatPoint2D, textAnchor: String) -> String {
 		result := "<g desc='" << tickValue toString() >> "'>\n"
 		if (this gridOn)
@@ -125,10 +114,8 @@ Axis: class {
 		result = result & Shapes text(position + numberOffset, Float absolute(tickValue - tickValue round()) < 0.001 ? tempTickInt : tempTick, this fontSize - 4, textAnchor)
 		tempTick free()
 		tempTickInt free()
-		result = result >> "</g>\n"
-		result
+		result >> "</g>\n"
 	}
-
 	getFirstTickValue: func -> Float {
 		result: Float
 		if ((this min < this tick && this min > 0.0f) || (this min > this tick && this min < 1.0f))
@@ -137,19 +124,16 @@ Axis: class {
 			result = this min - Float modulo(this min, this tick)
 		result
 	}
-
 	getRequiredMargin: func (fontSize: Int) -> Float {
 		result: Float
 		if (this fontSize == 0)
 			this fontSize = fontSize
-
 		if (this orientation == Orientation Horizontal)
 			result = 2.5f * this fontSize as Float
 		else
 			result = (this precision + 2) * (this fontSize as Float - 4)
 		result
 	}
-
 	roundEndpoints: func {
 		if (this roundAxisEndpoints) {
 			this min = Float roundToValueDigits(this min, 2, false)
--- a/source/plot/Axis.ooc
+++ b/source/plot/Axis.ooc
@@ -1,6 +1,7 @@
 use ooc-math
 import svg/Shapes
 import math
+use ooc-draw
 
 Orientation: enum {
 	Horizontal
@@ -115,9 +116,9 @@ Axis: class {
 	getTickSvg: func (tickValue, radix: Float, position, numberOffset, tickMarkerOnOtherSideOffset, tickMarkerEndOffset: FloatPoint2D, textAnchor: String) -> String {
 		result := "<g desc='" << tickValue toString() >> "'>\n"
 		if (this gridOn)
-			result = result & Shapes line(position, position + tickMarkerOnOtherSideOffset, 1, 255.0f, "grey", FloatPoint2D new(5, 5))
-		result = result & Shapes line(position, position + tickMarkerEndOffset, 1, 255.0f, "black")
-		result = result & Shapes line(position + tickMarkerOnOtherSideOffset, position + tickMarkerOnOtherSideOffset - tickMarkerEndOffset, 1, 255.0f, "black")
+			result = result & Shapes line(position, position + tickMarkerOnOtherSideOffset, 1, ColorBgra new(128, 128, 128, 255), FloatPoint2D new(5, 5))
+		result = result & Shapes line(position, position + tickMarkerEndOffset, 1, ColorBgra new(0, 0, 0, 255))
+		result = result & Shapes line(position + tickMarkerOnOtherSideOffset, position + tickMarkerOnOtherSideOffset - tickMarkerEndOffset, 1, ColorBgra new(0, 0, 0, 255))
 		tickValue = radix >= pow(10, this precision - 1) || radix <= pow(10, - this precision) ? (tickValue / radix) : tickValue
 		tempTick := tickValue toString()
 		tempTickInt := tickValue as Int toString()
--- a/source/plot/LinePlotData2D.ooc
+++ b/source/plot/LinePlotData2D.ooc
@@ -12,24 +12,36 @@ LineStyle: enum {
 
 LinePlotData2D: class extends PlotData2D {
 	lineStyle: LineStyle { get set }
+	dataSeries: VectorList<FloatPoint2D> { get set }
 
 	init: func ~default (lineStyle := LineStyle Solid) {
-		super()
+		super("", ColorBgra new())
+		this dataSeries = VectorList<FloatPoint2D> new()
 		this lineStyle = lineStyle
 	}
 
-	init: func ~dataSeries (dataSeries: VectorList<FloatPoint2D>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(dataSeries, label, colorBgra)
+	init: func ~dataSeries (=dataSeries, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
+		super(label, colorBgra)
 		this lineStyle = lineStyle
 	}
 
-	init: func ~color (dataSeries: VectorList<FloatPoint2D>, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
-		this init(dataSeries, "", colorBgra, lineStyle)
+	init: func ~color (=dataSeries, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
+		super("", colorBgra)
+		this lineStyle = lineStyle
 	}
 
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(xSeries, ySeries, label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count) {
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
+		}
+		super(label, colorBgra)
 		this lineStyle = lineStyle
+	}
+
+	free: override func {
+		this dataSeries free()
+		super()
 	}
 
 	getSvg: func (transform: FloatTransform2D) -> String {
@@ -64,6 +76,30 @@ LinePlotData2D: class extends PlotData2D {
 				result = Shapes line(start, end, this lineWidth, this opacity, this color)
 		}
 		result = result & Shapes text(FloatPoint2D new(end x, end y + fontSize / 3), this label, fontSize, this opacity, this color)
+		result
+	}
+
+	minValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this dataSeries empty)
+			result = FloatPoint2D new()
+		else {
+			result = this dataSeries[0]
+			for (i in 1 .. this dataSeries count)
+				result = result minimum(this dataSeries[i])
+		}
+		result
+	}
+
+	maxValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this dataSeries empty)
+			result = FloatPoint2D new()
+		else {
+			result = this dataSeries[0]
+			for (i in 1 .. this dataSeries count)
+				result = result maximum(this dataSeries[i])
+		}
 		result
 	}
 }
--- a/source/plot/LinePlotData2D.ooc
+++ b/source/plot/LinePlotData2D.ooc
@@ -9,41 +9,33 @@ LineStyle: enum {
 	Dashed
 	Dotted
 }
-
 LinePlotData2D: class extends PlotData2D {
 	lineStyle: LineStyle { get set }
 	dataSeries: VectorList<FloatPoint2D> { get set }
-
 	init: func ~default (lineStyle := LineStyle Solid) {
 		super("", ColorBgra new())
 		this dataSeries = VectorList<FloatPoint2D> new()
 		this lineStyle = lineStyle
 	}
-
 	init: func ~dataSeries (=dataSeries, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
 		super(label, colorBgra)
 		this lineStyle = lineStyle
 	}
-
 	init: func ~color (=dataSeries, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
 		super("", colorBgra)
 		this lineStyle = lineStyle
 	}
-
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		this dataSeries = VectorList<FloatPoint2D> new()
-		for (i in 0 .. ySeries count) {
-			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
-		}
 		super(label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count)
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
 		this lineStyle = lineStyle
 	}
-
 	free: override func {
 		this dataSeries free()
 		super()
 	}
-
 	getSvg: func (transform: FloatTransform2D) -> String {
 		result := ""
 		if (!this dataSeries empty) {
@@ -62,7 +54,6 @@ LinePlotData2D: class extends PlotData2D {
 		}
 		result
 	}
-
 	getSvgLegend: func (legendCount, fontSize: Int) -> String {
 		result: String
 		start := FloatPoint2D new(this legendOffset, this legendOffset + (fontSize * legendCount - fontSize / 2) as Float)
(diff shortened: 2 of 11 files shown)


source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+		result = result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this colorBgra)
+		result
+	}
+
+	minValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this _shapes empty)
+			result = FloatPoint2D new()
+		else {
+			object := this _shapes[0] // Convert rvalue to lvalue
```
marcusnaslund: This is a bad thing about rock, I agree, but these comments are superfluous.

marcusnaslund:
Magic complains.

source/plot/LinePlotData2D.ooc
```diff
@@ -12,30 +12,42 @@ LineStyle: enum {
 
 LinePlotData2D: class extends PlotData2D {
 	lineStyle: LineStyle { get set }
+	dataSeries: VectorList<FloatPoint2D> { get set }
```
marcusnaslund: I don't like it, but we do have a `FloatPoint2DVectorList`. Any reason for not using it?

source/plot/LinePlotData2D.ooc
```diff
@@ -12,30 +12,42 @@ LineStyle: enum {
...
+		this lineStyle = lineStyle
 	}
 
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(xSeries, ySeries, label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count) {
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
+		}
+		super(label, colorBgra)
```
marcusnaslund: I think it's good practice to call `super` in the constructor first, and last when `freeing`.

source/plot/LinePlotData2D.ooc
```diff
@@ -57,13 +69,37 @@ LinePlotData2D: class extends PlotData2D {
...
+				result = Shapes line(start, end, this lineWidth, this colorBgra, FloatPoint2D new(this lineWidth as Float, this lineWidth as Float))
 			case => // LineStyle Solid
-				result = Shapes line(start, end, this lineWidth, this opacity, this color)
+				result = Shapes line(start, end, this lineWidth, this colorBgra)
+		}
+		result = result & Shapes text(FloatPoint2D new(end x, end y + fontSize / 3), this label, fontSize, this colorBgra)
+		result
+	}
+
+	minValues: override func -> FloatPoint2D {
```
marcusnaslund: Don't change this here, but for a future PR, feel free to add this to `FloatPoint2DVectorList` instead.

source/plot/LinePlotData2D.ooc
davidpiuva: Unless I need the specific functionality, this is easier to replace with another type as I did with free shapes.

marcusnaslund: There is nothing special in that class yet (but stuff like `minValues` etc. further down would fit well there). Let's leave it for now unless @davidhesselbom objects.

davidpiuva: FloatPoint2DVectorList is more like a temporary speed hack to go around memcpy in generics than something to make permanent use of.

marcusnaslund: I'm afraid there's nothing temporary about it.

davidhesselbom: It's not temporary, and it's not for speed - it's because every function like `sum`, `sort`, `max`, `min` etc. needs to be implemented explicitly for each `T`. You can't add those functions to `VectorList`. So if you have a data series of `FloatPoint2D`, chances are you will have use of the functions defined in `FloatPoint2DVectorList`. If you don't, then fine, just don't go implementing your own `sum` etc. functions here (in case you had any intention of doing so).

source/plot/ScatterPlotData2D.ooc
```diff
@@ -57,12 +63,36 @@ ScatterPlotData2D: class extends PlotData2D {
...
+		result: FloatPoint2D
+		if (this dataSeries empty)
+			result = FloatPoint2D new()
+		else {
+			result = this dataSeries[0]
+			for (i in 1 .. this dataSeries count)
+				result = result minimum(this dataSeries[i])
+		}
+		result
+	}
```
davidhesselbom: So, here it is, duplicate code. This is already implemented in `FloatPoint2DVectorList`.

davidhesselbom: Haha, oops, no, it isn't! But it should be, so move this function there, at least the part inside the `else` clause. I'm not sure what should be returned in case of an empty list...

marcusnaslund: I commented about this just three lines up. :)

marcusnaslund:
@davidpiuva Magic still complains here.

davidpiuva:
Something changed so that the test can't even run anymore because of a missing constructor.

source/plot/LinePlotData2D.ooc
```diff
@@ -12,30 +12,42 @@ LineStyle: enum {
...
+		super("", colorBgra)
+		this lineStyle = lineStyle
 	}
 
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(xSeries, ySeries, label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count) {
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
+		}
```
sebastianbaginski: Brackets can be removed

source/plot/ScatterPlotData2D.ooc
```diff
@@ -4,29 +4,35 @@ use ooc-draw
...
+	init: func ~color (=dataSeries, colorBgra: ColorBgra) {
+		super("", colorBgra)
 	}
 
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new()) {
-		super(xSeries, ySeries, label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count) {
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
+		}
```
sebastianbaginski: brackets can be removed

source/plot/LinePlotData2D.ooc
davidpiuva: Then I should cherry-pick d81e0904 from the latter PR and rebase again.

source/plot/ScatterPlotData2D.ooc
```diff
@@ -57,12 +63,36 @@ ScatterPlotData2D: class extends PlotData2D {
...
-				result = result & Shapes circle(FloatPoint2D new(start x + halfLineHeight, start y), size / 2.0f, this opacity, this color)
+				result = result & Shapes circle(FloatPoint2D new(start x + halfLineHeight, start y), size / 2.0f, this colorBgra)
 			case Shape Square =>
-				result = result & Shapes rect(FloatPoint2D new(start x, start y - halfLineHeight), FloatPoint2D new(size, size), this opacity, this color)
+				result = result & Shapes rect(FloatPoint2D new(start x, start y - halfLineHeight), FloatPoint2D new(size, size), this colorBgra)
 			case =>
 		}
-		result = result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this opacity, this color)
+		result = result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this colorBgra)
+		result
```
sebastianbaginski: I think this line is not needed, the `=` operator returns the left operand. I may be wrong @davidhesselbom @marcusnaslund ?

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+
+	free: override func {
+		this _shapes free()
+		super()
+	}
+
+	getSvg: func (transform: FloatTransform2D) -> String {
+		result := ""
+		if (!this _shapes empty) {
+			noneColor := ColorBgra new(0, 0, 0, 0)
```
sebastianbaginski: @davidhesselbom will probably say something about this variable. Maybe `nullColor` or `emptyColor` ?

source/plot/ScatterPlotData2D.ooc
davidhesselbom: Instead of 
```ooc
result = result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this colorBgra)
result
```
I think
```ooc
result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this colorBgra)
```
should work just as well.

davidpiuva: How do the compiler know that it is a return value then?

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+
+	minValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this _shapes empty)
+			result = FloatPoint2D new()
+		else {
+			object := this _shapes[0] // Convert rvalue to lvalue
+			result = object region leftTop
+			for (i in 1 .. this _shapes count) {
+				object := this _shapes[i] // Convert rvalue to lvalue
```
sebastianbaginski: I agree with @marcusnaslund , comments can be removed.

source/plot/ScatterPlotData2D.ooc
marcusnaslund: Whatever is at the end is the return value, as far as I know

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+		}
+		result
+	}
+
+	maxValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this _shapes empty)
+			result = FloatPoint2D new()
+		else {
+			object := this _shapes[0] // Convert rvalue to lvalue
```
sebastianbaginski: here too

source/plot/ScatterPlotData2D.ooc
davidhesselbom: Right.
```ooc
foo: func -> Int {
    5
}
```
will return `5`.

source/plot/SvgPlot.ooc
```diff
@@ -118,17 +114,17 @@ SvgPlot: class {
...
+		this colorList add(ColorBgra new(100, 0, 0, 255))
+		this colorList add(ColorBgra new(0, 0, 0, 255))
+		this colorList add(ColorBgra new(50, 0, 50, 255))
+		this colorList add(ColorBgra new(0, 50, 0, 255))
+		this colorList add(ColorBgra new(0, 20, 40, 255))
+		this colorList add(ColorBgra new(100, 50, 0, 255))
+		this colorList add(ColorBgra new(200, 0, 100, 255))
+		this colorList add(ColorBgra new(0, 255, 0, 255))
+		this colorList add(ColorBgra new(128, 255, 0, 255))
+		this colorList add(ColorBgra new(0, 200, 0, 255))
```
sebastianbaginski: I think it could be nice to have a static properties in Color* classes with predefined colors, so this one could be 
```
    this colorList add(ColorBgra crimson)
    this colorList add(ColorBgra black)
    this colorList add(ColorBgra darkMagneta)
```
etc.
I don't mean to do that here in this PR, of course. @davidhesselbom what do you think ?

davidhesselbom: He just removed something very similar to, and I'm honestly not sure why.

source/plot/svg/Shapes.ooc
```diff
@@ -1,31 +1,45 @@
 use ooc-math
+use ooc-draw
 
 Shapes: abstract class {
-	line: static func (start, end: FloatPoint2D, lineWidth: Float, opacity: Float, color: String) -> String {
-		"<line x1='" << start x toString() >> "' y1='" & start y toString() >> "' x2='" & end x toString() >> "' y2='" & end y toString() >> "' stroke-width='" & lineWidth toString() >> "' stroke='" >> color >> "' stroke-opacity='" & opacity toString() >> "'/>\n"
+	getColor: static func (bgra: ColorBgra) -> String {
+		"rgb(" clone() & bgra bgr red toString() & "," clone() & bgra bgr green toString() & "," clone() & bgra bgr blue toString() & ")" clone()
```
sebastianbaginski: I'm not familiar with String operators, but this one looks interesting. Why do you need to create a String and the immediately call `clone()` on it ?

source/plot/Axis.ooc
```diff
@@ -115,9 +116,9 @@ Axis: class {
 	getTickSvg: func (tickValue, radix: Float, position, numberOffset, tickMarkerOnOtherSideOffset, tickMarkerEndOffset: FloatPoint2D, textAnchor: String) -> String {
 		result := "<g desc='" << tickValue toString() >> "'>\n"
 		if (this gridOn)
-			result = result & Shapes line(position, position + tickMarkerOnOtherSideOffset, 1, 255.0f, "grey", FloatPoint2D new(5, 5))
-		result = result & Shapes line(position, position + tickMarkerEndOffset, 1, 255.0f, "black")
-		result = result & Shapes line(position + tickMarkerOnOtherSideOffset, position + tickMarkerOnOtherSideOffset - tickMarkerEndOffset, 1, 255.0f, "black")
+			result = result & Shapes line(position, position + tickMarkerOnOtherSideOffset, 1, ColorBgra new(128, 128, 128, 255), FloatPoint2D new(5, 5))
```
sebastianbaginski: What's with the indentation here ?

source/plot/svg/Shapes.ooc
marcusnaslund: Looks weird to me.

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+use ooc-math
+use ooc-draw
+import math
+import PlotData2D
+import svg/Shapes
+
+AnyShape: cover {
+	shape: Shape
+	region: FloatBox2D
+	colorBgra: ColorBgra
```
sebastianbaginski: Shouldn't this be called simply `color` ?

source/plot/SvgPlot.ooc
davidpiuva: The names are specific to SVG and had to be abstracted into the SVG folder. I also need the color values to extract alpha for opacity.

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+
+	getSvg: func (transform: FloatTransform2D) -> String {
+		result := ""
+		if (!this _shapes empty) {
+			noneColor := ColorBgra new(0, 0, 0, 0)
+			for (i in 0 .. this _shapes count) {
+				object := this _shapes[i]
+				pointA := transform * (object region leftTop)
+				pointB := transform * (object region rightBottom)
+				area := FloatBox2D new(pointA, pointB)
```
sebastianbaginski: Maybe just `area := FloatBox2D new(transform * (object region leftTop), transform * (object region rightBottom))` ? @marcusnaslund @davidhesselbom ?

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+			}
+		}
+		result
+	}
+
+	getSvgLegend: func (legendCount, fontSize: Int) -> String {
+		result := ""
+		start := FloatPoint2D new(this legendOffset as Float, this legendOffset + (fontSize * legendCount) as Float - (fontSize as Float) / 2.0f)
+		size := (fontSize as Float) * 0.8f
+		halfLineHeight := (fontSize as Float) / 2.0f
```
sebastianbaginski: You could avoid casts if you write it like
```
size := 0.8f * fontSize
halfLineHeight := 0.5f * fontSize
```
But I don't know if you should. @marcusnaslund ?

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,92 @@
...
+						result = result >> ""
+				}
+			}
+		}
+		result
+	}
+
+	getSvgLegend: func (legendCount, fontSize: Int) -> String {
+		result := ""
+		start := FloatPoint2D new(this legendOffset as Float, this legendOffset + (fontSize * legendCount) as Float - (fontSize as Float) / 2.0f)
```
sebastianbaginski: I think first cast is not needed.

source/plot/svg/Shapes.ooc
davidhesselbom: This was written before we had the `>>` and `<<` operators. `&` frees the operators on both sides (unlike `+`, which frees neither), so in order for the literal strings here not to be freed, they had to be cloned. Trust me, it was even worse when we only had `+` - every intermediary `String` created had to be freed on its own line. 

This line can now be written as
```ooc
"rgb(" << bgra bgr red toString() >> "," & bgra bgr green toString() >> "," & bgra bgr blue toString() >> ")"
```
and that's the reason we now have the `Text` class, which we should use instead of changing the code to what I just wrote.

source/plot/LinePlotData2D.ooc
sebastianbaginski: I'm just trying guess what @davidhesselbom would say :)

source/plot/ShapePlotData2D.ooc
marcusnaslund: I think that looks better, then again I don't feel like trusting rock's somewhat weird typing preferences. I leave this one to @davidhesselbom

source/plot/svg/Shapes.ooc
sebastianbaginski: Thanks for explanation.

source/plot/ShapePlotData2D.ooc
sebastianbaginski: I think it is a simple as "operator result type is the same as type of the first operand".

source/plot/svg/Shapes.ooc
davidpiuva: I did not write the code. It leaked and crashed whenever I touched it.

source/plot/ShapePlotData2D.ooc
davidhesselbom: > operator result type is the same as type of the first operand

Correct.

davidhesselbom: Why is `fontSize` an `Int` to begin with?

source/plot/svg/Shapes.ooc
davidhesselbom: I know.

sebastianbaginski:
I think I'm done reviewing this. I'm not sure about the logic as I don't really know what we need from the plotting module. I will do a validation on this one.

source/plot/Axis.ooc
davidpiuva: Looks normal to me when the red text is removed.

sebastianbaginski: Sure, it is ok, sorry.

source/plot/ShapePlotData2D.ooc
davidpiuva: Maybe

source/plot/Axis.ooc
marcusnaslund: Yes, it replaces line 118, so this is ok.

source/plot/ShapePlotData2D.ooc
marcusnaslund: If that works, yes.

source/plot/ShapePlotData2D.ooc
davidpiuva: It would be a risk of having to look into the default constructor every time and check if alpha is 0 or 255 when trying to understand the code.

source/plot/ShapePlotData2D.ooc
davidpiuva: I get pulled in both directions by people who want more or less lines.

sebastianbaginski:
Validated (with magic etc.), but I can't really test for memory leaks, as the plotter test does not free the data (not that I blame @davidpiuva , it probably was like that before).
@davidhesselbom @marcusnaslund

source/plot/ShapePlotData2D.ooc
davidhesselbom: I think Sebastian was commenting on the name, not the variable's location.

source/plot/LinePlotData2D.ooc
davidhesselbom: Nice work!

source/plot/ShapePlotData2D.ooc
marcusnaslund: Who said to create `pointA` and `pointB`?

davidpiuva: Bettina complained about a similar syntax in Autocut but not this one.

davidhesselbom:
```ooc
area := FloatBox2D new(transform * object region leftTop, transform * object region rightBottom)
```
should suffice, you don't need the extra parentheses.

davidpiuva: ok

davidhesselbom: But, does `object region rightBottom` actually describe the size of `region`?

davidhesselbom: Nevermind, both of them are points. It should work, then.

davidpiuva: It is not a size, it is the (FloatPoint2D, FloatPoint2D) constructor.

source/plot/LinePlotData2D.ooc
```diff
@@ -50,20 +54,40 @@ LinePlotData2D: class extends PlotData2D {
...
 			case LineStyle Dotted =>
-				result = Shapes line(start, end, this lineWidth, this opacity, this color, FloatPoint2D new(this lineWidth as Float, this lineWidth as Float))
+				result = Shapes line(start, end, this lineWidth, this colorBgra, FloatPoint2D new(this lineWidth as Float, this lineWidth as Float))
 			case => // LineStyle Solid
-				result = Shapes line(start, end, this lineWidth, this opacity, this color)
+				result = Shapes line(start, end, this lineWidth, this colorBgra)
+		}
+		result & Shapes text(FloatPoint2D new(end x, end y + fontSize / 3), this label, fontSize, this colorBgra)
+	}
+	minValues: override func -> FloatPoint2D {
```
simonmika: `minimumValues`

source/plot/LinePlotData2D.ooc
```diff
@@ -50,20 +54,40 @@ LinePlotData2D: class extends PlotData2D {
...
+		if (this dataSeries empty)
+			result = FloatPoint2D new()
+		else {
+			result = this dataSeries[0]
+			for (i in 1 .. this dataSeries count)
+				result = result minimum(this dataSeries[i])
+		}
+		result
+	}
+	maxValues: override func -> FloatPoint2D {
```
simonmika: `maximumValues`

source/plot/ShapePlotData2D.ooc
simonmika: It should be:
```ooc
area := transform * object region
```

source/plot/SvgPlot.ooc
simonmika: I agree, we should have it, but first we should find a good way to represent color.

simonmika:
While this PR includes a lot of clean up that certainly can be welcome I am questioning the whole idea of adding a general drawing API to plots and I am definitely opposed of the API this PR introduces.

simonmika:
So:

marcusnaslund:
So, close this?

bettinaselig:
My suggestion is to create an issue to clean up the whole plot module and another issue to discuss the changes that would be introduced by this pull request and https://github.com/cogneco/ooc-kean/pull/664. Then to close both pull requests.
@simonmika @marcusnaslund @davidhesselbom ?

There is a pull request in the other project that is dependent on this one. But that will be handled separately.

marcusnaslund:
Sounds good to me.

davidhesselbom:
I still think https://github.com/davidpiuva/ooc-kean/commit/c799f0eb41c824af5d3451b44383e979713c5ee9 can be useful, though.

davidhesselbom:
@bettinaselig Will you create the issues you mentioned so I can close this?

bettinaselig:
Yes. I am on it already :-)

bettinaselig:
I opened the issues https://github.com/cogneco/ooc-kean/issues/800 and https://github.com/cogneco/ooc-kean/issues/801.

davidhesselbom:
Thanks.

PR #664, opened 2015-11-02
Plot round

davidpiuva:
Allowing plotting with round shapes. Lines are also round now.

### DIFF
--- a/source/plot/LinePlotData2D.ooc
+++ b/source/plot/LinePlotData2D.ooc
@@ -12,24 +12,36 @@ LineStyle: enum {
 
 LinePlotData2D: class extends PlotData2D {
 	lineStyle: LineStyle { get set }
+	dataSeries: VectorList<FloatPoint2D> { get set }
 
 	init: func ~default (lineStyle := LineStyle Solid) {
-		super()
+		super("", ColorBgra new())
+		this dataSeries = VectorList<FloatPoint2D> new()
 		this lineStyle = lineStyle
 	}
 
-	init: func ~dataSeries (dataSeries: VectorList<FloatPoint2D>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(dataSeries, label, colorBgra)
+	init: func ~dataSeries (=dataSeries, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
+		super(label, colorBgra)
 		this lineStyle = lineStyle
 	}
 
-	init: func ~color (dataSeries: VectorList<FloatPoint2D>, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
-		this init(dataSeries, "", colorBgra, lineStyle)
+	init: func ~color (=dataSeries, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
+		super("", colorBgra)
+		this lineStyle = lineStyle
 	}
 
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(xSeries, ySeries, label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count) {
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
+		}
+		super(label, colorBgra)
 		this lineStyle = lineStyle
+	}
+
+	free: override func {
+		this dataSeries free()
+		super()
 	}
 
 	getSvg: func (transform: FloatTransform2D) -> String {
@@ -64,6 +76,30 @@ LinePlotData2D: class extends PlotData2D {
 				result = Shapes line(start, end, this lineWidth, this opacity, this color)
 		}
 		result = result & Shapes text(FloatPoint2D new(end x, end y + fontSize / 3), this label, fontSize, this opacity, this color)
+		result
+	}
+
+	minValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this dataSeries empty)
+			result = FloatPoint2D new()
+		else {
+			result = this dataSeries[0]
+			for (i in 1 .. this dataSeries count)
+				result = result minimum(this dataSeries[i])
+		}
+		result
+	}
+
+	maxValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this dataSeries empty)
+			result = FloatPoint2D new()
+		else {
+			result = this dataSeries[0]
+			for (i in 1 .. this dataSeries count)
+				result = result maximum(this dataSeries[i])
+		}
 		result
 	}
 }
--- a/source/plot/LinePlotData2D.ooc
+++ b/source/plot/LinePlotData2D.ooc
@@ -9,41 +9,33 @@ LineStyle: enum {
 	Dashed
 	Dotted
 }
-
 LinePlotData2D: class extends PlotData2D {
 	lineStyle: LineStyle { get set }
 	dataSeries: VectorList<FloatPoint2D> { get set }
-
 	init: func ~default (lineStyle := LineStyle Solid) {
 		super("", ColorBgra new())
 		this dataSeries = VectorList<FloatPoint2D> new()
 		this lineStyle = lineStyle
 	}
-
 	init: func ~dataSeries (=dataSeries, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
 		super(label, colorBgra)
 		this lineStyle = lineStyle
 	}
-
 	init: func ~color (=dataSeries, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
 		super("", colorBgra)
 		this lineStyle = lineStyle
 	}
-
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		this dataSeries = VectorList<FloatPoint2D> new()
-		for (i in 0 .. ySeries count) {
-			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
-		}
 		super(label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count)
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
 		this lineStyle = lineStyle
 	}
-
 	free: override func {
 		this dataSeries free()
 		super()
 	}
-
 	getSvg: func (transform: FloatTransform2D) -> String {
 		result := ""
 		if (!this dataSeries empty) {
@@ -62,7 +54,6 @@ LinePlotData2D: class extends PlotData2D {
 		}
 		result
 	}
-
 	getSvgLegend: func (legendCount, fontSize: Int) -> String {
 		result: String
 		start := FloatPoint2D new(this legendOffset, this legendOffset + (fontSize * legendCount - fontSize / 2) as Float)
@@ -75,10 +66,8 @@ LinePlotData2D: class extends PlotData2D {
 			case => // LineStyle Solid
 				result = Shapes line(start, end, this lineWidth, this colorBgra)
 		}
-		result = result & Shapes text(FloatPoint2D new(end x, end y + fontSize / 3), this label, fontSize, this colorBgra)
-		result
+		result & Shapes text(FloatPoint2D new(end x, end y + fontSize / 3), this label, fontSize, this colorBgra)
 	}
-
 	minValues: override func -> FloatPoint2D {
 		result: FloatPoint2D
 		if (this dataSeries empty)
@@ -90,7 +79,6 @@ LinePlotData2D: class extends PlotData2D {
 		}
 		result
 	}
-
 	maxValues: override func -> FloatPoint2D {
 		result: FloatPoint2D
 		if (this dataSeries empty)
--- a/source/plot/LinePlotData2D.ooc
+++ b/source/plot/LinePlotData2D.ooc
@@ -56,7 +56,7 @@ LinePlotData2D: class extends PlotData2D {
 	}
 	getSvgLegend: func (legendCount, fontSize: Int) -> String {
 		result: String
-		start := FloatPoint2D new(this legendOffset, this legendOffset + (fontSize * legendCount - fontSize / 2) as Float)
+		start := FloatPoint2D new(this legendOffset, this legendOffset + (fontSize * legendCount as Float) - (0.5f * fontSize))
 		end := FloatPoint2D new(this legendOffset + fontSize, start y)
 		match (this lineStyle) {
 			case LineStyle Dashed =>
--- a/source/plot/LinePlotData2D.ooc
+++ b/source/plot/LinePlotData2D.ooc
@@ -47,7 +47,7 @@ LinePlotData2D: class extends PlotData2D {
 	getSvg: func (transform: FloatTransform2D) -> String {
 		result := ""
 		if (!this dataSeries empty) {
-			result = result & "<path stroke='" + this color >> "' stroke-opacity='" & this opacity toString() >> "' fill='none' stroke-width='" & this lineWidth toString() >> "' d='M " & ((transform * this dataSeries[0]) x) toString() >> " " & ((transform * this dataSeries[0]) y) toString() >> " L "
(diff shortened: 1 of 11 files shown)


davidpiuva:
How can the test fail at 492bf0d and then randomly work again at f26d41f?

marcusnaslund:
For the first one, magic had complaints.
`Vector2D` seems to be slightly magical on Windows.

marcusnaslund:
Sorry.

source/plot/LinePlotData2D.ooc
```diff
@@ -12,30 +12,42 @@ LineStyle: enum {
...
-	init: func ~dataSeries (dataSeries: VectorList<FloatPoint2D>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(dataSeries, label, colorBgra)
+	init: func ~dataSeries (=dataSeries, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
+		super(label, colorBgra)
 		this lineStyle = lineStyle
 	}
 
-	init: func ~color (dataSeries: VectorList<FloatPoint2D>, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
-		this init(dataSeries, "", colorBgra, lineStyle)
+	init: func ~color (=dataSeries, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
```
marcusnaslund: If you put the `label` argument last, this constructor would be identical to the one above.

source/plot/LinePlotData2D.ooc
```diff
@@ -12,30 +12,42 @@ LineStyle: enum {
...
+		super("", colorBgra)
+		this lineStyle = lineStyle
 	}
 
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
-		super(xSeries, ySeries, label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count) {
```
marcusnaslund: These brackets are not needed

source/plot/ScatterPlotData2D.ooc
```diff
@@ -4,29 +4,35 @@ use ooc-draw
...
-	init: func ~color (dataSeries: VectorList<FloatPoint2D>, colorBgra: ColorBgra) {
-		super(dataSeries, "", colorBgra)
+	init: func ~color (=dataSeries, colorBgra: ColorBgra) {
+		super("", colorBgra)
 	}
 
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new()) {
-		super(xSeries, ySeries, label, colorBgra)
+		this dataSeries = VectorList<FloatPoint2D> new()
+		for (i in 0 .. ySeries count) {
```
marcusnaslund: These brackets are not needed

source/plot/ScatterPlotData2D.ooc
```diff
@@ -4,29 +4,35 @@ use ooc-draw
...
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
```
marcusnaslund: `1.0f + i` gets rid of the explicit cast, I believe.

source/plot/ScatterPlotData2D.ooc
```diff
@@ -57,12 +63,36 @@ ScatterPlotData2D: class extends PlotData2D {
 		halfLineHeight := (fontSize as Float) / 2.0f
 		match (this shape) {
 			case Shape Circle =>
-				result = result & Shapes circle(FloatPoint2D new(start x + halfLineHeight, start y), size / 2.0f, this opacity, this color)
+				result = result & Shapes circle(FloatPoint2D new(start x + halfLineHeight, start y), size / 2.0f, this colorBgra)
 			case Shape Square =>
-				result = result & Shapes rect(FloatPoint2D new(start x, start y - halfLineHeight), FloatPoint2D new(size, size), this opacity, this color)
+				result = result & Shapes rect(FloatPoint2D new(start x, start y - halfLineHeight), FloatPoint2D new(size, size), this colorBgra)
 			case =>
```
marcusnaslund: Not introduced here, but this looks dangerous. What happens? Please put a `raise("Unsupported shape in...")` or something here.

source/plot/ScatterPlotData2D.ooc
```diff
@@ -57,12 +63,36 @@ ScatterPlotData2D: class extends PlotData2D {
...
 		}
-		result = result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this opacity, this color)
+		result = result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this colorBgra)
+		result
+	}
+
+	minValues: override func -> FloatPoint2D {
```
marcusnaslund: Since these functions crop up in several places, add these to `FloatPoint2DVectorList` and use that instead of `VectorList<FloatPoint2D>`.

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,100 @@
...
+	addRectangle: func (box: FloatBox2D, colorBgra := ColorBgra new()) {
+		this _shapes add(AnyShape new(Shape Rectangle, box, colorBgra))
+	}
+	addCircle: func (center: FloatPoint2D, radius: Float, colorBgra := ColorBgra new()) {
+		this addEllipse(center, radius, radius, colorBgra)
+	}
+	addEllipse: func (center: FloatPoint2D, radiusX, radiusY: Float, colorBgra := ColorBgra new()) {
+		this _shapes add(AnyShape new(Shape Ellipse, FloatBox2D new(center x - radiusX, center y - radiusY, radiusX * 2.0f, radiusY * 2.0f), colorBgra))
+	}
+
```
marcusnaslund: Several empty lines between functions here, we don't usually have that. Please remove.

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,100 @@
...
+		result = result & Shapes text(FloatPoint2D new(start x + fontSize as Float, start y + halfLineHeight), this label, fontSize, this colorBgra)
+		result
+	}
+
+	minValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this _shapes empty)
+			result = FloatPoint2D new()
+		else {
+			object := this _shapes[0] // Convert rvalue to lvalue
```
marcusnaslund: These comments are not needed. (same fore line 79, 91, 94)

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,100 @@
...
+			result = object region leftTop
+			for (i in 1 .. this _shapes count) {
+				object := this _shapes[i] // Convert rvalue to lvalue
```
marcusnaslund: Does this work? You're allocating `object` which already exists from line 76.

source/plot/ShapePlotData2D.ooc
```diff
@@ -0,0 +1,100 @@
...
+
+	maxValues: override func -> FloatPoint2D {
+		result: FloatPoint2D
+		if (this _shapes empty)
+			result = FloatPoint2D new()
+		else {
+			object := this _shapes[0] // Convert rvalue to lvalue
+			result = object region rightBottom
+			for (i in 1 .. this _shapes count) {
+				object := this _shapes[i] // Convert rvalue to lvalue
```
marcusnaslund: Same here.

source/plot/SvgPlot.ooc
```diff
@@ -105,12 +105,8 @@ SvgPlot: class {
 		if (!datasets empty) {
 			noneColor := ColorBgra new(0, 0, 0, 0)
 			for (j in 0 .. datasets count) {
-				if (noneColor != datasets[j] colorBgra) {
-					datasets[j] color = datasets[j] colorBgra svgRGBToString()
-					datasets[j] opacity = ((datasets[j] colorBgra svgRGBAlpha()) as Float) / 255.0f
-				} else {
-					datasets[j] color = colorList[this colorCount % this colorList count]
-					datasets[j] opacity = 1
+				if (noneColor == datasets[j] colorBgra) {
```
marcusnaslund: What about the `!=` case, do we not care about it anymore?

source/plot/ShapePlotData2D.ooc
davidpiuva: I don't see any allocation but I can use basic assignments after the first one to reduce the variable count.

source/plot/LinePlotData2D.ooc
```diff
@@ -12,30 +12,42 @@ LineStyle: enum {
...
+			this dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
+		}
+		super(label, colorBgra)
```
marcusnaslund: Place the `super` call first.

source/plot/LinePlotData2D.ooc
```diff
@@ -57,13 +69,37 @@ LinePlotData2D: class extends PlotData2D {
...
+				result = Shapes line(start, end, this lineWidth, this colorBgra, FloatPoint2D new(this lineWidth as Float, this lineWidth as Float))
 			case => // LineStyle Solid
-				result = Shapes line(start, end, this lineWidth, this opacity, this color)
+				result = Shapes line(start, end, this lineWidth, this colorBgra)
+		}
+		result = result & Shapes text(FloatPoint2D new(end x, end y + fontSize / 3), this label, fontSize, this colorBgra)
+		result
+	}
+
+	minValues: override func -> FloatPoint2D {
```
marcusnaslund: Should be in `FloatPoint2DVectorList` as commented below.

source/plot/SvgPlot.ooc
davidpiuva: We only override non existing colors. The others are already defined.

source/plot/ShapePlotData2D.ooc
marcusnaslund: Yes, `=` instead of `:=`. I'm confused about why this works.

davidpiuva: Because { } creates a new local scope like in C99.

marcusnaslund: Oh you're right. Anyway, `=` is more appropriate here.

source/plot/LinePlotData2D.ooc
davidpiuva: The result for empty lists might cause arguments and breaking changes in Kean. I would return positive infinity for minimum of nothing so that comparisons can still work.

marcusnaslund: There are no breaking changes since noone else uses these functions but you right now. :) Also, if you want something "strange", you should handle that in this class. This function still belongs in `FloatPoint2DVectorList`. In other words, here you should do something like:
`result := (myList empty ? Float positiveInfinity : myList minimumValue)`

source/plot/LinePlotData2D.ooc
davidpiuva: That makes sense but I don't want to bother Niklas with breaking Kean changes. Is there another way?

marcusnaslund: No, let's leave it for now then.

davidpiuva:
Moving minimum and maximum to Kean broke the test case somehow.

davidpiuva:
Rebased and ready for peer review.

test/plot/SvgPlotTest.ooc
```diff
@@ -40,14 +36,22 @@ scatterData := ScatterPlotData2D new(scatter, "Random numbers")
...
-trigonometryPlot addDataset(cosData)
-trigonometryPlot addDataset(sinMinusCosData)
-trigonometryPlot xAxis label = "x"
-trigonometryPlot yAxis label = "y"
+shapesData := ShapePlotData2D new("Mixed shapes", ColorBgra new(0, 0, 0, 128))
+shapesData addRectangle(FloatBox2D new(-4, 2, 8, 1)) // Default
+shapesData addRectangle(FloatBox2D new(-4, -3, 8, 1)) // Default
+shapesData addCircle(FloatPoint2D new(-3.1416, 0), 1, ColorBgra new(0, 0, 200, 128)) // Red
+shapesData addCircle(FloatPoint2D new(0, 0), 1, ColorBgra new(0, 200, 0, 128)) // Green
+shapesData addCircle(FloatPoint2D new(3.1416, 0), 1, ColorBgra new(200, 0, 0, 128)) // Blue
```
sebastianbaginski: Why not `Float pi` ?

davidpiuva: I can try that.

source/plot/LinePlotData2D.ooc
```diff
@@ -9,33 +9,37 @@ LineStyle: enum {
...
 	}
-
-	init: func ~color (dataSeries: VectorList<FloatPoint2D>, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
-		this init(dataSeries, "", colorBgra, lineStyle)
+	init: func ~color (=dataSeries, colorBgra: ColorBgra, lineStyle := LineStyle Solid) {
+		super("", colorBgra)
+		this lineStyle = lineStyle
 	}
-
 	init: func ~twoFloatSeries (xSeries, ySeries: VectorList<Float>, label := "", colorBgra := ColorBgra new(), lineStyle := LineStyle Solid) {
```
sebastianbaginski: So this constructor copies the input while the other two takes ownership of the input.
Shouldn't we do the same thing (copy) in all of them ? Or at least leave a comment about this, as it is not clear  from the API when user is supposed to free the data. @davidhesselbom ?

davidpiuva: I don't really know how it works. I think a student wrote it.

sebastianbaginski: Oh, I see.

davidhesselbom: Yes, it should be consistent, so `xSeries` and `ySeries` should be freed here, I think.

source/plot/PlotData2D.ooc
```diff
@@ -3,67 +3,17 @@ use ooc-collections
...
-		for (i in 0 .. ySeries count) {
-			dataSeries add(FloatPoint2D new(xSeries != null ? xSeries[i] : (i + 1) as Float, ySeries[i]))
-		}
-		this init(dataSeries, label, colorBgra)
-	}
-
+	label: String
+	lineWidth: Float = 1.0f
+	legendOffset: Float = 5.0f
+	colorBgra: ColorBgra
```
sebastianbaginski: Isn't this PR going to conflict with your other one ? I'm pretty sure I saw this change yesterday in your other PR

source/plot/LinePlotData2D.ooc
sebastianbaginski: So you could not reuse, for example, the same `X` data with different `Y` values for two different plots ? But then you will have to free the data manually. I don't know what's best, up to you.

source/plot/svg/Shapes.ooc
```diff
@@ -1,31 +1,45 @@
...
 	}
+	text: static func ~simple (position: FloatPoint2D, text: String, fontSize: Int, color := ColorBgra new(0, 0, 0, 255)) -> String {
+		This text(position, text, fontSize, "start", color)
+	}
+	text: static func ~textAnchor (position: FloatPoint2D, text: String, fontSize: Int, textAnchor: String, color := ColorBgra new(0, 0, 0, 255)) -> String {
+		"<text id='" + text >> "' x='" & position x toString() >> "' y='" & position y toString() >> "' font-size='" & fontSize toString() >> "px' text-anchor='" >> textAnchor >> "' fill='" >> This getColor(color) >> "' fill-opacity='" >> This getOpacity(color) >> "'>" >> text >> "</text>\n"
+	}
+}
+
+Shape: enum {
```
sebastianbaginski: Could it be possible to use a base class and subclasses for every shape instead of `enum` like this ? Just asking because @davidhesselbom likes object-oriented style :)

source/plot/LinePlotData2D.ooc
davidpiuva: Since it is for testing purposes only, I rather have an accidental memory leak than an accidental invalid free.

sebastianbaginski:
Looks ok, magic ok, builds & runs.

source/plot/svg/Shapes.ooc
davidpiuva: Then we have a new allocation for each element which makes finding memory leaks in the real code harder if it starts to leak.

marcusnaslund: In any case, it seems more reasonable that this enum should be called `Shapes`, and the class above called `Shape`.

source/plot/PlotData2D.ooc
davidpiuva: It is rebased on plotShape.

davidhesselbom:
Builds on a not-ready PR (#663), so, not ready.

bettinaselig:
I opened the issues https://github.com/cogneco/ooc-kean/issues/800 and https://github.com/cogneco/ooc-kean/issues/801.

davidhesselbom:
Thanks.

PR #665, opened 2015-11-02
Simple event processing loop and handlers for mouse clicks in DisplayWindow

sebastianbaginski:
I think it's good enough for now, as I can implement the other thing needed by @bettinaselig on top of it.
For now there is no abstraction around NativeEvent, EventLoop, EventDispatcher etc. but I will happily extend this later.
No version for windows, for now.

### DIFF
--- a/source/widgets/DisplayWindow.ooc
+++ b/source/widgets/DisplayWindow.ooc
@@ -15,13 +15,23 @@
  * along with this software. If not, see <http://www.gnu.org/licenses/>.
  */
 
+use ooc-base
 use ooc-draw
 use ooc-math
 import UnixWindow
 import Win32DisplayWindow
 
 DisplayWindow: abstract class {
+	_mousePressHandler: Event1<IntPoint2D>
+	_mouseReleaseHandler: Event1<IntPoint2D>
 	init: func (size: IntSize2D, title: String)
+	free: override func {
+		if (this _mousePressHandler)
+			this _mousePressHandler free()
+		if (this _mouseReleaseHandler)
+			this _mouseReleaseHandler free()
+		super()
+	}
 	draw: abstract func (image: Image)
 	refresh: virtual func
 	create: static func (size: IntSize2D, title: String) -> This {
@@ -31,5 +41,18 @@ DisplayWindow: abstract class {
 			return Win32DisplayWindow new(size, title)
 		raise("Platform not supported (DisplayWindow)")
 		null
+	}
+	processEvents: abstract func
+	addMousePressHandler: func (handler: Event1<IntPoint2D>) {
+		if (!this _mousePressHandler)
+			this _mousePressHandler = handler
+		else
+			this _mousePressHandler = this _mousePressHandler add(handler)
+	}
+	addMouseReleaseHandler: func (handler: Event1<IntPoint2D>) {
+		if (!this _mouseReleaseHandler)
+			this _mouseReleaseHandler = handler
+		else
+			this _mouseReleaseHandler = this _mouseReleaseHandler add(handler)
 	}
 }
--- a/source/widgets/UnixWindow.ooc
+++ b/source/widgets/UnixWindow.ooc
@@ -21,6 +21,7 @@ use ooc-math
 use ooc-draw
 use ooc-opengl
 use ooc-x11
+import include/x11
 
 version(unix || apple) {
 UnixWindowBase: class extends DisplayWindow {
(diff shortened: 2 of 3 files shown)


source/widgets/DisplayWindow.ooc
```diff
@@ -19,9 +19,19 @@ use ooc-draw
 use ooc-math
 import UnixWindow
 import Win32DisplayWindow
+import Event
```
davidhesselbom: No, `use ooc-base`.

source/widgets/UnixWindow.ooc
```diff
@@ -43,5 +44,15 @@ UnixWindow: class extends DisplayWindow {
...
+	processEvents: override func {
+		while (XPending(this _xWindow display)) {
+			event: XEventOoc
+			XNextEvent(this _xWindow display, event&)
+			if ((event type == ButtonPress) && this _mousePressHandler)
+				this _mousePressHandler call(IntPoint2D new(event xkey x, event xkey y))
+			if ((event type == ButtonRelease) && this _mouseReleaseHandler)
+				this _mouseReleaseHandler call(IntPoint2D new(event xkey x, event xkey y))
+		}
+	}
```
davidhesselbom: Doesn't this belong in X11Window?

sebastianbaginski: You are right, but this will be gone from both the `UnixWindow` and `X11Window`, as it in fact belongs to the, yet non-exisiting, `X11Application` class, as events processing can happen even without the window dispplayed on screen. I want to extend this soon, this is only a temporary solution for Bettina's problem in the other project.

davidhesselbom: Is there an issue for this? I don't like leaving things where they don't belong, and I especially don't like doing it without leaving a note about it.

sebastianbaginski: I understand. Is this enough for now https://github.com/cogneco/ooc-kean/issues/705 ?

davidhesselbom: Okay, but the problem here is that you're doing things specific to X11 in the class called `UnixWindow`, and you're also importing `inlude/x11`, which is part of `ooc-ui`.

sebastianbaginski: It was like that before - every `UnixWindow` is based on `X11Window`, even the one with OpenGL backend.

davidhesselbom: Okay, but you didn't need `include/x11` until now. Can't you move the things in this function into a new one in `X11Window` and let this function here call that one?

sebastianbaginski: Could be done that way, but I will have to move the mouse event handlers there as well and create functions to forward then into the xWindow from DisplayWindow. This will add some work later when I will clean this up anyway.

marcusnaslund:
Validated @davidhesselbom

bettinaselig:
What about this pull request? Can you come to an agreement?

davidhesselbom:
My only problem, really, is that a class in `widgets` uses an include from `ui`. It doesn't seem right to me. Maybe @simonmika can tell me whether I should care...

sebastianbaginski:
It will be gone soon, I just don't want to write code which will be removed anyway.

sebastianbaginski:
And `include/X11.ooc` should not be located in `ui` module anyway.

simonmika:
Why is there as `widgets` namespace? I suggest putting it all into `ui` and probably rename it to `gui`. Not all user interfaces are based on windows and mouse movements (and not all graphical user interfaces do in fact either). The `cli` namespace contains stuff for a totally different user interface paradigm.

davidhesselbom:
@sebastian can explain this, but it had to do with the x11 window being dependent on opengl, and this was the workaround. It was never meant to be a permanent solution in the first place, though.

sebastianbaginski:
Exactly, ui module has opengl dependency, and we needed to have a ui without opengl.

bettinaselig:
@simonmika and @sebastianbaginski, please find and agreement.

sebastianbaginski:
@bettinaselig : this is just a temporary solution (as suggested by @simonmika ) to enable you to use the mouse in the other project. I don't understand why it is a problem, it was much worse before this (hardcoded X11 event loop in `main()` ) and I will clean this up later.

davidhesselbom:
The hard-coded event loop was really bad, yeah (even when we hadn't considered being able to run without OpenGL or X11). That's why it never made it into the master branch :)

bettinaselig:
@simonmika and @sebastianbaginski will create a new pull request.

sebastianbaginski:
I'm gonna redo it anyway, so closing this one to clean up the queue.

davidhesselbom:
Appreciated / sorry.

sebastianbaginski:
No problem, I guess it's even better to first clean up the widgets / ui stuff.

PR #667, opened 2015-11-03
Implemented additional mutex types

sebastianbaginski:
It will be needed to simplify the implementation of thread unsafe reference counters.

### DIFF
--- a/source/sdk/threading/Thread.ooc
+++ b/source/sdk/threading/Thread.ooc
@@ -110,6 +110,12 @@ Thread: abstract class {
 
 }
 
+MutexType: enum {
+	Safe,
+	Unsafe,
+	Global
+}
+
 /**
  * A mutex is a mechanism used for synchronizing threads.
  *
@@ -120,23 +126,24 @@ Mutex: abstract class {
 
     /**
        :return: an intialized mutex, unlocked.
-
-       IMPORTANT: mutexes are special beasts. Don't attempt to access
-       the class of a mutex. Depending on the underlying implementation,
-       it may not be a real ooc object but only a pointer.
      */
-    new: static func -> This {
-
-        version (unix || apple) {
-            return MutexUnix new()
-        }
-        version (windows) {
-            return MutexWin32 new()
-        }
-
-        Exception new(This, "Unsupported platform!\n") throw()
-        null
-
+    new: static func (mutexType := MutexType Safe) -> This {
+		result: This
+		match (mutexType) {
+			case MutexType Safe => {
+				version (unix || apple)
+					result = MutexUnix new()
+				version (windows)
+					result = MutexWin32 new()
+				if (result == null)
+					Exception new(This, "Unsupported platform!\n") throw()
+			}
+			case MutexType Unsafe =>
+				result = MutexUnsafe new()
+			case =>
+				result = MutexGlobal new()
+		}
+        result
     }
 
     /**
@@ -163,6 +170,26 @@ Mutex: abstract class {
         unlock()
     }
 
+}
+
+MutexUnsafe: class extends Mutex {
+	init: func
+	lock: override func
+	unlock: override func
+}
+
+MutexGlobal: class extends Mutex {
+	_globalMutex := static Mutex new(MutexType Safe)
+	init: func
+	lock: override func {
+		This _globalMutex lock()
+	}
+	unlock: override func {
+		This _globalMutex unlock()
+	}
+	cleanup: static func {
+		This _globalMutex free()
+	}
 }
 
 /**


source/sdk/threading/Thread.ooc
```diff
@@ -165,6 +172,23 @@ Mutex: abstract class {
 
 }
 
+MutexUnsafe: class extends Mutex {
+	init: func
+	lock: override func
```
marcusnaslund: Since `lock` and `unlock` are abstract, these aren't overrides..?

source/sdk/threading/Thread.ooc
```diff
@@ -165,6 +172,23 @@ Mutex: abstract class {
 
 }
 
+MutexUnsafe: class extends Mutex {
```
marcusnaslund: What is this class supposed to do?

sebastianbaginski: Nothing, in fact.

marcusnaslund: That surely *is* unsafe. :)

sebastianbaginski: Hence the name :D
It is used to implement reference counters which do not use a real mutex lock but still needs the mutex interface for the sake of simplicity.

source/sdk/threading/Thread.ooc
sebastianbaginski: Yes, `lock/unlock` are abstract and these are overrides.

source/sdk/threading/Thread.ooc
marcusnaslund: Ah, that makes sense, alright.

source/sdk/threading/Thread.ooc
marcusnaslund: If they were `virtual`, you would have overrides. Since they are `abstract` and thus not implemented in the base class, they are not.

marcusnaslund: Of course, rock doesn't seem to mind at all.

sebastianbaginski: Is there any point of having `abstract` function in base class which is *not* virtual ?

marcusnaslund: I just don't think we use the `override` keyword for `abstract` functions anywhere else. Better have @davidhesselbom say something about this.

sebastianbaginski: We do, check in the other project.
Looks like it works kind of as `override` in C++, like a safety mechanism to make sure that you are implementing correct interface in the derived class. So if you provide an `override` specifier for a function that does not exist in the base class, you will get a compiler error.

davidhesselbom: I'll just leave this here:
```ooc
Foo: abstract class {
    init: func
    hey: abstract func
}
Bar: class extends Foo {
    init: func
    hey: override func { println("hey") }
}
Baz: class extends Bar {
    init: func
    hey: override func { println("hey you") } // won't work, because `hey` in `Bar` isn't virtual
}
```
and
```ooc
Foo: abstract class {
    init: func
    hey: abstract func
}
Bar: class extends Foo {
    init: func
    hey: func { println("hey") }
}
Baz: class extends Bar {
    init: func
    hey: override func { println("hey you") }

bar := Bar new()
bar hey()
baz := Baz new()
baz hey()
```
both print
```
hey
hey you
```
but
```ooc
Foo: abstract class {
    init: func
    hey: abstract func
}
Bar: class extends Foo {
    init: func
    hey: virtual func { println("hey") }
}
Baz: class extends Bar {
    init: func
    hey: override func { println("hey you") }
}

bar := Bar new()
bar hey()
baz := Baz new()
baz hey()
```
prints
```
Child received signal 11 (Segmentation fault)
```
Perhaps @fredrikbryntesson can explain why the `override` keyword in `Bar` is optional?

source/sdk/threading/Thread.ooc
```diff
@@ -165,6 +172,23 @@ Mutex: abstract class {
...
+MutexGlobal: class extends Mutex {
+	_globalMutex := static Mutex new(MutexType Safe)
+	init: func
+	lock: override func {
+		This _globalMutex lock()
+	}
+	unlock: override func {
+		This _globalMutex unlock()
+	}
+}
```
davidhesselbom: It would be nice to have a cleanup function here, in case, for whatever reason (valgrind logs or who knows what), someone wants to be able to destroy the `_globalMutex`.

sebastianbaginski: That's true, fixed.

source/sdk/threading/Thread.ooc
```diff
@@ -165,6 +172,26 @@ Mutex: abstract class {
...
+MutexGlobal: class extends Mutex {
+	_globalMutex := static Mutex new(MutexType Safe)
+	init: func
+	lock: override func {
+		This _globalMutex lock()
+	}
+	unlock: override func {
+		This _globalMutex unlock()
+	}
+	cleanup: static func {
```
marcusnaslund: I think `dispose` might be better and more in line with similar things, for example `Profiler`.

sebastianbaginski: In `ByteBuffer` it is called `clean`. Is `Profiler` class the only example of using `dispose` ?

marcusnaslund: It is, unless you count the `dispose()` calls to every `Closure`. 
To be fair, I find only two examples in kean (`ByteBuffer` and `RecycleBin`) of using `clean`. :)
@davidhesselbom ?

davidhesselbom: Before `free`, there was `dispose` (something we "inherited" from the C# code base). I think it makes sense to let the static functions be called something other than `free` and use `free` for class instances. *But* I do agree we shouldn't have both `dispose` and `clean` and `cleanup`.

marcusnaslund: I think `dispose` sounds better here.

sebastianbaginski: I like `dispose` too, but it is already used in Closure to free the `instances` of Closures. Maybe we should use something different to clean up `instances` and `classes`.

davidhesselbom: In `Closure` is should probably be `free`. At some point, we had an ambition to name cover destructors `dispose`. I don't remember why, though (and I'm also not sure why `Closure` is a cover).

marcusnaslund: I find "I suggest that we turn Closure from a cover (struct) into a class." in a very big issue from June 2012: https://github.com/fasterthanlime/rock/issues/439

davidhesselbom: No one hates rock the way Amos does :)

source/sdk/threading/Thread.ooc
davidhesselbom: @fredrikbryntesson ?

source/sdk/threading/Thread.ooc
davidhesselbom: I know it's in line with the existing `MutexUnix` and `MutexWin32`, but these names are funny looking, like `ImageRasterMonochrome` or so.

source/sdk/threading/Thread.ooc
davidhesselbom: So, basically, Fredrik's reply is that "the test is written a little strangely". I guess that means a fix is on the way.

source/sdk/threading/Thread.ooc
davidhesselbom: https://github.com/cogneco/ooc-kean/issues/727 will have to do for now.

PR #645, opened 2015-10-30
Moved out LinkedList from sdk

marcusnaslund:
Cleaned it up to make magic happy and added a simple test.

PR #674, opened 2015-11-03
Moved Terminal color+attribute support out of the SDK

### LINKED ISSUE #673: Move Terminal colouring and attributes from the sdk to source/io
Never used anywhere in the sdk, and we might want it for Fixture, so easy to fix.

marcusnaslund:
Fixes #673 

Removed the Terminal handler in the SDK.

The Win32 part has been removed since we only need to use ASCII escape sequences. They work fine on Linux and should work fine on Windows where we use `msys2` and Appveyor.


marcusnaslund:
Waiting on #645

marcusnaslund:
This is OK now.

PR #675, opened 2015-11-03
Removed Base64 encoding/decoding from sdk

marcusnaslund:
I can imagine a veeeery small usefulness for the networking part, but since we're not even using that, let's throw this away.


sebastianbaginski:
not so long ago @simonmika mentioned something about networking being useful in the future, I guess you should ask him about it

marcusnaslund:
@simonmika says remove.

simonmika:
We might have situations where we want to pick this up again but at the moment let us remove it. It most likely is very leaky.

PR #676, opened 2015-11-04
Updated distance method in raster image classes

sebastianbaginski:
Cleaned up `distance` methods in raster images for https://github.com/cogneco/ooc-kean/issues/591 .
Since no one has a better idea, I suggest that for two images with different sizes we return "maximum" distance. It is better than previously returned zero (*="images are identical"*) anyway.

### DIFF
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -63,13 +63,13 @@ RasterBgr: class extends RasterPacked {
 	}
 	distance: func (other: Image) -> Float {
 		result := 0.0f
-		if (!other)
+		if (!other || (this size != other size))
 			result = Float maximumValue
-//		else if (!other instanceOf?(This))
-//			FIXME
-//		else if (this size != other size)
-//			FIXME
-		else {
+		else if (!other instanceOf?(This)) {
+			converted := This convertFrom(other as RasterImage)
+			result = this distance(converted)
+			converted referenceCount decrease()
+		} else {
 			for (y in 0 .. this size height)
 				for (x in 0 .. this size width) {
 					c := this[x, y]
--- a/source/draw/RasterUv.ooc
+++ b/source/draw/RasterUv.ooc
@@ -61,13 +61,13 @@ RasterUv: class extends RasterPacked {
 
 	distance: func (other: Image) -> Float {
 		result := 0.0f
-		if (!other)
+		if (!other || (this size != other size))
 			result = Float maximumValue
-//		else if (!other instanceOf?(This))
-//			FIXME
-//		else if (this size != other size)
-//			FIXME
-		else {
+		else if (!other instanceOf?(This)) {
+			converted := This convertFrom(other as RasterImage)
+			result = this distance(converted)
+			converted referenceCount decrease()
+		} else {
 			for (y in 0 .. this size height)
 				for (x in 0 .. this size width) {
 					c := this[x, y]
--- a/source/draw/RasterYuvPlanar.ooc
+++ b/source/draw/RasterYuvPlanar.ooc
@@ -67,12 +67,8 @@ RasterYuvPlanar: abstract class extends RasterPlanar {
 	}
 	distance: func (other: Image) -> Float {
 		result := 0.0f
-		if (!other)
+		if (!other || (this size != other size) || !other instanceOf?(This))
 			result = Float maximumValue
-//		else if (!other instanceOf?(This))
-//			FIXME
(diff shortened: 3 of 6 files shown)


source/draw/RasterBgr.ooc
```diff
@@ -63,13 +63,13 @@ RasterBgr: class extends RasterPacked {
...
 		result := 0.0f
-		if (!other)
+		if (!other || (this size != other size))
 			result = Float maximumValue
-//		else if (!other instanceOf?(This))
-//			FIXME
-//		else if (this size != other size)
-//			FIXME
-		else {
+		else if (!other instanceOf?(This)) {
```
marcusnaslund: Isn't it enough that they're instances of `RasterImage`?

sebastianbaginski: Unfortunately no, as `RasterImage` does not have a generic method to access pixel value at (x, y).

marcusnaslund:
> It is better than previously returned zero (="images are identical") anyway.

Agreed. This does not break anything?

sebastianbaginski:
No, this method is used only in tests anyway.

marcusnaslund:
Validated. @davidhesselbom

source/draw/RasterBgr.ooc
davidhesselbom: We could add one, I suppose, but if we want it to return its "native" color (i.e. RGB or YUV or what have you), we'd still need functions from comparing any color to any color, and at any rate, it's pretty much guaranteed to run much slower than this will.

source/draw/RasterYuvPlanar.ooc
```diff
@@ -67,12 +67,8 @@ RasterYuvPlanar: abstract class extends RasterPlanar {
 	}
 	distance: func (other: Image) -> Float {
 		result := 0.0f
-		if (!other)
+		if (!other || (this size != other size) || !other instanceOf?(This))
 			result = Float maximumValue
```
davidhesselbom: So... don't even bother converting from the other format here?

source/draw/RasterYuvSemiplanar.ooc
```diff
@@ -68,12 +68,8 @@ RasterYuvSemiplanar: abstract class extends RasterPlanar {
 	}
 	distance: func (other: Image) -> Float {
 		result := 0.0f
-		if (!other)
+		if (!other || (this size != other size) || !other instanceOf?(This))
 			result = Float maximumValue
```
davidhesselbom: Or here?

source/draw/RasterUv.ooc
```diff
@@ -61,13 +61,13 @@ RasterUv: class extends RasterPacked {
...
-//		else if (!other instanceOf?(This))
-//			FIXME
-//		else if (this size != other size)
-//			FIXME
-		else {
+		else if (!other instanceOf?(This)) {
+			converted := This convertFrom(other as RasterImage)
+			result = this distance(converted)
+			converted referenceCount decrease()
+		} else {
```
davidhesselbom: This part of the function looks identical in 4 classes now. Is there any way we can move it to some function in `RasterImage`?

source/draw/RasterYuvPlanar.ooc
sebastianbaginski: No. `RasterYuvPlanar` is an abstract class, it does not have the `convertFrom` method.

source/draw/RasterYuvSemiplanar.ooc
sebastianbaginski: Same as above

source/draw/RasterUv.ooc
sebastianbaginski: I don't know. I have the same problem with the bilinear rescaling function I pushed recently, it will look almost the same in every class, but I don't know how to generalize it yet.

davidhesselbom: Alright, another one preserved for posterity, I guess...

source/draw/RasterYuvPlanar.ooc
davidhesselbom: Okay, so the reason we don't move this function into the non-abstract classes is to avoid the code duplication it would result in?

Anyway, I just noticed now that `distance` is 100% identical in this class and in `RasterYuvSemiplanar`. I guess that's better than 99%.

sebastianbaginski: Probably that's the reason, I never thought about it.

PR #678, opened 2015-11-04
FloatBox2D distance measures and projection of all corners

tomasmahlberg:
To be used in the other project. Any suggestions for a better name for `separatedDistance`? It calculates the distance from the box edge to a point in x and y separately.


source/math/FloatBox2D.ooc
```diff
@@ -111,6 +111,20 @@ FloatBox2D: cover {
...
 	}
+	separatedDistance: func (point: FloatPoint2D) -> FloatPoint2D {
+		distanceX := Float absolute(point x - this center x) - this width / 2.f
+		distanceY := Float absolute(point y - this center y) - this height / 2.f
+		FloatPoint2D new(distanceX, distanceY) maximum(FloatPoint2D new(0.f, 0.f))
+	}
+	distance: func (point: FloatPoint2D) -> Float {
+		this separatedDistance(point) norm
+	}
+	maximumSeparatedDistance: func (points: VectorList<FloatPoint2D>) -> FloatPoint2D {
```
davidhesselbom: Why not a `FloatPoint2DVectorList`?

source/math/FloatBox2D.ooc
```diff
@@ -111,6 +111,20 @@ FloatBox2D: cover {
 				result add(i)
 		result
 	}
+	separatedDistance: func (point: FloatPoint2D) -> FloatPoint2D {
```
davidhesselbom: This should return a `FloatSize2D`.

source/math/FloatBox2D.ooc
```diff
@@ -111,6 +111,20 @@ FloatBox2D: cover {
...
+		distanceX := Float absolute(point x - this center x) - this width / 2.f
+		distanceY := Float absolute(point y - this center y) - this height / 2.f
+		FloatPoint2D new(distanceX, distanceY) maximum(FloatPoint2D new(0.f, 0.f))
```
davidhesselbom: If you add an `absolute` property to `FloatSize2D`
```ooc
absolute ::= This new(Float absolute(this width), Float absolute(this height))
```
, these three lines can be reduced to
```ooc
((point - this center) absolute - this size / 2.f) maximum(FloatSize new())
```

davidhesselbom: It seems a little strange to want to know the distance in 2 dimensions, but not the direction, though...

source/math/FloatBox2D.ooc
tomasmahlberg: Is this advised? I'm not interested in using any methods or operators in `FloatPoint2DVectorList`.

source/math/FloatBox2D.ooc
tomasmahlberg: Strange it is, but this is what we need...

tomasmahlberg: A `FloatPoint2D` minus a `FloatSize2D` returns a `FloatPoint2D` so I'm not sure adding an `absolute` property in `FloatSIze2D` solves this.

source/math/FloatBox2D.ooc
davidhesselbom: Okay, leave it then.

source/math/FloatBox2D.ooc
davidhesselbom: Oh, right. I don't think it should, though... Hm. Let me investigate and get back to you.

davidhesselbom: Yeah... it should return a `FloatPoint2D`. Fixing that will probably have tons of side effects, so I'll let it slide for now.

davidhesselbom: However, you can do `(- this center + point) absolute`. Doesn't look very good, but that's kind of the point.

source/math/FloatTransform3D.ooc
```diff
@@ -21,6 +21,7 @@ import FloatPoint2D
 import FloatPoint3D
 import FloatBox2D
 import FloatTransform2D
+import VectorList
```
davidhesselbom: No. `use ooc-collections`.

tomasmahlberg:
Updated.

source/math/FloatSize2D.ooc
```diff
@@ -30,6 +30,10 @@ FloatSize2D: cover {
 	basisY: static This { get { This new(0, 1) } }
 	init: func@ (=width, =height)
 	init: func@ ~square (length: Float) { this width = this height = length }
+	init: func@ ~fromFloatPoint2D (point: FloatPoint2D) {
+		this width = point x
+		this height = point y
+	}
```
davidhesselbom: You could have used `toFloatSize2D` in `FloatPoint2D`, you know.

tomasmahlberg: DIdn't see it! Change?

davidhesselbom: Yes, please use the existing function and don't add this one.

source/math/FloatBox2D.ooc
```diff
@@ -111,6 +111,15 @@ FloatBox2D: cover {
 				result add(i)
 		result
 	}
+	distance: func (point: FloatPoint2D) -> FloatSize2D {
+		((point - this center) toFloatSize2D() absolute - this size / 2.f) maximum(FloatSize2D new())
```
marcusnaslund: Do you mind adding a pair of parentheses here to make it more clear what you want to divide with `.2f`?

davidhesselbom: I disagree here, `*` has higher precedence than `-` and everyone should know that.

davidhesselbom: If the `/` was a `-` so you had two of them, I'd agree with you.

marcusnaslund: nvm then. :)

marcusnaslund:
In that case I'm happy here @davidhesselbom

davidhesselbom:
Maybe there's an explanation in the comments somewhere, but what became of `separatedDistance`?

niklasborwell:
I thought it was a confusing name so I asked to rename it to simply distance. The return type already indicates that you will get a two-dimensional value, so adding ´separated´ isn't necessary.

davidhesselbom:
No argument there! I see now that the other `distance` function added earlier has been removed - nice move.

PR #679, opened 2015-11-04
Fixed bug when converting Pen color to Y and UV textures.

emilwestergren:


source/draw/gpu/GpuCanvas.ooc
```diff
@@ -45,8 +45,8 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		set(value) {
 			this _pen = value
 			yuv := value color toYuv()
-			this _target y canvas pen = Pen new(ColorBgra new(yuv y, 0, 0, 255), this pen width)
-			this _target uv canvas pen = Pen new(ColorBgra new(yuv u, yuv v, 0, 255), this pen width)
```
davidhesselbom: At this point, `this pen` should already have been set to `value`. I don't mind merging this, but why doesn't it work already?

emilwestergren: That's what I thought, but it's not the case. `this pen` is still unintialized here. Maybe `this _pen` would have worked as well but I used `value` to make sure.

davidhesselbom: Strange.

PR #680, opened 2015-11-04
sync check tool.

ziquanyu:


source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
...
+		for (i in 0 .. vectorLength) {
+			upperPart += ((this[i] - this mean) * (other[i] - other mean))
+			lowerPart += ((this[i] - this mean) * (this[i] - this mean))
+			lowerPart2 += ((other[i] - other mean) * (other[i] - other mean))
+		}
+		upperPart / sqrt(lowerPart * lowerPart2)
+	}
+	shift: func (offset: Int, filledValue := 0.0f) -> This {
+		result := This new(this count)
+		if (offset < 0) {
```
marcusnaslund: You don't need any brackets in this function.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
...
+		printf("calculating.")
+		for (offset in offsetStart .. offsetEnd) {
+			crossCorrelation add(this shift(offset) crossCorrelation(other, other count))
+			printf(".")
+		}
+		(index, maxValue) := crossCorrelation findMaxIndex()
+		println("finished")
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		result := CorrelationData new(index, maxValue, crossCorrelation)
```
marcusnaslund: Don't think you need the `result := ` at the start.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
...
+	}
+	//input: how many points there will be mroe between two origin points.
+	// this is a Linear interpolation way to have highier precision
+	higherPrecision: func (numberOfPoint: Int) -> This {
+		if (numberOfPoint != 0) {
+			newCount := this count + numberOfPoint * (this count - 1)
+			result := This new(newCount)
+			precision := (numberOfPoint + 1) as Float
+			for (ind in 0 .. newCount) {
```
marcusnaslund: These brackets are not needed.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
 			result add(this[i] clamp(floor, ceiling))
 		result
 	}
+	findMaxIndex: func -> (Int, Float) {
+		result := Float negativeInfinity
+		index := 0
+		for (i in 0 .. this count)
+			if (result < this[i]) {
```
marcusnaslund:
```ooc
index := 0
for (i in 1 .. this count)
    if (this[index] < this[i])
        index = i
(index, this[index])

File: test/math/FloatVectorListTest.ooc [hunk 5]
```diff
@@ -294,6 +294,48 @@ FloatVectorListTest: class extends Fixture {
...
+			list add(7.0f)
+			tempLists0 := list shift(-2)
+			tempLists1 := list shift(2)
+			tempLists2 := list shift(0)
+
+			temp0 := list findCrossCorrelationOffset(tempLists0, -5, 5)
+			temp1 := list findCrossCorrelationOffset(tempLists1, -5, 5)
+			temp2 := list findCrossCorrelationOffset(tempLists2, -5, 5)
+
+			expect(temp0 index == -2)
```
marcusnaslund: `expect(temp0 index, is equal to(-2))` and the same for the next two lines.

test/math/FloatVectorListTest.ooc
```diff
@@ -294,6 +294,48 @@ FloatVectorListTest: class extends Fixture {
...
+			list := FloatVectorList new()
+			list add(2.0f)
+			list add(1.0f)
+			list add(6.0f)
+			list add(4.0f)
+			list add(7.0f)
+			numberOfPointsBetween := 1
+			tempLists := list higherPrecision(numberOfPointsBetween)
+
+			expect(tempLists count == (list count + numberOfPointsBetween * (list count - 1)))
```
marcusnaslund: `expect(tempLists count, is equal to(list count + numberOfPointsBetween * (list count - 1)))`

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
...
+			crossCorrelation add(this shift(offset) crossCorrelation(other, other count))
+			printf(".")
+		}
+		(index, maxValue) := crossCorrelation findMaxIndex()
+		println("finished")
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		result := CorrelationData new(index, maxValue, crossCorrelation)
+	}
+	//input: how many points there will be mroe between two origin points.
```
marcusnaslund: no `mroe`

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
...
+	// this is a Linear interpolation way to have highier precision
+	higherPrecision: func (numberOfPoint: Int) -> This {
```
marcusnaslund: `numberOfPoints`

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
...
+			printf(".")
+		}
+		(index, maxValue) := crossCorrelation findMaxIndex()
+		println("finished")
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		result := CorrelationData new(index, maxValue, crossCorrelation)
+	}
+	//input: how many points there will be mroe between two origin points.
+	// this is a Linear interpolation way to have highier precision
```
marcusnaslund: `higher`

source/math/FloatVectorList.ooc
marcusnaslund: Also, `index`, not `ind`

source/math/FloatVectorList.ooc
ziquanyu: this is better, thanks

source/math/FloatVectorList.ooc
ziquanyu: that's true

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,88 @@ FloatVectorList: class extends VectorList<Float> {
...
+					result add(filledValue)
+				else
+					result add(this[i - offset])
+			}
+		}
+		result
+	}
+	// offset < 0 means it shifts to left
+	findCrossCorrelationOffset: func (other: This, offsetStart, offsetEnd: Int) -> CorrelationData {
+		println("There are " + this count toString() + " points, please wait for calculating...")
```
marcusnaslund: Why are we printing stuff here?

test/math/FloatVectorListTest.ooc
ziquanyu: sure

source/math/FloatVectorList.ooc
ziquanyu: This part takes long time if there are too many points, I thought it's better to let user know it's still running well.

marcusnaslund: @davidhesselbom ?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,80 @@ FloatVectorList: class extends VectorList<Float> {
...
+			if (this[index] < this[i])
+				index = i
+		(index, this[index])
+	}
+	crossCorrelation: func (other: This, vectorLength: Int) -> Float {
+		upperPart := 0.0f
+		lowerPart := 0.0f
+		lowerPart2 := 0.0f
+		for (i in 0 .. vectorLength) {
+			upperPart += ((this[i] - this mean) * (other[i] - other mean))
```
davidhesselbom: `this mean` and `other mean` will be calculated every time you call them. Call these properties once each, before the loop, and store them in variables.

source/math/FloatVectorList.ooc
ziquanyu: why do i have this in ci?
SvgPlotTest [TODO: Not implemented as a fixture!??

source/math/FloatVectorList.ooc
ziquanyu: okay

davidhesselbom: `this[i]` causes an expensive memcpy operation. Before starting the loop, create the following variables:
```ooc
thisPointer := (this pointer as Float*)
otherPointer := (other pointer as Float*)
```
and access those instead:
`(thisPointer[i] - thisMean) * (otherPointer[i] - otherMean)`
etc. This should improve your performance enough to make your printouts below unnecessary.

source/math/FloatVectorList.ooc
davidhesselbom: Let's see what the optimizations I suggested can do about the performance. If these printouts are still deemed necessary, they should at the very least be `Debug print`.

davidhesselbom: > SvgPlotTest [TODO: Not implemented as a fixture!??

To remind us to reimplement the tests as fixtures. The idea is that someone will get so tired of seeing those messages that they eventually get around to doing it :hammer:

ziquanyu: It's much much faster now. Thanks!

davidhesselbom: Great, I'm glad my eagerness to optimize things finally helped someone :)

ziquanyu: :+1: :+1:

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,80 @@ FloatVectorList: class extends VectorList<Float> {
 			result add(this[i] clamp(floor, ceiling))
 		result
 	}
+	findMaxIndex: func -> (Int, Float) {
```
marcusnaslund: I would prefer if this only returned the index, as the name suggests

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,80 @@ FloatVectorList: class extends VectorList<Float> {
...
+			lowerPart2 += ((otherPointer[i] - otherMean) * (otherPointer[i] - otherMean))
+		}
+		upperPart / sqrt(lowerPart * lowerPart2)
+	}
+	shift: func (offset: Int, filledValue := 0.0f) -> This {
+		result := This new(this count)
+		if (offset < 0)
+			for (i in 0 .. this count)
+				if ((i - offset) <= this count)
+					result add(this[i - offset])
```
marcusnaslund: Do you need the performance of `thisPointer` as in the above function?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,80 @@ FloatVectorList: class extends VectorList<Float> {
...
+	//input: how many points there will be between two origin points.
+	// this is a Linear interpolation way to have higher precision
+	higherPrecision: func (numberOfPoints: Int) -> This {
+		if (numberOfPoints != 0) {
+			newCount := this count + numberOfPoints * (this count - 1)
+			result := This new(newCount)
+			precision := (numberOfPoints + 1) as Float
+			for (index in 0 .. newCount)
+				if (Float modulo(index as Float, precision) == 0.0f)
+					result add(this[index / precision])
```
marcusnaslund: Again, will you need `thisPointer` or not here?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,80 @@ FloatVectorList: class extends VectorList<Float> {
...
+				if (i < offset)
+					result add(filledValue)
+				else
+					result add(this[i - offset])
+		result
+	}
+	// offset < 0 means it shifts to left
+	findCrossCorrelationOffset: func (other: This, offsetStart, offsetEnd: Int) -> CorrelationData {
+		crossCorrelation := This new(this count)
+		for (offset in offsetStart .. offsetEnd) {
```
marcusnaslund: You don't need these brackets

source/math/FloatVectorList.ooc
ziquanyu: There is property "maxValue" that returns value, but it doesn't index. I was wondering if it's possible to return both. What do you think?

source/math/FloatVectorList.ooc
ziquanyu: you mean use pointer instead for this[i]?

source/math/FloatVectorList.ooc
marcusnaslund: Sure, but `findMaxIndex()` claims it is just returning an index.
It seems just as easy to do
```ooc
maxIndex := myList findMaxIndex()
maxValue := myList[index]
```

source/math/FloatVectorList.ooc
marcusnaslund: Yeah, if you need the performance.

ziquanyu: yes, thisPointer make the process very fast.

marcusnaslund: Then please change :)

source/math/FloatVectorList.ooc
ziquanyu: sure, just want to drag attention about what I noticed:)
will fix it.

test/math/FloatVectorListTest.ooc
```diff
@@ -294,6 +294,54 @@ FloatVectorListTest: class extends Fixture {
 			list free()
 			filtered free()
 		})
+		this add("higher Precision", func {
+			list := FloatVectorList new()
+			list add(2.0f)
```
marcusnaslund: `list add(2.0f) . add(1.0f) . add(6.0f) . add(4.0f) . add(7.0f)` works too.

ziquanyu: should i change all in the test file?

marcusnaslund: Nah, leave it

source/math/FloatVectorList.ooc
marcusnaslund: Will you change this also?

test/math/FloatVectorListTest.ooc
ziquanyu: okay

source/math/FloatVectorList.ooc
ziquanyu: have changed already.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,82 @@ FloatVectorList: class extends VectorList<Float> {
...
+	// this is a Linear interpolation way to have higher precision
+	higherPrecision: func (numberOfPoints: Int) -> This {
+		if (numberOfPoints != 0) {
+			newCount := this count + numberOfPoints * (this count - 1)
+			result := This new(newCount)
+			precision := (numberOfPoints + 1) as Float
+			thisPointer := (this pointer as Float*)
+			for (index in 0 .. newCount)
+				if (Float modulo(index as Float, precision) == 0.0f)
+					result add(this[index / precision])
```
marcusnaslund: thisPointer

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,82 @@ FloatVectorList: class extends VectorList<Float> {
...
+				else {
+					ratio := Float modulo(index as Float, precision) / precision
+					difference := thisPointer[index / precision + 1] - thisPointer[index / precision]
```
marcusnaslund: `index / precision` is not an integer, that's why `CircleCI` fails.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,82 @@ FloatVectorList: class extends VectorList<Float> {
...
+		index := 0
+		for (i in 1 .. this count)
+			if (this[index] < this[i])
+				index = i
+		index
+	}
+	crossCorrelation: func (other: This, vectorLength: Int) -> Float {
+		upperPart := 0.0f
+		lowerPart := 0.0f
+		lowerPart2 := 0.0f
```
emilwestergren: Why `lowerPart2`? Isn't there a more descriptive name for this?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,82 @@ FloatVectorList: class extends VectorList<Float> {
...
+		thisMean := this mean
+		otherMean := other mean
+		thisPointer := (this pointer as Float*)
+		otherPointer := (other pointer as Float*)
+		for (i in 0 .. vectorLength) {
+			upperPart += ((thisPointer[i] - thisMean) * (otherPointer[i] - otherMean))
```
emilwestergren: I would do something like this:
```ooc
thisValue := thisPointer[i]
otherValue := otherPointer[i]
lowerPart += ((thisValue - thisMean) * (thisValue - thisMean))
lowerPart2 += ((otherValue - otherMean) * (otherValue - otherMean))
```

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,82 @@ FloatVectorList: class extends VectorList<Float> {
...
+			lowerPart += ((thisPointer[i] - thisMean) * (thisPointer[i] - thisMean))
+			lowerPart2 += ((otherPointer[i] - otherMean) * (otherPointer[i] - otherMean))
+		}
+		upperPart / sqrt(lowerPart * lowerPart2)
+	}
+	shift: func (offset: Int, filledValue := 0.0f) -> This {
+		result := This new(this count)
+		if (offset < 0)
+			for (i in 0 .. this count)
+				if ((i - offset) <= this count)
```
emilwestergren:
```ooc
result add((i - offset) <= this count ? this[i - offset] : filledValue)
```

source/math/FloatVectorList.ooc
ziquanyu: have fixed it, yet have some problem about git. i will make a new PR instead.

source/math/FloatVectorList.ooc
davidhesselbom: This reads clearly enough to me.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,85 @@ FloatVectorList: class extends VectorList<Float> {
...
+		thisPointer := (this pointer as Float*)
+		otherPointer := (other pointer as Float*)
+		for (i in 0 .. vectorLength) {
+			upperPart += ((thisPointer[i] - thisMean) * (otherPointer[i] - otherMean))
+			lowerPart1 += ((thisPointer[i] - thisMean) * (thisPointer[i] - thisMean))
+			lowerPart2 += ((otherPointer[i] - otherMean) * (otherPointer[i] - otherMean))
+		}
+		upperPart / sqrt(lowerPart1 * lowerPart2)
+	}
+	shift: func (offset: Int, filledValue := 0.0f) -> This {
```
emilwestergren: Should be simplified to something like this:
```ooc
	shift: func (offset: Int, filledValue := 0.0f) -> This {
		result := This new(this count)
		thisPointer := (this pointer as Float*)
		for (i in 0 .. this count) {
			index := i + offset
			value := (index >= 0 && index < this count) ? thisPointer[index] : filledValue
			result add(value)
		}
		result
	}
```

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,85 @@ FloatVectorList: class extends VectorList<Float> {
...
+	higherPrecision: func (numberOfPoints: Int) -> This {
+		if (numberOfPoints != 0) {
+			newCount := this count + numberOfPoints * (this count - 1)
+			result := This new(newCount)
+			precision := (numberOfPoints + 1) as Float
+			thisPointer := (this pointer as Float*)
+			for (index in 0 .. newCount) {
+				newIndex := (index / precision) as Int
+				if (Float modulo(index as Float, precision) == 0.0f)
+					result add(this[newIndex])
```
marcusnaslund: use `thisPointer` here as well

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,85 @@ FloatVectorList: class extends VectorList<Float> {
 			result add(this[i] clamp(floor, ceiling))
 		result
 	}
+	findMaxIndex: func -> Int {
```
emilwestergren: Why MaxIndex and not MaxValue?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,85 @@ FloatVectorList: class extends VectorList<Float> {
...
+	}
+	//input: how many points there will be between two origin points.
+	// this is a Linear interpolation way to have higher precision
+	higherPrecision: func (numberOfPoints: Int) -> This {
+		if (numberOfPoints != 0) {
+			newCount := this count + numberOfPoints * (this count - 1)
+			result := This new(newCount)
+			precision := (numberOfPoints + 1) as Float
+			thisPointer := (this pointer as Float*)
+			for (index in 0 .. newCount) {
```
tomasmahlberg: This whole loop could be simplified to something like
```ooc
for (index1 in 0 .. this count - 1) {
	result add(this[index1])
	for (index2 in 1 .. numberOfPoints + 1)
		result add(this[index1] + (this[index1 + 1] - this[index1]) * index2 / (numberOfPoints + 1))
}
result add(l[this count - 1])
```
Of course with some added optimizations.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,85 @@ FloatVectorList: class extends VectorList<Float> {
...
+
+		index := crossCorrelation findMaxIndex()
+		maxValue := this[index]
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		CorrelationData new(index, maxValue, crossCorrelation)
+	}
+	//input: how many points there will be between two origin points.
+	// this is a Linear interpolation way to have higher precision
+	higherPrecision: func (numberOfPoints: Int) -> This {
```
tomasmahlberg: Maybe `interpolate` would be a more descriptive name for this function?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,85 @@ FloatVectorList: class extends VectorList<Float> {
...
+					ratio := Float modulo(index as Float, precision) / precision
+					difference := thisPointer[newIndex + 1] - thisPointer[newIndex]
+					tempResult := thisPointer[newIndex] + ratio * difference
+					result add(tempResult)
+				}
+			}
+			result
+		}
+		else
+			this
```
tomasmahlberg: I think it's advised to not have more than one exit point in a function.

source/math/FloatVectorList.ooc
emilwestergren: I agree.

source/math/FloatVectorList.ooc
davidhesselbom: Correct.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,85 @@ FloatVectorList: class extends VectorList<Float> {
...
+		index := crossCorrelation findMaxIndex()
+		maxValue := this[index]
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		CorrelationData new(index, maxValue, crossCorrelation)
+	}
+	//input: how many points there will be between two origin points.
+	// this is a Linear interpolation way to have higher precision
+	higherPrecision: func (numberOfPoints: Int) -> This {
+		if (numberOfPoints != 0) {
```
emilwestergren: so `-5` is ok?

source/math/FloatVectorList.ooc
ziquanyu: it's changed, but why it's still here?

source/math/FloatVectorList.ooc
davidhesselbom: Like, using `thisPointer` instead of `this`.

ziquanyu: @davidhesselbom do we need to change this to thisPointer for the whole class?

source/math/FloatVectorList.ooc
sebastianbaginski: You should be able to use `memcpy` for the overlapping part

ziquanyu: I started with memcpy, yet I must have done something wrong. Then I made it like this so as to save time. Because optimization can be done after the function is working.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,79 @@ FloatVectorList: class extends VectorList<Float> {
...
+		for (offset in offsetStart .. offsetEnd)
+			crossCorrelation add(this shift(offset) crossCorrelation(other, other count))
+
+		index := crossCorrelation findMaxIndex()
+		maxValue := crossCorrelation[index]
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		CorrelationData new(index, maxValue, crossCorrelation)
+	}
+	//input: how many points there will be between two origin points.
```
davidhesselbom: There should be a space after the `//`

source/math/FloatVectorList.ooc
marcusnaslund: Certainly not in this PR.

ziquanyu: It will be in a separate PR for sure. I'm just wondering if we need to do that or not.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,79 @@ FloatVectorList: class extends VectorList<Float> {
...
+				result add(thisPointer[index1])
+				for (index2 in 1 .. numberOfPoints + 1)
+					result add(thisPointer[index1] + (thisPointer[index1 + 1] - thisPointer[index1]) * index2 as Float / (numberOfPoints + 1) as Float)
+			}
+			result add(thisPointer[this count - 1])
+		}
+		else {
+			if (numberOfPoints < 0)
+				println("numberOfPoints should be greater than or equal to 0, the original list is returned!")
+			this
```
marcusnaslund: You should not return in two places (here and line 445).

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,78 @@ FloatVectorList: class extends VectorList<Float> {
...
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		CorrelationData new(index, maxValue, crossCorrelation)
+	}
+	// input: how many points there will be between two origin points.
+	// this is a Linear interpolation way to have higher precision
+	interpolate: func (numberOfPoints: Int) -> This {
+		result := this
+		if (numberOfPoints < 0)
+			println("numberOfPoints should be greater than or equal to 0, the original list is returned!")
```
tomasmahlberg: Why is this printout necessary? Haven't really seen anything similar anywhere else.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,78 @@ FloatVectorList: class extends VectorList<Float> {
...
+		if (numberOfPoints > 0) {
```
tomasmahlberg: Maybe this should be `numberOfPoints > 0 && this count > 1`?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,78 @@ FloatVectorList: class extends VectorList<Float> {
...
+			newCount := this count + numberOfPoints * (this count - 1)
```
tomasmahlberg: No need to assign this to a variable, right? Just put it inside `new()` below.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,78 @@ FloatVectorList: class extends VectorList<Float> {
...
+			result = This new(newCount)
+			precision := (numberOfPoints + 1) as Float
```
tomasmahlberg: This is not used anywhere.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +381,78 @@ FloatVectorList: class extends VectorList<Float> {
...
+			thisPointer := (this pointer as Float*)
+			for (index1 in 0 .. this count - 1) {
+				result add(thisPointer[index1])
+				for (index2 in 1 .. numberOfPoints + 1)
+					result add(thisPointer[index1] + (thisPointer[index1 + 1] - thisPointer[index1]) * index2 as Float / (numberOfPoints + 1) as Float)
```
tomasmahlberg: I'm pretty sure these explicit `Float` casts are unnecessary.

source/math/FloatVectorList.ooc
niklasborwell: Yes, it should either be an exception or simply ignored. We don't use prints in the code like this, especially since people wouldn't know from the message where it was being printed. In this case I would just remove this line.

emilwestergren: If this is necessary, I'd rather do something like this:
```ooc
if (numberOfPoints > 0) { ... }
else
   Debug raise("numberOfPoints should be greather than 0 in FloatVectorList interpolate!")
```

source/math/FloatVectorList.ooc
davidhesselbom: Depends. If `index2 / (numberOfPoints + 1)` is evaluated before the rest of the expression, it will result in an `Int` division. I still don't trust rock do to the right thing here.

davidhesselbom: The second cast *is* unnecessary, though.

tomasmahlberg: Ah, alright. But `index2 / (numberOfPoints + 1.f)` would work though, right?

davidhesselbom: No, rock uses the type of the left operand, unlike e.g. C. In C, `Int / Float -> Float` and `Float / Int -> Float`, but in ooc (sadly) `Int / Float -> Int` and `Float / Int -> Float`.

tomasmahlberg: Good to know!

source/math/FloatVectorList.ooc
ziquanyu: removed

source/math/FloatVectorList.ooc
marcusnaslund: Is this a bug or a feature?

ziquanyu: I tested, it is not necessary. but it wasn't like this before.
I removed them anyway.

ziquanyu: I made a test like this. I got 0.5 from both??
```ooc
x := 1.0f / 2
y := 1 / 2.0f
println("x: " + x toString())
println("y: " + y toString())
```

davidhesselbom: Alright, maybe someone fixed it, then?

ziquanyu: It's good to know it's working as C now.

niklasborwell: I don't think it was changed, I just think it's just poorly coded, or rather inconsistently coded. Unless you tested this very thoroughly, I just made some quick assumptions based on a few cases. I think it's better if we are overly explicit when it comes to this. It could also potentially save us some headache in the future if it did get fixed.

davidhesselbom: Agreed, I don't trust it to have been fixed completely until I find a commit in rock that mentions it explicitly.

ziquanyu: @davidhesselbom  what should I do? Cast Int to Float in case something goes wrong with rock. Or leave it like this and solve the future problems in the future?!

test/math/FloatVectorListTest.ooc
```diff
@@ -293,6 +293,54 @@ FloatVectorListTest: class extends Fixture {
...
+			list add(4.0f)
+			list add(7.0f)
+			list add(4.0f)
+			list add(7.0f)
+			tempLists0 := list shift(2)
+			tempLists1 := list shift(-1)
+			tempLists2 := list shift(0)
+
+			temp0 := list findCrossCorrelationOffset(tempLists0, -3, 3)
+			temp1 := list findCrossCorrelationOffset(tempLists1, -3, 3)
```
niklasborwell: tempLists should probably be called shiftedList or similar, and temp0 called correlationOffset or something similar. A name that implies what the variable contains, so that someone reading this doesn't have to read the previous code to understand what happens later. Same with the first test.

ziquanyu: Right, they've been changed.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,76 @@ FloatVectorList: class extends VectorList<Float> {
...
+		}
+		result
+	}
+}
+CorrelationData: cover {
+	index: Int
+	maxValue: Float
+	crossCorrelation: FloatVectorList
+	init: func@ (=index, =maxValue, =crossCorrelation)
 }
```
niklasborwell: Sorry if I'm bringing up some old conversation here, but what is the purpose of this cover? What does index mean? Is it something obvious that I should understand?

davidhesselbom: It could have been a tuple...

marcusnaslund: should*?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,76 @@ FloatVectorList: class extends VectorList<Float> {
...
+	findCrossCorrelationOffset: func (other: This, offsetStart, offsetEnd: Int) -> CorrelationData {
+		crossCorrelation := This new(this count)
+		for (offset in offsetStart .. offsetEnd)
+			crossCorrelation add(this shift(offset) crossCorrelation(other, other count))
+
+		index := crossCorrelation findMaxIndex()
+		maxValue := crossCorrelation[index]
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+
+		CorrelationData new(index, maxValue, crossCorrelation)
```
davidhesselbom: Just `(index, maxValue, crossCorrelation)` works fine. Then the function returns not a `CorrelationData` but `(Int, Float, This)`.

source/math/FloatVectorList.ooc
ziquanyu: I could not return more than 2 variables, that's the reason to have this cover. But I think it's possible to optimize this part and take away this cover.

source/math/FloatVectorList.ooc
niklasborwell: If I understand correctly, `index` seems to be the index of the maximum value, which then of course gives us `maxValue`. But FloatVectorList already has functions to calculate these things, so I don't see the need to return anything but the FloatVectorList itself. If I understood it correctly.

source/math/FloatVectorList.ooc
davidhesselbom: What do you mean you can't return more than 2 variables? `RasterYuvPlanar _allocate` returns 3.

source/math/FloatVectorList.ooc
davidhesselbom: You're right, it seems `findMaxIndex` is only used here to get the maximum value. Does it have other uses?

niklasborwell: We discussed a bit IRL and realized that maybe we don't have to return the vectorlist at all here, since we are actually trying to calculate the offset.

source/math/FloatVectorList.ooc
ziquanyu: Alright, I will remove it and change it a bit anyway.

source/math/FloatVectorList.ooc
davidhesselbom: That would agree more with the function's name, at least...

ziquanyu: @davidhesselbom `indMaxIndex` will return maxValue's index, because we do not have function to return maxValue's index.

ziquanyu: @niklasborwell  What if someone wants to know how good / bad correlation these two signals have. If we do not show the plot of two signals or plot doesn't say much. We need numbers to show it's 40% correlated or 90% correlated. That's the reason why this `findCrossCorrelationOffset` returns `index` and `maxValue of CrosCorrelation`  and crosscorrelation of each shift. The last one is actually not necessary if we return the maxValue instead.

ziquanyu: This is a debug tool, it's better to return more information than less, I think.

niklasborwell: Well, the issue is that
a) The function says calculateCorrelationOffset, but it returns a struct that contains more than that. If we want to return more information it should be called something else, like calculateCorrelationData.
b) If we want to return all the information, actually all we need to return is the crossCorrelation vectorlist itself, because it already has information about the index and max value embedded.

niklasborwell: And if we only need the index and maxValue, we can return a tuple here instead.

ziquanyu: You are totally right about this. I was thinking about this name "calculateCorrelationData" too.

The reason why I did not return vectorlist is this
 `index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)`
Because index for vectorlist doesn't have minus index, and the offsetIndex can be anything, which means when we use this function we must remember this otherwise the index function returns will mean nothing.

ziquanyu: I returned verctolist at the end for debug plot purpose.

ziquanyu: @niklasborwell 
1, let's say if offsetStart is -3,  offsetEnd is 3. 
2, let's it's calculate auto correlation. 
3, index (correlationOffset) = 0 based on (-3, 3). But index to the vectorlist is 4. 
4, we either calculate once in the function and return the maxValue or do it twice outside function.

tomasmahlberg: I think I accidentally brought this discussion to another PR somewhere else, sorry. Was kind of hard to follow what happened here...

My suggestion would be to have one function that returns two `FloatVectorList`s: The calculated cross-correlations and the corresponding lags, and one function that finds the lag with the maximum correlation.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,69 @@ FloatVectorList: class extends VectorList<Float> {
...
+		maxValue := crossCorrelation[index]
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+		(index, maxValue, crossCorrelation)
+	}
+	// input: how many points there will be between two origin points.
+	// this is a Linear interpolation way to have higher precision
+	interpolate: func (numberOfPoints: Int) -> This {
+		result := this
+		if (numberOfPoints < 0)
+			Debug raise("numberOfPoints should be greather than 0 in FloatVectorList interpolate!")
```
niklasborwell: Actually, I think this check should be removed. We don't check parameters on the way in in functions.

davidhesselbom: It could go in a `safe` block...

niklasborwell: Good idea.

ziquanyu: :+1:  It will be UInt, i will adapt the rest.

davidhesselbom: Okay, so what happens if someone tries to call the function with a negative `Int`? Will it refuse to build, or will it just behave strangely?

marcusnaslund: The `Int` will be interpreted as a `UInt` and will become a very large number instead.

sebastianbaginski: Good compiler should warn about it, but I guess `rock` won't and the negative number will be simply cast to `UInt` :)

source/math/FloatVectorList.ooc
tomasmahlberg: I just realized there is a `Float linearInterpolation` function. It would probably be advised to replace
```ooc
thisPointer[index1] + (thisPointer[index1 + 1] - thisPointer[index1]) * index2 as Float / (numberOfPoints + 1) as Float
```
with
```ooc
Float linearInterpolation(thisPointer[index1], thisPointer[index1 + 1], index2 as Float / (numberOfPoints + 1) as Float)
```
Sorry for misleading you earlier...

source/math/FloatVectorList.ooc
marcusnaslund: You're right @sebastianbaginski - a good compiler would. Hence rock won't. :)

sebastianbaginski: Yes, builds without a single warning:
```
testuint: func (value: UInt) {
	value toString() println()
}

testuint(-1)

```
output:
> 4294967295

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,69 @@ FloatVectorList: class extends VectorList<Float> {
...
+	}
+	// offset < 0 means it shifts to left
+	calculateCorrelationData: func (other: This, offsetStart, offsetEnd: Int) -> (Int, Float, This) {
+		crossCorrelation := This new(this count)
+		for (offset in offsetStart .. offsetEnd)
+			crossCorrelation add(this shift(offset) crossCorrelation(other, other count))
+		index := crossCorrelation findMaxIndex()
+		maxValue := crossCorrelation[index]
+		index = (index < (offsetEnd - offsetStart) / 2) ? (index + offsetStart) : (index - offsetEnd)
+		(index, maxValue, crossCorrelation)
```
niklasborwell: I'm bringing this up again, sorry...
First of all, maybe index should be called correlationOffset instead, if we keep it. Index is kind of general.
But I still don't think we need to return index here. You calculate index with `crossCorrelation findMaxIndex()`, and then you return both `index` and `crossCorrelation`. It's not necessary, the caller could just call `crossCorrelation findMaxIndex()` themselves. And in the same way, after that, the caller can call `crossCorrelation[index]` to get `maxValue`. So the same information is basically returned twice here, which I think we should avoid.

source/math/FloatVectorList.ooc
davidhesselbom: Now I remember why we aren't using `UInt` more...

source/math/FloatVectorList.ooc
niklasborwell: We had a discussion IRL and decided to change a bit according to Tomas idea.

source/math/FloatVectorList.ooc
ziquanyu: Don't worry about it. :)

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,68 @@ FloatVectorList: class extends VectorList<Float> {
...
 	}
+	findMaxIndex: func -> Int {
+		index := 0
+		thisPointer := (this pointer as Float*)
+		for (i in 1 .. this count)
+			if (thisPointer[index] < thisPointer[i])
+				index = i
+		index
+	}
+	crossCorrelation: func (other: This, vectorLength: Int) -> Float {
```
tomasmahlberg: This function should be called just `correlation`.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,68 @@ FloatVectorList: class extends VectorList<Float> {
...
+		thisPointer := (this pointer as Float*)
+		for (i in 0 .. this count) {
+			index := i + offset
+			value := (index >= 0 && index < this count) ? thisPointer[index] : filledValue
+			result add(value)
+		}
+		result
+	}
+	// offset < 0 means it shifts to left
+	calculateCorrelation: func (other: This, offsetRange: Range, calculateOffset := true) -> (This, This) {
```
tomasmahlberg: This should be called `crossCorrelation`.

tomasmahlberg: `calculateOffset`should be renamed to `calculateOffsets`.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,68 @@ FloatVectorList: class extends VectorList<Float> {
...
+		crossCorrelation := This new(offsetRange count)
+		for (offset in offsetRange min .. offsetRange max)
```
tomasmahlberg: `offsetRange min .. offsetRange max` => `offsetRange min .. offsetRange max + 1`.

source/math/FloatVectorList.ooc
davidhesselbom: Is there a difference between the terms?

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,68 @@ FloatVectorList: class extends VectorList<Float> {
...
+			crossCorrelation add(this shift(offset) crossCorrelation(other, other count))
+		xValue := This new()
+		xValue add(offsetRange min as Float) . add(offsetRange max as Float)
+		(crossCorrelation, xValue interpolate((offsetRange max - offsetRange min) - 1))
```
tomasmahlberg: I would rename `xValue` to `offsets` and you should make the function return `null` for the second tuple value instead of `offsets` if `calculateOffsets` is `false`.

source/math/FloatVectorList.ooc
ziquanyu: I forgot the if check. :P

source/math/FloatVectorList.ooc
tomasmahlberg: Correlation is a measure of similarity between two data sets and is a scalar. Cross-correlation is a set of correlations where one of the data sets is shifted with different lags.

source/math/FloatVectorList.ooc
ziquanyu: two exits?

source/math/FloatVectorList.ooc
davidhesselbom: Ranges can be created like this: `range := minimum .. maximum` so it's a real shame you can't just do this `for (offset in offsetRange)`...

source/math/FloatVectorList.ooc
tomasmahlberg: Actually, I'm not sure if correlation is the scientifically correct term here... but it definitely isn't cross-correlation.

source/math/FloatVectorList.ooc
tomasmahlberg: No, never.

source/math/FloatVectorList.ooc
ziquanyu: it's good to know it can be created like that.

marcusnaslund: `for (offset in offsetRange)` would leak memory like similar iterators, no?

source/math/FloatVectorList.ooc
niklasborwell: Either way, I say remove this check altogether. There is no reason why people would try to call this function with a negative number.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,69 @@ FloatVectorList: class extends VectorList<Float> {
...
+		thisPointer := (this pointer as Float*)
+		for (i in 0 .. this count) {
+			index := i + offset
+			value := (index >= 0 && index < this count) ? thisPointer[index] : filledValue
+			result add(value)
+		}
+		result
+	}
+	// offset < 0 means it shifts to left
+	calculateCrossCorrelation: func (other: This, offsetRange: Range, calculateOffsets := true) -> (This, This) {
```
tomasmahlberg: I still think a better name for this function is just `crossCorrelation`.

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,69 @@ FloatVectorList: class extends VectorList<Float> {
...
+		crossCorrelation := This new(offsetRange count)
+		for (offset in offsetRange min .. offsetRange max + 1)
+			crossCorrelation add(this shift(offset) correlation(other, other count))
+		offsets := This new()
+		if (calculateOffsets) {
+			offsets add(offsetRange min as Float) . add(offsetRange max as Float)
+			offsets = offsets interpolate((offsetRange max - offsetRange min) - 1)
```
tomasmahlberg: This will probably leak since you overwrite `offsets` with new data without freeing its original content.

source/math/FloatVectorList.ooc
sebastianbaginski: @tomasmahlberg https://github.com/vidhance/ooc-vidproc/pull/933#discussion_r44663122 :)

source/math/FloatVectorList.ooc
niklasborwell: I would suggest running memcheck to see where there are memory leaks.

source/math/FloatVectorList.ooc
niklasborwell: I can't see Sebastian's comment here for some reason, but I got it in my mail. Anyway he is right, functions to should be verbs so calculateCrossCorrelation is more correct here.

sebastianbaginski: Sorry I just deleted it, as it contained invalid link.

tomasmahlberg: In that case, I have no objections! Sorry, I'm still new here. :)

source/math/FloatVectorList.ooc
tomasmahlberg: This is one of two spots where valgrind reports leaks when calling `calculateCrossCorrelation`. Maybe change this block to:
```ooc
offsets: This = null
if (calculateOffsets) {
	temporary := This new() . add(offsetRange min as Float) . add(offsetRange max as Float)
	offsets = temporary interpolate((offsetRange max - offsetRange min) - 1)
	temporary free()
}
```

source/math/FloatVectorList.ooc
```diff
@@ -381,4 +382,69 @@ FloatVectorList: class extends VectorList<Float> {
...
+			value := (index >= 0 && index < this count) ? thisPointer[index] : filledValue
+			result add(value)
+		}
+		result
+	}
+	// offset < 0 means it shifts to left
+	calculateCrossCorrelation: func (other: This, offsetRange: Range, calculateOffsets := true) -> (This, This) {
+		crossCorrelation := This new(offsetRange count)
+		for (offset in offsetRange min .. offsetRange max + 1)
+			crossCorrelation add(this shift(offset) correlation(other, other count))
```
tomasmahlberg: This leaks. the result of `this shift(offset)` needs to be put in a temporary variable and freed after it's been used. Other than that, I think it's time for the "real" review to begin! :+1:

source/math/FloatVectorList.ooc
davidhesselbom: lol, `temporary` is a `FloatVectorList` with 2 elements that you call `interpolate` on?

tomasmahlberg: `offsets` could definitely be populated inside the for loop a few lines up instead...

source/math/FloatVectorList.ooc
ziquanyu: Does that mean if we use it in this way, it will leak?  If so, why does it support this type of function call?

davidhesselbom: ooc was designed with a garbage collector. We turned the garbage collector off, and then everything started leaking, everywhere. It's not like the compiler will tell us about it...

source/math/FloatVectorList.ooc
ziquanyu: Thanks for the suggestions anyway!

source/math/FloatVectorList.ooc
ziquanyu: How can we avoid it? Can we ask compiler to tell us about this information?

davidhesselbom: I guess the compiler should be able to tell us when we create things and lose track of them in the same scope, but it would never be reliable enough. You'll have to keep using Valgrind, I'm afraid.

ziquanyu: It's good to know. I will be careful about it. Thanks!

ziquanyu:
https://github.com/cogneco/ooc-kean/issues/771

niklasborwell:
You should try to re-push this to see of you can get the appveyor test to work.

marcusnaslund:
We're not using appveyor anymore, so don't bother with it.

davidhesselbom:
But if you do re-push, the appveyor status should go away.

niklasborwell:
That was kind of my point too. I felt like no one was looking at this because of the red cross over here.

davidhesselbom:
I wasn't looking because I thought you were still discussing something. Are we done with the `crossCorrelation` thing yet, or is there something more left to fix? I have other PRs to merge, so if you still need to fix something, this is your chance.

tomasmahlberg:
I had some comments a few hours ago, maybe after @ziquanyu has had the time to look at those it's about ready.

davidhesselbom:
When you're done, mark it "ready" and "peer reviewed".

ziquanyu:
Do we have tag which is "ready"?

davidhesselbom:
That's the one. Thanks.

source/math/FloatVectorList.ooc
```diff
@@ -382,4 +383,68 @@ FloatVectorList: class extends VectorList<Float> {
 			result add(this[i] clamp(floor, ceiling))
 		result
 	}
+	findMaxIndex: func -> Int {
```
davidhesselbom: `findMaximumIndex`. Although it's easy to assume this actually returns `this count - 1`... So `findIndexOfMaximum` would be more specific.

ziquanyu: will change it.

source/math/FloatVectorList.ooc
```diff
@@ -382,4 +383,68 @@ FloatVectorList: class extends VectorList<Float> {
...
+	// input: how many points there will be between two origin points.
+	// this is a Linear interpolation way to have higher precision
+	interpolate: func (numberOfPoints: Int) -> This {
+		if (numberOfPoints > 0 && this count > 1) {
+			result := This new(this count + numberOfPoints * (this count - 1))
+			thisPointer := (this pointer as Float*)
+			for (index1 in 0 .. this count - 1) {
+				result add(thisPointer[index1])
+				for (index2 in 1 .. numberOfPoints + 1)
+					result add(Float linearInterpolation(thisPointer[index1], thisPointer[index1 + 1], index2 as Float / (numberOfPoints + 1) as Float))
```
davidhesselbom: The last `as Float` should not be necessary.

source/math/FloatVectorList.ooc
```diff
@@ -382,4 +383,68 @@ FloatVectorList: class extends VectorList<Float> {
...
+			}
+			result add(thisPointer[this count - 1])
+		} else
+			result := this copy()
+	}
```
davidhesselbom: I don't see anything being returned anwhere...

davidhesselbom: @simonmika Again, I'm not sure whether a `result` variable declared outside the `if/else` is necessary here.

ziquanyu: line 448 and line 446 will be return at the end separately. It works well so far, I've checked the result.

source/math/FloatVectorList.ooc
ziquanyu: Can i keep it there just in case rock goes banana somehow.

source/math/FloatVectorList.ooc
marcusnaslund: > I'm not sure whether a result variable declared outside the if/else is necessary here.

Isn't that how we *always* do it?

davidhesselbom: @ziquanyu It's a matter of coding style, it doesn't affect the behavior.

PR #687, opened 2015-11-04
Only print Backtrace error messages once

marcusnaslund:
Implements updates found in https://github.com/fasterthanlime/rock/commit/19bfc01a7904676f49a9cbb2626e7803df6484fb to only *once* print the error message `"[lang/Backtrace] No backtrace extension nor execinfo - use a debugger!\n"`. 

Cleans up the output in Appveyor and CircleCI.

### DIFF
--- a/source/sdk/lang/Backtrace.ooc
+++ b/source/sdk/lang/Backtrace.ooc
@@ -12,6 +12,7 @@ BacktraceHandler: class {
 
     // constants
     BACKTRACE_LENGTH := static 128
+    WARNED_ABOUT_FALLBACK := static false
     
     // singleton
     instance: static This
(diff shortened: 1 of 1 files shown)


source/sdk/lang/Backtrace.ooc
```diff
@@ -52,13 +53,19 @@ BacktraceHandler: class {
         } else {
             // fall back on execinfo? still informative
             version (linux || apple) {
-                stderr write("[lang/Backtrace] Falling back on execinfo.. (build extension if you want fancy backtraces)\n")
+                if (!WARNED_ABOUT_FALLBACK) {
```
davidhesselbom: `!This WARNED_ABOUT_FALLBACK` this time around ,too, right?

source/sdk/lang/Backtrace.ooc
```diff
@@ -52,13 +53,19 @@ BacktraceHandler: class {
...
+                    stderr write("[lang/Backtrace] Falling back on execinfo.. (build extension if you want fancy backtraces)\n")
+                    This WARNED_ABOUT_FALLBACK = true
+                }
                 length := backtrace(buffer, BACKTRACE_LENGTH)
                 return Backtrace new(buffer, length)
             }
 
             // no such luck, use a debugger :(
-            stderr write("[lang/Backtrace] No backtrace extension nor execinfo - use a debugger!\n")
+            if (!WARNED_ABOUT_FALLBACK) {
```
davidhesselbom: And here

source/sdk/lang/Backtrace.ooc
sebastianbaginski: Btw. does it have to be `NAMED_LIKE_THIS` ? :)

davidhesselbom: You know, I have no idea. I figured it's because of some dependency in another library, but it should be easy to test (just make the program crash and see if the backtrace still works).

marcusnaslund: I'm pretty sure the name is arbitrary. However, I felt that we should strive to be as close to the original SDK as possible in files we're not messing around in (read: improving) yet.

PR #690, opened 2015-11-05
Cleaned code in shader classes.

emilwestergren:


source/draw/gpu/opengl/Map/OpenGLMap.ooc
```diff
@@ -47,10 +57,8 @@ OpenGLMap: abstract class extends GpuMap {
 		super()
 	}
 	use: override func {
-		currentIndex := this _context getCurrentIndex()
-		if (this _program[currentIndex] == null)
-			this _program[currentIndex] = this _context as OpenGLContext _backend createShaderProgram(this _vertexSource, this _fragmentSource)
-		this _program[currentIndex] use()
+		program := this program
+		program use()
```
davidhesselbom: Would `this program use()` work?

emilwestergren:
Updated.

PR #693, opened 2015-11-05
Greatly reduced memory leaks in HashDictionary/HashBag/HashMap.

emilwestergren:


emilwestergren:
HashBag still leaks.

emilwestergren:
Do you think there is anything of value here or should we just close it? I don't really care since I don't need HashDictionary anymore and we probably want to make a new implementation from scratch anyway.

marcusnaslund:
We do need `HashBag` for stuff in the sdk, so my plan is to fix it up even more and eventually move it out of the sdk along with its dependencies.
If this does improve things, I see no reason to throw it away.

emilwestergren:
Alright, I think at least it improves the code quality and fixes some memory leaks, but it definitely needs more work. You decide if you want to go on from here or start from scratch.

marcusnaslund:
That's good. Thanks for creating an issue on this btw.

marcusnaslund:
Also: You should change the flag on this PR if you're ready.

marcusnaslund:
I like how this PR is both high and low.

PR #696, opened 2015-11-05
Removed acceses out of range from Vector/Vector2D tests

tomasmahlberg:
Also fixed the range check in `Vector2D`.


sebastianbaginski:
Does it work if you run `./test.sh collections/Vector2D -Dsafe` and `./test.sh collections/Vector -Dsafe` now ?

tomasmahlberg:
Yes, it works now. But I also removed all places where we try to access an element out of range.

sebastianbaginski:
That's great, so it shouldn't crash on windows/appveyor now.

tomasmahlberg:
I hope so!

PR #699, opened 2015-11-05
Implemented the Quickhull algorithm for FloatConvexHull2D

### LINKED ISSUE #562: FloatConvexHull2D does not actually compute the convex hull
The constructor merely sets the given points (which are hopefully a convex set), and applying a transformation does not recompute the convex hull either, as it should.

There is also a limitaton mentioned that `// Points must be in clockwise order` - we might also want to take care of this.

marcusnaslund:
`FloatConvexHull2D` now actually computes the convex hull, if asked to do so in the constructor or by calling `computeHull()`.

I've set the default value to `false` to not accidentally impact performance anywhere, but I'll gladly change that if/when it's OK.

Fixes #562


source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,126 @@ import FloatTransform2D
...
+		}
+	}
 	init: func ~fromBox (box: FloatBox2D) {
 		this _points = VectorList<FloatPoint2D> new(4)
+		this _hull = VectorList<FloatPoint2D> new(4)
 		this _points add(box leftTop)
 		this _points add(box leftBottom)
 		this _points add(box rightBottom)
 		this _points add(box rightTop)
+		this _hull = this _points copy()
```
davidhesselbom: But... this will overwrite the `VectorList` you created on line 40 and cause a memory leak. Right?

source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,126 @@ import FloatTransform2D
...
 		result := true
-		for (i in 0 .. this count - 1)
-			if (This _isOnLeft(this _points[i], this _points[i + 1], point)) {
+		for (i in 0 .. this _hull count - 1)
+			if (result && This _isOnLeft(this _hull[i], this _hull[i + 1], point))
 				result = false
-				break
-			}
-		if (This _isOnLeft(this _points[this count - 1], this _points[0], point))
+		if (This _isOnLeft(this _hull[this _hull count - 1], this _hull[0], point))
```
davidhesselbom: Instead of the double `if`, why not loop from 0 to `this _hull count` and check the index `i % this _hull count`?

source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,126 @@ import FloatTransform2D
...
-	_isOnLeft: static func (leftPoint, rightPoint, queryPoint: FloatPoint2D) -> Bool {
-		(rightPoint x - leftPoint x) * (queryPoint y - leftPoint y) > (rightPoint y - leftPoint y) * (queryPoint x - leftPoint x)
+	computeHull: func {
+		if (this _hull)
+			this _hull clear()
+		else
+			this _hull = VectorList<FloatPoint2D> new()
+		
+		if (this _points count <= 2)
+			this _hull = this _points copy()
```
davidhesselbom: Here, too, `this _hull` will be overwritten, which will cause a memory leak.

source/math/FloatConvexHull2D.ooc
marcusnaslund: Using `%` on every iteration is much less efficient. Isn't this clear enough?

source/math/FloatConvexHull2D.ooc
marcusnaslund: Right, will fix both.

source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,126 @@ import FloatTransform2D
...
+		else {
+			// Uses the Quickhull algorithm if more than two points, average complexity O(n log n)
```
davidhesselbom: If performance is a concern (considering you added the expected complexity), remember that each `this _points[]` access results in a memcpy. You can avoid this by creating a variable `thisPoints := (this _points pointer as FloatPoint2D*)` or similar and accessing `thisPoints[]` instead.

source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,126 @@ import FloatTransform2D
...
+				if (i != maximumDistanceIndex)
+					if (This _isOnLeft(leftPoint, maximumDistancePoint, currentSet[i]))
+						outsideLeft add(currentSet[i])
+					else if (This _isOnLeft(maximumDistancePoint, rightPoint, currentSet[i]))
+						outsideRight add(currentSet[i])
+			this _findHull(outsideLeft, leftPoint, maximumDistancePoint)
+			this _hull add(maximumDistancePoint)
+			this _findHull(outsideRight, maximumDistancePoint, rightPoint)
+		}
+		currentSet free()
```
davidhesselbom: As a rule, I would strongly advise you not to free input variables here, but in the function where they were created.

source/math/FloatConvexHull2D.ooc
marcusnaslund: I honestly don't know how much performance we need here. Is it OK to leave it like this until we know?

source/math/FloatConvexHull2D.ooc
marcusnaslund: Probably a smart idea.

source/math/FloatConvexHull2D.ooc
marcusnaslund: Yes. Fixed.

source/math/FloatConvexHull2D.ooc
davidhesselbom: If you're concerned about efficiency, you shouldn't access `this _hull[]` like that (especially not in a loop, see my other comment), and the same goes for `this _hull count` which actually calls a get function on the `count` property each iteration of the loop.

marcusnaslund: I'll go for the mod option then.

davidhesselbom: The other things I mentioned are still a performance problem, of course - I just think `%` is a small expense compared to those things.

marcusnaslund:
Updated. I will look up if performance with this is going to be important later and if so create an issue about it.

source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,124 @@ import FloatTransform2D
...
+	hull ::= this _hull
 	count ::= this _points count
-	init: func ~fromPoints (=_points) // Note: Does not compute the convex hull, only assigns it
+	init: func ~fromPoints (=_points, computeHull := false) { //TODO: Compute the convex hull per default
+		// Note, if hull is not computed, points must be in clockwise order
+		if (computeHull)
+			this computeHull()
+		else {
+			this _hull = VectorList<FloatPoint2D> new(this _points count)
+			this _hull = this _points copy()
```
davidhesselbom: Here's another leak.

source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,124 @@ import FloatTransform2D
 import text/StringTokenizer
 
 FloatConvexHull2D: class {
-	// Points must be in clockwise order
-	// TODO: Implement computation of convex hull from a set of points
 	_points: VectorList<FloatPoint2D>
+	_hull: VectorList<FloatPoint2D> = null
```
davidhesselbom: Does it need setting to `null` here?

marcusnaslund:
Validated @davidhesselbom

source/math/FloatConvexHull2D.ooc
```diff
@@ -21,55 +21,122 @@ import FloatTransform2D
 import text/StringTokenizer
 
 FloatConvexHull2D: class {
-	// Points must be in clockwise order
-	// TODO: Implement computation of convex hull from a set of points
 	_points: VectorList<FloatPoint2D>
+	_hull: VectorList<FloatPoint2D>
 	points ::= this _points
+	hull ::= this _hull
 	count ::= this _points count
```
tomasmahlberg: Wouldn't it make more sense to have the property `count` return `this _hull count` now instead? Or is it even necessary to keep the point set saved in the class at all? Can't the hull points just be saved to the `points` list?

marcusnaslund: Probably, I just didn't want to break the API now. @niklasborwell ?

tomasmahlberg: But you're copying `points` to `hull` if `computeHull` is `false` so nothing would break, right?

marcusnaslund: True. Will fix.

marcusnaslund:
@tomasmahlberg Please review this again if you have the time and set the peer review flag if you're happy.

source/math/FloatConvexHull2D.ooc
```diff
@@ -33,42 +35,99 @@ FloatConvexHull2D: class {
...
+			this _findHull(rightSet, rightEndpoint, leftEndpoint)
+			points free()
+			leftSet free()
+			rightSet free()
+		}
 	}
-	free: override func {
-		this _points free()
-		super()
+	_findHull: func (currentSet: VectorList<FloatPoint2D>, leftPoint, rightPoint: FloatPoint2D) {
```
tomasmahlberg: This name is a bit confusing, it sounds like it would return a hull or something but it actually adds points to the current hull. I don't think this is a big deal though.

tomasmahlberg: Other than that, I have also verified this is memory leak free and calculating correct results in `computeHull`.

marcusnaslund: Thanks :) Good point about the name, but it's a private method, so I think it's OK anyhow.

source/math/FloatConvexHull2D.ooc
```diff
@@ -33,42 +35,99 @@ FloatConvexHull2D: class {
...
+			this _points add(currentSet[0])
+		else if (currentSet count > 1) {
+			maximumDistance := Float negativeInfinity
+			maximumDistanceIndex := -1
+			for (i in 0 .. currentSet count) {
+				distance := This _pointsLinePseudoDistance(leftPoint, rightPoint, currentSet[i])
+				if (distance > maximumDistance) {
+					maximumDistance = distance
+					maximumDistanceIndex = i
+				}
```
davidhesselbom: I wonder how many times finding the maximum value of some collection has been implemented in our code base. Feels like it's done, like, at least once in every file :/

source/math/FloatConvexHull2D.ooc
```diff
@@ -33,42 +35,99 @@ FloatConvexHull2D: class {
...
+					else if (This _isOnLeft(maximumDistancePoint, rightPoint, currentSet[i]))
+						outsideRight add(currentSet[i])
+			this _findHull(outsideLeft, leftPoint, maximumDistancePoint)
+			this _points add(maximumDistancePoint)
+			this _findHull(outsideRight, maximumDistancePoint, rightPoint)
+			outsideLeft free()
+			outsideRight free()
+		}
+	}
+	_pointsLinePseudoDistance: static func (leftPoint, rightPoint, queryPoint: FloatPoint2D) -> Float {
```
davidhesselbom: What's a "pseudoDistance"? I think this function deserves a brief explanation...

davidhesselbom: Also, if what it does is compute the distance between a point a line between two other points, it *may* belong in the `FloatPoint2D` class, should it be useful elsewhere.

marcusnaslund: I don't think this belongs in `FloatPoint2D`. Maybe one which calculates an accurate (Euclidean) distance, but we don't need that much calculation here. Also, the name is not madeup, see https://en.wikipedia.org/wiki/Pseudometric_space

marcusnaslund: I think this is clear enough if one follows the link above.

niklasborwell: Are you talking about this link: http://www.cse.yorku.ca/~aaw/Hang/quick_points/Algorithm.html ?
Because it seems to be dead.

marcusnaslund: Oh, thanks. I will replace it

source/math/FloatConvexHull2D.ooc
```diff
@@ -33,42 +35,99 @@ FloatConvexHull2D: class {
...
+	free: override func {
+		this _points free()
+		super()
+	}
 	contains: func ~Point (point: FloatPoint2D) -> Bool {
 		result := true
-		for (i in 0 .. this count - 1)
-			if (This _isOnLeft(this _points[i], this _points[i + 1], point)) {
+		for (i in 0 .. this count)
+			if (result && This _isOnLeft(this _points[i], this _points[(i + 1) % this count], point))
```
emilwestergren: Why `result &&` and remove `break`? Looks like it will always iterate `count` even though it's not necessary.

source/math/FloatConvexHull2D.ooc
```diff
@@ -33,42 +35,99 @@ FloatConvexHull2D: class {
...
-			}
-		if (This _isOnLeft(this _points[this count - 1], this _points[0], point))
-			result = false
 		result
 	}
 	contains: func ~ConvexHull (other: This) -> Bool {
 		result := true
 		for (i in 0 .. other count)
-			if (!this contains(other points[i])) {
+			if (result && !this contains(other _points[i]))
```
emilwestergren: Same thing here.

source/math/FloatConvexHull2D.ooc
marcusnaslund: I was blinded by my line-minimizing thinking.

source/math/FloatConvexHull2D.ooc
```diff
@@ -33,42 +35,99 @@ FloatConvexHull2D: class {
...
 			newPoints add(transform * this _points[i])
 		This new(newPoints)
 	}
-	_isOnLeft: static func (leftPoint, rightPoint, queryPoint: FloatPoint2D) -> Bool {
-		(rightPoint x - leftPoint x) * (queryPoint y - leftPoint y) > (rightPoint y - leftPoint y) * (queryPoint x - leftPoint x)
+	computeHull: func {
+		if (this count > 2) {
+			// Uses the Quickhull algorithm if more than two points, average complexity O(n log n)
+			// https://en.wikipedia.org/wiki/Quickhull
+			points := this _points copy()
```
emilwestergren: This can be done without copying:
```ooc
points := this _points
this _points = VectorList<FloatPoint2D> new()
```

source/math/FloatConvexHull2D.ooc
```diff
@@ -57,18 +61,77 @@ FloatConvexHull2D: class {
...
+			this _points clear()
+			leftMostIndex := 0
+			rightMostIndex := 0
+			for (i in 0 .. points count)
```
emilwestergren: `for (i in 1 .. points count)` No branch condition will be true the first iteration.

source/math/FloatConvexHull2D.ooc
marcusnaslund: This feels safer.

davidhesselbom: But you're already doing `this _points clear` on the second line, so if anyone else is using this list, they're in trouble already.

source/math/FloatConvexHull2D.ooc
```diff
@@ -57,18 +61,77 @@ FloatConvexHull2D: class {
...
+				if (points[i] x < points[leftMostIndex] x)
```
emilwestergren: Save the `leftMostValue` and `rightMostValue` instead of looking it up every iteration. I would also read `points` only once per iteration.

source/math/FloatConvexHull2D.ooc
marcusnaslund: Done. :)

source/math/FloatConvexHull2D.ooc
marcusnaslund: I talked to @niklasborwell about the need to optimize this right now and he told me I didn't have to.

source/math/FloatConvexHull2D.ooc
```diff
@@ -57,18 +61,77 @@ FloatConvexHull2D: class {
...
+					leftMostIndex = i
+				else if (points[i] x > points[rightMostIndex] x)
+					rightMostIndex = i
+			leftEndpoint := points[leftMostIndex]
+			rightEndpoint := points[rightMostIndex]
+			leftSet := VectorList<FloatPoint2D> new()
+			rightSet := VectorList<FloatPoint2D> new()
+			for (i in 0 .. points count)
+				if (i != leftMostIndex && i != rightMostIndex)
```
emilwestergren: Can't you just remove these elements from the list and then you don't have to check this?

source/math/FloatConvexHull2D.ooc
```diff
@@ -57,18 +61,77 @@ FloatConvexHull2D: class {
...
+					if (This _isOnLeft(points[leftMostIndex], points[rightMostIndex], points[i]))
```
emilwestergren: Why not use `leftEndpoint` and `rightEndpoint` instead of again accessing them from the list?

source/math/FloatConvexHull2D.ooc
```diff
@@ -57,18 +61,77 @@ FloatConvexHull2D: class {
...
+						leftSet add(points[i])
+					else
```
emilwestergren: If you follow my suggestions, this loop could be written like this:
```ooc
	for (i in 0 .. points count) {
		point := points[i]
		(This _isOnLeft(leftEndpoint, rightEndpoint, point) ? leftSet : rightSet) add(point)
	}
```

source/math/FloatConvexHull2D.ooc
marcusnaslund: Not sure why this is an outdated diff now.

source/math/FloatConvexHull2D.ooc
marcusnaslund: Nice, thanks.

source/math/FloatConvexHull2D.ooc
```diff
@@ -57,18 +61,75 @@ FloatConvexHull2D: class {
...
 	}
-	_isOnLeft: static func (leftPoint, rightPoint, queryPoint: FloatPoint2D) -> Bool {
-		(rightPoint x - leftPoint x) * (queryPoint y - leftPoint y) > (rightPoint y - leftPoint y) * (queryPoint x - leftPoint x)
+	computeHull: func {
+		if (this count > 2) {
+			// Uses the Quickhull algorithm if more than two points, average complexity O(n log n)
+			// https://en.wikipedia.org/wiki/Quickhull
+			points := this _points
+			this _points = VectorList<FloatPoint2D> new()
+			leftMostIndex := 0
```
emilwestergren: Still not satisifed:
```ooc
	leftMostIndex := 0
	rightMostIndex := 0
	leftEndpoint := points[0]
	rightEndpoint := leftEndpoint
	for (i in 1 .. points count) {
		point := points[i]
		if (point x < leftEndpoint x) {
			leftEndpoint = point
			leftMostIndex = i
		}
		else if (point x > rightEndpoint x) {
			rightEndpoint = point
			rightMostIndex = i
		}
	}
```

davidhesselbom: Happy now?

source/math/FloatConvexHull2D.ooc
davidhesselbom: Did you?

marcusnaslund: Yes, it's now a wikipedia link. Don't expect that to go down anytime soon.

PR #700, opened 2015-11-05
Fixed incorrect function signatures in VectorList and added some tests

thomasfanell:
* Adapted ```VectorList map()``` for rock 1.0.11
* Added test for ```VectorList sort()```
* Style fix in ```VectorList map()```

### DIFF
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -125,9 +125,8 @@ VectorList: class <T> {
 	}
 	map: func <S> (function: Func(T) -> S) -> This<S> {
 		result := This<S> new(this count)
-		for (i in 0 .. this count) {
+		for (i in 0 .. this count)
 			result add(function(this[i]))
-		}
 		result
 	}
 	fold: func <T, S> (function: Func(T, S) -> S, initial: S) -> S {
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -123,7 +123,7 @@ VectorList: class <T> {
 		for (i in 0 .. this count)
 			this[i] = function(this[i])
 	}
-	map: func <T, S> (function: Func(T) -> S) -> This<S> {
+	map: func <S> (function: Func(T) -> S) -> This<S> {
 		result := This<S> new(this count)
 		for (i in 0 .. this count) {
 			result add(function(this[i]))
--- a/source/collections/VectorList.ooc
+++ b/source/collections/VectorList.ooc
@@ -129,7 +129,7 @@ VectorList: class <T> {
 			result add(function(this[i]))
 		result
 	}
-	fold: func <T, S> (function: Func(T, S) -> S, initial: S) -> S {
+	fold: func <S> (S: Class, function: Func(T, S) -> S, initial: S) -> S {
 		for (i in 0 .. this count)
 			initial = function(this[i], initial)
 		initial
--- a/test/collections/VectorListTest.ooc
+++ b/test/collections/VectorListTest.ooc
@@ -225,7 +225,13 @@ VectorListTest: class extends Fixture {
 			list free()
 		})
 		this add("VectorList sort", func {
-			//TODO Current way of sorting not supported by Rock
+			list := VectorList<Int> new()
+			list add(8); list add(16); list add(32)
+			sortedList := list copy()
+			sortedList sort(|v1, v2| v1 < v2)
+			count := list count
+			for (value in sortedList)
+				expect(value, is equal to(list[--count]))
 		})
 		this add("VectorList fold", func {
 			//TODO Current way of folding not supported by Rock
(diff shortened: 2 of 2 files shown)


test/collections/VectorListTest.ooc
```diff
@@ -225,7 +225,13 @@ VectorListTest: class extends Fixture {
 			list free()
 		})
 		this add("VectorList sort", func {
-			//TODO Current way of sorting not supported by Rock
+			list := VectorList<Int> new()
+			list add(8); list add(16); list add(32)
+			sortedList := list copy()
+			sortedList sort(|v1, v2| v1 < v2)
```
davidhesselbom: Can you try to create a closure that does this instead? I'm pretty sure `| ... |` will cause a memory leak, in which case this isn't how we need to be doing it where such things matter (i.e. outside of tests).

marcusnaslund:
So sort will work now, but not fold?

test/collections/VectorListTest.ooc
thomasfanell:
```ooc
sortedList sort(func(left, right: Int) -> Bool { left < right })
```
fails with
```ooc
call sortedList sort((left: Int, right: Int) -> Bool) waiting on VectorList sort(greaterThan: Func (T, T) -> Bool) to resolve.
```

marcusnaslund: What about
```ooc
sortFunc: Func (left, right: Int) -> Bool { left < right }
sortedList sort(sortFunc)
```
or even
```
```ooc
sortFunc: Func (left, right: Int) -> Bool { left < right ? true : false }
sortedList sort(sortFunc)
```

source/collections/VectorList.ooc
```diff
@@ -123,11 +123,10 @@ VectorList: class <T> {
...
-	map: func <T, S> (function: Func(T) -> S) -> This<S> {
+	map: func <S> (function: Func(T) -> S) -> This<S> {
 		result := This<S> new(this count)
-		for (i in 0 .. this count) {
+		for (i in 0 .. this count)
 			result add(function(this[i]))
-		}
 		result
 	}
 	fold: func <T, S> (function: Func(T, S) -> S, initial: S) -> S {
```
marcusnaslund: `func <S>` here as well then?

test/collections/VectorListTest.ooc
davidhesselbom: Well then, I guess https://github.com/fasterthanlime/rock/issues/923 is still relevant.

thomasfanell: @marcusnaslund same

source/collections/VectorList.ooc
thomasfanell: It's still the same problem: ```error Not enough info to resolve return type S of function call```. Which could be solved by qualifying ```S: Class```, but that is not acceptable in my opinion.

thomasfanell: Anyway, the rock-devs are checking out some issues as we speak. Let's wait a day or two and see what they come up with.

thomasfanell: So it appears that we for now have to live with the fact that we must qualify ```S``` under certain circumstances, such as when ```S``` is used in an argument in a ```Func```, as is the case with ```VectorList fold()```.

marcusnaslund: If that *works*, it's better than before, which didn't generally work at all, even if you specified `S`. Is that the case now?

test/collections/VectorListTest.ooc
marcusnaslund: Still true?

test/collections/VectorListTest.ooc
```diff
@@ -225,10 +225,19 @@ VectorListTest: class extends Fixture {
 			list free()
 		})
 		this add("VectorList sort", func {
-			//TODO Current way of sorting not supported by Rock
+			list := VectorList<Int> new()
+			list add(8); list add(16); list add(32)
```
marcusnaslund: This isn't consistent with the above tests, but please do `list add(8) . add(16) . add(32)` instead.

test/collections/VectorListTest.ooc
```diff
@@ -225,10 +225,19 @@ VectorListTest: class extends Fixture {
...
+			sortedList := list copy()
+			sortedList sort(|v1, v2| v1 < v2)
+			count := list count
+			for (value in sortedList)
+				expect(value, is equal to(list[--count]))
 		})
 		this add("VectorList fold", func {
-			//TODO Current way of folding not supported by Rock
+			list := VectorList<Int> new()
+			list add(1); list add(2); list add(3)
```
marcusnaslund: `list add(1) . add(2) . add(3)`

davidhesselbom:
So... all good to go?

thomasfanell:
Yup :+1:

marcusnaslund:
I was expecting something to fail without this, but that's not the case.

thomasfanell:
That was the case before we changed ```func <S>```, since rock segfaulted when we attempted to redefine ```T```. I included the segfault fix in the latest rock version, so the title of this PR is now obsolete.

marcusnaslund:
Validated @davidhesselbom

marcusnaslund:
I'm happy with this, it passes validation, and there are no side-effects. @davidhesselbom

Issue #692, opened 2015-11-05
Clean files in sdk/io


marcusnaslund:
Add `this` where appropriate, remove not needed code.
We won't be using these classes forever, but for the forseeable future, so making them easier to maintain is a big plus. Throw out whatever isn't used in these files.

- [x] BinarySequence.ooc
- [x] BufferReader.ooc, BufferWriter.ooc
- [x] File.ooc
- [x] FileReader.ooc, FileWriter.ooc
- [x] Reader.ooc, Writer.ooc

thomasfanell:
Should we keep the convention of having classes that are specific to a certain platform under the ```native``` namespace? I vote yes.

Example: ```source/io/native/FileUnix.ooc```

marcusnaslund:
Hmm, `source/widgets` disagrees.

thomasfanell:
I could live without it, but I think it's much cleaner since most often one only need to look at the parent.

marcusnaslund:
I'd like to hear the opinions of @davidhesselbom and @simonmika because the same answer applies to moving `threading` as well.

davidhesselbom:
I guess it would keep things clearer... but then again, should we do the same thing in `ooc/widgets`?

simonmika:
Keep and where necessary introduce `native` folders.

simonmika:
When it comes to `io` we should build heavily on the concepts used in https://github.com/cogneco/Kean. They work really well.

marcusnaslund:
It doesn't seem to be possible to move these files because of dependencies in the sdk. Changed the issue to do some general cleanup so they're easier to maintain for the time being. If we ever get around to rewriting the io component from scratch we can discuss thinning it out even further.

PR #701, opened 2015-11-05
Cleaned up files in sdk/io 

marcusnaslund:
Before you throw this out the window and yell at me, append `?w=1` :)

It's not completely magic compliant yet, but it's better to get all of these whitespace fixes out of the way first.

No functionality has changed. Part of the work on #692

### DIFF
--- a/source/sdk/io/File.ooc
+++ b/source/sdk/io/File.ooc
@@ -28,13 +28,13 @@ File: abstract class {
 	nameWithoutExtension: String { get { getNameWithoutExtension() } }
 	pathWithoutExtension: String { get { getPathWithoutExtension() } }
 	rectifySeparator: func {
-		if (this dir?() && !this path endsWith?(File separator)) {
+		if (this dir?() && !this path endsWith?(This separator)) {
 			oldPath := this path
-			this path = oldPath + File separator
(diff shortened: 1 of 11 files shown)


source/sdk/io/File.ooc
```diff
@@ -18,604 +18,603 @@ import text/StringTokenizer
...
+
+	extension: String { get { getExtension() } }
+	nameWithoutExtension: String { get { getNameWithoutExtension() } }
+	pathWithoutExtension: String { get { getPathWithoutExtension() } }
+	rectifySeparator: func {
+		if (this dir?() && !this path endsWith?(This separator)) {
 			oldPath := this path
-			this path = oldPath + File separator
+			this path = oldPath + This separator
 			oldPath free()
```
davidhesselbom: `this path = this path >> This separator` should work as a replacement to the 3 lines here.

davidhesselbom: But, it can be fixed later.

davidhesselbom:
> append ?w=1

Nifty, innit? :)

source/sdk/io/File.ooc
marcusnaslund: I'll write it up.

PR #707, opened 2015-11-06
Cleaned up sdk/math.ooc

marcusnaslund:
Again, with `?w=1` this is a very small PR.
Only style fixes to make magic happy, and some comments removed.

### DIFF
--- a/source/sdk/math.ooc
+++ b/source/sdk/math.ooc
@@ -3,8 +3,8 @@ use math
 
 PI := 3.14159_26535_89793_23846_26433_83279
 FLT_EPSILON: extern Float
-constantE: extern(M_E) Double
-constantPi: extern(M_PI) Double
+constantE: extern (M_E) Double
+constantPi: extern (M_PI) Double
 
 abs: extern func (Int) -> Int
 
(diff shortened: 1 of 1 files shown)


source/sdk/math.ooc
```diff
@@ -86,36 +86,31 @@ extend Int {
...
-	minimum: static func ~two(first: This, second: This) -> This {
+	minimum: static func ~two (first: This, second: This) -> This {
 		first < second ? first : second
 	}
 	// TODO: Avoid using this, consider removing it.
-	minimum: static func ~multiple(value: This, values: ...) -> This {
+	minimum: static func ~multiple (value: This, values: ...) -> This {
 		values each(|v|
-			if ((v as This*)@ < value)
+			if ((v as This *)@ < value)
```
davidhesselbom: If you've done this before, I've missed it. Does a space really belong here?

marcusnaslund: It looks weird, I know, but magic complains if it's not there.
I had to do the same thing for the `net` classes.

davidhesselbom: @thomasfanell Can we make magic understand this isn't a multiplication operator?

davidhesselbom: Actually, I found plenty of examples of `type*` in another project, and magic didn't mind that. None of them were `This*`, though. Magic bug?

marcusnaslund: You're right, magic wants a space after every `This` for some reason. It's not because magic believes it's a multiplication operator (since it doesn't complain about there being no space between `*` and `)`).

thomasfanell: Yeah I can fix that.

thomasfanell: And it's not a bug, it's just related to the fact that ```*``` is disregarded since it can mean different things, as we see here. The violated rule here is the one that deals with keyword spacing (in this case, ```This```). I've added ```*``` to the ignore list for this particular rule and should have a release up in a few minutes.

davidhesselbom: Great, thanks.

marcusnaslund: I will fix this then.

marcusnaslund:
There we go.

source/sdk/math.ooc
```diff
@@ -86,34 +86,29 @@ extend Int {
...
 		)
 		value
 	}
-	minimum: static func ~two(first: This, second: This) -> This {
+	minimum: static func ~two (first: This, second: This) -> This {
 		first < second ? first : second
 	}
 	// TODO: Avoid using this, consider removing it.
-	minimum: static func ~multiple(value: This, values: ...) -> This {
+	minimum: static func ~multiple (value: This, values: ...) -> This {
```
davidhesselbom: Doesn't belong in this PR, but, when you have the time, could you investigate the possibility of removing this and the corresponding `maximum` function (the ones with `...`)?

marcusnaslund: It's on my list :)

marcusnaslund: Are we not interested in `VarArgs` at all?

davidhesselbom: I don't think they can be used without leaking memory. They also have absolutely 0 type safety - the arguments in there could be *anything*.

marcusnaslund: So let's throw them out of the SDK, then?

davidhesselbom: I guess so. But not in this PR :)

marcusnaslund: Once they're not used anymore, I mean.

PR #708, opened 2015-11-06
Added missing free() function in Event classes

sebastianbaginski:


davidhesselbom:
Hm, how come we haven't had need of these before?

sebastianbaginski:
Good question, I think because class `Event1` is used most of the time. I just think I could need the other ones for the upcoming mouse event cleanup.

PR #709, opened 2015-11-06
profiler fix.

ziquanyu:


source/base/Profiler.ooc
```diff
@@ -65,6 +65,7 @@ Profiler: class {
 	}
 	resetAll: static func { This _profilers apply(func (profiler: This) { profiler reset() }) }
 	dispose: static func {
+		This _profilers[0] printResults()
```
marcusnaslund: This will only print the result of one, not all (if you have had several).
```ooc
This _profilers apply(func (profiler: This) { profiler printResults() })
```

ziquanyu: it will print all, if you have run any of profiler printResults once.

marcusnaslund: `This printResults()`

marcusnaslund: But this is not what I have in mind when I think of a `dispose()` method, though. @davidhesselbom ?

source/base/Profiler.ooc
```diff
@@ -65,6 +65,7 @@ Profiler: class {
 	}
 	resetAll: static func { This _profilers apply(func (profiler: This) { profiler reset() }) }
 	dispose: static func {
+		This printResults()
```
emilwestergren: Should this really be here?

marcusnaslund: Hmm I think not. I would expect anyone who uses `Profiler` to call `printResults` explicitly.

niklasborwell: I agree.

emilwestergren: Exactly, move it to a better place.

marcusnaslund: Since that's the only change in this PR we should close this then. Any objections @ziquanyu ?

ziquanyu: I don't have any objection!

PR #711, opened 2015-11-09
Calculate vector mean using given list of indices

sebastianbaginski:
Used to avoid the need of creating temporary vector just to calculate the mean for given vector subset.
Labelled with `performance` as it is used to optimize something in other project.

### DIFF
--- a/source/math/FloatPoint2DVectorList.ooc
+++ b/source/math/FloatPoint2DVectorList.ooc
@@ -69,6 +69,23 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
 		sortedY free()
 		result
 	}
+	getMean: func (indices: VectorList<Int>) -> FloatPoint2D {
+		result := FloatPoint2D new()
+		indicesCount := indices count
+		if (indicesCount > 0) {
+			buffer := this pointer as FloatPoint2D*
+			indicesBuffer := indices pointer as Int*
+			for (i in 0 .. indicesCount) {
+				index := indicesBuffer[i]
+				element := buffer[index]
+				result x = result x + element x
+				result y = result y + element y
+			}
+			result x = result x / indicesCount
+			result y = result y / indicesCount
+		}
+		result
+	}
 	operator + (value: FloatPoint2D) -> This {
 		result := This new()
 		for (i in 0 .. this _count)


source/math/FloatPoint2DVectorList.ooc
```diff
@@ -69,6 +69,18 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
 		sortedY free()
 		result
 	}
+	getMean: func (indices: VectorList<Int>) -> FloatPoint2D {
+		result := FloatPoint2D new()
+		if (indices count > 0) {
+			for (i in 0 .. indices count) {
+				element := this[indices[i]]
```
marcusnaslund: Something wrong with `result = result + this[indices[i]]`?

sebastianbaginski: Yes, creating three copies of `FloatPoint2D` (which means calling at least six functions) when you can just access the floats directly.

sebastianbaginski: And it should probably be coded like 
```
buffer := this buffer pointer
element := buffer[i] as FloatPoint2D
```
(or something) to avoid calling `operator []`

marcusnaslund: You might want to do the division on line 80 per float as well.

sebastianbaginski: That's true, thanks.

davidhesselbom:
```ooc
buffer := this buffer pointer
element := buffer[i] as FloatPoint2D
```
This goes for both `this[]` and `indices[]`, mind you - both are `VectorList`, both suffer the same memcpy penalty.

sebastianbaginski: Updated. Thanks for suggestions, both of you.

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -69,6 +69,22 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
 		sortedY free()
 		result
 	}
+	getMean: func (indices: VectorList<Int>) -> FloatPoint2D {
+		result := FloatPoint2D new()
+		if (indices count > 0) {
+			buffer := this pointer as FloatPoint2D*
+			indicesBuffer := indices pointer as Int*
+			for (i in 0 .. indices count) {
```
davidhesselbom: Remember that `count` is a property - a function that's called once for every iteration of the loop.

sebastianbaginski: Updated. This one gets me every time...

davidhesselbom: It's not an expensive function, but since you mentioned this being function being a bottleneck...

PR #714, opened 2015-11-09
Made TimerTest a Fixture instance and fixed issues on Windows

### LINKED ISSUE #715: Timer and Profiler do not work on Windows
As I was writing a test for `Timer` and `ClockTimer` (#714) I noticed that it kept failing on AppVeyor. I have now confirmed that `Timer` does indeed not work on Windows at all - which means `Profiler` also doesn't work.

The reason is that `Time runTimeMicro()` always returns `UINT_MAX` on Windows.

`ClockTimer` works well because it uses a different implementation.

marcusnaslund:
Not a great test, but certainly better than the original one which didn't really do anything.

Fixes #715

### DIFF
--- a/source/sdk/os/Time.ooc
+++ b/source/sdk/os/Time.ooc
@@ -145,6 +145,9 @@ Time: class {
             gettimeofday(tv&, null)
             return ((tv tv_usec + tv tv_sec * 1_000_000) - __time_millisec_base) as UInt
         }
+		version(windows) {
+            return This runTime() * (1000 as UInt)
+        }
         return -1
 	}
 
--- a/source/base/Timer.ooc
+++ b/source/base/Timer.ooc
@@ -31,6 +31,7 @@ Timer: class {
 		this _min = INFINITY
 		this _max = 0.0
 	}
+	// Note: On Windows, there is only millisecond precision
 	start: virtual func { this _startTime = (Time runTimeMicro() as Double) }
 	stop: virtual func -> Double {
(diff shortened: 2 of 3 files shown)


marcusnaslund:
Interesting. It would appear that `Timer` does not actually work on Windows.

source/sdk/os/Time.ooc
```diff
@@ -145,6 +145,9 @@ Time: class {
             gettimeofday(tv&, null)
             return ((tv tv_usec + tv tv_sec * 1_000_000) - __time_millisec_base) as UInt
         }
+		version(windows) {
+            return This runTime() * (1000 as UInt)
+        }
```
davidhesselbom: The indentation is messed up here, but since it's in the SDK, I'm not too bothered.

marcusnaslund: It's at least the same as three lines above. :)

davidhesselbom: No, line 148 uses tabs, line 150 uses spaces.

marcusnaslund: Yes, but compare with line 143 and 147.

davidhesselbom: Oh, right. Well, both are wrong, but I'm still not bothered.

davidhesselbom:
@sebastianbaginski Any objections?

sebastianbaginski:
@davidhesselbom no, it's better than before

PR #717, opened 2015-11-09
Removed specialized code for when using GC

marcusnaslund:
We can't even build our stuff with garbage collection, so keeping specialized code for that version flag is strange.

For the next step, we should replace all `gc_*` calls with their "direct" counterparts and then remove the `gc_*` function declarations completely,


source/sdk/lang/Memory.ooc
```diff
@@ -1,53 +1,17 @@
-
-// as goofy as it sounds, memory-related routines are actually in string.h
-// yay C. (Also, Google Native Client doesn't have memory.h)
 include string
 version(windows) { include malloc }
```
marcusnaslund: Removing the curly brackets here makes rock **always** try to import this file. I don't know if this is a known bug.

sebastianbaginski: Oh wow, so you can use version blocks around includes, I thought it does not work at all. Nice to know.

marcusnaslund: Yes you can, but the brackets are mandatory.

davidhesselbom:
Also, it would be nice to fix rock so that `--gc=off` is the default...

marcusnaslund:
Oh definitely.

source/sdk/lang/Memory.ooc
```diff
@@ -1,53 +1,20 @@
...
+gc_malloc: func (size: SizeT) -> Pointer { gc_calloc(1, size) }
+gc_malloc_atomic: extern(malloc) func (size: SizeT) -> Pointer
+gc_strdup: extern(strdup) func (str: CString) -> CString
+gc_realloc: extern(realloc) func (ptr: Pointer, size: SizeT) -> Pointer
+gc_calloc: extern(calloc) func (nmemb: SizeT, size: SizeT) -> Pointer
+gc_free: extern(free) func (ptr: Pointer)
+
+//TODO Remove the above declarations once they're no longer used - only use the following:
+malloc: extern(malloc) func (size: SizeT) -> Pointer
+calloc: extern(calloc) func (nmemb: SizeT, size: SizeT) -> Pointer
```
davidhesselbom:
```ooc
malloc: extern func (SizeT) -> Pointer
calloc: extern func (SizeT, SizeT) -> Pointer
```
would do just as nicely. If you want parameter names, please use something more descriptive than `nmemb`.

source/sdk/lang/Memory.ooc
```diff
@@ -1,53 +1,20 @@
-
-// as goofy as it sounds, memory-related routines are actually in string.h
-// yay C. (Also, Google Native Client doesn't have memory.h)
```
davidhesselbom: I think this comment should stay, or a more serious one should replace it. It's far from obvious to the next person who looks at this file that we need `string` for anything.

source/sdk/lang/types.ooc
```diff
@@ -15,10 +15,8 @@ Object: abstract class {
     __destroy__: func {}
 
     free: virtual func {
-      version(!gc) {
-        this __destroy__()
-        gc_free(this)
-      }
+      this __destroy__()
```
davidhesselbom: Soon, this should be removed too, because we shouldn't have any `__destroy__` functions anymore.

source/sdk/threading/native/ThreadUnix.ooc
```diff
@@ -112,14 +112,8 @@ version(unix || apple) {
...
 
-	version (gc) {
-		pthread_create: extern (GC_pthread_create) func (threadPtr: PThread*, attrPtr: Pointer, startRoutine: Pointer, userArgument: Pointer) -> Int
-		pthread_join: extern (GC_pthread_join) func (thread: PThread, retval: Pointer*) -> Int
-	}
-	version (!gc) {
-		pthread_create: extern func (threadPtr: PThread*, attrPtr: Pointer, startRoutine: Pointer, userArgument: Pointer) -> Int
-		pthread_join: extern func (thread: PThread, retval: Pointer*) -> Int
-	}
+	pthread_create: extern func (threadPtr: PThread*, attrPtr: Pointer, startRoutine: Pointer, userArgument: Pointer) -> Int
```
davidhesselbom: There's room for cleanup here, too: `threadPointer: PThread*, attributePointer, startRoutine, userArgument: Pointer) -> Int`

source/sdk/lang/types.ooc
marcusnaslund: Ah, yes, `__destroy__` doesn't do anything at all here. Will fix in a future PR.

source/sdk/lang/Memory.ooc
marcusnaslund: Replaced by one that doesn't have the word "goofy".

source/sdk/threading/native/ThreadUnix.ooc
davidhesselbom: Or at the very least, remove the extra `Pointer` - you only need 1.

marcusnaslund:
Updated according to comments.

source/sdk/lang/Memory.ooc
```diff
@@ -1,59 +1,26 @@
...
+gc_malloc: func (size: SizeT) -> Pointer { gc_calloc(1, size) }
+gc_malloc_atomic: extern(malloc) func (size: SizeT) -> Pointer
+gc_strdup: extern(strdup) func (str: CString) -> CString
+gc_realloc: extern(realloc) func (ptr: Pointer, size: SizeT) -> Pointer
+gc_calloc: extern(calloc) func (nmemb: SizeT, size: SizeT) -> Pointer
+gc_free: extern(free) func (ptr: Pointer)
+
+//TODO Remove the above declarations once they're no longer used - only use the following:
+malloc: extern(malloc) func (SizeT) -> Pointer
+calloc: extern(calloc) func (SizeT, SizeT) -> Pointer
```
davidhesselbom: Like I said, this could be 
```ooc
malloc: extern func (SizeT) -> Pointer
calloc: extern func (SizeT, SizeT) -> Pointer
```
Did that not work?

marcusnaslund: Sorry, I saw only the removal of argument names.

davidhesselbom:
@simonmika, you had objections about a similar pull request, so I just want to run this by you first.

simonmika:
I'm fine with this but do wrap these raw c-functions in an object oriented API as we discussed last time.

davidhesselbom:
The raw C functions were already exposed, this PR only removes a ton of function we weren't using anyway. Except the one place where `gc_free` is replaced by `free`.

simonmika:
Go ahead with this PR and let's do the object oriented API later.

tomasmahlberg:
Apart from having conflicts, there is a `version(!gc)` in `ByteBufferSliceTest` that this PR might want to take care of?

marcusnaslund:
Thanks @tomasmahlberg, will fix

marcusnaslund:
There we go.

source/sdk/threading/native/ThreadWin32.ooc
```diff
@@ -68,13 +68,7 @@ version(windows) {
...
 	// CreateThread: extern func (...) -> Handle
 
-	version (gc) {
-		// Use Boehm-provided replacement for _beginthreadex
-		_beginthreadex: extern (GC_beginthreadex) func (security: Pointer, stackSize: UInt, startAddress: Pointer, arglist: Pointer, initflag: UInt, thrdaddr: UInt*) -> Handle
-	} else {
-		_beginthreadex: extern func (security: Pointer, stackSize: UInt, startAddress: Pointer, arglist: Pointer, initflag: UInt, thrdaddr: UInt*) -> Handle
-	}
-
+	_beginthreadex: extern func (security: Pointer, stackSize: UInt, startAddress: Pointer, arglist: Pointer, initflag: UInt, thrdaddr: UInt*) -> Handle
```
davidhesselbom: At the very least, `startAddress, arglist: Pointer`

marcusnaslund: Oh I think I did this but then it got lost in my merge

marcusnaslund: Will fix

marcusnaslund: What do you mean "at the very least", is there anything else I could reasonably to?

davidhesselbom: Sorry for the tone - I just don't like the `arglist` and especially the `thrdaddr` thing. It can wait, though.

davidhesselbom: I don't remember right now whether they have to be the same as in the library being called...

marcusnaslund: I fixed `startAddress, arglist: Pointer` :)

davidhesselbom: Yeah, good enough for now.

PR #718, opened 2015-11-09
Changed gc_free to free

marcusnaslund:
Changed all `gc_free` to `free`.


marcusnaslund:
Since we won't be doing it this way, I'm closing this.

Issue #643, opened 2015-10-30
Text toNumeric needs to handle leading whitespace


davidhesselbom:
This 
```ooc
text := t" 1.0"
println(text toFloat() toString())
string := text toString()
println(string toFloat() toString())
```
prints
```
0.00
1.00
```
because `String` uses the standard library function `strtof` under the hood, and it can handle leading whitespace, while the conversion functions in `Text` cannot. This has lead to #575 limiting parse functions to work only when the input has no whitespace separating the components of e.g. a `FloatPoint2D`.

The fix is simple: implement a `trim` function in `Text` (could be useful for other reasons, too) and use it on `Text` instances before trying to convert it to a numeric type, or, better yet, let the conversion functions call `trim` before they do anything else.

davidhesselbom:
`strtof` can also handle a leading `+`, for some reason. @sebastianbaginski, do you think it would be a better idea to use `strtof` or to mimic its behavior?

sebastianbaginski:
I vote for implementing it without calling external functions, `trim` can be useful anyway.

davidhesselbom:
Fine by me. Are you up for it?

sebastianbaginski:
Sorry, not at the moment, I still have some PR debts to pay here and in the other project:)

davidhesselbom:
Alright then.

thomasfanell:
I did actually implement trim (left, right and both) for Text quite some time ago. I think I may still have it at home. Otherwise I'll rewrite and PR this evening or tomorrow.

thomasfanell:
Was this fixed by #744?

davidhesselbom:
It was. I think maybe the `toFloat` etc. function should be able to handle leading `+` and maybe even `+ `, but that belongs in another issue, if it turns out we need it.

Issue #689, opened 2015-11-04
Cell toString() has no overload for Text


thomasfanell:
This will cause ```Fixture``` to crash if a test fails and the value being tested is a ```Text```

marcusnaslund:
Since we do exactly that, this sounds like a bug to me.

PR #719, opened 2015-11-09
Added Text trim() and test in preparation for #643

thomasfanell:
Motivation: #643 
* added ```Text trim()``` and associated test
* ~~added guard in ```slice()``` to prevent unnecessary slicing if the resulting slice would be an identical copy~~

```trim()``` could be implemented using only one loop (attacking both left and right), but that would mean less readable code and the performance gain is only really noticeable for very large strings where the non-whitespace characters are near center (I did a comparison a while back).

I'm tagging this as 'not ready' until we decide what to do here.

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -135,10 +135,15 @@ Text: cover {
 		this _buffer free(criteria)
 	}
 	slice: func (start, distance: Int) -> This {
-		result := This new(this _buffer slice(start, distance))
-		if (this _buffer owner == Owner Receiver)
-			result = result copy() // TODO: Could we be smarter here?
-		this free(Owner Receiver)
+		result: This
+		if (start == 0 && distance == this count)
+			result = this
+		else {
+			result = This new(this _buffer slice(start, distance))
+			if (this _buffer owner == Owner Receiver)
+				result = result copy() // TODO: Could we be smarter here?
+			this free(Owner Receiver)
+		}
 		result
 	}
 	split: func ~character (separator: Char) -> VectorList<This> {
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -169,6 +169,16 @@ Text: cover {
 		separator free(Owner Receiver)
 		result
 	}
+	trim: func -> This {
+		length := this count
+		leftPosition := 0
+		rightPosition := length - 1
+		while (leftPosition < length && this _buffer[leftPosition] whitespace?())
+			++leftPosition
+		while (rightPosition > leftPosition && this _buffer[rightPosition] whitespace?())
+			--rightPosition
+		this slice(leftPosition, rightPosition - leftPosition + 1)
+	}
 	toString: func -> String {
 		result := String new(this _buffer raw, this count)
 		this free(Owner Receiver)
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -135,15 +135,10 @@ Text: cover {
 		this _buffer free(criteria)
 	}
 	slice: func (start, distance: Int) -> This {
-		result: This
-		if (start == 0 && distance == this count)
-			result = this
-		else {
-			result = This new(this _buffer slice(start, distance))
-			if (this _buffer owner == Owner Receiver)
-				result = result copy() // TODO: Could we be smarter here?
-			this free(Owner Receiver)
-		}
+		result := This new(this _buffer slice(start, distance))
+		if (this _buffer owner == Owner Receiver)
+			result = result copy() // TODO: Could we be smarter here?
+		this free(Owner Receiver)
 		result
 	}
 	split: func ~character (separator: Char) -> VectorList<This> {
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -170,10 +170,9 @@ Text: cover {
 		result
 	}
 	trim: func -> This {
-		length := this count
 		leftPosition := 0
-		rightPosition := length - 1
-		while (leftPosition < length && this _buffer[leftPosition] whitespace?())
+		rightPosition := this count - 1
+		while (leftPosition < this count && this _buffer[leftPosition] whitespace?())
 			++leftPosition
 		while (rightPosition > leftPosition && this _buffer[rightPosition] whitespace?())
 			--rightPosition
--- a/test/base/TextTest.ooc
+++ b/test/base/TextTest.ooc
@@ -166,6 +166,13 @@ TextTest: class extends Fixture {
 			text free()
 			text2 free()
 		})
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by davidhesselbom** [comment]
This looks good enough to me. I have no reason to believe a double loop will improve performance noticeably at this time.

**Comment by thomasfanell** [comment]
@simonmika did you have some input regarding ```slice()``` before I mark this as ready?

File: test/base/TextTest.ooc [hunk 1]
```diff
@@ -166,6 +166,13 @@ TextTest: class extends Fixture {
 			text free()
 			text2 free()
 		})
+		this add("trim", func {
+			paddedText := t"  \t test \n test \r\n\t "
+			trimmedText := paddedText trim()
+			expect(trimmedText, is equal to(t"test \n test"))
```
marcusnaslund: Isn't/wasn't there a problem with `is equal to(` and `Text`?

davidhesselbom: The problem was with `String`, wasn't it?

thomasfanell: It is a problem if it fails, since ```Cell``` has no overload for ```Text```, and we can't implement one as long ```Cell``` is in the SDK.

thomasfanell: Should I change this test to a ```==``` instead?

marcusnaslund: Having the whole test crash if something's wrong with `Text` isn't very good :) So yes, until my war against the SDK has been won.

marcusnaslund: For reference: #689

davidhesselbom: Hm, would it be possible to write an extension for `Cell<Text>`?

thomasfanell:
Fixed.

test/base/TextTest.ooc
marcusnaslund: I'm actually very close to pulling `Cell` out, so maybe don't bother with that now.

davidhesselbom: Alright.

thomasfanell: @davidhesselbom I tried that, no go.

source/base/Text.ooc
```diff
@@ -164,6 +169,16 @@ Text: cover {
 		separator free(Owner Receiver)
 		result
 	}
+	trim: func -> This {
+		length := this count
```
simonmika: Stop adding temporary variables that don't do anything good.

test/base/TextTest.ooc
```diff
@@ -166,6 +166,13 @@ TextTest: class extends Fixture {
 			text free()
 			text2 free()
 		})
+		this add("trim", func {
+			paddedText := t"  \t test \n test \r\n\t "
+			trimmedText := paddedText trim()
+			expect(trimmedText == t"test \n test")
+			paddedText free()
+			trimmedText free()
```
simonmika: Double free.

test/base/TextTest.ooc
```diff
@@ -166,6 +166,13 @@ TextTest: class extends Fixture {
 			text free()
 			text2 free()
 		})
+		this add("trim", func {
+			paddedText := t"  \t test \n test \r\n\t "
+			trimmedText := paddedText trim()
+			expect(trimmedText == t"test \n test")
+			paddedText free()
```
simonmika: It is already freed by the comparison and on top of this these are literals so no need to free at all.

source/base/Text.ooc
thomasfanell: @simonmika what it does is that it removes a function call from the loop condition, how is that bad?

simonmika: It reduces readability in favor for a potential performance improvement that is not needed. It adds one line of code and by that makes maintenance harder.

source/base/Text.ooc
```diff
@@ -135,10 +135,15 @@ Text: cover {
...
 	}
 	slice: func (start, distance: Int) -> This {
-		result := This new(this _buffer slice(start, distance))
-		if (this _buffer owner == Owner Receiver)
-			result = result copy() // TODO: Could we be smarter here?
-		this free(Owner Receiver)
+		result: This
+		if (start == 0 && distance == this count)
+			result = this
+		else {
```
davidhesselbom: @simonmika, since the discussion about saving lines vs performance came up in this PR, why don't we (generally) do something like this?
```ooc
result := this
if (start != 0 || distance != this count) {
    result = This new(this _buffer slice(start, distance))
    if (this _buffer owner == Owner Receiver)
        result = result copy() // TODO: Could we be smarter here?
    this free(Owner Receiver)
}
result
```
It saves 2 lines, and the first assignment, while admittedly unnecessary, has *got* to be cheaper than a function call caused by accessing `this count`.

source/base/Text.ooc
```diff
@@ -135,10 +135,15 @@ Text: cover {
...
+			result = This new(this _buffer slice(start, distance))
+			if (this _buffer owner == Owner Receiver)
+				result = result copy() // TODO: Could we be smarter here?
+			this free(Owner Receiver)
+		}
 		result
 	}
```
davidhesselbom: @simonmika, since the discussion about saving lines vs performance came up in this PR, why don't we (generally) do something like this?
```ooc
slice: func (start, distance: Int) -> This {
    result := this
    if (start != 0 || distance != this count) {
        result = This new(this _buffer slice(start, distance))
        if (this _buffer owner == Owner Receiver)
            result = result copy() // TODO: Could we be smarter here?
        this free(Owner Receiver)
    }
    result
}
```
It saves 2 lines, and the first assignment, while admittedly unnecessary, has *got* to be cheaper than a function call caused by accessing `this count`.

simonmika: I would say that Thomas implementation is a bit easier to read but I don't dislike your somewhat shorter solution either. The question I ask is why at all do we need this special case at all?

marcusnaslund: It's not like we need amazing performance out of `Text`, right?

simonmika: As far as I can see this change is not needed at all.

davidhesselbom: > It's not like we need amazing performance out of Text, right?

Not now, I think. I guess optimizations like these can wait until we actually need them.

thomasfanell:
Yet another amend... ```slice()``` is back the way it was.

davidhesselbom:
That's not an amendment in Git terminology, but thanks.

thomasfanell:
@davidhesselbom I didn't mean it like that either ;-)

davidhesselbom:
A really old PR is waiting for this one, so I'll try to get it merged ASAP.

PR #721, opened 2015-11-10
Removed HashMap imports

marcusnaslund:
Simplifies the SDK dependencies.


marcusnaslund:
Ouch, I can't do this. :(

Issue #544, opened 2015-10-13
::= does not work for rvalue covers because a pointer is expected.


davidpiuva:
When a FloatBox2D is a value type given from a collection look-up or function call, rightBottom can't be accessed without storing the box to make it a referrable cover because the ::= operand generates a getter that expects a pointer.

A workaround is to store rvalues to make them into cloned lvalues.

This can be solved by getting lvalues from [] operands since rvalues are slow to pass around.
At least, the error should be explained by the ooc compiler instead of failing in the C compiler.

davidhesselbom:
Can you post a non-working code example?

davidpiuva:
// Not working
result = myCovers[i] myAttribute

// Working
myCover := myCovers[i] // Convert rvalue to lvalue
result = myCover myAttribute

davidhesselbom:
Ah, I see what you mean. So collections are not suitable as get properties, then.

davidhesselbom:
Would it help to wrap parentheses around `myCovers[i]`?

davidpiuva:
Parentheses did not help.

davidhesselbom:
How about `result := (myCovers[i] as MyCover) attribute`?

davidpiuva:
Failed too.

davidhesselbom:
Interesting...

marcusnaslund:
I doubt we're going to change rock regarding this. Could you post an issue about it there, though? https://github.com/fasterthanlime/rock/issues
Should we change these properties (potentially quite a lot)? Otherwise, this is not really a bug in kean.

davidpiuva:
It is probably too wide for ooc since .NET has the same dangerous bug/limitation.

marcusnaslund:
You're right, C# doesn't do this either.
But I agree with 

> At least, the error should be explained by the ooc compiler instead of failing in the C compiler.

so that would warrant an issue in rock. Nothing to do here though, it seems.

sebastianbaginski:
Maybe this is a good place to ask. Can you guys tell me what's so good about properties anyway ? Coming from the C++ background, the whole *property* idea seems strange to me and I still can't really get used to it. So basically you have something that looks like a variable ...
```ooc
Example: class {
    property ::= this something + 29
}

var ::= object property
```
... but is in fact a function :)
In C++, you have a const methods instead
```C++
class Example {
public:
    int property() const { return this->something + 29; }
};
auto var = object.property();
```
[...]

PR #722, opened 2015-11-10
Added convolve function in euclid transform.

niklasborwell:
In preparation for a coming pull request.

### DIFF
--- a/source/math/FloatEuclidTransform.ooc
+++ b/source/math/FloatEuclidTransform.ooc
@@ -14,11 +14,14 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use ooc-collections
 import math
 import FloatPoint3D
 import FloatRotation3D
 import FloatTransform2D
 import FloatTransform3D
+import FloatVectorList
+import Quaternion
 
 FloatEuclidTransform: cover {
 	rotation: FloatRotation3D
@@ -42,6 +45,22 @@ FloatEuclidTransform: cover {
 		this translation == other translation &&
 		this rotation == other rotation &&
 		this scaling == other scaling
+	}
+	convolveCenter: static func (euclidTransforms: VectorList<This>, kernel: FloatVectorList) -> This {
+		result := This new()
+		if (euclidTransforms count > 0) {
+			result scaling = 0.0f
+			quaternions := VectorList<Quaternion> new(euclidTransforms count)
+			for (i in 0 .. euclidTransforms count) {
+				result translation = result translation + euclidTransforms[i] translation * kernel[i]
+				result scaling = result scaling + euclidTransforms[i] scaling * kernel[i]
+				rotation := euclidTransforms[i] rotation
+				quaternions add(rotation _quaternion)
+			}
+			result rotation = FloatRotation3D new(Quaternion weightedMean(quaternions, kernel))
+			quaternions free()
+		}
+		result
 	}
 	toString: func -> String {
 		"Translation: " << this translation toString() >> " Rotation: " & this rotation toString() >> " Scaling: " & this scaling toString()


source/math/FloatEuclidTransform.ooc
```diff
@@ -43,6 +46,22 @@ FloatEuclidTransform: cover {
...
+	convolveCenter: static func (euclidTransforms: VectorList<This>, kernel: FloatVectorList) -> This {
+		result := This new()
+		if (euclidTransforms count > 0) {
+			result scaling = 0.0f
+			quaternions := VectorList<Quaternion> new(euclidTransforms count)
+			for (i in 0 .. euclidTransforms count) {
+				result translation = result translation + euclidTransforms[i] translation * kernel[i]
+				result scaling = result scaling + euclidTransforms[i] scaling * kernel[i]
+				rotation := euclidTransforms[i] rotation
+				quaternions add(rotation _quaternion)
```
marcusnaslund: `quaternions add(euclidTransforms[i] rotation _quaternion)`?

niklasborwell: Nope. Tried a bunch of stuff to get that to work but the compiler doesn't accept it. 
Tried quaternions add((euclidTransforms[i] rotation) _quaternion), quaternions add((euclidTransforms[i] rotation) as FloatRotation3D _quaternion), same error.

marcusnaslund: @sebastianbaginski Since we just discussed something similar - do you know why this wouldn't work?

sebastianbaginski: No idea, sorry :) 
It happened to me few times as well, when I wanted to create a complicated property like
```ooc
property ::= object member anotherMember something()
```
rock refused to compile. I remember having to create few more properties like
```ooc
tempProperty ::= object member 
anotherProperty ::= tempProperty anotherMember
property ::= anotherProperty something()
```
I've never actually took the time to debug it so I don't really know what's the reason behind this.

marcusnaslund: Oh, this is probably related to #544

source/math/FloatEuclidTransform.ooc
```diff
@@ -43,6 +46,22 @@ FloatEuclidTransform: cover {
 		this rotation == other rotation &&
 		this scaling == other scaling
 	}
+	convolveCenter: static func (euclidTransforms: VectorList<This>, kernel: FloatVectorList) -> This {
+		result := This new()
+		if (euclidTransforms count > 0) {
+			result scaling = 0.0f
+			quaternions := VectorList<Quaternion> new(euclidTransforms count)
+			for (i in 0 .. euclidTransforms count) {
+				result translation = result translation + euclidTransforms[i] translation * kernel[i]
```
davidhesselbom: I'm tempted to optimize this (each `VectorList` access will result in a memcpy, and you're accessing the same elements 3 and 2 times in each iteration of the loop), but I guess it can wait until we see how much it actually affects performance...

niklasborwell: I agree we shouldn't do it now. This function will only be called once per frame with 7 elements at the most.

davidhesselbom: Oh. Well, in that case, nevermind, I guess.

marcusnaslund:
I'm happy here @davidhesselbom

davidhesselbom:
A test would have been nice... can you make a PR debt issue for it, and refer to this PR in its description?

niklasborwell:
Done.

source/math/FloatEuclidTransform.ooc
emilwestergren: I agree though, can't you just make sure you access `euclidTransforms` and `kernel` once? I can fix it for you if you don't want to.

davidhesselbom:
Thanks.

PR #724, opened 2015-11-10
Added cover for handling memory.

simonmika:
This PR is for illustrating a concept.

### DIFF
--- a/source/base/Memory.ooc
+++ b/source/base/Memory.ooc
@@ -1,0 +1,85 @@
+/*
+* Copyright (C) 2015 - Simon Mika <simon@mika.se>
+*
+* This sofware is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License as published by the Free Software Foundation; either
+* version 2.1 of the License, or (at your option) any later version.
+*
+* This software is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


source/base/Memory.ooc
```diff
@@ -0,0 +1,85 @@
...
+		this init(null, 0)
+	}
+	init: func@ ~fromSize (size: Int) {
+		this init(gc_malloc(size), size)
+	}
+	init: func@ ~fromData (=_pointer, =_size)
+	free: func@ -> Bool {
+		result: Bool
+		if (result = this _pointer != null)
+			gc_free(this _pointer)
```
davidhesselbom: And if we want to get rid of `gc_free`?

simonmika: That is just a name and as long as we don't change the compiler we will need to define it.

davidhesselbom: What do you mean we need to define it? Are you really sure about that?

marcusnaslund: Not true. We need some kind of free function of course, but we can name it however we like.

marcusnaslund:
I will close this for now - and revisit/reopen when we're not in a PR rush and when we've decided how we actually want to solve this (either with a `Memory` cover or by extending `Pointer`, which I think would prefer)

simonmika:
@sebastianbaginski You might be interested in this.

PR #725, opened 2015-11-10
Moved getStandardEquals() to a separate file

marcusnaslund:
These functions didn't belong in `HashMap.ooc` as an original comment mentioned.

Furthermore, this (along with other sdk PRs here) allows us to move `Hash*` classes out of the sdk, meaning we can move `Cell` out of the sdk, meaning we can solve the problem with `is equal to(Text)` in `Fixture`.

### DIFF
--- a/source/sdk/lang/equalities.ooc
+++ b/source/sdk/lang/equalities.ooc
@@ -1,8 +1,6 @@
 // The following functions were moved from HashMap.ooc
 
-/* this function seem is called by List */
 getStandardEquals: func <T> (T: Class) -> Func <T> (T, T) -> Bool {
-    // choose comparing function for key type
     if(T == String) {
         stringEquals
     } else if(T == CString) {
@@ -26,7 +24,6 @@ stringEquals: func <K> (k1, k2: K) -> Bool {
 cstringEquals: func <K> (k1, k2: K) -> Bool {
     k1 as CString == k2 as CString
 }
-
 
 pointerEquals: func <K> (k1, k2: K) -> Bool {
     k1 as Pointer == k2 as Pointer
--- a/source/sdk/lang/equalities.ooc
+++ b/source/sdk/lang/equalities.ooc
@@ -1,0 +1,47 @@
+// The following functions were moved from HashMap.ooc
+
+/* this function seem is called by List */
+getStandardEquals: func <T> (T: Class) -> Func <T> (T, T) -> Bool {
+    // choose comparing function for key type
+    if(T == String) {
+        stringEquals
+    } else if(T == CString) {
+        cstringEquals
+    } else if(T size == Pointer size) {
+        pointerEquals
+    } else if(T size == UInt size) {
+        intEquals
+    } else if(T size == Char size) {
+        charEquals
+    } else {
+        genericEquals
+    }
+}
+
+stringEquals: func <K> (k1, k2: K) -> Bool {
+    assert(K == String)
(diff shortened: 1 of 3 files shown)


source/sdk/lang/equalities.ooc
```diff
@@ -0,0 +1,47 @@
+// The following functions were moved from HashMap.ooc
```
davidhesselbom: I think the filename should have a capital `E`, but even then, wouldn't `Equals` make more sense?

davidhesselbom:
Forgive me, but, what does this have to do with `Cell`?

source/sdk/lang/equalities.ooc
```diff
@@ -0,0 +1,47 @@
...
+
+/* this function seem is called by List */
```
davidhesselbom: This comment...

source/sdk/lang/equalities.ooc
marcusnaslund: I chose not to since it isn't a class. This is the case in `source/net` for instance (although we didn't write that of course). Would you prefer a capital `E` anyways?

davidhesselbom: The only other examples I've been able to find (except things that were obviously not code) are either wrapper files (files that specify functions in other libraries, and I'm not sure it's correct in that case), shaders (not sure why, but I remember we decided to do that), data files (large, dumb arrays and no actual functions or anything). @simonmika, input please...

marcusnaslund:
Since we can't move `List`, we have to remove the dependancy on `HashMap` for us to move the `Hash*` classes (which use `Cell`).
Although I'm no longer sure we actually want `HashBag` or `Bag`, as they seem to be mostly wrappers around `HashMap` anyway (and that is what we have our `HashDictionary` for.

davidhesselbom:
`HashDictionary` uses `HashBag`... :) But, alright. Thanks.

marcusnaslund:
Yes - but it should use `HashMap` instead, otherwise we have a wrapper around a wrapper, sort of.

marcusnaslund:
Waiting on #693 to fix that though.

source/sdk/lang/equalities.ooc
marcusnaslund: Haha I just copied. I can remove it.

davidhesselbom:
#693 was merged, in case it helps.

marcusnaslund:
I won't fix it in this PR anyhow.

davidhesselbom:
Alright, just letting you know.

source/sdk/lang/equalities.ooc
davidhesselbom: @simonmika , a friendly reminder. Should the filename start with `E` or `e`?

simonmika: All folder names shall always start with lower case letters. Files named after the containing class are named with an initial upper case letter and the rest with a lower case letter. Does this make sense?

davidhesselbom: ... and since there are no classes here, just global functions, that means you want the `e`, not `E`. Fine.

marcusnaslund:
Updated per comments. This should be good to go then.

source/sdk/lang/equalities.ooc
simonmika: Yes, you are right. But then we don't want global functions. In this case let it slide as we don't want any of this code in the long run.

PR #728, opened 2015-11-10
Whitespace cleanup in Thread.ooc

marcusnaslund:
I don't have to remind you, but with `?w=1` you'll see that the only real changes are removal of empty lines and comments. Magic is now happy with this file.

### DIFF
--- a/source/sdk/threading/Thread.ooc
+++ b/source/sdk/threading/Thread.ooc
@@ -3,111 +3,49 @@ import native/[MutexUnix, MutexWin32]
 import native/[ThreadLocalUnix, ThreadLocalWin32]
 import native/[ConditionUnix, ConditionWin32]
 
-/**
- * A thread is a thread of execution in a program. Multiple threads
- * can run concurrently, allowing parallel computations to occur.
- *
- * However, since such threads are preemptible, synchronization issues
- * can occur, such as race conditions (two threads access the same memory
- * location in an interleaved manner, leaving an inconsistent state)
- * or
- */
 Thread: abstract class {
+	_code: Func
 
-    _code: Func
+	new: static func (._code) -> This {
+		version (unix || apple) {
+			return ThreadUnix new(_code) as This
+		}
+		version (windows) {
+			return ThreadWin32 new(_code) as This
+		}
+		Exception new(This, "Unsupported platform!\n") throw()
+		null
+	}
 
-    /**
-     * Create a new thread that will run a given function.
-     * Note that this only creates the thread - the start() method
-     * can be used to effectively start execution of this thread.
-     *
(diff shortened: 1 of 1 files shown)


source/sdk/threading/Thread.ooc
```diff
@@ -3,111 +3,49 @@ import native/[MutexUnix, MutexWin32]
...
-
-    /**
-     * Blocks until this thread has finished running (ie. join this thread).
-     *
-     * @return true if the thread has successfully joined, ie. if it's finished running.
-     * otherwise, false - the thread is not joinable, is already dead, a deadlock was
-     * detected, etc.
-     */
-    wait: abstract func -> Bool
+	new: static func (._code) -> This {
```
marcusnaslund: Why the dot? I find no documentation on this.

davidhesselbom: There's this: http://oocmanual.cogneco.com/language/constructors/#toc_1

Admittedly, no one uses that site. It should probably updated (with proper syntax highlighting and stuff)...

source/sdk/threading/Thread.ooc
```diff
@@ -3,111 +3,49 @@ import native/[MutexUnix, MutexWin32]
...
+		version (unix || apple) {
+			return ThreadUnix new(_code) as This
```
davidhesselbom: If `this` was enforced when referring to member variables, it would be more obvious that this refers to the incoming argument to the function, and not the member variable.

source/sdk/threading/Thread.ooc
marcusnaslund: That makes sense!

marcusnaslund:
Do you feel like any of the comments removed here should be kept @sebastianbaginski ?

sebastianbaginski:
@marcusnaslund I don't think so. This one is quite low-level, so anyone using this should know what he's doing and how things work. It's nice to have the comment about `RecursiveMutex`, though.

marcusnaslund:
Thanks. Then we're good to go here @davidhesselbom - no side-effects since there is no actual change in code.

source/sdk/threading/Thread.ooc
```diff
@@ -255,56 +135,45 @@ RecursiveMutex: abstract class extends Mutex {
...
+	}
 
-    /**
-        Set the data.
-     */
-    set: abstract func (value: T)
+	set: abstract func (value: T)
 
-    /**
-        Get the data, obviously.
```
davidhesselbom: This is what happens when people learn Java in school...

marcusnaslund: Haha :)

PR #730, opened 2015-11-11
Updated reference counter with safe/unsafe version

sebastianbaginski:
To change behaviour of existing ReferenceCounter:
`referenceCounter isSafe = true / false`
For now `safe` is default, as it was before.

### DIFF
--- a/source/base/ReferenceCounter.ooc
+++ b/source/base/ReferenceCounter.ooc
@@ -1,20 +1,30 @@
 import threading/Thread
 import Synchronized
 
-ReferenceCounter: class extends Synchronized {
+ReferenceCounter: class {
 	_target: Object
 	count: static Int = 0
 	_count: Int = 0
 	_kill := false
-	init: func (target: Object) {
-		super()
+	_lock: Mutex
+	isSafe: Bool {
+		get { !this _lock instanceOf?(MutexUnsafe) }
+		set(value) {
+			if (this isSafe != value) {
+				this _lock free()
+				this _lock = Mutex new(value ? MutexType Safe : MutexType Unsafe)
+			}
+		}
+	}
+	init: func (target: Object, mutexType := MutexType Safe) {
 		This count += 1
 		this _target = target
+		this _lock = Mutex new(mutexType)
 //		"  RC, #{This count}, #{this _target class name}" println()
 	}
 	update: func (delta: Int) {
 		if (delta != 0) {
-			this lock()
+			this _lock lock()
 			target := null
 			if (!this _kill) {
 				this _count += delta
@@ -23,7 +33,7 @@ ReferenceCounter: class extends Synchronized {
 				if (this _kill)
 					target = this _target
 			}
-			this unlock()
+			this _lock unlock()
 			if (target != null) {
 				this _count = 0
 				this _kill = false
@@ -34,6 +44,7 @@ ReferenceCounter: class extends Synchronized {
 	free: override func {
 		This count -= 1
 //		"--RC, #{This count}, #{this _target class name}" println()
+		this _lock free()
 		super()
 	}
 	increase: func { this update(1) }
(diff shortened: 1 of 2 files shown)


source/base/ReferenceCounter.ooc
```diff
@@ -1,20 +1,29 @@
...
 	count: static Int = 0
 	_count: Int = 0
 	_kill := false
+	_lock := Mutex new(MutexType Safe)
+	isSafe: Bool {
+		get { !this _lock instanceOf?(MutexUnsafe) }
+		set(value) {
+			if (this isSafe != value) {
+				this _lock free()
+				this _lock = Mutex new(value ? MutexType Safe : MutexType Unsafe)
```
davidhesselbom: Maybe this is a dumb question (still brewing my coffee), but, would anyone ever want to use the global mutex here?

source/base/ReferenceCounter.ooc
```diff
@@ -1,20 +1,29 @@
 import threading/Thread
 import Synchronized
 
-ReferenceCounter: class extends Synchronized {
+ReferenceCounter: class {
```
davidhesselbom: I take it Simon is in on this?

source/base/ReferenceCounter.ooc
sebastianbaginski: Maybe for testing. But I don't think it's worth to include this in production code.

davidhesselbom: Alright, just checking.

source/base/ReferenceCounter.ooc
sebastianbaginski: Yes, we discussed this.

source/base/ReferenceCounter.ooc
```diff
@@ -1,20 +1,29 @@
...
 	_target: Object
 	count: static Int = 0
 	_count: Int = 0
 	_kill := false
+	_lock := Mutex new(MutexType Safe)
```
davidhesselbom: So... if you want an unsafe one, you need to first create the `ReferenceCounter`, then call `isSafe(false)`?

sebastianbaginski: Yes, `referenceCounter isSafe = false`

davidhesselbom: How about we modify the constructor so it takes a bool that defaults to true instead? That way the Mutex won't have to be created just to be destroyed immediately afterwards...

sebastianbaginski: Why not. I changed it to accept the `MutexType`, so now it's possible to use global mutex easily just by changing the default value in constructor.

davidhesselbom: Okay, that's even better. Thanks!

source/base/ReferenceCounter.ooc
```diff
@@ -1,20 +1,30 @@
...
 	_count: Int = 0
 	_kill := false
-	init: func (target: Object) {
-		super()
+	_lock: Mutex
+	isSafe: Bool {
+		get { !this _lock instanceOf?(MutexUnsafe) }
+		set(value) {
+			if (this isSafe != value) {
+				this _lock free()
```
davidhesselbom: I just realized that... now that this class can be initialized with a `MutexGlobal`, if anyone actually *does* initialize one with a `MutexGlobal`, and then calls `isSafe = true` *or* `isSafe =false`, this line might cause problems... I'm not sure what happens if you try to call `free` on a `MutexGlobal`.

sebastianbaginski: Nothing, look here
```
MutexGlobal: class extends Mutex {
	_globalMutex := static Mutex new(MutexType Safe)
	init: func
	lock: override func {
		This _globalMutex lock()
	}
	unlock: override func {
		This _globalMutex unlock()
	}
	cleanup: static func {
		This _globalMutex free()
	}
}
```
`free` function does nothing with the global instance of the mutex

davidhesselbom: You're right. A very strange function...

PR #731, opened 2015-11-11
Cleanup in Image interface, method to apply 2D transform to RasterBgr

sebastianbaginski:
Cleaned up unused functions from `Image` interface.
Added method to transform raster image with `FloatTransform2D` with test cases.

### DIFF
--- a/source/draw/Image.ooc
+++ b/source/draw/Image.ooc
@@ -73,18 +73,18 @@ Image: abstract class {
 		this _canvas = null
 		super()
 	}
-	resizeWithin: func (restriction: IntSize2D) -> This {
-		this resizeTo(((this size toFloatSize2D()) * Float minimum(restriction width as Float / this size width as Float, restriction height as Float / this size height as Float)) toIntSize2D())
-	}
+	create: virtual func (size: IntSize2D) -> This { raise("Image type not implemented."); null }
+	copy: abstract func -> This
+	distance: virtual abstract func (other: This) -> Float
+	equals: func (other: This) -> Bool { this size == other size && this distance(other) < 10 * Float epsilon }
 	resizeTo: abstract func (size: IntSize2D) -> This
 	resizeTo: virtual func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
 		this resizeTo(size)
 	}
-	create: virtual func (size: IntSize2D) -> This { raise("Image type not implemented."); null }
-	copy: abstract func -> This
-	copy: abstract func ~fromParams (size: IntSize2D, transform: FloatTransform2D) -> This
-	distance: virtual abstract func (other: This) -> Float
-	equals: func (other: This) -> Bool { this size == other size && this distance(other) < 10 * Float epsilon }
+	transformed: virtual func (transform: FloatTransform2D, method := TransformMethod Smooth) -> This {
+		raise("transformed() not implemented!")
+		null
+	}
 	isValidIn: func (x, y: Int) -> Bool {
 		return (x >= 0 && x < this size width && y >= 0 && y < this size height)
 	}
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -153,6 +153,26 @@ RasterBgr: class extends RasterPacked {
 		result swapRedBlue()
 		result
 	}
+	transformed: func (transform: FloatTransform2D, method := TransformMethod Smooth) -> This {
+		result: This
+		if (transform isIdentity)
+			result = this copy()
+		else {
+			inverse := transform inverse
+			box := FloatBox2D new(FloatPoint2D new(), this size toFloatSize2D())
+			topLeft := transform * box leftTop
+			topRight := transform * box rightTop
+			bottomLeft := transform * box leftBottom
+			bottomRight := transform * box rightBottom
+			result = This new(IntSize2D new(topLeft distance(topRight), topLeft distance(bottomLeft)))
+			for (row in 0 .. result height)
+				for (column in 0 .. result width) {
+					position := (inverse * FloatPoint2D new(column, row)) toIntPoint2D()
+					result[column, row] = this[position x, position y]
+				}
+		}
+		result
+	}
 	_createCanvas: override func -> Canvas { BgrRasterCanvas new(this) }
 	kean_draw_rasterBgr_new: static unmangled func (width, height, stride: Int, data: Void*) -> This {
 		result := This new(IntSize2D new(width, height), stride)
--- a/source/draw/RasterImage.ooc
+++ b/source/draw/RasterImage.ooc
@@ -38,26 +38,6 @@ RasterImage: abstract class extends Image {
 		result
 	}
 	copy: abstract func -> This
-	copy: func ~fromParams (size: IntSize2D, transform: FloatTransform2D) -> This {
-		transform = (this transform toFloatTransform2D()) * transform * (this transform toFloatTransform2D()) inverse
-		mappingTransform := FloatTransform2D createTranslation(this size width / 2, this size height / 2) * transform
-		upperLeft := mappingTransform * FloatPoint2D new(-size width / 2, -size width / 2)
-		upperRight := mappingTransform * FloatPoint2D new(size width / 2, -size width / 2)
-		downLeft := mappingTransform * FloatPoint2D new(-size width / 2, size width / 2)
-		downRight := mappingTransform * FloatPoint2D new(size width / 2, size width / 2)
-		source := FloatBox2D bounds([upperLeft, upperRight, downLeft, downRight])
-		mappingTransformInverse := mappingTransform inverse
-		upperLeft = mappingTransformInverse * source leftTop
-		upperRight = mappingTransformInverse * source rightTop
-		downLeft = mappingTransformInverse * source leftBottom
-		downRight = mappingTransformInverse * source rightBottom
-		this copy(size toFloatSize2D(), source, FloatPoint2D new(), FloatPoint2D new(), FloatPoint2D new())
-	}
-	copy: func ~fromMoreParams (size: FloatSize2D, source: FloatBox2D, upperLeft, upperRight, lowerLeft: FloatPoint2D) -> This {
-		result := RasterBgra new(size ceiling() toIntSize2D())
-//		TODO: The stuff
-		result
-	}
 	open: static func ~unknownType (filename: String) -> This {
 		x, y, imageComponents: Int
 		data := StbImage load(filename, x&, y&, imageComponents&, 0)
--- a/test/draw/RasterBgrTest.ooc
+++ b/test/draw/RasterBgrTest.ooc
@@ -63,6 +63,25 @@ RasterBgrTest: class extends Fixture {
 			image2 free()
 			output free()
 		})
+		this add("transformed", func {
+			outputScaled := "test/draw/output/RasterBgrTest_rescaled.png"
+			outputMirrored := "test/draw/output/RasterBgrTest_mirrored.png"
+			outputRotated := "test/draw/output/RasterBgrTest_rotated.png"
+			image := RasterBgr open(this sourceFlower)
+			result := image transformed(FloatTransform2D createScaling(0.55, 1.5))
+			result save(outputScaled)
+			result referenceCount decrease()
+			result = image transformed(FloatTransform2D createTranslation(image width / 2, image height / 2) * FloatTransform2D createReflectionY() * FloatTransform2D createTranslation(-image width / 2, -image height / 2))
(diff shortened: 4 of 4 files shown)


source/draw/RasterBgr.ooc
```diff
@@ -153,6 +153,26 @@ RasterBgr: class extends RasterPacked {
...
+			topLeft := transform * box leftTop
+			topRight := transform * box rightTop
+			bottomLeft := transform * box leftBottom
+			bottomRight := transform * box rightBottom
+			result = This new(IntSize2D new(topLeft distance(topRight), topLeft distance(bottomLeft)))
+			for (row in 0 .. result height)
+				for (column in 0 .. result width) {
+					position := (inverse * FloatPoint2D new(column, row)) toIntPoint2D()
+					result[column, row] = this[position x, position y]
+				}
```
sebastianbaginski: I know it's not optimized. This is only to cleanup and extend the interface for raster images. I will be working on this more.

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/574

source/draw/RasterImage.ooc
```diff
@@ -38,26 +38,6 @@ RasterImage: abstract class extends Image {
...
-		mappingTransformInverse := mappingTransform inverse
-		upperLeft = mappingTransformInverse * source leftTop
-		upperRight = mappingTransformInverse * source rightTop
-		downLeft = mappingTransformInverse * source leftBottom
-		downRight = mappingTransformInverse * source rightBottom
-		this copy(size toFloatSize2D(), source, FloatPoint2D new(), FloatPoint2D new(), FloatPoint2D new())
-	}
-	copy: func ~fromMoreParams (size: FloatSize2D, source: FloatBox2D, upperLeft, upperRight, lowerLeft: FloatPoint2D) -> This {
-		result := RasterBgra new(size ceiling() toIntSize2D())
-//		TODO: The stuff
```
marcusnaslund: This is my favourite comment

test/draw/RasterBgrTest.ooc
```diff
@@ -63,6 +63,25 @@ RasterBgrTest: class extends Fixture {
 			image2 free()
 			output free()
 		})
+		this add("transformed", func {
+			output := "test/draw/output/RasterBgrTest_rescaled.png"
+			outputScaled := "test/draw/output/RasterBgrTest_mirrored.png"
```
marcusnaslund: Your filenames don't seem to reflect the variable names.. is this correct?

test/draw/RasterBgrTest.ooc
```diff
@@ -63,6 +63,25 @@ RasterBgrTest: class extends Fixture {
...
+			outputRotated := "test/draw/output/RasterBgrTest_rotated.png"
+			image := RasterBgr open(this sourceFlower)
+			result := image transformed(FloatTransform2D createScaling(0.55, 1.5))
+			result save(output)
+			result referenceCount decrease()
+			result = image transformed(FloatTransform2D createTranslation(image width / 2, image height / 2) * FloatTransform2D createReflectionY() * FloatTransform2D createTranslation(- image width / 2, - image height / 2))
```
marcusnaslund: No space after `-`.

test/draw/RasterBgrTest.ooc
```diff
@@ -63,6 +63,25 @@ RasterBgrTest: class extends Fixture {
...
+			result save(outputScaled)
+			result referenceCount decrease()
+			result = image transformed(FloatTransform2D createTranslation(image width / 2, image height / 2) * FloatTransform2D createZRotation(3.14 / 4) * FloatTransform2D createTranslation(- image width / 2, - image height / 2))
```
marcusnaslund: Same here.

test/draw/RasterBgrTest.ooc
sebastianbaginski: No :)

source/draw/RasterImage.ooc
davidhesselbom: I'll take the credit for that ^^

source/draw/Image.ooc
```diff
@@ -73,18 +73,18 @@ Image: abstract class {
...
 	create: virtual func (size: IntSize2D) -> This { raise("Image type not implemented."); null }
 	copy: abstract func -> This
-	copy: abstract func ~fromParams (size: IntSize2D, transform: FloatTransform2D) -> This
 	distance: virtual abstract func (other: This) -> Float
 	equals: func (other: This) -> Bool { this size == other size && this distance(other) < 10 * Float epsilon }
+	resizeTo: abstract func (size: IntSize2D) -> This
+	resizeTo: virtual func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
+		this resizeTo(size)
+	}
+	transfrormed: virtual func (transform: FloatTransform2D, method := TransformMethod Smooth) -> This {
```
marcusnaslund: `transfrormed`?

source/draw/Image.ooc
```diff
@@ -73,18 +73,18 @@ Image: abstract class {
...
+		raise("transfrormed() not implemented!")
```
marcusnaslund: Incidentally, the same misspelling here

sebastianbaginski: Haha autocompletion is not awlays that good :)

source/draw/Image.ooc
```diff
@@ -73,18 +73,18 @@ Image: abstract class {
...
 	create: virtual func (size: IntSize2D) -> This { raise("Image type not implemented."); null }
 	copy: abstract func -> This
-	copy: abstract func ~fromParams (size: IntSize2D, transform: FloatTransform2D) -> This
 	distance: virtual abstract func (other: This) -> Float
 	equals: func (other: This) -> Bool { this size == other size && this distance(other) < 10 * Float epsilon }
+	resizeTo: abstract func (size: IntSize2D) -> This
+	resizeTo: virtual func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
+		this resizeTo(size)
+	}
+	transformed: virtual func (transform: FloatTransform2D, method := TransformMethod Smooth) -> This {
```
emilwestergren: Shouldn't this be called `transform`?

davidhesselbom: There's already a property called `transform`...

sebastianbaginski: ... and this function returns a copy. `transform` sounds to me more like a verb (or a synonym for matrix).
`newImage := image transformed(matrix)` or `newImage := image transform(matrix)`, I think first one looks better.

emilwestergren: Yes I know, but a function named `transformed` ? Then `copy` should be `copied`, `resizeTo` `resizedTo` etc. Isn't there anything better?

emilwestergren: Just looking for consistency.

sebastianbaginski: I understand. Actually I do think that `resizeTo` should be called `scaled` :)

sebastianbaginski: I don't have any better ideas, so I welcome any suggestions.

marcusnaslund: `transformUsing`?

emilwestergren: I actually don't like the current `transform` property. It looks like its only purpose is to define the coordinate system. Shouldn't it be more specific like `coordinateTransform` or something? It would differentiate it from the `transform` in `Canvas` as well. In that case, `transform` would be available for this function.

Other suggestions:
* `applyTransform`
* `getTransformed`

sebastianbaginski: In your above examples I think `copy` should be left alone:
- `newImage := oldImage transformed(matrix)` as in "newImage *is* oldImage transformed by the matrix"
- `newImage := oldImage resizedTo(size)` as in "newImage *is* oldImage resizedTo to size"
- `newImage := oldImage copy()` as in "newImage *is* a copy of oldImage"
or did I misunderstood the meaning of `copied` ?
Or are we supposed to use verbs here ?

sebastianbaginski: @emilwestergren the transform property is used only on Android. Do you think you could get rid of it ?

emilwestergren: Why remove it? Then we can't handle images with different coordinate systems.

davidhesselbom: English is a terrible language for programming, all things considered.

sebastianbaginski: Well you still have the other property in the Image. And this is actually something Simon has suggested when I asked about what should I do with it. He said something like "remove it and let them worry about it". I just changed it to read only then.

emilwestergren: Oh I understand. Yes there is no need for this transform, the Android code needs to use the `coordinateSystem` enum instead, that's all.

emilwestergren: It's used in the `Raster` code also though, so someone should create a PR which removes it and adapts all the code.

davidhesselbom: So... my understanding is, we merge it even if it's stupid, we make an issue about fixing it later, and we fix it later?

emilwestergren: That's your call, but API definitions are probably the sort of things that shouldn't apply to the "merge and fix later" mentality since it can get messy to change in the future. A bad internal variable name is much easier to fix than a function used in 100 different places.

davidhesselbom: That's why I want to get those things right the first time, but apparently it takes too much time.

emilwestergren: @simonmika help

simonmika: For the purpose of creating a new image that is a transformed version of the previous one should create a target image and render to it's canvas. Something along these lines:
```
(result := image create()) canvas render(image, transform)
```

source/draw/Image.ooc
```diff
@@ -73,18 +73,18 @@ Image: abstract class {
 		this _canvas = null
 		super()
 	}
-	resizeWithin: func (restriction: IntSize2D) -> This {
```
simonmika: This is a very useful function, good for saving down a small image and keep the aspect ratio intact. I see no reason to remove it and to keep `resizeTo`.

source/draw/Image.ooc
```diff
@@ -73,18 +73,18 @@ Image: abstract class {
...
-		this resizeTo(((this size toFloatSize2D()) * Float minimum(restriction width as Float / this size width as Float, restriction height as Float / this size height as Float)) toIntSize2D())
-	}
-	resizeTo: abstract func (size: IntSize2D) -> This
-	resizeTo: virtual func ~withMethod (size: IntSize2D, method: TransformMethod) -> This {
```
simonmika: When rendering with a interpolation method it should be done using a canvas. `resizeTo` and `resizeWithin` are only crude potentially slow methods for simplifying debugging.

source/draw/RasterImage.ooc
```diff
@@ -38,26 +38,6 @@ RasterImage: abstract class extends Image {
 		result
 	}
 	copy: abstract func -> This
-	copy: func ~fromParams (size: IntSize2D, transform: FloatTransform2D) -> This {
```
simonmika: No copy with transform, remove.

sebastianbaginski: I don't understand the comment, it is removed.

source/draw/Image.ooc
sebastianbaginski: That's quite unintuitive api, if you ask me. I don't know how to implement it.

source/draw/Image.ooc
sebastianbaginski: OK, but in the end both `canvas` and `resizeTo` would use the same algorithm for CPU images, or ? How could one be faster than the other ? Is there something about the canvas that I'm missing

source/draw/Image.ooc
simonmika: If you want to write it on many lines:
```ooc
result := image create()
result canvas render(image, transform)

source/draw/Image.ooc
simonmika: They could have differently optimized methods of doing it. Let's just remove all resize methods and put them into canvas for now.

source/draw/RasterImage.ooc
simonmika: Yes, good.

source/draw/Image.ooc
sebastianbaginski: So the api for resizing an image will be ... ... instead of `newImage := image resizeTo(size)` (can you fill the dots)

bettinaselig:
Just a reminder that the original milestone of this pull request was "Clean up".
@sebastianbaginski

sebastianbaginski:
Closing this one, we're gonna split it up a little bit.

PR #733, opened 2015-11-11
Fixed default color for Pen

sebastianbaginski:
This fixes overlay display in `gpuOff` version of the other project.

### DIFF
--- a/test/draw/gpu/GpuImageRotationTest.ooc
+++ b/test/draw/gpu/GpuImageRotationTest.ooc
@@ -45,6 +45,7 @@ GpuImageRotationTest: class extends Fixture {
 		this add("GPU rotation small X (BGRA)", func {
 			correctImage := RasterBgra open("test/draw/gpu/correct/rotation_small_bgra_X.png")
 			gpuImage := gpuContext createBgra(sourceImage size)
+			gpuImage canvas pen = Pen new(ColorBgra new())
 			gpuImage canvas clear()
 			gpuImage canvas focalLength = focalLength
 			gpuImage canvas transform = FloatTransform3D createRotationX(smallRotation)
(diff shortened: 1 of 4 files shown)


sebastianbaginski:
Yeah so looks like GPU images expects black pen with 0 alpha as default ? @emilwestergren

emilwestergren:
Alpha is not implemented yet for YUV so it doesn't matter.

sebastianbaginski:
OK, but almost every GPU image test failed here. Does it have anything to do with default values for `Pen` ?

emilwestergren:
The tests probably expect black as default, yes.

sebastianbaginski:
So we set black as default then ? I only need it to have non zero alpha channel anyway.

emilwestergren:
Sure

sebastianbaginski:
OK, updated.

sebastianbaginski:
Something is still wrong with the tests, even if the default color is set to zero. The only difference now is the alpha channel default value for Pen. Should I explicitly assign the Pen with zero alpha in the GPU tests, or is there a better solution ?

emilwestergren:
That's strange. You could run the tests locally and see how the output differs from the correct image.

emilwestergren:
Oh wait, I forgot the tests use BGRA images, that's why it fails. Are you sure we want white as default color?

sebastianbaginski:
I that case I will update the gpu tests.

sebastianbaginski:
Updated. It is ready.

bettinaselig:
@jonathanudd is doing peer review.

test/draw/gpu/GpuImageRotationTest.ooc
```diff
@@ -45,6 +45,7 @@ GpuImageRotationTest: class extends Fixture {
 		this add("GPU rotation small X (BGRA)", func {
 			correctImage := RasterBgra open("test/draw/gpu/correct/rotation_small_bgra_X.png")
 			gpuImage := gpuContext createBgra(sourceImage size)
+			gpuImage canvas pen = Pen new(ColorBgra new())
```
emilwestergren: How did this make any difference? I thought the pen was already defaulted to `ColorBgra new()` in the Canvas?

jonathanudd:
The code looks good, I have nothing to add.

test/draw/gpu/GpuImageRotationTest.ooc
sebastianbaginski: Maybe because Canvas defaults to `Pen new() // no alpha` which now defaults to `Pen new(ColorBgra new(0, 0, 0, 255)) // alpha == 1.0` ?

PR #734, opened 2015-11-11
Automatically add correct flag for nogpu

### LINKED ISSUE #726: test.sh should add DgpuOff to flags when nogpu
Something like 
```
if [[ $1 == "nogpu" ]]; then
	FEATURES=$FEATURES" -DgpuOff"
fi
```
is missing from the script.

marcusnaslund:
Not sure if this should have the `bug` flag, but this fixes #726

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -20,6 +20,10 @@ case "$1" in
 		TESTS="./test/$1"
 		;;
 esac
+FEATURES=""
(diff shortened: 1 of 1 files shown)


marcusnaslund:
We could add `-DgpuOff` to `FLAGS` as well, but that's not how we do it in the other project for some reason.

PR #736, opened 2015-11-11
Added backend map for GpuMap

emilwestergren:
* GpuMaps now have a backend map for binding values to shader variables
* This PR only adds the functionality, adapted code using the map will come later
* This PR should be fully compatible with existing code

### DIFF
--- a/source/draw/gpu/GpuMap.ooc
+++ b/source/draw/gpu/GpuMap.ooc
@@ -14,19 +14,75 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this _program. If not, see <http://www.gnu.org/licenses/>.
 use ooc-math
+use ooc-collections
 
+_MapEntry: class {
+	key: String
+	value: Object
+	init: func (=key, =value)
+	free: override func {
+		if (this value instanceOf?(Cell))
+			this value free()
+		super()
+	}
+}
+
+_Map: class {
+	_entries := VectorList<_MapEntry> new()
+	init: func
+	free: override func {
+		this _entries free()
+		super()
+	}
+	add: func (key: String, value: Object) {
+		entry := this get(key)
+		if (entry != null) {
+			if (entry value instanceOf?(Cell))
+				entry value free()
+			entry value = value
+		}
+		else
+			this _entries add(_MapEntry new(key, value))
+	}
+	get: func (key: String) -> _MapEntry {
+		result: _MapEntry = null
+		for (i in 0 .. this _entries count)
+			if (this _entries[i] key == key)
+				result = this _entries[i]
+		result
+	}
+	apply: func (action: Func (String, Object)) {
+		for (i in 0 .. this _entries count) {
+			entry := this _entries[i]
+			action(entry key, entry value)
+		}
+	}
+}
 version(!gpuOff) {
 GpuMap: abstract class {
 	model: FloatTransform3D { get set }
 	view: FloatTransform3D { get set }
 	projection: FloatTransform3D { get set }
 	textureTransform: FloatTransform3D { get set }
+	_bindings := _Map new()
 	init: func {
 		this model = FloatTransform3D identity
 		this view = FloatTransform3D identity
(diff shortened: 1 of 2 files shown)


source/draw/gpu/GpuMap.ooc
```diff
@@ -14,19 +14,76 @@
...
+	}
+	get: func (key: String) -> _MapEntry {
+		result: _MapEntry = null
+		for (i in 0 .. this _entries count) {
+			if (this _entries[i] key == key)
+				result = this _entries[i]
+		}
+		result
+	}
+	foreach: func (action: Func (String, Object)) {
```
davidhesselbom: Isn't this what's usually called `apply`?

emilwestergren: Sure, it's probably more consistent.

emilwestergren:
Updated.

source/draw/gpu/GpuMap.ooc
```diff
@@ -14,19 +14,76 @@
...
+			if (entry value instanceOf?(Cell))
+				entry value free()
+			entry value = value
+		}
+		else
+			this _entries add(_MapEntry new(key, value))
+	}
+	get: func (key: String) -> _MapEntry {
+		result: _MapEntry = null
+		for (i in 0 .. this _entries count) {
```
marcusnaslund: You don't need these brackets

source/draw/gpu/opengl/Map/OpenGLMap.ooc
```diff
@@ -56,7 +56,44 @@ OpenGLMap: abstract class extends GpuMap {
...
 	}
-	use: override func { this program use() }
+	use: override func {
+		textureCount := 0
+		action := func (key: String, value: Object) {
+			program := this program
+			if (value instanceOf?(Cell)) {
+				cell := value as Cell
+				match (cell T) {
+					case Int => program setUniform(key, cell as Cell<Int> get())
```
marcusnaslund: Do you really need the explicit `as Cell<Int>` cast?

marcusnaslund: Maybe `cell get() as Int` would look better in that case

emilwestergren: I don't think `get()` works unless the type is known.

davidhesselbom: I agree `cell get() as Int` looks better. Does it work?

marcusnaslund: You're right, nvm

source/draw/gpu/opengl/Map/OpenGLMap.ooc
```diff
@@ -56,7 +56,44 @@ OpenGLMap: abstract class extends GpuMap {
...
+					case FloatPoint3D => program setUniform(key, cell as Cell<FloatPoint3D> get())
+					case FloatPoint4D => program setUniform(key, cell as Cell<FloatPoint4D> get())
+					case FloatSize2D => program setUniform(key, cell as Cell<FloatSize2D> get())
+					case FloatSize3D => program setUniform(key, cell as Cell<FloatSize3D> get())
+					case FloatTransform2D => program setUniform(key, cell as Cell<FloatTransform2D> get())
+					case FloatTransform3D => program setUniform(key, cell as Cell<FloatTransform3D> get())
+					case => Debug raise("Invalid cover type in OpenGLMap use!")
+				}
+			}
+			else {
```
marcusnaslund: Either remove the brackets or do `} else {`

niklasborwell:
I have peer reviewed this pull request. The issue with these kind of very specific pull requests is that I don't really grasp what it will do for the code. What I can do is basically check for smaller errors, so that's what I did. Looks good to me.

marcusnaslund:
Validated.

davidhesselbom:
Hard to say how well this will work, but I don't see a problem so far.

PR #737, opened 2015-11-11
First version of new CLI argument parser

jonathanudd:


marcusnaslund:
Is there going to be something else in `source/cli` or are we making a folder for just one class?

test/cli/ArgumentParserTest.ooc
```diff
@@ -0,0 +1,97 @@
...
+			expect(valuea, is equal to(t"a"))
+			expect(valueb, is equal to(t"b"))
+			inputList free()
+			parser free()
+			valuea free()
+			valueb free()
+		})
+		this add("With parameter", func {
+			inputList := VectorList<Text> new()
+			inputList add(t"--arga")
```
marcusnaslund: Actually, `Fixture` doesn't support `is equal to()` with a `Text` argument. If it's not equal to, the whole test will crash. Better to use `==` for now.

ooc-cli.use
```diff
@@ -0,0 +1,6 @@
+Name: ooc CLI
+Description: CLI helpers
+SourcePath: source/cli
+Requires: ooc-collections
```
marcusnaslund: What does this do? Do you still need `use ooc-collections` in `ArgumentParser.ooc`?

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,109 @@
...
+				}
+				else
+					argument _action1 call()
+			}
+		}
+		parameters free()
+	}
+	_findArgument: func (identifier: Text) -> Argument{
+		result: Argument
+		for (i in 0 .. this _arguments count) {
```
marcusnaslund: You don't need these brackets

test/cli/ArgumentParserTest.ooc
```diff
@@ -0,0 +1,97 @@
...
+			inputList add(t"-b")
+			parser := ArgumentParser new()
+			valuea: Text
+			valueb: Text
+			parser add(t"arga", t"a", 0, Event new(func { valuea = t"a" }))
+			parser add(t"argb", t"b", 0, Event new(func { valueb = t"b" }))
+			parser parse(inputList)
+			expect(valuea, is equal to(t"a"))
```
marcusnaslund: Actually, `Fixture` doesn't support `is equal to()` with a `Text` argument. If it's not equal, the whole test will crash. Better to use `==` for now.

test/cli/ArgumentParserTest.ooc
```diff
@@ -0,0 +1,97 @@
...
+			inputList add(t"c1")
+			inputList add(t"c2")
+			inputList add(t"c3")
+			parser := ArgumentParser new()
+			valuea1: Text
+			valuea2: Text
+			valueb: Text
+			valuec1: Text
+			valuec2: Text
+			valuec3: Text
```
marcusnaslund: Please find better variable names than these. :)

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,109 @@
...
+					argument _action2 call(input[i + 1])
+					i = i + 1
+				}
+				else
+					argument _action1 call()
+			}
+		}
+		parameters free()
+	}
+	_findArgument: func (identifier: Text) -> Argument{
```
marcusnaslund: Magic should catch this, @thomasfanell 
Anyway, you're missing a space before `{`

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,109 @@
...
+		parameters := VectorList<Text> new()
+		for (i in 0 .. input count) {
+			identifier = t""
+			argument = null
+			parameters clear()
+			if (input[i] beginsWith(t"--"))
+				identifier = input[i] slice(2, input[i] count-2)
+			else if (input[i] beginsWith(t"-"))
+				identifier = input[i] slice(1, input[i] count-1)
+			else {
```
marcusnaslund: No brackets needed here

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,109 @@
...
+		this _arguments free()
+		super()
+	}
+	add: func (longIdentifier: Text, shortIdentifier: Text, parameters: Int, action: Event) {
+		this _arguments add(Argument new(longIdentifier, shortIdentifier, parameters, action))
+	}
+	add: func ~parameter (longIdentifier: Text, shortIdentifier: Text, parameters: Int, action: Event1<Text>) {
+		this _arguments add(Argument new(longIdentifier, shortIdentifier, parameters, action))
+	}
+	add: func ~parameters (longIdentifier: Text, shortIdentifier: Text, parameters: Int, action: Event1<VectorList<Text>>) {
```
marcusnaslund: `longIdentifier, shortIdentifier: Text`

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,109 @@
...
+* along with this software. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use ooc-collections
+use ooc-base
+import Event
+
+Argument: class {
+	_longIdentifier: Text
+	_shortIdentifier: Text
```
marcusnaslund: These don't seem to ever be freed

jonathanudd:
At the moment we are making a folder for only one class.

ooc-cli.use
jonathanudd: The only thing it does for us is indicating what other module this module depends on. I think it is used by a tool called [sam](https://ooc-lang.org/docs/tools/sam/).

jonathanudd:
@bettinaselig is going to do a peer review.

source/base/Text.ooc
```diff
@@ -134,6 +134,13 @@ Text: cover {
 	free: func@ ~withCriteria (criteria: Owner) -> Bool {
 		this _buffer free(criteria)
 	}
+	slice: func ~untilEnd (start: Int) -> This {
+		result := This new(this _buffer slice(start))
+		if (this _buffer owner == Owner Receiver)
+			result = result copy() // TODO: Could we be smarter here?
```
bettinaselig: We try to avoid having TODOs in the code. 
@marcusnaslund @sebastianbaginski Is that good enough?

marcusnaslund: This comment isn't new (look at the old line 140), so I'm OK with it.

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,112 @@
...
+use ooc-base
+import Event
+
+Argument: class {
+	_longIdentifier: Text
+	_shortIdentifier: Char
+	_parameters: Int
+	_action1: Event
+	_action2: Event1<Text>
+	_action3: Event1<VectorList<Text>>
```
bettinaselig: We usually do not have numbered variable names. 
What does @simonmika say about this?

source/base/Text.ooc
bettinaselig: Ok. Why not make an issue out if it, if it is a TODO?

marcusnaslund: Agreed. I will create one.

source/cli/ArgumentParser.ooc
marcusnaslund: Or @davidhesselbom ?

ooc-cli.use
davidhesselbom: I think we should very seriously consider removing this clause from all of our `use` files.

marcusnaslund: I'll put it on my list and we'll discuss it later

source/cli/ArgumentParser.ooc
davidhesselbom: We do in `Event`, obviously, but I think we can do better here, since the number doesn't describe the number of parameters, but a degree of complexity. So, how about `action`, `textAction`, `listAction`?

jonathanudd: Good suggestions on names, they are changed now.

source/base/Text.ooc
marcusnaslund: #747

davidhesselbom: One smart thing to do would be to give the other `slice` function a default value. That way, you wouldn't need two functions...

marcusnaslund: Either that or call `this slice(start, this count - start)` from here. @davidhesselbom 's suggestion is probably better though.

davidhesselbom: Also, in C# we had an implementation that worked like in PHP, which was pretty neat:
http://php.net/manual/en/function.substr.php

Obviously, don't implement it in this PR.

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,112 @@
...
+		this _arguments add(Argument new(longIdentifier claim(), shortIdentifier, parameters, action))
+	}
+	addDefault: func (action: Event1<Text>) {
+		this _defaultArgument = Argument new(action)
+	}
+	parse: func (input: VectorList<Text>) {
+		argument: Argument
+		parameters := VectorList<Text> new()
+		for (i in 0 .. input count) {
+			argument = null
```
marcusnaslund: This is not neccessary

source/base/Text.ooc
jonathanudd: Would something like this be a good solution?
```
slice: func (start: Int, distance := 0) -> This {
		distance = distance > 0 ? distance : this count - start
		result := This new(this _buffer slice(start, distance))
		if (this _buffer owner == Owner Receiver)
			result = result copy() // TODO: Could we be smarter here?
		this free(Owner Receiver)
		result
	}
```

marcusnaslund: Doesn't this work?
```ooc
slice: func (start: Int, distance := this count - start) -> This {
```

source/cli/ArgumentParser.ooc
davidhesselbom: It might... lines 82 and 83 may be the reason. But since `argument` isn't used anywhere outside the loop, line 70 can be moved here and replace line 73.

source/base/Text.ooc
davidhesselbom: I remember someone noticed default values are only computed once, when the program starts... but I'm not sure.

source/cli/ArgumentParser.ooc
marcusnaslund: Whatever comes from `_findArgument` is going to replace the contents of this variable, so setting it to `null` here (which it already should be btw) makes no difference.

davidhesselbom: Hm, you're probably right. So this line can simply be removed, then, as you said.

davidhesselbom: But then, so could `argument: Argument`. All you need is `argument := _findArgument`, which, by the way, should be `argument := this _findArgument`, and actually, you should even be able to do `if ((argument := this _findArgument))`.

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,112 @@
...
+				current = current slice(1)
+			else
+				this _defaultArgument _textAction call(current)
+			argument = _findArgument(current)
+			if (argument) {
+				if (argument _parameters > 1) {
+					for (j in 0 .. argument _parameters)
+						parameters add(input[i + j + 1] take())
+					argument _listAction call(parameters)
+					i = i + argument _parameters
```
marcusnaslund: Moving this line to after line 95 lets you remove line 92, 93 and the bracket on 90. Boom, linecount -= 3 again.

source/cli/ArgumentParser.ooc
jonathanudd: Yea you are right :) I will fix it.

source/cli/ArgumentParser.ooc
davidhesselbom: Also, `i += argument _parameters`.

source/base/Text.ooc
jonathanudd: For some reason I get a segmentation fault on a different line when I try with
```ooc
slice: func (start: Int, distance := this count - start) -> This {
```

source/cli/ArgumentParser.ooc
marcusnaslund: Sorry, -= 2. But still.

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,112 @@
...
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with this software. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use ooc-collections
+use ooc-base
+import Event
+
+Argument: class {
```
davidhesselbom: Is this class used anywhere outside of this file? If not, it should be `_Argument`.

source/base/Text.ooc
davidhesselbom: Yeah, I think maybe the default value is only evaluated once, and not every time you call the function. So you probably can't do it like that.

marcusnaslund: :(

jonathanudd:
Rebased and updated based on comments.

source/cli/ArgumentParser.ooc
```diff
@@ -0,0 +1,107 @@
...
+				i += argument _parameters
+			}
+		}
+		for (i in 0 .. input count)
+			input[i] free(Owner Receiver)
+		parameters free()
+	}
+	_findArgument: func (identifier: Text) -> _Argument {
+		result: _Argument
+		for (i in 0 .. this _arguments count) {
```
davidhesselbom: You don't need these brackets.

jonathanudd: I quite sure I removed those brackets once... oh well I will remove them again.

ooc-cli.use
simonmika: Wouldn't it be better if the requires line reflected the requirement in that component. We should never have a use in the code of a component that is not in the `Requires:` line of its use-file.

davidhesselbom: You mean the compiler should protest if you try?

davidhesselbom:
`TextTest` and `TextLiteralTest` are failing. They probably expect `slice` to something different.

jonathanudd:
Ok, sorry I didn't check the test properly. I will fix them tomorrow.

davidhesselbom:
Alright. Maybe the default `0` isn't a good idea, maybe it is it, but it needs to be clear what's supposed to happen when you only give the function one argument - do you get the first x characters, or everything after character x?

simonmika:
```ooc
slice := t"abcdef" slice(2)
silice == t"cdef"
```

marcusnaslund:
Yeah, `TextBuffer` and `OwnedBuffer` seem to have decided that question already.

simonmika:
Implement it after:
http://php.net/manual/en/function.substr.php

simonmika:
which means you can't default to `0`.

ooc-cli.use
davidhesselbom: Because people are never going to remember to keep these files up to date, otherwise.

jonathanudd:
And I can't have `-1` as default then because you can pass negative distances too. But I should be able to use `INT_MAX` as default right?

jonathanudd:
It seems to work, I have pushed a new version.

source/base/Text.ooc
```diff
@@ -134,8 +134,8 @@ Text: cover {
 	free: func@ ~withCriteria (criteria: Owner) -> Bool {
 		this _buffer free(criteria)
 	}
-	slice: func (start, distance: Int) -> This {
-		result := This new(this _buffer slice(start, distance))
+	slice: func (start: Int, distance := INT_MAX) -> This {
+		result := This new(this _buffer slice(start, distance == INT_MAX ? this count - start : distance))
```
davidhesselbom: So... what should happen if `distance` is negative?

marcusnaslund: Nothing here. The change @simonmika talks about above would have to be implemented in the buffers.

davidhesselbom: Oh, I guess `this _buffer` will handle it. Okay then.

jonathanudd: Yes the buffer will handle it.

PR #738, opened 2015-11-11
IntVectorList and IntBox2D union~point()

bettinaselig:
- Added the class `IntVectorList`, since I needed the functions `sort()` and `compress()` (open for naming suggestions) for `VectorList<Int>`.
- Added `IntBox2D union~point()`, which corresponds to `FloatBox2D union~point()`.

### DIFF
--- a/source/math/IntBox2D.ooc
+++ b/source/math/IntBox2D.ooc
@@ -84,6 +84,13 @@ IntBox2D: cover {
 		height := Int maximum~two(0, Int maximum(this bottom, other bottom) - top)
 		This new(left, top, width, height)
 	}
+	union: func ~point (point: IntPoint2D) -> This {
+		left := Int minimum~two(this left, point x)
+		top := Int minimum~two(this top, point y)
+		width := Int maximum~two(0, Int maximum~two(this right, point x) - left)
+		height := Int maximum~two(0, Int maximum~two(this bottom, point y) - top)
+		This new(left, top, width, height)
+	}
 	contains: func (point: IntPoint2D) -> Bool {
 		this left <= point x && point x < this right && this top <= point y && point y < this bottom
 	}
--- a/source/math/IntVectorList.ooc
+++ b/source/math/IntVectorList.ooc
@@ -20,13 +20,13 @@ import FloatVectorList
 
 IntVectorList: class extends VectorList<Int> {
 	init: func ~default {
-		this super()
+		super()
 	}
 	init: func ~heap (capacity: Int) {
 		super(capacity)
 	}
 	init: func ~fromVectorList (other: VectorList<Int>) {
-		this super(other _vector)
+		super(other _vector)
 		this _count = other count
 	}
 	copy: func -> This {
--- a/source/math/IntVectorList.ooc
+++ b/source/math/IntVectorList.ooc
@@ -49,9 +49,9 @@ IntVectorList: class extends VectorList<Int> {
 			for (i in 0 .. result count - 1)
 				if (result[i] > result[i + 1]) {
 					inOrder = false
-					tmp := result[i]
+					temporary := result[i]
 					result[i] = result[i + 1]
-					result[i + 1] = tmp
+					result[i + 1] = temporary
 				}
 		}
 		result
--- a/source/math/IntVectorList.ooc
+++ b/source/math/IntVectorList.ooc
@@ -29,33 +29,43 @@ IntVectorList: class extends VectorList<Int> {
 		this super(other _vector)
 		this _count = other count
 	}
+	copy: func -> This {
+		result := This new(this _count)
+		for (i in 0 .. this _count)
+			result add(this[i])
+		result
+	}
 	operator [] <T> (index: Int) -> T {
 		this as VectorList<Int> _vector[index]
 	}
 	operator []= (index: Int, item: Int) {
 		this _vector[index] = item
 	}
-	sort: func {
+	sort: func -> This {
+		result := this copy()
 		inOrder := false
 		while (!inOrder) {
 			inOrder = true
-			for (i in 0 .. this count - 1)
-				if (this[i] > this[i + 1]) {
+			for (i in 0 .. result count - 1)
+				if (result[i] > result[i + 1]) {
 					inOrder = false
-					tmp := this[i]
-					this[i] = this[i + 1]
-					this[i + 1] = tmp
+					tmp := result[i]
+					result[i] = result[i + 1]
+					result[i + 1] = tmp
 				}
 		}
+		result
 	}
 	compress: func -> This {
+	// Returns a new, sorted list of unique elements, e.g. [1,2,1,5,2,7,6,7,7] -> [1,2,5,6,7]
 		result := This new()
 		if (this count > 0) {
-			this sort()
-			result add(this[0])
-			for (i in 1 .. this count)
-				if (this[i] != result[result count - 1])
-					result add(this[i])
+			sortedList := this sort()
+			result add(sortedList[0])
+			for (i in 1 .. sortedList count)
+				if (sortedList[i] != result[result count - 1])
+					result add(sortedList[i])
+			sortedList free()
 		}
 		result
 	}
--- a/source/math/IntVectorList.ooc
+++ b/source/math/IntVectorList.ooc
@@ -30,10 +30,7 @@ IntVectorList: class extends VectorList<Int> {
 		this _count = other count
 	}
 	copy: func -> This {
-		result := This new(this _count)
-		for (i in 0 .. this _count)
-			result add(this[i])
-		result
+		this as VectorList<Int> copy() as This
 	}
 	operator [] <T> (index: Int) -> T {
 		this as VectorList<Int> _vector[index]
@@ -56,8 +53,8 @@ IntVectorList: class extends VectorList<Int> {
 		}
 		result
 	}
-	compress: func -> This {
 	// Returns a new, sorted list of unique elements, e.g. [1,2,1,5,2,7,6,7,7] -> [1,2,5,6,7]
+	compress: func -> This {
 		result := This new()
 		if (this count > 0) {
 			sortedList := this sort()
--- a/source/math/IntVectorList.ooc
+++ b/source/math/IntVectorList.ooc
@@ -1,0 +1,68 @@
+/*
+* Copyright (C) 2014 - Simon Mika <simon@mika.se>
+*
+* This sofware is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License as published by the Free Software Foundation; either
+* version 2.1 of the License, or (at your option) any later version.
+*
+* This software is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with this software. If not, see <http://www.gnu.org/licenses/>.
+*/
+use ooc-collections
+import math
+import FloatVectorList
+
+IntVectorList: class extends VectorList<Int> {
+	init: func ~default {
+		this super()
+	}
+	init: func ~heap (capacity: Int) {
+		super(capacity)
+	}
+	init: func ~fromVectorList (other: VectorList<Int>) {
+		this super(other _vector)
+		this _count = other count
+	}
+	operator [] <T> (index: Int) -> T {
+		this as VectorList<Int> _vector[index]
+	}
+	operator []= (index: Int, item: Int) {
+		this _vector[index] = item
+	}
+	sort: func {
+		inOrder := false
+		while (!inOrder) {
+			inOrder = true
+			for (i in 0 .. this count - 1)
+				if (this[i] > this[i + 1]) {
+					inOrder = false
+					tmp := this[i]
+					this[i] = this[i + 1]
+					this[i + 1] = tmp
+				}
+		}
+	}
+	compress: func -> This {
+		result := This new()
+		if (this count > 0) {
+			this sort()
+			result add(this[0])
+			for (i in 1 .. this count)
+				if (this[i] != result[result count - 1])
+					result add(this[i])
+		}
+		result
+	}
+	toString: func -> String {
+		result := ""
+		for (i in 0 .. this _count)
+			result = result >> this[i] toString() >> "\n"
+		result
+	}
+}
--- a/test/math/IntVectorListTest.ooc
+++ b/test/math/IntVectorListTest.ooc
@@ -1,0 +1,75 @@
+//
+// Copyright (c) 2011-2015 Simon Mika <simon@mika.se>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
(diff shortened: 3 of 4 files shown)


source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,68 @@
...
+			for (i in 0 .. this count - 1)
+				if (this[i] > this[i + 1]) {
+					inOrder = false
+					tmp := this[i]
+					this[i] = this[i + 1]
+					this[i + 1] = tmp
+				}
+		}
+	}
+	compress: func -> This{
```
marcusnaslund: Missing a space after `This`

bettinaselig:
Validation script for kean? Or does the validation script for the other code base also check kean?

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,68 @@
...
+		inOrder := false
+		while (!inOrder) {
+			inOrder = true
+			for (i in 0 .. this count - 1)
+				if (this[i] > this[i + 1]) {
+					inOrder = false
+					tmp := this[i]
+					this[i] = this[i + 1]
+					this[i + 1] = tmp
+				}
```
marcusnaslund: 1) How many elements will we use and what's the performance needed? You're doing a lot of `memcpy` here.
2) Linear complexity would be awesome, but this won't actually guarantee a sorted vectorlist.

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+	compress: func -> This {
```
marcusnaslund: So `compress` is going to rearrange the elements in this `VectorList`, that's probably at least good to note.

marcusnaslund:
Yes, the validation script also runs kean tests.
Although - you don't actually have an `IntVectorList` test here. Either make one or create an issue for that in Pull request debt.

source/math/IntVectorList.ooc
bettinaselig: We definitely need to talk about the sorting algorithms. But right now it's only me who uses this class and there are about 4 elements in the list to be sorted.

source/math/IntVectorList.ooc
bettinaselig: I think the real name of this function is `getSet()` from set theory. But I did not think that that was a good name either.

source/math/IntVectorList.ooc
marcusnaslund: Alright, but it still won't always sort.
A double loop would solve the problem (this would be Bubblesort, with `O(n^2)` complexity). With 4-ish elements, that's good enough. 
```ooc
for (j in 0 .. this count - 1)
    for (i in 0 .. this count - 1)
        if (this[i] > this[i + 1]) {
            inOrder = false
            tmp := this[i]
            this[i] = this[i + 1]
            this[i + 1] = tmp
    }
```

source/math/IntVectorList.ooc
marcusnaslund: What is it supposed to do?

bettinaselig: `[1,1,2,2,5,6,7,7,7] -> [1,2,5,6,7]`

marcusnaslund: Right. `getUniqueElements` or `getUniques`, maybe? @davidhesselbom ?

source/math/IntVectorList.ooc
bettinaselig: Now I don't get it. When wouldn't it sort?

source/math/IntVectorList.ooc
marcusnaslund: Either way: If this function affects the elements in `this` (which it doesn't really have to, just use a temporary `VectorList`) - this should be commented here.

source/math/IntVectorList.ooc
sebastianbaginski: @marcusnaslund  you can start the second loop from `j + 1`. And the previous one will sort anyway.

source/math/IntVectorList.ooc
bettinaselig: I can add a comment. I just wanted to have a quick unique function.

davidhesselbom: > getSet

Yeah, not clear enough, unless you are already in a set theory mindset.

> getUniqueElements or getUniques, maybe?

`compress` is actually not bad, if you are familiar with the term, but this suggestion works for people who aren't, too. If you want to keep the name, though, you should at least leave a comment explaining what it does - `[1,1,2,2,5,6,7,7,7] -> [1,2,5,6,7]` is very pedagogical, I think, but perhaps
```ooc
// Returns a new list without duplicate elements, e.g. [1,1,2,2,5,6,7,7,7] -> [1,2,5,6,7]
```
Still, if you call it `getUniqueElements` (or `uniqueElements`, I'm not sure where we draw the line for a property to be honest), you won't need a comment. Up to @marcusnaslund .

source/math/IntVectorList.ooc
marcusnaslund: Right, thanks @sebastianbaginski

marcusnaslund: Oooooh I'm sorry, that was the purpose of the while-loop. :(
Nvm all of this.

marcusnaslund: I have studied CS, I promise. :)

source/math/IntVectorList.ooc
davidhesselbom: If you do leave the name as is, maybe this comment would be even better:
```ooc
// Returns a new, sorted list of unique elements, e.g. [1,2,1,5,7,2,6,7,7] -> [1,2,5,6,7]
```

marcusnaslund: `// Note: Also sorts the current VectorList`

source/math/IntVectorList.ooc
bettinaselig: Haha! But we anyway need to talk about sorting algorithms. @davidhesselbom and I recently chatted about it. 
In another class, bubble sort was exchanged by quick sort, where I was not that happy with.

davidhesselbom: Since you're talking about complexity... if you want to sort 4 elements, you will gain more by
```ooc
thisPointer := this pointer as Int*
```
and accessing `thisPointer[i]` instead of `this[i]`. This has been benchmarked to run about 40 times faster...

sebastianbaginski: ... :( I'm sooo sad right now, Bettina is not happy with my quicksort ...

source/math/IntVectorList.ooc
marcusnaslund: I'm OK with `compress` if you include @davidhesselbom 's comment

source/math/IntVectorList.ooc
bettinaselig: I am happy with your quick sort... I just would have liked to have a `bubblesort()` left in the class... Or the question answered if it actually make a difference for our use case...

marcusnaslund: I would like to see some performance tests for quicksort vs bubblesort in ooc.
Anyway, this is fine for now, we obviously don't do major stuff here.

davidhesselbom: The only sorting algorithm I know of that isn't fast enough for 4 elements is Shotgun Sort.

sebastianbaginski: @marcusnaslund : https://github.com/cogneco/ooc-kean/pull/330

marcusnaslund: Awesome, thanks. :)

marcusnaslund:
Validated.

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,78 @@
...
+	}
+	sort: func -> This {
+		result := this copy()
+		inOrder := false
+		while (!inOrder) {
+			inOrder = true
+			for (i in 0 .. result count - 1)
+				if (result[i] > result[i + 1]) {
+					inOrder = false
+					tmp := result[i]
```
marcusnaslund: `temporary`, not `tmp`

bettinaselig: Done.

bettinaselig:
Why do we set pull requests to high, if one anyway works with different ones?

marcusnaslund:
They're still good, especially for David. I'm just looking through others while waiting for replies/tests.

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,78 @@
...
+		result := this copy()
+		inOrder := false
+		while (!inOrder) {
+			inOrder = true
+			for (i in 0 .. result count - 1)
+				if (result[i] > result[i + 1]) {
+					inOrder = false
+					temporary := result[i]
+					result[i] = result[i + 1]
+					result[i + 1] = tmp
```
marcusnaslund: `temporary` here as well. :)

bettinaselig:
Ok.

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,78 @@
...
+
+IntVectorList: class extends VectorList<Int> {
+	init: func ~default {
+		this super()
+	}
+	init: func ~heap (capacity: Int) {
+		super(capacity)
+	}
+	init: func ~fromVectorList (other: VectorList<Int>) {
+		this super(other _vector)
```
marcusnaslund: Oh, just one more thing before I'm happy. `super`, not `this super`. You probably copied this from `FloatVectorList` - I have a PR which fixes that over there.

marcusnaslund: Same on line 23.

bettinaselig:
Done. 

Does this one need a peer review or can you merge it right away?

marcusnaslund:
I'm happy with it, it passes validation, and there are no side-effects. @davidhesselbom

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,78 @@
...
+					inOrder = false
+					temporary := result[i]
+					result[i] = result[i + 1]
+					result[i + 1] = temporary
+				}
+		}
+		result
+	}
+	compress: func -> This {
+	// Returns a new, sorted list of unique elements, e.g. [1,2,1,5,2,7,6,7,7] -> [1,2,5,6,7]
```
davidhesselbom: The comment and the function declaration should switch places.

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,78 @@
...
+		super(capacity)
+	}
+	init: func ~fromVectorList (other: VectorList<Int>) {
+		super(other _vector)
+		this _count = other count
+	}
+	copy: func -> This {
+		result := This new(this _count)
+		for (i in 0 .. this _count)
+			result add(this[i])
```
davidhesselbom: Why not a memcopy?

source/math/IntVectorList.ooc
marcusnaslund: Haha, I said to do it like this. We're not exactly consistent.

source/math/IntVectorList.ooc
davidhesselbom: Actually, this function already exists in `VectorList`. Do you even need to add it here?

marcusnaslund: Probably copied (heheheh) from `FloatVectorList`. Guess it shouldn't be there either.

marcusnaslund: I'll make a note of it.

davidhesselbom: Okay, so, either get rid of it or replace it with
```ooc
    // copied from VectorList
	result := This new(this _count)
	memcpy(result pointer, this pointer, this _count * T size)
	result _count = this _count
	result
```
at the first convenience

bettinaselig: Does this work `this copy() as IntVectorList` ?

davidhesselbom: In that case, so should `this copy() as This`... but maybe it doesn't.

marcusnaslund: `this as VectorList<Int> copy() as This`

source/math/IntVectorList.ooc
davidhesselbom: Where else does a function start with a comment (except for in the SDK, which I take 0 responsibility for)?

source/math/IntVectorList.ooc
davidhesselbom: If you had written a test for this class, it'd be pretty easy to find out!

source/math/IntVectorList.ooc
marcusnaslund: `Promise free` :)

marcusnaslund: Or every `take`, `give` and `claim` method

marcusnaslund: Anyway, you decide.

davidhesselbom: Those are either `TODO` or on the same line as the function declaration.

bettinaselig:
Updated `copy()` and created a test class. I think this pull request is ready to be merged :-)

source/math/IntVectorList.ooc
```diff
@@ -0,0 +1,75 @@
...
+	}
+	init: func ~fromVectorList (other: VectorList<Int>) {
+		super(other _vector)
+		this _count = other count
+	}
+	copy: func -> This {
+		this as VectorList<Int> copy() as This
+	}
+	operator [] <T> (index: Int) -> T {
+		this as VectorList<Int> _vector[index]
```
marcusnaslund: Do you really have to cast like this?

test/math/IntVectorListTest.ooc
```diff
@@ -0,0 +1,76 @@
...
+			copy := list copy()
+			expect(list count == copy count)
+			for (i in 0 .. list count)
+				expect(copy[i], is equal to(list[i]))
+			list free()
+			copy free()
+		})
+		this add("compress", func {
+			list := IntVectorList new()
+			list add(1)
```
marcusnaslund: You don't have to change this now - but for the future remember you can do `list add(1) . add(2) . add(1) . add(5)` etc. to save some lines.

source/math/IntVectorList.ooc
bettinaselig: Copy-pasted it from `FloatVectorList`.

test/math/IntVectorListTest.ooc
bettinaselig: Copy-pasted it from `FloatVectorListTest`. But yes, I try to remember for the future :-)

source/math/IntVectorList.ooc
marcusnaslund: That one is not without its problems either. :)

marcusnaslund: But this works obviously so I'm OK with it for now.

source/math/IntBox2D.ooc
```diff
@@ -84,6 +84,13 @@ IntBox2D: cover {
 		height := Int maximum~two(0, Int maximum(this bottom, other bottom) - top)
 		This new(left, top, width, height)
 	}
+	union: func ~point (point: IntPoint2D) -> This {
+		left := Int minimum~two(this left, point x)
+		top := Int minimum~two(this top, point y)
+		width := Int maximum~two(0, Int maximum~two(this right, point x) - left)
+		height := Int maximum~two(0, Int maximum~two(this bottom, point y) - top)
+		This new(left, top, width, height)
```
davidhesselbom:
```ooc
This new(this leftTop minimum(point), this rightBottom maximum(point))
```
should do the same thing...

bettinaselig: Probably... but seriously... shall I spend the time for rewriting and testing it? I thought we are in a "merging pull request" sprint. This change does not fix a bug, it _only_ makes the code more compact. It does not even increase the readability. `union()` is done everywhere like this. If this is a problem, please make an issue.

davidhesselbom: > It does not even increase the readability.

I definitely don't agree there.

> If this is a problem, please make an issue.

Fine: #753

bettinaselig:
Thanks. Can we now merge this pull request and be friends again? :-)

PR #743, opened 2015-11-11
Make properties and unary operators free owned buffers.

simonmika:
Updates all users of `OwnedBuffer` so that method calls behave the same as properties and operators.


simonmika:
@tomasmahlberg could you please help me fix the double free in the quartinons test.

tomasmahlberg:
@simonmika  simonmika/ooc-kean#6. Might have do a PR to the other project as well, let me have a look.

marcusnaslund:
@simonmika Magic complains, please fix.

simonmika:
Fixed.

source/base/Text.ooc
```diff
@@ -21,8 +21,20 @@ import math
...
 	_buffer: TextBuffer
-	count ::= this _buffer count
-	isEmpty ::= this count == 0
+	count: Int {
+		get {
+			result := this _buffer count
+			this free(Owner Receiver)
+			result
+		}
+	}
```
davidhesselbom: Am I the only one who prefers 
```ooc
count: Int { get {
    result := this _buffer count
    this free(Owner Receiver)
    result
}}
```
?

marcusnaslund: Wouldn't work well with a setter though. (or?)
But with only a getter, yes, I like that too.

davidhesselbom:
```ooc
count ::= {
    result := this _buffer count
    this free(Owner Receiver)
    result
}
```
would be cool too. I don't think that's supported in rock, but it shouldn't be too hard to make it possible, in that case.

simonmika: when only having a getter I prefer the way @davidhesselbom suggested first.

simonmika: Should we consider it a violation that blocks a PR from being accepted?

marcusnaslund: Seems easy enough to fix and saves us a lot of lines, so I vote yes.

simonmika:
Does this look better to you?

davidhesselbom:
The double indentation is more correct, technically, I guess, but it also makes my fingers twitch and it looks like a mistake.

marcusnaslund:
Same here. Don't like the double indent.

davidhesselbom:
Magic seems to agree.

simonmika:
What about:
```ooc
} }
```
or
```ooc
}}
```

davidhesselbom:
Both look like mistakes :-/ I guess I like `}}` better though.

simonmika:
Next try.

test/math/FloatMatrixTest.ooc
```diff
@@ -13,127 +14,126 @@ FloatMatrixTest: class extends Fixture {
...
+			m := this createMatrix(3, 3, [1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f, 1.0f])
+			m take()[0, 0] = 42.0f
+			expect(m[0, 0], is equal to(42.0f) within(this precision))
 		})
 
 		this add("print columns", func {
-			A := createMatrix(3, 3, [1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f, 9.0f]) take()
-			column := A getColumn(1)
+			a := this createMatrix(3, 3, [1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f, 9.0f]) take()
+			column := a take() getColumn(1)
```
marcusnaslund: Doesn't make a difference here, but just so I understand, you don't need to `take()` here since you did it on the above line.. right?

simonmika: Correct, I'll remove it.

test/base/TextTest.ooc
```diff
@@ -166,6 +166,23 @@ TextTest: class extends Fixture {
...
+this add("text literal toString", func {
+			text := t"Hello World"
+			string := text toString()
+			expect(text == string)
+		})
+		this add("add operator", func {
+			correct := t"Hello World"
+			text := t"Hello" + t" World"
+			expect(text take() count == correct count)
+			for (i in 0 .. text take() count)
```
marcusnaslund: Also here, you don't need to take all the time?

source/math/FloatMatrix.ooc
```diff
@@ -282,59 +330,76 @@ FloatMatrix : cover {
...
+		for (i in 0 .. this take() dimensions area)
 			this elements[i] -= other elements[i]
 		other free(Owner Receiver)
 	}
 
 	operator * (other: Float) -> This {
-		result := This new(this dimensions)
-		for (i in 0 .. this dimensions area)
+		result := This new(this take() dimensions)
+		for (i in 0 .. this take() dimensions area)
```
marcusnaslund: double `take`

simonmika: this is necessary if i don't assign it to a variable. `take()` only affects what it returns.

test/base/TextTest.ooc
simonmika: I do in this case.

source/base/Text.ooc
```diff
@@ -21,8 +21,16 @@ import math
 
 Text: cover {
 	_buffer: TextBuffer
-	count ::= this _buffer count
-	isEmpty ::= this count == 0
+	count: Int { get {
```
emilwestergren: Is this the way we should format properties?

marcusnaslund: Yes.

source/base/Text.ooc
```diff
@@ -174,34 +192,22 @@ Text: cover {
 		this slice(leftPosition, rightPosition - leftPosition + 1)
 	}
 	toString: func -> String {
-		result := String new(this _buffer raw, this count)
-		this free(Owner Receiver)
-		result
+		this _buffer toString()
```
emilwestergren: I like to put these sort of functions on a single line but I guess it's optional.
`toString: func -> String { this _buffer toString() }`

source/base/Text.ooc
davidhesselbom: Those that won't fit on one line, yes.

marcusnaslund:
Conflicts @simonmika

davidhesselbom:
If this were made by anyone but you, I'd mark it "too complex", but I don't think that's going to work on you.

PR #750, opened 2015-11-12
Removed texture binding from public API.

emilwestergren:
* Texture binding has been removed from the public API
* Breaks API

### DIFF
--- a/source/draw/gpu/GpuCanvas.ooc
+++ b/source/draw/gpu/GpuCanvas.ooc
@@ -77,6 +77,8 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target uv canvas fill()
 	}
 	draw: override func ~mesh (image: GpuImage, mesh: GpuMesh) {
+		if (!image instanceOf?(GpuYuv420Semiplanar))
+			Debug raise("Invalid input format in GpuYuv420SemiplanarCanvas::draw~mesh")
 		yuv := image as GpuYuv420Semiplanar
 		this _target y canvas draw(yuv y, mesh)
 		this _target uv canvas draw(yuv uv, mesh)
--- a/source/draw/gpu/GpuCanvas.ooc
+++ b/source/draw/gpu/GpuCanvas.ooc
@@ -39,7 +39,6 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 			this _target uv canvas focalLength = value / 2
 		}
 	}
-
 	pen: Pen {
 		get { this _pen }
 		set(value) {
@@ -59,6 +58,11 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target y canvas draw(gpuImage y, source, destination, map)
 		this _target uv canvas draw(gpuImage uv, IntBox2D new(source leftTop / 2, source size / 2), IntBox2D new(destination leftTop / 2, destination size / 2), map)
 	}
+	draw: override func ~ImageSourceDestination (image: Image, source, destination: IntBox2D) {
+		gpuImage := image as GpuYuv420Semiplanar
+		this _target y canvas draw(gpuImage y, source, destination)
+		this _target uv canvas draw(gpuImage uv, IntBox2D new(source leftTop / 2, source size / 2), IntBox2D new(destination leftTop / 2, destination size / 2))
+	}
 	drawLines: override func (pointList: VectorList<FloatPoint2D>) {
 		this _target y canvas drawLines(pointList)
 		uvLines := VectorList<FloatPoint2D> new()
@@ -72,7 +76,7 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target y canvas fill()
 		this _target uv canvas fill()
 	}
-	draw: override func ~mesh (image: Image, mesh: GpuMesh) {
+	draw: override func ~mesh (image: GpuImage, mesh: GpuMesh) {
 		yuv := image as GpuYuv420Semiplanar
 		this _target y canvas draw(yuv y, mesh)
 		this _target uv canvas draw(yuv uv, mesh)
--- a/source/draw/gpu/GpuMap.ooc
+++ b/source/draw/gpu/GpuMap.ooc
@@ -14,14 +14,16 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this _program. If not, see <http://www.gnu.org/licenses/>.
 use ooc-math
+use ooc-base
 use ooc-collections
 
 _MapEntry: class {
 	key: String
 	value: Object
-	init: func (=key, =value)
+	_ownsObject: Bool
+	init: func (=key, =value, =_ownsObject)
 	free: override func {
-		if (this value instanceOf?(Cell))
+		if (this _ownsObject)
 			this value free()
 		super()
 	}
@@ -34,15 +36,15 @@ _Map: class {
 		this _entries free()
 		super()
 	}
-	add: func (key: String, value: Object) {
+	add: func (key: String, value: Object, ownsObject: Bool) {
 		entry := this get(key)
 		if (entry != null) {
-			if (entry value instanceOf?(Cell))
+			if (entry _ownsObject)
 				entry value free()
 			entry value = value
 		}
 		else
-			this _entries add(_MapEntry new(key, value))
+			this _entries add(_MapEntry new(key, value, ownsObject))
 	}
 	get: func (key: String) -> _MapEntry {
 		result: _MapEntry = null
@@ -78,9 +80,9 @@ GpuMap: abstract class {
 	use: virtual func
 	add: func <T> (key: String, value: T) {
 		if (T inheritsFrom?(Object))
-			this _bindings add(key, value as Object)
+			this _bindings add(key, value as Object, false)
 		else
-			this _bindings add(key, Cell<T> new(value))
+			this _bindings add(key, Cell<T> new(value), true)
 	}
 	get: func (key: String) -> Object { this _bindings get(key) value }
 	apply: func (action: Func (String, Object)) { this _bindings apply(action) }
--- a/source/draw/gpu/GpuSurface.ooc
+++ b/source/draw/gpu/GpuSurface.ooc
@@ -19,7 +19,7 @@ use ooc-draw
 use ooc-collections
 use ooc-base
 
-import GpuContext, GpuMap, GpuImage, GpuMesh
+import GpuContext, GpuMap, GpuImage, GpuMesh, GpuYuv420Semiplanar
 
 version(!gpuOff) {
 GpuSurface: abstract class extends Canvas {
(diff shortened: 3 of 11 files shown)


davidpiuva:
I would place everything that breaks backward compability in another pull requests that is taken in when nothing in upstream is using the old texture binding.

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -59,6 +58,11 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target y canvas draw(gpuImage y, source, destination, map)
 		this _target uv canvas draw(gpuImage uv, IntBox2D new(source leftTop / 2, source size / 2), IntBox2D new(destination leftTop / 2, destination size / 2), map)
 	}
+	draw: override func ~ImageSourceDestination (image: Image, source, destination: IntBox2D) {
```
davidpiuva: Are these boxes in upper left base zero or reference coordinates?
I prefer base 0 for general purpose calculations to treat textures like arrays as in compute shaders.
Reference coordinates in integers would not even be mathematically possible for odd dimensions.

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -72,7 +76,7 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target y canvas fill()
 		this _target uv canvas fill()
 	}
-	draw: override func ~mesh (image: Image, mesh: GpuMesh) {
+	draw: override func ~mesh (image: GpuImage, mesh: GpuMesh) {
 		yuv := image as GpuYuv420Semiplanar
```
davidpiuva: What happens if image is not GpuYuv420Semiplanar?

source/draw/gpu/GpuCanvas.ooc
emilwestergren: base 0

davidpiuva: Good

source/draw/gpu/GpuCanvas.ooc
emilwestergren: Undefined, probably a crash somewhere. It's an ugly assumption but can be found in many places in the code. This function is currently only used in **one** place, and I expect that to be true for a long time.

source/draw/gpu/GpuSurface.ooc
```diff
@@ -68,25 +68,31 @@ GpuSurface: abstract class extends Canvas {
 		translation * scaling
 	}
 	_getDefaultMap: virtual func (image: Image) -> GpuMap { this _defaultMap }
+	_createImage: func (image: Image) -> GpuImage {
+		match (image) {
+			case (i: RasterImage) => this _context createImage(i)
+			case (i: GpuImage) => i
+			case => Debug raise("Invalid image type in GpuSurface!")
+		}
```
davidpiuva: This looks better.

emilwestergren:
Sure, I'll save the commit which removes the API until later.

source/draw/gpu/GpuCanvas.ooc
davidpiuva: I have wanted to use something like that a few times before it was implemented.

emilwestergren: Ok, but in this case you are drawing to a GpuYuv420Semiplanar image, so I think the assumption is not that bad. Of course it would be nice to support any format, but that's the sort of thing we probably want to implement when it's needed.

davidpiuva: It should at least raise an exception before crashing.

emilwestergren:
It's still possible to bind the "old" way by accessing private variables. I tried to do as you said, but all these commits require that the old binding is not used at all and the corresponding PR in Vidproc depends on this PR. I'll take the responsibility if something breaks. That's the reason I wanted you to actually test this code since it's only "your" files that have been modified in the adaptation.

source/draw/gpu/GpuCanvas.ooc
emilwestergren: Sure, I can add it.

davidpiuva:
ok

davidpiuva:
I got this crash and no detected movement when running MotionInMotion.
```
(SIGSEGV) segmentation fault
[fancy backtrace]
0     BacktraceHandler backtrace()     in lang/Backtrace  (at /home/dpiuva/Versioned/ooc/ooc-kean/source/sdk/lang/Backtrace.ooc:50)           
1     Exception getCurrentBacktrace()  in lang/Exception  (at /home/dpiuva/Versioned/ooc/ooc-kean/source/sdk/lang/Exception.ooc:177)          
2     _signalHandler()                 in lang/Exception  (at /home/dpiuva/Versioned/ooc/ooc-kean/source/sdk/lang/Exception.ooc:237)          
3     killpg()                         in                 (at (null):0)                                                                       
4     Object instanceOf__quest()       in lang/types      (at /home/dpiuva/Versioned/ooc/ooc-kean/source/sdk/lang/types.ooc:31)               
5     _MapEntry_free_impl()            in GpuMap          (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/GpuMap.ooc:24)              
6     Object free()                    in lang/types      (at /home/dpiuva/Versioned/ooc/ooc-kean/source/sdk/lang/types.ooc:46)               
7     Vector _free_range()             in Vector          (at /home/dpiuva/Versioned/ooc/ooc-kean/source/collections/Vector.ooc:35)           
8     VectorList clear()               in VectorList      (at /home/dpiuva/Versioned/ooc/ooc-kean/source/collections/VectorList.ooc:71)       
9     VectorList free_impl()           in VectorList      (at /home/dpiuva/Versioned/ooc/ooc-kean/source/collections/VectorList.ooc:75)       
10    VectorList free()                in VectorList      (at /home/dpiuva/Versioned/ooc/ooc-kean/source/collections/VectorList.ooc:35)       
11    _Map_free_impl()                 in GpuMap          (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/GpuMap.ooc:35)              
12    _Map_free()                      in GpuMap          (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/GpuMap.ooc:69)              
13    GpuMap free_impl()               in GpuMap          (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/GpuMap.ooc:76)              
14    OpenGLMap free_impl()            in Map/OpenGLMap   (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/opengl/Map/OpenGLMap.ooc:58)
15    OpenGLMap free()                 in Map/OpenGLMap   (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/opengl/Map/OpenGLMap.ooc:47)
16    OpenGLWindow free_impl()         in OpenGLWindow    (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/opengl/OpenGLWindow.ooc:36) 
17    OpenGLWindow free()              in OpenGLWindow    (at /home/dpiuva/Versioned/ooc/ooc-kean/source/draw/gpu/opengl/OpenGLWindow.ooc:64) 
18    UnixWindow free_impl()           in UnixWindow      (at /home/dpiuva/Versioned/ooc/ooc-kean/source/widgets/UnixWindow.ooc:64)           
19    Object free()                    in lang/types      (at /home/dpiuva/Versioned/ooc/ooc-kean/source/sdk/lang/types.ooc:46)               
20    main()                           in                 (at /home/dpiuva/Versioned/ooc/ooc-vidproc/source/vidhance/sdk/pc/main.ooc:149)     
21    libc_start_main()                in                 (at /build/buildd/eglibc-2.19/csu/libc-start.c:321)                                 
22    _start()                         in                 (at (null):0)                                                                       
```

emilwestergren:
I've had that problem, looks like you're freeing some objects that are referenced in the map, so when the map is freed it tries to access these objects. The map doesn't free any objects *except* for Cells, but it still tries to access all objects.

emilwestergren:
I'll try to fix it.

davidpiuva:
The contrast is fully broken.

emilwestergren:
In what way?

davidpiuva:
The mapping curves are static so the luma analyzer is probably getting no texture input.

emilwestergren:
That's right, I've missed to add a binding there, good catch. This is why I need you to test it :)

davidpiuva:
Upstream works better with moving mappings but there seems to be a compiler/sync bug since I can only see the mapping in the shader when debug printing from LumaApplier.

davidpiuva:
I still have some code to read.

davidpiuva:
Syntax looks fine.

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -59,6 +58,11 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target y canvas draw(gpuImage y, source, destination, map)
 		this _target uv canvas draw(gpuImage uv, IntBox2D new(source leftTop / 2, source size / 2), IntBox2D new(destination leftTop / 2, destination size / 2), map)
 	}
+	draw: override func ~ImageSourceDestination (image: Image, source, destination: IntBox2D) {
+		gpuImage := image as GpuYuv420Semiplanar
+		this _target y canvas draw(gpuImage y, source, destination)
+		this _target uv canvas draw(gpuImage uv, IntBox2D new(source leftTop / 2, source size / 2), IntBox2D new(destination leftTop / 2, destination size / 2))
```
davidhesselbom: Looks like `IntBox2D` could use a division operator.

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -72,7 +76,9 @@ GpuCanvasYuv420Semiplanar: class extends GpuSurface {
 		this _target y canvas fill()
 		this _target uv canvas fill()
 	}
-	draw: override func ~mesh (image: Image, mesh: GpuMesh) {
+	draw: override func ~mesh (image: GpuImage, mesh: GpuMesh) {
+		if (!image instanceOf?(GpuYuv420Semiplanar))
+			Debug raise("Invalid input format in GpuYuv420SemiplanarCanvas::draw~mesh")
```
davidhesselbom: Is there a clean way to prevent the program from even compiling if anyone tries to call this function with anything but a `GpuYuv420Semiplanar`?

sebastianbaginski: Apart from changing the argument parameter, I don't think so, as the types are meant to be resolved at runtime.

source/draw/gpu/GpuMap.ooc
```diff
@@ -77,7 +79,9 @@ GpuMap: abstract class {
 	}
 	use: virtual func
 	add: func <T> (key: String, value: T) {
-		if (T inheritsFrom?(Object))
+		if (T inheritsFrom?(Cell))
+			Debug raise("Can't add Cell to GpuMap!")
```
davidhesselbom: Why, exactly?

source/draw/gpu/GpuCanvas.ooc
davidpiuva: We used to have static types for that.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
```diff
@@ -39,7 +39,7 @@ Gles3ShaderProgram: class extends GLShaderProgram {
 	setUniform: override func ~Int (name: String, value: Int) {
 		version(debugGL) { validateStart("ShaderProgram setUniform~Int") }
 		glUniform1i(glGetUniformLocation(this _backend, name), value)
-		version(debugGL) { validateEnd("ShaderProgram setUniform~Int") }
+		version(debugGL) { validateEnd("ShaderProgram setUniform~Int name:%s value:%d" format(name, value)) }
```
davidhesselbom: Why do you want to do this for `Int`, but not otherwise?

davidhesselbom:
> Texture binding has been removed from the public API

What's the occasion?

source/draw/gpu/GpuMap.ooc
davidhesselbom: Is it because you don't know whether you should free the `Cell` afterwards?

emilwestergren: Because I assume all the cells in the map have been created internally and therefore should be freed by the map itself. A bit ugly but it doesn't make sense to upload a Cell to the GPU. 

I could improve this now however, after my latest change where I added the `_ownsObject` bool instead of just checking if the object is a `Cell` when freeing. So only when you add a value type and an internal Cell is created by `GpuMap` this bool is set to true. That would be a better solution.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
emilwestergren: I just added it while debugging and decided to include it in this PR.

source/draw/gpu/GpuMap.ooc
davidhesselbom: Sounds like a better idea, yes. Can you fix it soon, or does it have to wait?

emilwestergren:
It shouldn't have been there in the first place. It's my fault due to misunderstanding what the purpose of the Gpu API was.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
davidhesselbom: Okay, just making sure it wasn't a mistake. So you're planning on adding it to the other functions too, then?

davidhesselbom:
Alright.

source/draw/gpu/GpuMap.ooc
emilwestergren: I'll fix it.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
emilwestergren: Yes it's useful information when debugging.

davidhesselbom:
@emilwestergren Is this ready to be merged, or it there anything else you need to fix?

emilwestergren:
This was actually the bug causing the problems: https://github.com/vidhance/ooc-vidproc/issues/972
I removed it from the history due to its evil nature, but might be worth looking into.

emilwestergren:
Ready to merge.

PR #758, opened 2015-11-13
Fixed operator + in Event classes

sebastianbaginski:
Fixes problem with `operator +` mentioned here https://github.com/cogneco/ooc-kean/issues/755
Removed `add` functions as they duplicated the functionality of `operator +`

### DIFF
--- a/source/base/Event.ooc
+++ b/source/base/Event.ooc
@@ -25,11 +25,16 @@ Event: class {
 	}
 	init: func (=_head)
 	init: func ~add (=_head, =_tail)
-	operator + (action: Func) -> This {
-		This new(action, this)
+	add: func (action: Func) -> This {
+		this add(This new(action, null))
 	}
-	operator + (event: This) -> This {
-		event != null ? This new(event _head, this) + event _tail : this
+	add: func ~withEvent (event: This) -> This {
+		if (event)
+			if (this _tail)
+				this _tail add(event)
+			else
+				this _tail = event
+		this
 	}
 	call: func {
 		if (this _tail != null)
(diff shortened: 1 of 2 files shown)


sebastianbaginski:
@simonmika @davidhesselbom Which API should it be to add two events :
- `operator +` looks nice in the client code but does not work for adding a Closure, so `add (Func)` is needed
- `add` works for both Event and Closures

for now I used `operator +` as Simon said "I like operators" :)

marcusnaslund:
Writing new code, are we? :)

davidhesselbom:
Will this fix memory leaks?

sebastianbaginski:
@marcusnaslund hehe, but there was one exception to this "no new code" rule :)
@davidhesselbom Yes.

davidhesselbom:
> operator + looks nice in the client code but does not work for adding a Closure

What do you mean it "doesn't work"? It doesn't build or it doesn't do what it's supposed to?

sebastianbaginski:
rock won't compile

source/base/Event.ooc
```diff
@@ -69,7 +63,14 @@ Event1: class <T> {
 	}
 	*/
```
marcusnaslund: I think you should remove this big comment (And the one further down) while you're at it.

davidhesselbom:
Oh, as in the comments, then. Hm. Something related was fixed, or about to be fixed, in rock, recently. For now, I guess we'll have to keep add, but we can open an issue in rock about the operator and hope for some sort of an ETA for when it can be fixed, maybe?

davidhesselbom:
Just create a minimal example to show what we want to do and that it won't compile. And use the official rock, obviously.

sebastianbaginski:
Yeah I just talked with Simon too. `add` it is :)

sebastianbaginski:
`operator +` replaced with `add`. Ready now.

test/base/EventTest.ooc
```diff
@@ -112,8 +110,8 @@ EventTest: class extends Fixture {
 			expect(string1, is equal to("bc"))
 			expect(string2, is equal to(""))
 			/* Not working */
-			//e += (func (delta: Int, s: String) { counter2 += delta; string2 += s})
-			e = e add(func (delta: Int, s: String) { counter2 += delta; string2 += s })
+			//e += (func (dela: Int, s: String) { counter2 += delta; string2 += s})
```
davidhesselbom: `dela`?

davidhesselbom: Anyway, you could just remove this line, right?

sebastianbaginski: Why is it even there, I removed the comments...

marcusnaslund:
I think the title of this PR is a bit misleading.

davidhesselbom:
Haha, yes...

PR #766, opened 2015-11-16
Fixed binding of YUV.

emilwestergren:


source/draw/gpu/GpuSurface.ooc
```diff
@@ -83,7 +83,13 @@ GpuSurface: abstract class extends Canvas {
...
 		else
 			Debug raise("Invalid image type in GpuSurface!")
-		map add("texture0", temporary)
+		if (temporary instanceOf?(GpuYuv420Semiplanar)) {
+			yuv := temporary as GpuYuv420Semiplanar
+			map add("texture0", yuv y)
+			map add("texture1", yuv uv)
+		}
+		else
+			map add("texture0", temporary)
```
niklasborwell: } else

davidhesselbom: @thomasfanell This one should be a piece of cake for magic.

davidhesselbom: Well, if it weren't for version blocks...

emilwestergren: Oh yes, the ugly `else` format rule, how could I forget... We probably want it in magic, I will never get used to it.

niklasborwell:
Is there a reason why the appveyor test has started failing again?

davidhesselbom:
Duno, but it seems to always have to do with `Promise`. Maybe something in the Windows implementation broke when the threading classes in the SDK were cleaned up, I don't know.

Issue #771, opened 2015-11-16
Tests for Promise and PromiseCollector fail (too) often on AppVeyor


marcusnaslund:
They run just fine when tested on a local Windows machine, but seem to crash often on AppVeyor. This comes and goes, so is likely dependant on varying levels of available resources. But this is of course not good anyway.

Any ideas on how to make sure the tests run fine on AppVeyor without limiting them too much locally are much appreciated.

simonmika:
Stop using AppVeyor?

marcusnaslund:
Are we not interested in Windows support anymore?

simonmika:
Windows support is not that important. That might change but if it costs us time now I prefer not spending that time on it. On the other side, if there is a quick fix then fix it but if it takes forever drop Windows.

marcusnaslund:
That would solve #760 as well.

marcusnaslund:
@davidhesselbom - I guess you're the only one who can turn it off.

davidhesselbom:
Done. I guess `appveyor.yml` should be deleted as well, then.

davidhesselbom:
So... we can close this, I take it?

marcusnaslund:
Place a PR that removes appveyor.yml and close it, I suggest.

PR #775, opened 2015-11-16
Yuva

davidpiuva:
Added yuva color needed for pull request 861 in Vidhance.
I noticed that 50% of the code in other colors is legacy that won't even work with covers or makes no difference.

### DIFF
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -114,8 +114,8 @@ ColorYuv: cover {
 	// TODO: use valid default colors
 	init: func@ ~default { this init(0, 0, 0) }
 	init: func@ ~int (y, u, v: Int) { this init(y as UInt8, u as UInt8, v as UInt8) }
-	init: func@ ~float (y, u, v: Float) { this init(y*255.0f clamp(0.0f, 255.0f) as UInt8, u*255.0f clamp(0.0f, 255.0f) as UInt8, v*255.0f clamp(0.0f, 255.0f) as UInt8) }
-	init: func@ ~double (y, u, v: Double) { this init(y*255.0f clamp(0.0f, 255.0f) as UInt8, u*255.0f clamp(0.0f, 255.0f) as UInt8, v*255.0f clamp(0.0f, 255.0f) as UInt8) }
+	init: func@ ~float (y, u, v: Float) { this init(y * 255.0f clamp(0.0f, 255.0f) as UInt8, u * 255.0f clamp(0.0f, 255.0f) as UInt8, v * 255.0f clamp(0.0f, 255.0f) as UInt8) }
+	init: func@ ~double (y, u, v: Double) { this init(y * 255.0f clamp(0.0f, 255.0f) as UInt8, u * 255.0f clamp(0.0f, 255.0f) as UInt8, v * 255.0f clamp(0.0f, 255.0f) as UInt8) }
 	copy: func -> This { This new(this y, this u, this v) }
 	toMonochrome: func -> ColorMonochrome { ColorMonochrome new(this y) }
 	toYuv: func -> This { this copy() }
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -71,7 +71,6 @@ ColorUv: cover {
 		this v = color v
 	}
 	init: func@ (=u, =v)
-	// TODO: use valid default colors
 	init: func@ ~default { this init(0, 0) }
 	init: func@ ~int (i: Int) { this init(i as UInt8) }
 	init: func@ ~float (f: Float) { this init(f*255.0f clamp(0.0f, 255.0f) as UInt8) }
@@ -111,7 +110,6 @@ ColorYuv: cover {
 		this v = color v
 	}
 	init: func@ (=y, =u, =v)
-	// TODO: use valid default colors
 	init: func@ ~default { this init(0, 0, 0) }
 	init: func@ ~int (y, u, v: Int) { this init(y as UInt8, u as UInt8, v as UInt8) }
 	init: func@ ~float (y, u, v: Float) { this init(y * 255.0f clamp(0.0f, 255.0f) as UInt8, u * 255.0f clamp(0.0f, 255.0f) as UInt8, v * 255.0f clamp(0.0f, 255.0f) as UInt8) }
@@ -150,10 +148,6 @@ ColorYuva: cover {
 		this yuv = ColorYuv new(y, u, v)
 		this alpha = a
 	}
-	init: func@ ~default { this init(0, 128, 128, 255) }
-	init: func@ ~int (y, u, v, a: Int) { this init(y as UInt8, u as UInt8, v as UInt8, a as UInt8) }
-	init: func@ ~float (y, u, v, a: Float) { this init(y*255.0f clamp(0.0f, 255.0f) as UInt8, u*255.0f clamp(0.0f, 255.0f) as UInt8, v*255.0f clamp(0.0f, 255.0f) as UInt8, a*255.0f clamp(0.0f, 255.0f) as UInt8) }
-	init: func@ ~double (y, u, v, a: Double) { this init(y*255.0f clamp(0.0f, 255.0f) as UInt8, u*255.0f clamp(0.0f, 255.0f) as UInt8, v*255.0f clamp(0.0f, 255.0f) as UInt8, a*255.0f clamp(0.0f, 255.0f) as UInt8) }
 	toMonochrome: func -> ColorMonochrome { ColorMonochrome new(this yuv y) }
 	toYuv: func -> ColorYuv { this yuv }
 	toYuva: func -> This { this }
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -41,6 +41,7 @@ ColorMonochrome: cover {
 	copy: func -> This { This new(this y) }
 	toMonochrome: func -> This { this copy() }
 	toYuv: func -> ColorYuv { ColorYuv new(this y, 128, 128) }
+	toYuva: func -> ColorYuva { ColorYuva new(this y, 128, 128, 255) }
 	toBgr: func -> ColorBgr { ColorConvert yuvToBgr(this toYuv()) }
 	toBgra: func -> ColorBgra { this toBgr() toBgra() }
 	blend: func (factor: Float, other: This) -> This {
@@ -70,6 +71,7 @@ ColorUv: cover {
 		this v = color v
 	}
 	init: func@ (=u, =v)
+	// TODO: use valid default colors
 	init: func@ ~default { this init(0, 0) }
 	init: func@ ~int (i: Int) { this init(i as UInt8) }
 	init: func@ ~float (f: Float) { this init(f*255.0f clamp(0.0f, 255.0f) as UInt8) }
@@ -77,6 +79,7 @@ ColorUv: cover {
 	copy: func -> This { This new(this u, this v) }
 	toMonochrome: func -> ColorMonochrome { ColorMonochrome new() }
 	toYuv: func -> ColorYuv { ColorYuv new(128, this u, this v) }
+	toYuva: func -> ColorYuva { ColorYuva new(128, this u, this v, 255) }
 	toBgr: func -> ColorBgr { ColorConvert yuvToBgr(this toYuv()) }
 	toBgra: func -> ColorBgra { this toBgr() toBgra() }
 	blend: func (factor: Float, other: This) -> This {
@@ -108,6 +111,7 @@ ColorYuv: cover {
 		this v = color v
 	}
 	init: func@ (=y, =u, =v)
+	// TODO: use valid default colors
 	init: func@ ~default { this init(0, 0, 0) }
 	init: func@ ~int (y, u, v: Int) { this init(y as UInt8, u as UInt8, v as UInt8) }
 	init: func@ ~float (y, u, v: Float) { this init(y*255.0f clamp(0.0f, 255.0f) as UInt8, u*255.0f clamp(0.0f, 255.0f) as UInt8, v*255.0f clamp(0.0f, 255.0f) as UInt8) }
@@ -115,6 +119,7 @@ ColorYuv: cover {
 	copy: func -> This { This new(this y, this u, this v) }
 	toMonochrome: func -> ColorMonochrome { ColorMonochrome new(this y) }
 	toYuv: func -> This { this copy() }
+	toYuva: func -> ColorYuva { ColorYuva new(this, 255) }
 	toBgr: func -> ColorBgr { ColorConvert yuvToBgr(this) }
 	toBgra: func -> ColorBgra { this toBgr() toBgra() }
 	blend: func (factor: Float, other: This) -> This {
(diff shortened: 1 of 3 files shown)


source/draw/Color.ooc
```diff
@@ -70,6 +71,7 @@ ColorUv: cover {
 		this v = color v
 	}
 	init: func@ (=u, =v)
+	// TODO: use valid default colors
```
marcusnaslund: Unless this is fixed very soon, please also create an issue.

source/draw/Color.ooc
```diff
@@ -137,6 +142,31 @@ ColorYuv: cover {
...
+	yuv: ColorYuv
+	alpha: UInt8
+	init: func@ (=yuv, =alpha)
+	init: func@ ~uint8 (y, u, v, a: UInt8) {
+		this yuv = ColorYuv new(y, u, v)
+		this alpha = a
+	}
+	init: func@ ~default { this init(0, 128, 128, 255) }
+	init: func@ ~int (y, u, v, a: Int) { this init(y as UInt8, u as UInt8, v as UInt8, a as UInt8) }
+	init: func@ ~float (y, u, v, a: Float) { this init(y*255.0f clamp(0.0f, 255.0f) as UInt8, u*255.0f clamp(0.0f, 255.0f) as UInt8, v*255.0f clamp(0.0f, 255.0f) as UInt8, a*255.0f clamp(0.0f, 255.0f) as UInt8) }
```
emilwestergren: Space before and after operator: `y * 255.0f`

source/draw/Color.ooc
```diff
@@ -137,6 +142,31 @@ ColorYuv: cover {
...
+ColorYuva: cover {
+	yuv: ColorYuv
+	alpha: UInt8
+	init: func@ (=yuv, =alpha)
+	init: func@ ~uint8 (y, u, v, a: UInt8) {
+		this yuv = ColorYuv new(y, u, v)
+		this alpha = a
+	}
+	init: func@ ~default { this init(0, 128, 128, 255) }
+	init: func@ ~int (y, u, v, a: Int) { this init(y as UInt8, u as UInt8, v as UInt8, a as UInt8) }
```
marcusnaslund: Do you actually need this constructor at all? I think rock might correctly match the `~uint8` constructor and implicitly cast if you used `Int` arguments.

source/draw/Color.ooc
marcusnaslund: To be fair, the same problem already exists in multiple locations in this file. No reason to not fix it here though.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
```diff
@@ -86,6 +87,31 @@ Gles3ShaderProgram: class extends GLShaderProgram {
 		glUniform4f(glGetUniformLocation(this _backend, name), value x, value y, value z, value w)
 		version(debugGL) { validateEnd("ShaderProgram setUniform~FloatPoint4D") }
 	}
+	setUniform: override func ~ColorBgr (name: String, value: ColorBgr) {
+		version(debugGL) { validateStart("ShaderProgram setUniform~ColorBgr") }
+		glUniform3f(glGetUniformLocation(this _backend, name), (value blue as Float) / 255.0f, (value green as Float) / 255.0f, (value red as Float) / 255.0f)
```
emilwestergren: Can't you implement a `normalized` property returning `Float/FloatPoint2D/FloatPoint3D/FloatPoint4D` for every `Color` instead? I've already implemented it for `ColorBgra`. Then all these could be simplified to:
```ooc
this setUniform(name, value normalized)
```
OR we keep this layer primitive and only allow Int and Float Size/Point/Array, which means the conversion needs to be done when binding the shader variable:
```ooc
map add("color", color normalized)
```

What do you think?

davidpiuva: It would be nice to avoid having every possible type in setUniform but Simon did not like when I casted a color to a point and we have no generic 4D vector for GPU upload.

source/draw/Color.ooc
davidhesselbom: I'm not even sure what this means?

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
emilwestergren: Yes we have, I created the beautiful `FloatPoint4D` for this reason

source/draw/Color.ooc
davidpiuva: (0,0) is way off the valid RGB values for any luma.

davidhesselbom: You're saying YUV can represent colors RGB cannot?

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
davidpiuva: Simon said that a point is not a color when I tried.

source/draw/Color.ooc
davidpiuva: Yes, this can preserve correct hue on oversaturated images so that the receiving television can scale it down.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
davidhesselbom: Btw, `FloatPoint4D` apparently has an invalid default constructor: `	init: func@ ~default { this init(0.0f, 0.0f, 0.0f, 1.0f) }`

Apparently you shouldn't assign anything to anything but 0 in default constructors for covers, because if you create an array of these things, you should be able to assume that they have all been "constructed", but they will be in an invalid state (alpha will be 0) if you do create an array and the default constructor looks like this. Something like that.

source/draw/Color.ooc
davidhesselbom: Interesting.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
emilwestergren: We should have simple Float2, Float3, Float4 covers for this purpose I guess.

davidpiuva: I predict a loud discussion ahead.

emilwestergren: I'm not even sure the UInt8 -> normalized float conversion should be applied here.

davidhesselbom: > We should have simple Float2, Float3, Float4 covers for this purpose I guess.

Agreed. They could probably be used for other things, too, like the Point/Size thing for translations that no one seems very happy about...

emilwestergren: This API will not be public for much longer anyway, so I'd rather keep this one as simple as possible with only Float/Int variations and move the conversions to either `GpuMap` or demand the user does it. In the end it is *your* shader variable so is it too much to ask that you convert it to the correct format yourself?

davidpiuva: I converted it to FloatPoint4D in the first version but Simon said no.

bettinaselig:
Why do the tests fail?

marcusnaslund:
> Why do the tests fail?

#771

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
emilwestergren: I agree the `FloatPoint4D` solution was ugly though, but if we change it to `Float2, Float3, Float4` or `Vector2, Vector3, Vector4`? @simonmika

simonmika: No, stop leaking OpenGL stuff. Use the ability to transfer a color and transfer points, sizes, transforms separately.

bettinaselig:
Thanks.

source/draw/Color.ooc
bettinaselig: Why do you talk about rgb here?

source/draw/Color.ooc
bettinaselig: magic? @thomasfanell

source/draw/Color.ooc
davidpiuva: Because it makes no sense to have a color that is (-100, 100, 0) in RGB.

bettinaselig: I thought we express everything in YUV and never convert it to RGB? Am I wrong?

davidpiuva: At some stage, someone have to convert to RGB before displaying on the monitor or CMYK for print.

bettinaselig: So it's only for displaying? @emilwestergren

davidpiuva: Going above 255 in RGB is okay but negative amount of light has no equivalence with reality.

source/draw/Color.ooc
davidpiuva: I can remove ~default, ~int, ~float and ~double to keep it simple.

source/draw/Color.ooc
davidhesselbom: The check for `*` was removed because magic couldn't tell a `Float*` from a `1.0f * 2.0f`, I think. It gave false positives.

bettinaselig: Aha. Did not know that we used pointers at all.

davidpiuva: Should I correct all colors?

source/draw/Color.ooc
```diff
@@ -137,6 +142,31 @@ ColorYuv: cover {
...
 
+ColorYuva: cover {
+	yuv: ColorYuv
+	alpha: UInt8
+	init: func@ (=yuv, =alpha)
+	init: func@ ~uint8 (y, u, v, a: UInt8) {
+		this yuv = ColorYuv new(y, u, v)
+		this alpha = a
+	}
+	init: func@ ~default { this init(0, 128, 128, 255) }
```
davidhesselbom: @simonmika you talked about this recently, and I seem to remember you saying that a default constructor shouldn't initialize to anything but 0. What should we do in a case like this?

marcusnaslund:
> I noticed that 50% of the code in other colors is legacy that won't even work with covers or makes no difference.

Will you create an issue about this? @davidpiuva

davidpiuva:
I can make an issue.

source/draw/Color.ooc
bettinaselig: Anyway... @davidpiuva can you make an issue concerning this?

source/draw/Color.ooc
davidhesselbom: No, don't make the PR bigger than it needs to be.

source/draw/Color.ooc
davidpiuva: Created issue #781.

bettinaselig: Thanks.

source/draw/Color.ooc
simonmika: For covers yes. Keep this one. I don't know how to handle it, especially not in this case.

bettinaselig:
https://github.com/cogneco/ooc-kean/issues/781

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
bettinaselig: @simonmika How shall things be done? *In this pull request*?

source/draw/Color.ooc
bettinaselig: @marcusnaslund @davidhesselbom Shall he remove them?

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
emilwestergren: I'm not talking about leaking it, I'm talking about moving the conversions to a higher level (but still not in the public API). We already have a OpenGL-specific type check for every shader variable upload value, any conversions to OpenGL could be made there. Then we keep the low level ShaderProgram API clean and primitive (as it should be).

simonmika: Here I agree with you @emilwestergren. Misunderstood the context of it.

source/draw/Color.ooc
bettinaselig: @simonmika @davidhesselbom @marcusnaslund Shall the comments be removed again?

source/draw/Color.ooc
marcusnaslund: If they don't serve a purpose, yes.

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
bettinaselig: What Emil describes is not in the scope of this pr. How shall it be done here?

emilwestergren: So how about this, `ShaderProgram setUniform` overloads (and same for integers):
```ooc
setUniform: func (value: Float)
setUniform: func (value1, value2: Float)
setUniform: func (value1, value2, value3: Float)
setUniform: func (value1, value2, value3, value4: Float)
setUniform: func (values: Float*, count: Int)
```

In OpenGLMap use() conversions from Imint standard to OpenGL standard can be added:
```ooc
			program := this program
			if (value instanceOf?(Cell)) {
				cell := value as Cell
				match (cell T) {
					case Int => program setUniform(key, cell as Cell<Int> get())
					case IntPoint2D => program setUniform(key, cell as Cell<IntPoint2D> get())
					case IntPoint3D => program setUniform(key, cell as Cell<IntPoint3D> get())
					case IntSize2D => program setUniform(key, cell as Cell<IntSize2D> get())
					case IntSize3D => program setUniform(key, cell as Cell<IntSize3D> get())
					case Float => program setUniform(key, cell as Cell<Float> get())
					case FloatPoint2D => program setUniform(key, cell as Cell<FloatPoint2D> get())
					case FloatPoint3D => program setUniform(key, cell as Cell<FloatPoint3D> get())
					case FloatPoint4D => program setUniform(key, cell as Cell<FloatPoint4D> get())
					case FloatSize2D => program setUniform(key, cell as Cell<FloatSize2D> get())
					case FloatSize3D => program setUniform(key, cell as Cell<FloatSize3D> get())
					case FloatTransform2D => program setUniform(key, cell as Cell<FloatTransform2D> get())
					case FloatTransform3D => program setUniform(key, cell as Cell<FloatTransform3D> get())
					case => Debug raise("Invalid cover type in OpenGLMap use!")
				}
			}
			else
				match (value) {
					case image: OpenGLPacked => {
						image backend bind(textureCount)
						program setUniform(key, textureCount)
						textureCount += 1
					}
					case image: OpenGLVolumeMonochrome => {
						image _backend bind(textureCount)
						program setUniform(key, textureCount)
						textureCount += 1
					}
					case => Debug raise("Invalid object type in OpenGLMap use: %s" format(value class name))
				}
```

emilwestergren: Just merge it and I can take care of it later...

bettinaselig: Thanks!

davidpiuva:
I should do more testing of the GPU upload.

davidpiuva:
Uploading to GPU seems to work.

PR #789, opened 2015-11-17
Renamed toStringFormat -> toString in DateTime and TimeSpan

sebastianbaginski:
@simonmika told me that we can make a PRs for the issues in "PR Debt" and "Cleanup" milestone in Kean.
Fixes https://github.com/cogneco/ooc-kean/issues/773

### DIFF
--- a/source/base/DateTime.ooc
+++ b/source/base/DateTime.ooc
@@ -94,9 +94,6 @@ DateTime: cover {
 		This _ticksToDateTimeHelper(this ticks) year
 	}
 
-	toString: func -> String {
-		this toStringFormat(This defaultFormat)
-	}
 	// <summary>
 	// Convert this object to string representation
 	// </summary>
@@ -112,7 +109,7 @@ DateTime: cover {
 	//	%ss - second
 	//	%zzzz - millisecond
 	// <param name="format">output format specification</param>
-	toStringFormat: func (format: String) -> String {
+	toString: func (format := This defaultFormat) -> String {
 		result := format
 		data := This _ticksToDateTimeHelper(this ticks)
 		result = result replaceAll("%yyyy", "%d" format(data year))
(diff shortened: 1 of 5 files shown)


source/base/TimeSpan.ooc
```diff
@@ -54,14 +54,14 @@ TimeSpan: cover {
 	//  %h - hours (<24)
 	//  %m - minutes (<60)
 	//  %s - seconds (<60)
-	//  %z - milliseconds (<1000)	
+	//  %z - milliseconds (<1000)
```
marcusnaslund: This is a copy of the line below for some reason, so just remove the whole thing.

sebastianbaginski: Updated.

source/base/DateTime.ooc
```diff
@@ -112,7 +112,7 @@ DateTime: cover {
 	//	%ss - second
 	//	%zzzz - millisecond
 	// <param name="format">output format specification</param>
-	toStringFormat: func (format: String) -> String {
+	toString: func ~format (format: String) -> String {
```
davidhesselbom: Hmmm... but why not
```ooc
toString: func (format := This defaultFormat) -> String {
```
?

marcusnaslund: Why? `toString()` would already do this.

davidhesselbom: Because then you don't need 2 overloads, all you need is the one with the default parameter.

marcusnaslund: Oh, right, then we don't need `toString`.

davidhesselbom: Or rather, we don't need `toString ~format`.

sebastianbaginski: True. Updated.

PR #791, opened 2015-11-17
Simplified union in IntBox2D

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/753

### DIFF
--- a/source/math/IntBox2D.ooc
+++ b/source/math/IntBox2D.ooc
@@ -86,11 +86,7 @@ IntBox2D: cover {
 		This new(left, top, width, height)
 	}
 	union: func ~point (point: IntPoint2D) -> This {
-		left := Int minimum~two(this left, point x)
-		top := Int minimum~two(this top, point y)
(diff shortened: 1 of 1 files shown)


marcusnaslund:
So, this function extends the box to contain a given point? Don't think `union` would be my first choice for the name. But that's not for this PR.

davidhesselbom:
Agreed. I think `engulf` or `swallow` would make sense (even if they sound silly). But then, what *is* the definition of a union of a box and a point?

sebastianbaginski:
> what is the definition of a union of a box and a point?

Good *point* :) Well, you can think of a point as a kind of retarded rectangle.

davidhesselbom:
> Well, you can think of a point as a kind of retarded rectangle.

I always thought of it as a very small circle... anyway, this one goes in my book of quotes.

marcusnaslund:
@sebastianbaginski I will let this through, but could you also create a similar PR for `FloatBox2D`?

sebastianbaginski:
Sure, no problem.

PR #795, opened 2015-11-17
Renamed temporary variable in parse methods to 'parts'

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/780


sebastianbaginski:
> marcusnaslund merged commit ..

Nice, so you can push the magic button now too :)

Issue #783, opened 2015-11-16
Use Size when it should be used.


simonmika:
Use `size` for translations not points. For example in `FloatEuclidTransform`. See https://github.com/cogneco/ooc-kean/pull/662 for details.

simonmika:
Let us also change the properties of all size covers into `x` and `y`.

bettinaselig:
Change `FloatSize2D` to a more intuitive name. @niklasborwell @sebastianbaginski @marcusnaslund @davidhesselbom

bettinaselig:
Suggestions: `FloatVector2D`, `FloatDirection2D`

simonmika:
What about `FloatDisplacement2D`?

davidhesselbom:
Are we keeping `FloatSize2D`? A `FloatBox2D` consisting of a `FloatPoint2D` and a `FloatDisplacement2D` doesn't seem very intuitive.

bettinaselig:
Just brain storming... `FloatTranslation2D`

bettinaselig:
Can collect first and then make pick the best. please?

marcusnaslund:
This has probably been brought up before but I'd like to point out that the `Point` covers have some properties that are generally associated (mathematically) to vectors, not points. So having a `FloatVector2D` and a `FloatPoint2D` would imply some consistency there which we don't have.

davidhesselbom:
`FloatDimensions2D` should cover both size and translations.

bettinaselig:
Ok. We can also discuss if `FloatPoint2D` can get a better name.

niklasborwell:
If we're gonna make this change, I would prefer to clean up the things @marcusnaslund mentioned. Otherwise we're just exchanging a half-weird system for another.

bettinaselig:
In computer graphics the terms are *location vector* and *direction vector*.

bettinaselig:
Position: https://en.wikipedia.org/wiki/Position_%28vector%29
Direction: https://en.wikipedia.org/wiki/Direction_vector

niklasborwell:
So I would suggest, remove all the things that are actually vector-related and move them into a general vector class, FloatVector2D. Keep the FloatPoints, but more specific. 
I think anybody would feel comfortable using a FloatVector2D, as it is a standard concept. A FloatDisplacement or FloatTranslation would just seem like special cases that would cause more confusion.

marcusnaslund:
[...]

PR #765, opened 2015-11-16
Math: Fixes and cleanup to Box and Size

marcusnaslund:
Replaces #610 because VSCode messed up my git history (at least that's what I'm blaming it on).

PR #796, opened 2015-11-17
Simplified union for FloatBox2D

sebastianbaginski:


source/math/FloatBox2D.ooc
```diff
@@ -87,18 +87,10 @@ FloatBox2D: cover {
 		This new(left, top, width, height)
 	}
 	union: func ~box (other: This) -> This { // Rock bug: Union without suffix cannot be used because the C name conflicts with keyword "union"
-		left := Float minimum(this left, other left)
-		top := Float minimum(this top, other top)
-		width := Float maximum(0, (Float maximum(this right, other right) - left))
-		height := Float maximum(0, (Float maximum(this bottom, other bottom) - top))
-		This new(left, top, width, height)
+		This new(this leftTop minimum(other leftTop), this rightBottom maximum(other rightBottom))
```
marcusnaslund: Is a `Size` cover guaranteed to be nonnegative? The previous version made sure width and height were >= 0, that doesn't happen here.

marcusnaslund: Whether we should do that is a different question, though.

marcusnaslund:
Letting this one wait until #765 is in.

source/math/FloatBox2D.ooc
sebastianbaginski: I think this is the same change as in original issue : https://github.com/cogneco/ooc-kean/issues/753

marcusnaslund: There you changed only the one with a `Point` argument, here you changed both.

sebastianbaginski: That's true. I don't know really. Tests passed, everything seems to work, let's carry on

marcusnaslund: The discussion in #783 seems to show that size *should be* nonnegative. And you're right, this seems to work just fine now.

marcusnaslund: Also, if `width` or `height` are negative, then `leftTop` is not actually the left top, so.

PR #797, opened 2015-11-17
Removed TODO in ThreadLocalWin32

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/777


marcusnaslund:
I guess I'll just let this slide in. Doesn't affect anything we do and testing it on Windows seems to work.

PR #799, opened 2015-11-17
Optimized FloatMatrix methods/operators by assigning elements pointer to variable

### LINKED ISSUE #578: FloatMatrix: get elements pointer once
As a result of #488, in `FloatMatrix`, each usage of `elements` results in a function call. This results in lots of function calls when elements are accessed in loops (which happens in a lot of functions in `FloatMatrix`. To fix this, create a variable for `this elements` first in the function, and then use that instead of `this elements`.

tomasmahlberg:
Fixes #578.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -65,8 +65,9 @@ FloatMatrix : cover {
 	identity: static func (order: Int) -> This {
 		result := This new(order, order)
 		resultWidth := result take() width
+		resultElements := result elements
 		for (i in 0 .. order)
-			result elements[i + resultWidth * i] = 1.0f
+			resultElements[i + resultWidth * i] = 1.0f
 		result
 	}
 	setVertical: func (xOffset, yOffset: Int, vector: FloatPoint3D) {
@@ -123,9 +124,11 @@ FloatMatrix : cover {
 	transpose: func -> This {
 		t := this take()
 		result := This new(t dimensions swap())
+		resultElements := result elements
+		thisElements := this elements
 		for (y in 0 .. t height)
 			for (x in 0 .. t width)
-				result elements[y + x * t height] = t elements[x + y * t width]
+				resultElements[y + x * t height] = thisElements[x + y * t width]
 		this free(Owner Receiver)
 		result
 	}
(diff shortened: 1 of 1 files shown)


source/math/FloatMatrix.ooc
```diff
@@ -141,13 +143,13 @@ FloatMatrix : cover {
 	}
 	swapRows: func@ (first, second: Int) {
 		t := this take()
-		order := t order
 		buffer: Float
+		thisElements := this elements
 		if (first != second)
-			for (i in 0 .. order) {
+			for (i in 0 .. t order) {
 				buffer = t elements[i + first * t width]
```
marcusnaslund: `thisElements` here also?

marcusnaslund:
Not as many elements are affected, but we want to do the same on line 69, no?

source/math/FloatMatrix.ooc
```diff
@@ -123,9 +123,11 @@ FloatMatrix : cover {
 	transpose: func -> This {
 		t := this take()
```
niklasborwell: I'm kind of new to this take and give thing, but is this something that we use as standard? I mean to use t as a variable name for take. Otherwise I would say it's a bit confusing.

tomasmahlberg:
I deliberately left out the one in `identity` but let's change it for consistency.

source/math/FloatMatrix.ooc
tomasmahlberg: I think it was SImon who first introduced the use of `t` in this way...

marcusnaslund: That's true, `Text` does it everywhere.

niklasborwell: I had a feeling that was the case. Fine by me then.

davidhesselbom: You don't mess with the Simon.

marcusnaslund:
Is this high because it's very urgent or a remaining habit from the PR sprint? @niklasborwell

niklasborwell:
Neither. The third reason for setting something high is that I want to make sure no one has to sit and wait for a long time while more important pull requests keep popping up and get handled. So this is just high basically because it was Tomas turn.

niklasborwell:
Oh wait, I didn't read your comment correctly.

niklasborwell:
Yes it is a remaining habit. I don't think anyone said we should stop? It was even up for discussion whether we should keep this system going.

marcusnaslund:
Right, just wanted to check. I will discuss this with @simonmika later today. In the mean time you may do as you please, but most of these cleanup/prd PRs are so small that you won't have to wait for long anyway (which means I might not look too closely at the high label).

At the moment I'm having trouble testing things on Android - that's why this PR is delayed - but it will get processed today.

PR #802, opened 2015-11-17
Renamed createFrom to convertFrom in RasterYuv422Semipacked and made it static

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/772

### DIFF
--- a/source/draw/RasterYuv422Semipacked.ooc
+++ b/source/draw/RasterYuv422Semipacked.ooc
@@ -34,35 +34,6 @@ RasterYuv422Semipacked: class extends RasterPacked {
 	init: func ~allocate (size: IntSize2D) { super~allocate(size) }
 	init: func ~fromByteBuffer (buffer: ByteBuffer, size: IntSize2D) { super(buffer, size, this bytesPerPixel * size width) }
 	init: func ~fromRasterImage (original: RasterImage) { super(original) }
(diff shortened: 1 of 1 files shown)


source/draw/RasterYuv422Semipacked.ooc
```diff
@@ -129,6 +100,30 @@ RasterYuv422Semipacked: class extends RasterPacked {
 		fileWriter write(this buffer pointer as Char*, this buffer size)
 		fileWriter close()
 	}
+	convertFrom: static func (original: RasterImage) -> This {
+		result := This new(original size)
```
marcusnaslund: The other `convertFrom` check
```ooc
if (original instanceOf?(This))
	result = (original as This) copy()
```

sebastianbaginski: That's true, thanks.

sebastianbaginski: Corrected.

marcusnaslund:
Can you create an issue for creating a test for this class?

sebastianbaginski:
Sure: https://github.com/cogneco/ooc-kean/issues/808

marcusnaslund:
Thanks. Looks good otherwise. If there's something wrong, we'll know with the test.

Issue #785, opened 2015-11-17
Text test reads from invalid memory and has invalid free/delete bug


sebastianbaginski:
Something is wrong with the `operator == (Text, Text)` method in Text class, it causes invalid memory reads.
To reproduce, run
```
./test.sh base/Text
```
and run valgrid on `Tests` executable.
Invalid free/delete is in the same method.

sebastianbaginski:
```
// class Text
	operator == (other: This) -> Bool {
		result := this _buffer == other _buffer
		if (this _buffer raw != other _buffer raw)
			other free(Owner Receiver)
		this free(Owner Receiver)
		result
	}

// class TextBuffer:
	raw: Char* { get {
		result := this _backend pointer as Char*
		this free(Owner Receiver)
		result
	}}
```
Looks like `if (this _buffer raw != other _buffer raw)` already calls ` free(Owner Receiver)` on both objects. @simonmika

tomasmahlberg:
@simonmika Are you looking at this or do you want me to take a look? It should be an easy fix. I think @davidhesselbom wants this into upstream ASAP as it is currently a bit broken.

tomasmahlberg:
So, I took the liberty to fix this: #803.

PR #803, opened 2015-11-17
Fixed bugs in + and == operators in Text

tomasmahlberg:
Issue #785.

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -51,7 +51,7 @@ Text: cover {
 	operator == (string: String) -> Bool { this == This new(string) }
 	operator == (other: This) -> Bool {
 		result := this _buffer == other _buffer
-		if (this _buffer raw != other _buffer raw)
+		if (this _buffer _backend pointer != other _buffer _backend pointer)
 			other free(Owner Receiver)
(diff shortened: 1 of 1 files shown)


source/base/Text.ooc
```diff
@@ -51,7 +51,7 @@ Text: cover {
 	operator == (string: String) -> Bool { this == This new(string) }
 	operator == (other: This) -> Bool {
 		result := this _buffer == other _buffer
-		if (this _buffer raw != other _buffer raw)
```
davidhesselbom: Haha, oops...

davidhesselbom:
@marcusnaslund Can you try to give this priority? It'd be awfully nice to get rid of that invalid free before people learn to ignore that sort of error.

marcusnaslund:
Yep, coming up.

simonmika:
@tomasmahlberg Shouldn't this be a part of `TextBuffer` instead?

tomasmahlberg:
@simonmika Might make more sense actually, since the operators work mostly with the `TextBuffer`s anyway. What do you think?

simonmika:
Do a PR.

PR #809, opened 2015-11-17
Cleaned ShaderProgram API.

emilwestergren:
Fixes https://github.com/cogneco/ooc-kean/issues/793

The ShaderProgram API has been trimmed to only accept primitive types like vectors and matrices. Conversions to these types are instead located in a higher abstraction layer (OpenGLMap). This loosens the coupling between the user layer and the backend OpenGL layer.


niklasborwell:
A short description would be nice. I looked at the issue, but it didn't help that much either.

source/draw/gpu/opengl/Map/OpenGLMap.ooc
```diff
@@ -65,23 +65,23 @@ OpenGLMap: abstract class extends GpuMap {
...
+					case FloatSize3D => size := cell as Cell<FloatSize3D> get(); program setUniform(key, size width, size height, size depth)
 					case FloatTransform2D => program setUniform(key, cell as Cell<FloatTransform2D> get())
 					case FloatTransform3D => program setUniform(key, cell as Cell<FloatTransform3D> get())
-					case ColorBgr => program setUniform(key, cell as Cell<ColorBgr> get())
-					case ColorBgra => program setUniform(key, cell as Cell<ColorBgra> get())
-					case ColorUv => program setUniform(key, cell as Cell<ColorUv> get())
-					case ColorYuv => program setUniform(key, cell as Cell<ColorYuv> get())
-					case ColorYuva => program setUniform(key, cell as Cell<ColorYuva> get())
+					case ColorBgr => color := cell as Cell<ColorBgr> get(); program setUniform(key, color blue as Float / 255, color green as Float / 255, color red as Float / 255)
+					case ColorBgra => color := cell as Cell<ColorBgra> get(); program setUniform(key, color blue as Float / 255, color green as Float / 255, color red as Float / 255, color alpha as Float / 255)
```
davidpiuva: Color vectors in GLSL (excluding texture packing) are always in RGBA order so it makes sense to reverse them here to prevent the red channel from being in color.b and blue channel from being in color.r.

davidpiuva:
I have reviewed it now.

source/draw/gpu/opengl/Map/OpenGLMap.ooc
emilwestergren: Sure

emilwestergren:
Updated.

PR #810, opened 2015-11-18
Test case for RasterYuv422Semipacked, fixed some memory leaks

sebastianbaginski:
- fixed memory leak in `RasterBgr open`
- fixed several memory leaks in `RasterYuv422Semipacked`
- test for `RasterYuv422Semipacked`

Fixes https://github.com/cogneco/ooc-kean/issues/808

### DIFF
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -117,7 +117,7 @@ RasterBgr: class extends RasterPacked {
 		x, y, imageComponents: Int
 		requiredComponents := 3
 		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
-		This new(ByteBuffer new(data as UInt8*, x * y * requiredComponents), IntSize2D new(x, y))
+		This new(ByteBuffer new(data as UInt8*, x * y * requiredComponents, true), IntSize2D new(x, y))
 	}
 	convertFrom: static func (original: RasterImage) -> This {
 		result: This
--- a/source/draw/RasterYuv422Semipacked.ooc
+++ b/source/draw/RasterYuv422Semipacked.ooc
@@ -46,7 +46,9 @@ RasterYuv422Semipacked: class extends RasterPacked {
 		result
 	}
 	apply: func ~bgr (action: Func(ColorBgr)) {
-		this apply(ColorConvert fromYuv(action))
+		convert := ColorConvert fromYuv(action)
+		this apply(convert)
+		(convert as Closure) dispose()
 	}
 	apply: func ~yuv (action: Func (ColorYuv)) {
 		row := this buffer pointer as UInt8*
@@ -63,7 +65,9 @@ RasterYuv422Semipacked: class extends RasterPacked {
 		}
 	}
 	apply: func ~monochrome (action: Func(ColorMonochrome)) {
-		this apply(ColorConvert fromYuv(action))
+		convert := ColorConvert fromYuv(action)
+		this apply(convert)
+		(convert as Closure) dispose()
 	}
 	operator [] (x, y: Int) -> ColorYuv {
 		result := ColorYuv new()
@@ -82,13 +86,10 @@ RasterYuv422Semipacked: class extends RasterPacked {
 		}
 	}
 	open: static func (filename: String) -> This {
-		x, y, imageComponents: Int
-		requiredComponents := 3
-		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
-		bgr := RasterBgr new(ByteBuffer new(data as UInt8*, x * y * requiredComponents), IntSize2D new(x, y))
-		result := This new(bgr)
+		bgr := RasterBgr open(filename)
+		result := This convertFrom(bgr)
 		bgr referenceCount decrease()
-		return result
+		result
 	}
 	save: func (filename: String) {
 		bgr := RasterBgr convertFrom(this)
--- a/test/draw/RasterYuv422SemipackedTest.ooc
+++ b/test/draw/RasterYuv422SemipackedTest.ooc
@@ -1,0 +1,40 @@
+use ooc-unit
+use ooc-draw
+use ooc-math
+import math
+
+RasterYuv422SemipackedTest: class extends Fixture {
+	_inputPath := "test/draw/input/Flower.png"
+	init: func {
+		super("RasterYuv422Semipacked")
+		this add("create and copy", func {
+			raster := RasterYuv422Semipacked open(this _inputPath)
+			rasterCopy := raster copy()
+			expect(raster width, is equal to(rasterCopy width))
+			expect(raster height, is equal to(rasterCopy height))
+			expect(raster stride, is equal to(rasterCopy stride))
+			rasterCopy referenceCount decrease()
+			outputPath := "test/draw/output/RasterYuv422Semipacked_test1.png"
+			raster save(outputPath)
+			outputPath free()
+			raster referenceCount decrease()
+			size := IntSize2D new(256, 256)
+			raster = RasterYuv422Semipacked new(size)
+			expect(size width, is equal to(raster width))
+			expect(size height, is equal to(raster height))
+			for (row in 0 .. size height)
+				for (column in 0 .. size width)
+					raster[column, row] = ColorBgr new(row, 0, column) toYuv()
+			outputPath = "test/draw/output/RasterYuv422Semipacked_test2.png"
+			raster save(outputPath)
+			raster referenceCount decrease()
+			outputPath free()
+		})
+	}
+	free: override func {
+		this _inputPath free()
+		super()
+	}
+}
+
+RasterYuv422SemipackedTest new() run() . free()
--- a/test/draw/RasterYuv422SemipackedTest.ooc
+++ b/test/draw/RasterYuv422SemipackedTest.ooc
@@ -5,6 +5,8 @@ import math
 
 RasterYuv422SemipackedTest: class extends Fixture {
 	_inputPath := "test/draw/input/Flower.png"
+	_outputPathFirst := "test/draw/output/RasterYuv422Semipacked_test1.png"
+	_outputPathSecond := "test/draw/output/RasterYuv422Semipacked_test2.png"
 	init: func {
 		super("RasterYuv422Semipacked")
 		this add("create and copy", func {
@@ -14,9 +16,7 @@ RasterYuv422SemipackedTest: class extends Fixture {
 			expect(raster height, is equal to(rasterCopy height))
 			expect(raster stride, is equal to(rasterCopy stride))
 			rasterCopy referenceCount decrease()
-			outputPath := "test/draw/output/RasterYuv422Semipacked_test1.png"
-			raster save(outputPath)
-			outputPath free()
+			raster save(this _outputPathFirst)
 			raster referenceCount decrease()
 			size := IntSize2D new(256, 256)
 			raster = RasterYuv422Semipacked new(size)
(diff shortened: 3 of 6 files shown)


test/draw/RasterYuv422SemipackedTest.ooc
```diff
@@ -0,0 +1,42 @@
...
+	}
+	free: override func {
+		this _inputPath free()
+		super()
+	}
+}
+
+test := RasterYuv422SemipackedTest new()
+test run()
+test free()
```
marcusnaslund: I noticed some `draw` tests do this, but I'd prefer if you follow the pattern of most tests. That would be:
`RasterYuv422SemipackedTest new() run()` and possibly ` . free()` after, as for `RasterBgrTest.ooc` for example.

sebastianbaginski: Oh yeah I forgot about this

test/draw/RasterYuv422SemipackedTest.ooc
```diff
@@ -0,0 +1,40 @@
...
+			for (row in 0 .. size height)
+				for (column in 0 .. size width)
+					raster[column, row] = ColorBgr new(row, 0, column) toYuv()
+			outputPath = "test/draw/output/RasterYuv422Semipacked_test2.png"
+			raster save(outputPath)
+			raster referenceCount decrease()
+			outputPath free()
+		})
+	}
+	free: override func {
```
marcusnaslund: All tests leak memory like crazy, and we don't really care, but OK. :)

sebastianbaginski: I *do* care, I think @davidhesselbom too ;)

source/draw/RasterBgr.ooc
```diff
@@ -117,7 +117,7 @@ RasterBgr: class extends RasterPacked {
 		x, y, imageComponents: Int
 		requiredComponents := 3
 		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
-		This new(ByteBuffer new(data as UInt8*, x * y * requiredComponents), IntSize2D new(x, y))
+		This new(ByteBuffer new(data as UInt8*, x * y * requiredComponents, true), IntSize2D new(x, y))
```
marcusnaslund: None of the other classes do this. Why here?

sebastianbaginski: By default this is set to false which means the data is not owned by the buffer. We want the returned image own the allocated data. It will leak otherwise.

marcusnaslund: So.. this should be fixed for all the other image classes too?

sebastianbaginski: Where exactly ?

marcusnaslund: `RasterBgra`, `RasterMonochrome`, `RasterUv` all make similar calls to `ByteBuffer` in their `open()` methods.

sebastianbaginski: Fixed now.

marcusnaslund: Awesome. If you can get someone relevant to peer review this I'd be happy. I'm still quite new to the graphics stuff here.

sebastianbaginski: Not much to review here, test basically look like any other for Raster classes and other changes are minor. You'll be fine :D

test/draw/RasterYuv422SemipackedTest.ooc
marcusnaslund: If we do care, that deserves an issue too. Probably in `Later`, though.

source/draw/RasterBgr.ooc
davidhesselbom: I think it would make more sense to invert the default and assume that unles you say otherwise, a `ByteBuffer` owns its data. But it's a matter of debate, and can be debated later.

sebastianbaginski:
@emilwestergren do you have time to review this ?

emilwestergren:
Sure

source/draw/RasterBgr.ooc
emilwestergren: My thoughts were that if you pass memory from outside, you have to explicitly pass the ownership. Since the ByteBuffer did not allocate the memory it makes sense that it doesn't take ownership by default. I agree though that we could switch if it's by far the most common scenario.

source/draw/RasterYuv422Semipacked.ooc
```diff
@@ -82,13 +86,10 @@ RasterYuv422Semipacked: class extends RasterPacked {
 		}
 	}
 	open: static func (filename: String) -> This {
-		x, y, imageComponents: Int
-		requiredComponents := 3
-		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
-		bgr := RasterBgr new(ByteBuffer new(data as UInt8*, x * y * requiredComponents), IntSize2D new(x, y))
-		result := This new(bgr)
+		bgr := RasterBgr open(filename)
```
emilwestergren: I know this was here before but should this class only support loading BGR?

source/draw/RasterBgr.ooc
davidhesselbom: You're right... I think opening image files with StbImage is the only time we can't read into an existing ByteBuffer - it simply has no API for it (which makes sense, since it's also responsible for figuring out how much memory we need to allocate: *It* tells *us* how many channels and pixels are in the image we're trying to open, not the other way around).

source/draw/RasterYuv422Semipacked.ooc
davidhesselbom: What else do you want support for? Monochrome images?

source/draw/RasterBgr.ooc
sebastianbaginski: I can't remember which function was it exactly, but there is one call in Windows API that you must do twice - first you pass a null buffer so it tells you how big buffer you need, then on second call you pass the non-null buffer pointer allocated with the size returned from first call. Simply beautiful isn't it ? :speak_no_evil:

test/draw/RasterYuv422SemipackedTest.ooc
```diff
@@ -0,0 +1,40 @@
...
+	init: func {
+		super("RasterYuv422Semipacked")
+		this add("create and copy", func {
+			raster := RasterYuv422Semipacked open(this _inputPath)
+			rasterCopy := raster copy()
+			expect(raster width, is equal to(rasterCopy width))
+			expect(raster height, is equal to(rasterCopy height))
+			expect(raster stride, is equal to(rasterCopy stride))
+			rasterCopy referenceCount decrease()
+			outputPath := "test/draw/output/RasterYuv422Semipacked_test1.png"
```
emilwestergren: Maybe you can have all paths as class members like you did with input?

sebastianbaginski: Sure we can keep all of them in the same place.

source/draw/RasterBgr.ooc
davidhesselbom: ... no.

test/draw/RasterYuv422SemipackedTest.ooc
sebastianbaginski: Updated.

source/draw/RasterYuv422Semipacked.ooc
emilwestergren: I just don't think it's obvious that `RasterYuv422Semipacked open` expects a BGR image. There is `openRaw` though, but this one seems more like the special case since it has a conversion. Anyway, it's not part of this PR so let's just leave it.

source/draw/RasterBgr.ooc
sebastianbaginski: I think it was this one : https://msdn.microsoft.com/en-us/library/windows/desktop/aa372603%28v=vs.85%29.aspx
`PdhEnumObjectsH`, we should make our API more like this. And remember to add `1` to the result, but only on Windows XP :D

emilwestergren: Haha, awesome.

source/draw/RasterYuv422Semipacked.ooc
davidhesselbom: Anyway, StbImage will load a PNG or JPG. JPG can only be BGR afaik, while PNG can also be BGRA and possibly monochrome as well. StbImage will then convert the image from whichever number of channels into 3 channels (since we're opening it as a BGR), which will then be converted by us into YUV. It wouldn't make sense to keep the alpha channel when decoding the image since YUV doesn't have such a channel, and the only downside to assuming the image is BGR and not monochrome is that StbImage will unnecessarily convert it from monochrome to BGR only to let us convert it back to a monochrome YUV.

davidhesselbom: And yeah, if you know the image you're trying to load is YUV (i.e. not a PNG or JPG), use `openRaw`.

davidhesselbom: Although, yes, we could check the file extension here and call `openRaw` from inside `open`.

emilwestergren: `openRaw` needs the size though

davidhesselbom: Ah, true.

marcusnaslund:
Noone tested this, obviously :) the `RasterMonochrome` test fails completely.

sebastianbaginski:
I did yesterday, everything was ok.

sebastianbaginski:
>  ./test.sh draw
>10:58:13 RasterYuv422Semipacked . done in 0.32s, total: 0.32s
> 10:58:13 RasterYuv420Semiplanar ........ done in 2.01s, total: 2.33s
> libEGL warning: failed to create a pipe screen for i965
> 10:58:15 GpuImageRotationTest ...... done in 0.49s, total: 2.82s
> 10:58:15 GraphicBufferAlign . done in 0.00s, total: 2.82s
> 10:58:15 GpuImageTranslationTest ... done in 0.21s, total: 3.03s
> 10:58:16 GpuImageReflectionTest ... done in 0.25s, total: 3.28s
> 10:58:16 GpuSurfaceTest ........ done in 0.69s, total: 3.97s
> CreateAlphaByteStringTest [TODO: Not implemented as a fixture!]
> 10:58:17 ImageFile .................... done in 3.60s, total: 7.57s
> 10:58:20 ColorConvert .... done in 0.00s, total: 7.57s
> 10:58:20 RasterMonochrome ....... done in 0.87s, total: 8.44s
> 10:58:21 RasterBgraTest ...... done in 1.63s, total: 10.07s
> 10:58:23 RasterCanvas ..... done in 1.84s, total: 11.91s
> 10:58:25 RasterBgrTest ...... done in 1.28s, total: 13.19s

marcusnaslund:
Right, CircleCI obviously worked. Hm. Something wrong on my end, then.

sebastianbaginski:
I just rebased, try again now.

marcusnaslund:
Seems to work better now, my bad. :) Please rebase #824 as well.

marcusnaslund:
Any updates here? Is the problem I'm getting related to #815 ?

sebastianbaginski:
I doubt that as I would get it as well and it would crash on CirlceCI. Valgring shows nothing on my machine, not a single invalid memory access (except for the known issues).

sebastianbaginski:
Here is the new one https://github.com/cogneco/ooc-kean/pull/841 with commits cherry-picked

sebastianbaginski:
I will close this one
https://github.com/cogneco/ooc-kean/issues/842

marcusnaslund:
Excellent. Hope to see it soon again.

sebastianbaginski:
You mean this one https://github.com/cogneco/ooc-kean/pull/841 ?

marcusnaslund:
I mean also the stuff that isn't in #841 :)

PR #816, opened 2015-11-18
Removed not needed return statements

marcusnaslund:
For consistency.

### DIFF
--- a/source/draw/RasterImage.ooc
+++ b/source/draw/RasterImage.ooc
@@ -23,7 +23,7 @@ import Color
 import StbImage
 
 RasterImage: abstract class extends Image {
-	distanceRadius: Int { get { 1 } }
+	distanceRadius ::= 1
 	stride: UInt { get }
 	init: func ~fromRasterImage (original: This) { super(original) }
 	init: func (size: IntSize2D) { super(size) }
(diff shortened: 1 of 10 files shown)


marcusnaslund:
Anything to say about this @davidhesselbom ?

source/draw/RasterImage.ooc
```diff
@@ -23,7 +23,7 @@ import Color
 import StbImage
 
 RasterImage: abstract class extends Image {
-	distanceRadius: Int { get { return 1; } }
+	distanceRadius: Int { get { 1 } }
```
davidhesselbom: While you're at it, can't this be `distanceRadius ::= 1`?

marcusnaslund: `::=` is read-only, right?

davidhesselbom: So is `get` with no `set`.

marcusnaslund: Thanks. I will change that in some subclasses as well.

davidhesselbom:
Looks good to me, but I don't have the time to validate it.

marcusnaslund:
> Looks good to me, but I don't have the time to validate it.

Oh you don't have to, just wanted another pair of eyes. Thanks.

PR #817, opened 2015-11-18
Removed opengl=>ui dependency

sebastianbaginski:
We need to remove `widgets` module (https://github.com/cogneco/ooc-kean/issues/813) and move everything from there into `ui` (and later rename to `gui`).
Currently it can't be done because `widgets` module depends on `opengl` and `opengl` depends on `ui` (so moving stuff from widgets into ui results in circular dependency `opengl <=> ui`).
This pull request removes the `opengl => ui` dependency and is the first step to remove `widgets` module.

### DIFF
--- a/source/widgets/UnixWindow.ooc
+++ b/source/widgets/UnixWindow.ooc
@@ -57,7 +57,7 @@ UnixWindow: class extends UnixWindowBase {
 	context ::= this _openGLWindow context
 	init: func (size: IntSize2D, title: String) {
 		super(size, title)
-		this _openGLWindow = OpenGLWindow new(this _xWindow)
+		this _openGLWindow = OpenGLWindow new(this _xWindow size, this _xWindow display, this _xWindow backend)
 	}
 	free: override func {
 		this _openGLWindow free()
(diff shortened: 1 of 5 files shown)


sebastianbaginski:
Another solution is to move `NativeWindow` abstract class out of `ui` module, but I don't think it makes sense to have it elsewhere.

sebastianbaginski:
@davidhesselbom can you do a review ? You are already familiar with the `ui / widgets` issues.

marcusnaslund:
Conflicts @sebastianbaginski

sebastianbaginski:
rebased

source/widgets/UnixWindow.ooc
```diff
@@ -57,7 +57,7 @@ UnixWindow: class extends UnixWindowBase {
 	context ::= this _openGLWindow context
 	init: func (size: IntSize2D, title: String) {
 		super(size, title)
-		this _openGLWindow = OpenGLWindow new(this _xWindow)
+		this _openGLWindow = OpenGLWindow new(this _xWindow size, this _xWindow display, this _xWindow backend)
```
davidhesselbom: Are these parameters specific to X11 windows? Or is that irrelevant, since we don't care about Windows anymore?

sebastianbaginski: > we don't care about Windows anymore

This and the fact that we do not support opengl on windows.

davidhesselbom: Right...

davidhesselbom:
I don't see a problem here.

PR #825, opened 2015-11-18
Fix for double free bug in Profiler cleanup / free

sebastianbaginski:
Fix for point 3 in https://github.com/cogneco/ooc-kean/issues/716 
> Calling free() on a specific instance, and later using the static dispose() will cause a segfault due to double freeing.

### DIFF
--- a/source/base/Profiler.ooc
+++ b/source/base/Profiler.ooc
@@ -44,6 +44,11 @@ Profiler: class {
 		})
 	}
 	free: override func {
+		for (i in 0 .. This _profilers count)
+			if (this == This _profilers[i]) {
+				This _profilers removeAt(i)
+				break
+			}
 		this _timer free()
 		this _timer = null
 		this _clockTimer free()
@@ -65,6 +70,8 @@ Profiler: class {
 	}
 	resetAll: static func { This _profilers apply(func (profiler: This) { profiler reset() }) }
 	dispose: static func {
+		while (This _profilers count > 0)
+			This _profilers remove() free()
 		This _profilers free()
 		This _profilers = null
 	}


source/base/Profiler.ooc
```diff
@@ -44,6 +44,14 @@ Profiler: class {
 		})
 	}
 	free: override func {
+		index := -1
+		for (i in 0 .. This _profilers count)
+			if (this == This _profilers[i]) {
+				index = i
```
marcusnaslund: Couldn't you free right here and avoid the `index` variable alltogether?

sebastianbaginski: I guess you mean `remove`, yes I think so

marcusnaslund: Ah, sorry, yes.

sebastianbaginski: Done

source/base/Profiler.ooc
```diff
@@ -65,6 +70,8 @@ Profiler: class {
 	}
 	resetAll: static func { This _profilers apply(func (profiler: This) { profiler reset() }) }
 	dispose: static func {
+		while (This _profilers count)
```
emilwestergren: Is this OK? I remember when we discussed if it was OK to use NULL as a bool and decided it wasn't, i.e. `if (pointer == null)` instead of `if (pointer)`. I don't have a problem with it but someone might.

sebastianbaginski: I can change this, but code like `if (pointer)` is currently used almost everywhere.
Do you remember any argument against it ?  I'm just curious

emilwestergren: Clarity I guess, @davidhesselbom @simonmika do we have a standard?

sebastianbaginski: I just changed it anyway

marcusnaslund: I would like to know a standard for the future anyhow.

emilwestergren: Yes, me too.

source/base/Profiler.ooc
```diff
@@ -65,6 +70,8 @@ Profiler: class {
 	}
 	resetAll: static func { This _profilers apply(func (profiler: This) { profiler reset() }) }
 	dispose: static func {
```
simonmika: free?

sebastianbaginski: Can't have static and member methods with the same name.

simonmika: To bad, that is a common design flaw in many programming languages in my opinion. Maybe call it `freeAll` then. It corresponds to the `resetAll`. In the end I would just like to state that this matter has nothing to do with this PR and should not block it from getting merged. We should never let the optimal solution stop an incremental improvement.

sebastianbaginski: > https://github.com/cogneco/ooc-kean/issues/727

Yeah it comes right after the "optional" `This` and `this`.

There is an issue created for the method name https://github.com/cogneco/ooc-kean/issues/727 . Maybe you could leave a comment there.

simonmika: Agreed, it is one of the bad things coming from making `this` and `This` optional.

sebastianbaginski:
@simonmika @emilwestergren @davidhesselbom can one of you officialy review this ?

PR #826, opened 2015-11-18
Reused memory in transpose when FloatMatrix is a vector

### LINKED ISSUE #579: FloatMatrix: reuse allocated memory
In `FloatMatrix`, instead of creating a new instance to store the result of e.g. `transpose` and freeing the original data, reuse the original data.

Also, if transposing a 1-dimensional `FloatMatrix`, just swap `this size` and return.

tomasmahlberg:
Fixes #579 (in part).

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -128,13 +128,19 @@ FloatMatrix : cover {
 	}
 	transpose: func -> This {
 		t := this take()
-		result := This new(t dimensions swap())
-		resultElements := result elements
-		thisElements := this elements
-		for (y in 0 .. t height)
-			for (x in 0 .. t width)
-				resultElements[y + x * t height] = thisElements[x + y * t width]
-		this free(Owner Receiver)
+		result: This
+		if (t isVector && this _elements owner == Owner Receiver) {
+			result = this
+			result _dimensions = result _dimensions swap()
+		} else {
+			result = This new(t dimensions swap())
+			resultElements := result elements
+			thisElements := this elements
+			for (y in 0 .. t height)
+				for (x in 0 .. t width)
+					resultElements[y + x * t height] = thisElements[x + y * t width]
+			this free(Owner Receiver)
+		}
 		result
 	}
 	trace: func -> Float {
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -47,6 +47,11 @@ FloatMatrix : cover {
 		this free(Owner Receiver)
 		result
 	}}
+	isVector: Bool { get {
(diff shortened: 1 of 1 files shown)


tomasmahlberg:
@davidhesselbom Review?

source/math/FloatMatrix.ooc
```diff
@@ -123,13 +128,19 @@ FloatMatrix : cover {
...
+		result: This
+		if (t isVector && this _elements owner == Owner Receiver) {
+			result = this
+			result _dimensions = result _dimensions swap()
+		} else {
+			result = This new(t dimensions swap())
+			resultElements := result elements
+			thisElements := this elements
+			for (y in 0 .. t height)
+				for (x in 0 .. t width)
```
davidhesselbom: `width` and `height` are actually a little more expensive to access now that they're 3 lines long, so in the interest of optimization, consider storing the result of `t height` and `t width` in variables just before the loop.

source/math/FloatMatrix.ooc
```diff
@@ -123,13 +128,19 @@ FloatMatrix : cover {
...
-				resultElements[y + x * t height] = thisElements[x + y * t width]
-		this free(Owner Receiver)
+		result: This
+		if (t isVector && this _elements owner == Owner Receiver) {
+			result = this
+			result _dimensions = result _dimensions swap()
+		} else {
+			result = This new(t dimensions swap())
+			resultElements := result elements
+			thisElements := this elements
```
davidhesselbom: does it matter if you use `this elements` or `t elements` here? I'm still not comfortable with `take` to know for sure.

tomasmahlberg: It does not. The `elements` property does not call `free` like the other properties. This is because it returns a pointer to the object, so it would not make much sense to free it immediately after. We may want to change this in the future though, for consistency with the other properties. But as of now, this is how it works.

source/math/FloatMatrix.ooc
tomasmahlberg: Good point. Will fix this.

source/math/FloatMatrix.ooc
davidhesselbom: Okay, just making sure.

tomasmahlberg:
Updated.

source/math/FloatMatrix.ooc
davidhesselbom: According to my profiler, this change will affect performance measurably, but nowhere near as much as the same change in the `operator *` would. Just in case you run out of other things to fix...

marcusnaslund:
@davidhesselbom - are you satisfied with this enough to set the green flag?

PR #828, opened 2015-11-18
Removed ~two suffix from Int minimum and maximum.

### LINKED ISSUE #821: Replace maximum~two and minimum~two => maximum and minimum
After #818 there is only one version of `maximum` and `minimum` and thus we can remove the `~two` suffix and clean up a lot of calls to those two functions.

niklasborwell:
Fixes #821.
Removed all ~two suffixes from function calls and from the function signatures.


marcusnaslund:
The PR you've been waiting for is in now, so you can move this to final review if you're ready. @niklasborwell

niklasborwell:
Roger

PR #832, opened 2015-11-19
Moved some operators from Text to TextBuffer

tomasmahlberg:
@simonmika Is this what you had in mind? Do we generally want to move the logic and freeing from `Text` to `TextBuffer` when possible? Right now it seems to be a bit of a mix. Also, for example `count` in `Text` is basically a wrapper for `count` in `TextBuffer` but they both `free`. This seems to be redundant (and potentially also causing a double free, did not check).

Should point out also that I have not been following the discussion of `Text`/`TextBuffer` closely so I'm not sure exactly what we're trying to accomplish with these.

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -49,13 +49,7 @@ Text: cover {
 	}
 	copyTo: func (buffer: TextBuffer) -> Int { this _buffer copyTo(buffer) }
 	operator == (string: String) -> Bool { this == This new(string) }
-	operator == (other: This) -> Bool {
-		result := this _buffer == other _buffer
-		if (this _buffer _backend pointer != other _buffer _backend pointer)
-			other free(Owner Receiver)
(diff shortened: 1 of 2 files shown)


simonmika:
Yes go ahead.

simonmika:
Please observe that adding the peer review label without formally asking somebody to assign a reviewer is not right.

tomasmahlberg:
Let's just go ahead with this PR for a start then. Also, rebased.

@davidhesselbom Can you review?

marcusnaslund:
Have you looked at #836 ?

tomasmahlberg:
@marcusnaslund #836 seems kind of related to this PR but not entirely. I can push another commit to this PR to fix the test if you want?

marcusnaslund:
No let's get this through first, I just wanted to make sure you were aware since you're doing related stuff. Smaller PRs are always preferable.

PR #833, opened 2015-11-19
Static method to create AbstractContext subclass

sebastianbaginski:
To reduce code duplication in other project.
@davidhesselbom can you review ?

### DIFF
--- a/source/draw/AbstractContext.ooc
+++ b/source/draw/AbstractContext.ooc
@@ -36,4 +36,6 @@ AbstractContext: abstract class {
 	alignWidth: virtual func (width: Int, align := AlignWidth Nearest) -> Int { width }
 	update: abstract func
 	isAligned: virtual func (width: Int) -> Bool { true }
+	_createContext := static func -> This { CpuContext new() }
+	create: static func -> This { This _createContext() }
 }
--- a/source/draw/gpu/opengl/OpenGLContext.ooc
+++ b/source/draw/gpu/opengl/OpenGLContext.ooc
@@ -182,5 +182,6 @@ OpenGLContext: class extends GpuContext {
 			vertices[i] = toGL * vertices[i]
 		OpenGLMesh new(vertices, textureCoordinates, this)
 	}
+	This _createContext = This new
 }
 }


bettinaselig:
We only use "peer review" and "final review" for pull requests. The other labels are for issues from now on.

sebastianbaginski:
ok

source/draw/AbstractContext.ooc
```diff
@@ -36,4 +36,6 @@ AbstractContext: abstract class {
 	alignWidth: virtual func (width: Int, align := AlignWidth Nearest) -> Int { width }
 	update: abstract func
 	isAligned: virtual func (width: Int) -> Bool { true }
+	_createContext := static func -> This { CpuContext new() }
```
davidhesselbom: `_createContext := static func`? This syntax is unfamiliar...

sebastianbaginski: It is just like `defaultValue := static 42`

davidhesselbom: Okay, so it's like `_createContext ::= This CpuContext new()`?

davidhesselbom: Wait, no, that's a property...

davidhesselbom: But how is it different from `_createContext: static func -> This { CpuContext new() }`?

sebastianbaginski: I wanted `_createContext` to be a static data member, Closure which you can reassign later. I think in your example above one cannot reassign anything to `_createContext`.

davidhesselbom: Oh, right. Got it. We don't use reassignable func members a lot, so I wasn't even sure what this did. Thanks.

sebastianbaginski: Sure.
This design is actually Simon's idea, but unfortunately I could not have done it like
```ooc
_createContext := CpuContext new
```
as he originally suggested. I needed to wrap it in a closure here.

davidhesselbom: Looks cool though!

marcusnaslund:
@davidhesselbom - are you satisfied with this enough to set the green flag?

davidhesselbom:
Looks good enough to me, but I haven't tested it.

sebastianbaginski:
Unfortunately this does not work as it should. I need to remove the flag for now.

sebastianbaginski:
It won't work this way. I'm closing this.

PR #834, opened 2015-11-19
Removed unused clock from Profiler

sebastianbaginski:
Fixes 1) in https://github.com/cogneco/ooc-kean/issues/716
@bettinaselig can you review ?

### DIFF
--- a/source/base/Profiler.ooc
+++ b/source/base/Profiler.ooc
@@ -26,15 +26,12 @@ Profiler: class {
 	_profilers := static VectorList<This> new(100)
 	_name: String
 	_timer := Timer new()
-	_clockTimer := ClockTimer new()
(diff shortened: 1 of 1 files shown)


bettinaselig:
Reviewed.

source/base/Profiler.ooc
```diff
@@ -51,8 +48,6 @@ Profiler: class {
 			}
 		this _timer free()
 		this _timer = null
```
marcusnaslund: Is there any reason to do this?

sebastianbaginski: None that I'm aware of.

marcusnaslund: In that case, since it's just a one-line removal, and this is a cleanup PR, please delete it here.

sebastianbaginski: Done.

PR #835, opened 2015-11-19
Test case for Profiler

sebastianbaginski:
Fixes 5) in https://github.com/cogneco/ooc-kean/issues/716
@davidhesselbom  can you review ?


test/base/ProfilerTest.ooc
```diff
@@ -0,0 +1,33 @@
+use ooc-base
+use ooc-unit
+import io/File
+
+ProfilerTest: class extends Fixture {
+	init: func {
+		super("Profiler")
+		this add("log", func {
+			outputFile := "test/test.log"
```
davidhesselbom: How about `test/profilerTest.log`?

sebastianbaginski: OK, but this file is removed in this test. Should we keep it for manual inspection ?

davidhesselbom: Also, I noticed in another test that you created strings as member variables and freed them in a `free` function. Any reason not to do that here?

davidhesselbom: > OK, but this file is removed in this test. Should we keep it for manual inspection ?

Oh, I missed that. Still, no harm in using more specific file names, is there?

davidhesselbom: And yes, we can keep it, why not.

sebastianbaginski: OK, updated

marcusnaslund:
I can take this one, it's simple enough.

marcusnaslund:
If you can think of a nice way to test `reset` and `resetAll` we would have a complete test. If you're busy with other stuff, this is good enough for now.

sebastianbaginski:
If it's possible then I'd prefer to do that later

marcusnaslund:
Fine with me.

PR #837, opened 2015-11-19
Added FloatTuple.

emilwestergren:
As part of https://github.com/cogneco/ooc-kean/issues/783 I have implemented FloatTuple which holds 2, 3 and 4 Floats. This will be used for cases where types like `FloatPoint`/`FloatVector` are not valid.

### DIFF
--- a/ooc-base.use
+++ b/ooc-base.use
@@ -20,3 +20,4 @@ Imports: OwnedBuffer
 Imports: TextBuffer
 Imports: Text
 Imports: TextBuilder
+Imports: FloatTuple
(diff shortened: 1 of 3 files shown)


emilwestergren:
@simonmika peer review?

simonmika:
This is good.

marcusnaslund:
Do we really want this in `base` and not in `math`?

emilwestergren:
Yes, it's not related to math.

marcusnaslund:
Alright, then.

PR #839, opened 2015-11-19
Changed use ooc-math to imports in FloatMatrix

tomasmahlberg:
This is more correct and also causing some minor issues.

PR #840, opened 2015-11-19
Implemented cubic spline interpolation in FloatVectorList

tomasmahlberg:
Needs #839 to be merged first.

### DIFF
--- a/source/math/FloatVectorList.ooc
+++ b/source/math/FloatVectorList.ooc
@@ -19,6 +19,7 @@ use ooc-base
 import math
 import FloatComplex
 import FloatComplexVectorList
+import FloatMatrix
 
 FloatVectorList: class extends VectorList<Float> {
 	init: func ~default {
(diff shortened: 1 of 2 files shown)


marcusnaslund:
#839 has been merged.

tomasmahlberg:
@niklasborwell

test/math/FloatVectorListTest.ooc
```diff
@@ -337,6 +337,29 @@ FloatVectorListTest: class extends Fixture {
...
+			expect(interpolatedList count, is equal to(list count + numberOfPointsBetween * (list count - 1)))
+			expect(interpolatedList[0], is equal to(list[0]) within(1.0e-5f))
+			expect(interpolatedList[1], is equal to(0.67634f) within(1.0e-5f))
+			expect(interpolatedList[2], is equal to(list[1]) within(1.0e-5f))
+			expect(interpolatedList[3], is equal to(3.72098f) within(1.0e-5f))
+			expect(interpolatedList[4], is equal to(list[2]) within(1.0e-5f))
+			expect(interpolatedList[5], is equal to(5.31473f) within(1.0e-5f))
+			expect(interpolatedList[6], is equal to(list[3]) within(1.0e-5f))
+			expect(interpolatedList[7], is equal to(4.77009f) within(1.0e-5f))
+			expect(interpolatedList[8], is equal to(list[4]) within(1.0e-5f))
```
niklasborwell: Is `within(tolerance)` too strict here?

source/math/FloatVectorList.ooc
```diff
@@ -432,9 +433,51 @@ FloatVectorList: class extends VectorList<Float> {
 		}
 		(crossCorrelation, offsets)
 	}
+	// Assumes equi-distant samples
+	// https://en.wikipedia.org/wiki/Spline_interpolation
+	interpolate: func ~cubicSpline (numberOfPoints: Int) -> This {
```
niklasborwell: We usually don't use the ~ suffix in this way, where the caller have to explicitly write it out. Normally I would just call them interpolateLinear and interpolateSpline. Or maybe interpolate: func (numberOfPoints: Int, mode: InterpolationMode).

niklasborwell:
Finished first review, awaiting fixes.

tomasmahlberg:
Updated.

test/math/FloatVectorListTest.ooc
tomasmahlberg: Yes, I was too lazy to find enough decimals... :)

niklasborwell:
In that way I'm done here.

source/math/FloatVectorList.ooc
```diff
@@ -432,9 +433,51 @@ FloatVectorList: class extends VectorList<Float> {
...
+	// Assumes equi-distant samples
+	// https://en.wikipedia.org/wiki/Spline_interpolation
+	interpolateSpline: func (numberOfPoints: Int) -> This {
+		result := This new(this count + numberOfPoints * (this count - 1))
+		thisPointer := this pointer as Float*
+		coefficientMatrix := This _generateCubicSplineCoefficients(this count)
+		b := FloatMatrix new(1, this count) take()
+		for (index in 0 .. this count)
+			b[0, index] = 3.f * (thisPointer[Int minimum(this count - 1, index + 1)] - thisPointer[Int maximum(0, index - 1)])
+		// TODO: Since coefficientMatrix is tri-diagonal, there are better methods for this. Both in terms of data storage and solving.
```
marcusnaslund: Create an issue for this. Either PR debt or Later depending on how important it is.

test/math/FloatVectorListTest.ooc
```diff
@@ -319,7 +319,7 @@ FloatVectorListTest: class extends Fixture {
 			divided := list divideByMaxValue()
 			expect(divided sum, is equal to(-0.5f) within(tolerance))
 		})
-		this add("linear interpolation for list", func {
+		this add("interpolate~linear", func {
```
marcusnaslund: You don't need the `~` anymore

test/math/FloatVectorListTest.ooc
```diff
@@ -337,6 +337,29 @@ FloatVectorListTest: class extends Fixture {
 			list free()
 			interpolatedList free()
 		})
+		this add("interpolate~cubicSpline", func {
```
marcusnaslund: Same here, no `~`

source/math/FloatVectorList.ooc
```diff
@@ -432,9 +433,51 @@ FloatVectorList: class extends VectorList<Float> {
 		}
 		(crossCorrelation, offsets)
 	}
+	// Assumes equi-distant samples
+	// https://en.wikipedia.org/wiki/Spline_interpolation
+	interpolateSpline: func (numberOfPoints: Int) -> This {
+		result := This new(this count + numberOfPoints * (this count - 1))
+		thisPointer := this pointer as Float*
+		coefficientMatrix := This _generateCubicSplineCoefficients(this count)
+		b := FloatMatrix new(1, this count) take()
```
marcusnaslund: At least `rhs`. If you can think of a better name, do that. But not `b`.

source/math/FloatVectorList.ooc
```diff
@@ -432,9 +433,51 @@ FloatVectorList: class extends VectorList<Float> {
 		}
 		(crossCorrelation, offsets)
 	}
+	// Assumes equi-distant samples
+	// https://en.wikipedia.org/wiki/Spline_interpolation
+	interpolateSpline: func (numberOfPoints: Int) -> This {
```
marcusnaslund: Cubic splines are not the only splines. Maybe `interpolateCubic`?

test/math/FloatVectorListTest.ooc
marcusnaslund: Most other tests have `tolerance := 1.0e-5f`, so for me it's OK to change it here and use `tolerance` to make it easier to read.

source/math/FloatVectorList.ooc
tomasmahlberg: Issue #858.

source/math/FloatVectorList.ooc
tomasmahlberg: I discussed this with @niklasborwell before and we agreed that `b` is a sufficient name here and it is also what we use in some other places (which I can not talk about here). I can change it if you want though.

niklasborwell: If we're gonna change it to `rhs` i think `rightHandSide` would be better.

marcusnaslund: I don't like `b` generally, and especially not since you use another `b` just a few lines down. `rightHandSide` is also fine.

tomasmahlberg:
Updated.

PR #844, opened 2015-11-19
Updated RasterBgr test to use reference counters for images

sebastianbaginski:
part of https://github.com/cogneco/ooc-kean/issues/842
@marcusnaslund can you review and test this one ? maybe we should do it step by step to see where the error comes from

### DIFF
--- a/test/draw/RasterBgrTest.ooc
+++ b/test/draw/RasterBgrTest.ooc
@@ -14,7 +14,7 @@ RasterBgrTest: class extends Fixture {
 			image2 := RasterBgr open(this sourceSpace)
 			expect(image1 equals(image1))
 			expect(image1 equals(image2), is false)
-			image1 free(); image2 free()
+			image1 referenceCount decrease(); image2 referenceCount decrease()
 		})
 		this add("equals 2", func {
 			output := "test/draw/output/RasterBgr_test.png"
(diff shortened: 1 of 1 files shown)


marcusnaslund:
No problem. Maybe fix this class by class (one PR for each) and update each corresponding test to use referencecounter instead of free at the same time? You can move them to `final review` directly.

marcusnaslund:
This one runs fine, btw.

sebastianbaginski:
OK, good to know. Updated the tests to use reference counter.

marcusnaslund:
No, sorry, don't know if I missed something earlier. This segfaults for me in the same way in `RasterMonochromeTest` part 5.

marcusnaslund:
Interesting, this branch alone works, but if I merge it with upstream/master, the test fails.

sebastianbaginski:
@marcusnaslund can you test again now after https://github.com/cogneco/ooc-kean/pull/850 is merged ?

sebastianbaginski:
@marcusnaslund what are we going to do with this one ?

marcusnaslund:
I still have no idea why this fails for me I'm afraid. There's obviously something with the ByteBuffer or the use of it.

sebastianbaginski:
I have removed the commit with modified ByteBuffer ownership. This one only updates the test with reference counters now.

marcusnaslund:
Good. We can do the same for most other `Raster*` tests as well in the future, whenever you have the time.

sebastianbaginski:
@marcusnaslund can you review it and set the flag ?

PR #848, opened 2015-11-19
Consistency in Modulo operator

### LINKED ISSUE #508: Int modulo

### LINKED ISSUE #768: Inconsistent modulo operator
The `modulo` function in `Int`, `Int64`, and `Float` (and possibly others) are all inconsistent with each other. They should be corrected to comply with the following C# code from the original kean:
```csharp
public static int Modulo(int dividend, int divisor)
{
	if (dividend < 0)
		dividend += Integer.Ceiling(Integer.Absolute(dividend) / (float)divisor) * divisor;
	return dividend % divisor;
}
```
See the original discussion here: #508 

The ooc methods should *not* be static, though.

marcusnaslund:
Fixes #768

### DIFF
--- a/source/plot/Axis.ooc
+++ b/source/plot/Axis.ooc
@@ -116,9 +116,9 @@ Axis: class {
 	getFirstTickValue: func -> Float {
 		result: Float
 		if ((this min < this tick && this min > 0.0f) || (this min > this tick && this min < 1.0f))
-			result = this min + (this tick - (this min modulo(this tick)))
+			result = this min + this tick - this min modulo(this tick)
 		else
-			result = this min - (this min modulo(this tick))
+			result = this min - this min modulo(this tick)
 		result
 	}
 	getRequiredMargin: func (fontSize: Int) -> Float {
--- a/source/plot/Axis.ooc
+++ b/source/plot/Axis.ooc
@@ -116,9 +116,9 @@ Axis: class {
 	getFirstTickValue: func -> Float {
 		result: Float
 		if ((this min < this tick && this min > 0.0f) || (this min > this tick && this min < 1.0f))
-			result = this min + (this tick - Float modulo(this min, this tick))
+			result = this min + (this tick - (this min modulo(this tick)))
 		else
-			result = this min - Float modulo(this min, this tick)
+			result = this min - (this min modulo(this tick))
 		result
 	}
 	getRequiredMargin: func (fontSize: Int) -> Float {
--- a/source/plot/SvgWriter2D.ooc
+++ b/source/plot/SvgWriter2D.ooc
@@ -79,7 +79,7 @@ SvgWriter2D: class {
 					numPlotsX = 2
 				else
 					numPlotsX = 3
-				numPlotsY = Int modulo(this svgPlots count, numPlotsX) ? 1 + this svgPlots count / numPlotsX : this svgPlots count / numPlotsX
+				numPlotsY = this svgPlots count modulo(numPlotsX) ? 1 + this svgPlots count / numPlotsX : this svgPlots count / numPlotsX
 			} else {
 				numPlotsX = numberOfPlotsHorizontally
 				numPlotsY = ceil(this svgPlots count as Float / numPlotsX as Float) as Int
@@ -87,7 +87,7 @@ SvgWriter2D: class {
 			plotSize := FloatVector2D new(this size x / numPlotsX, this size y / numPlotsY)
 			position := FloatPoint2D new()
 			for (i in 0 .. this svgPlots count) {
-				position x = plotSize x * Int modulo(i, numPlotsX)
+				position x = plotSize x * (i modulo(numPlotsX))
 				position y = plotSize y * (i / numPlotsX)
 				result = result >> "<svg desc='Plot " & (i + 1) toString() >> "' x='" & position x toString() >> "' y='" & position y toString() >> "' width='" & plotSize x toString() >> "' height='" & plotSize y toString() >> "'>\n"
 				result = result & svgPlots[i] getSvg(plotSize, fontSize)
--- a/source/plot/SvgWriter2D.ooc
+++ b/source/plot/SvgWriter2D.ooc
@@ -79,7 +79,7 @@ SvgWriter2D: class {
 					numPlotsX = 2
 				else
 					numPlotsX = 3
-				numPlotsY = this svgPlots count modulo(numPlotsX) ? 1 + this svgPlots count / numPlotsX : this svgPlots count / numPlotsX
+				numPlotsY = this svgPlots count modulo(numPlotsX) != 0 ? 1 + this svgPlots count / numPlotsX : this svgPlots count / numPlotsX
 			} else {
 				numPlotsX = numberOfPlotsHorizontally
 				numPlotsY = ceil(this svgPlots count as Float / numPlotsX as Float) as Int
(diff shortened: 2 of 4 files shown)


marcusnaslund:
@simonmika or @davidhesselbom , will one of you peer review this?

simonmika:
Could we have tests on this?

simonmika:
I can peer review this.

marcusnaslund:
Test added.

test/sdk/MathTest.ooc
```diff
@@ -0,0 +1,30 @@
+use ooc-unit
+import math
+
+MathTest: class extends Fixture {
+	init: func {		
+		super("Math")
+		tolerance := 1.0e-5f		
+		this add("Modulo", func {
+			intValuePos := 22
```
simonmika: I would love to also have tests where the dividend is closer to zero then the divider.

simonmika:
I'm done with the peer review.

test/sdk/MathTest.ooc
marcusnaslund: Alright, I will add some more.

marcusnaslund:
Ran into some rebasing problems earlier. Everything fixed now, but please peer review again @simonmika

source/plot/Axis.ooc
```diff
@@ -116,9 +116,9 @@ Axis: class {
 	getFirstTickValue: func -> Float {
 		result: Float
 		if ((this min < this tick && this min > 0.0f) || (this min > this tick && this min < 1.0f))
-			result = this min + (this tick - Float modulo(this min, this tick))
+			result = this min + (this tick - (this min modulo(this tick)))
```
simonmika: Method calls have higher precedence then operators, should be:
```ooc
 result = this min + (this tick - this min modulo(this tick))
```

simonmika: By the way, somebody should fix the names `minimum` and `maximum`, but that is hardly the responsibility if this PR.

source/plot/Axis.ooc
```diff
@@ -116,9 +116,9 @@ Axis: class {
...
 		else
-			result = this min - Float modulo(this min, this tick)
+			result = this min - (this min modulo(this tick))
```
simonmika: Same as above.

source/plot/SvgWriter2D.ooc
```diff
@@ -79,7 +79,7 @@ SvgWriter2D: class {
 					numPlotsX = 2
 				else
 					numPlotsX = 3
-				numPlotsY = Int modulo(this svgPlots count, numPlotsX) ? 1 + this svgPlots count / numPlotsX : this svgPlots count / numPlotsX
+				numPlotsY = this svgPlots count modulo(numPlotsX) ? 1 + this svgPlots count / numPlotsX : this svgPlots count / numPlotsX
```
simonmika: You didn't introduce it but we should probably not do it the C way and assume that integers have a boolean value. It is probably enough to do the lazy version for not null.

simonmika: `numberOfPlotsY` or something. Again hardly something for this PR.

source/plot/SvgWriter2D.ooc
```diff
@@ -87,7 +87,7 @@ SvgWriter2D: class {
 			plotSize := FloatVector2D new(this size x / numPlotsX, this size y / numPlotsY)
 			position := FloatPoint2D new()
 			for (i in 0 .. this svgPlots count) {
-				position x = plotSize x * Int modulo(i, numPlotsX)
+				position x = plotSize x * (i modulo(numPlotsX))
```
simonmika: To many parentheses.

source/plot/SvgWriter2D.ooc
marcusnaslund: You mean `this svgPlots count modulo(numPlotsX) != 0`? I agree.

source/plot/SvgWriter2D.ooc
```diff
@@ -87,7 +87,7 @@ SvgWriter2D: class {
...
 				position y = plotSize y * (i / numPlotsX)
```
simonmika: Here there are to many parenthesis again although not exactly on you (again).

source/sdk/math.ooc
```diff
@@ -60,7 +60,7 @@ extend Short {
 }
 
 extend Int64 {
-	modulo: func (divisor: This) -> This { 
+	modulo: func (divisor: This) -> This {
 		result := this - (this / divisor) * divisor
```
simonmika: To many parenthesis again.

source/sdk/math.ooc
```diff
@@ -68,14 +68,10 @@ extend Int64 {
 
 extend Int {
 	modulo: func (divisor: This) -> This {
-		if (divisor < 0) {
-			this *= -1
-			divisor *= -1
-		}
-		result := this % divisor
+		result := this
```
simonmika: It is not strictly necessary to introduce `result` as you could as well modify `this` to contain your result as this is a call by value member of a cover but this way is probably much clearer.

source/sdk/math.ooc
```diff
@@ -306,7 +299,7 @@ extend Float {
 		value
 	}
 	minusPiToPiOverTwo: static func (value: This) -> This {
```
simonmika: Again, not on you, but if we really should have this method it should probably be called `minusPiToPiHalf` and not be static, which is true for the methods above as well.

test/sdk/MathTest.ooc
```diff
@@ -0,0 +1,31 @@
...
+	init: func {
+		super("Math")
+		tolerance := 1.0e-5f
+		this add("Modulo", func {
+			// Note: Negative values are in parentheses because of bug in rock
+			expect(22 modulo(5), is equal to(2))
+			expect((-7) modulo(3), is equal to(2))
+			expect((-7) modulo(1), is equal to(0))
+			expect(0 modulo(3), is equal to(0))
+			expect(3 modulo(4), is equal to(3))
```
simonmika: Thanks for the additional tests.

marcusnaslund:
Cleaned up some parentheses. The rest of the comments will be taken care of in a separate cleanup PR of the entire `math.ooc` file.

simonmika:
Thanks, peer review done.

PR #849, opened 2015-11-19
Adaptation to new rock version that uses full paths.

simonmika:
Makes error messages work when building from Code.


simonmika:
@thomasfanell Can you review this?

thomasfanell:
@simonmika confirmed as working

PR #850, opened 2015-11-19
Removed public forceFree for ByteBuffer

emilwestergren:
* In `_RecoverableByteBuffer` replaced `forceFree` by returning a `Bool` in the `recover` function which determines if the object shall be freed.
* In `_RecyclableByteBuffer` `forceFree` is now private and signaled when freeing and `this _size == 0`

Fixes https://github.com/cogneco/ooc-kean/issues/698

### DIFF
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -29,7 +29,6 @@ ByteBuffer: class {
 	_referenceCount: ReferenceCounter
 	referenceCount ::= this _referenceCount
 	_owner: Bool
-	_forceFree := false
 
 	init: func (=_pointer, =_size, owner := false) {
 		this _referenceCount = ReferenceCounter new(this)
@@ -43,10 +42,6 @@ ByteBuffer: class {
 			gc_free(this _pointer)
 		this _pointer = null
 		super()
-	}
-	forceFree: func {
-		this _forceFree = true
-		this free()
 	}
 	zero: func ~whole { memset(this _pointer, 0, _size) }
 	zero: func ~range (offset, length: Int) { memset(this _pointer + offset, 0, length) }
@@ -96,13 +91,17 @@ _RecoverableByteBuffer: class extends ByteBuffer {
 }
 _RecyclableByteBuffer: class extends ByteBuffer {
 	init: func (pointer: UInt8*, size: Int) { super(pointer, size, true) }
+	_forceFree: func {
+		this _size = 0
+		this free()
+	}
 	free: override func {
-		if (!this _forceFree) {
+		if (this size > 0) {
 			This _lock lock()
 			bin := This _getBin(this size)
 			while (bin count > 20) {
 				version(debugByteBuffer) { Debug print("ByteBuffer bin full; freeing one ByteBuffer") }
-				bin remove(0) forceFree()
+				bin remove(0) _forceFree()
 			}
 			this referenceCount _count = 0
 			bin add(this)
@@ -137,7 +136,7 @@ _RecyclableByteBuffer: class extends ByteBuffer {
 	}
 	_cleanList: static func (list: VectorList<This>) {
 		while (list count > 0)
-			list remove(0) forceFree()
+			list remove(0) _forceFree()
 	}
 	_clean: static func {
 		This _cleanList(This _smallRecycleBin)
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -28,17 +28,17 @@ ByteBuffer: class {
 	size ::= this _size
 	_referenceCount: ReferenceCounter
 	referenceCount ::= this _referenceCount
-	_owner: Bool
+	_ownsMemory: Bool
 
-	init: func (=_pointer, =_size, owner := false) {
+	init: func (=_pointer, =_size, ownsMemory := false) {
 		this _referenceCount = ReferenceCounter new(this)
-		this _owner = owner
+		this _ownsMemory = ownsMemory
 	}
 	free: override func {
 		if (this _referenceCount != null)
 			this _referenceCount free()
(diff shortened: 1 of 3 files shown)


emilwestergren:
@simonmika peer review

simonmika:
@sebastianbaginski Can you review this? You will have to live with that `free` not always calls `super`. But at least the flags are gone although there is a nasty trick with size.

source/base/ByteBuffer.ooc
```diff
@@ -28,26 +28,21 @@ ByteBuffer: class {
 	size ::= this _size
 	_referenceCount: ReferenceCounter
 	referenceCount ::= this _referenceCount
-	_owner: Bool
-	_forceFree := false
+	_ownsMemory: Bool
 
-	init: func (=_pointer, =_size, owner := false) {
+	init: func (=_pointer, =_size, ownsMemory := false) {
```
sebastianbaginski: @davidhesselbom @emilwestergren did you come to an agreement about the default value for `ownsMemory` ?

source/base/ByteBuffer.ooc
```diff
@@ -85,12 +80,10 @@ _SlicedByteBuffer: class extends ByteBuffer {
...
 }
 _RecoverableByteBuffer: class extends ByteBuffer {
-	_recover: Func (ByteBuffer)
+	_recover: Func (ByteBuffer) -> Bool
 	init: func (pointer: UInt8*, size: Int, =_recover) { super(pointer, size) }
 	free: override func {
-		if (!this _forceFree)
-			this _recover(this)
-		else {
+		if (this _recover(this)) {
```
sebastianbaginski: I understand what you are doing here, but the code like 
```ooc
if (this recover())
    delete(this)
```
is quite confusing (*if recover then delete*) :) Maybe invert the flag ?

source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -75,15 +75,14 @@ AndroidContext: class extends OpenGLContext {
 		fence := this createFence()
 		fence sync()
 		eglImage := gpuRgba as EGLBgra
-		sourcePointer := eglImage buffer lock(false)
+		sourcePointer := eglImage buffer lock(false) as UInt8*
```
sebastianbaginski: What happened, why you need to cast now ? Or should it be there from the start ?

sebastianbaginski:
>  You will have to live with that free not always calls super. 

Yeah, but at least this PR cleans up the public interface.
I can live with nasty tricks as long as they are hidden. This PR makes things better so I think it should go upstream.

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
```diff
@@ -35,8 +35,7 @@ GraphicBufferYuv420Semiplanar: class extends RasterYuv420Semiplanar {
 		ptr := _buffer lock()
 		_buffer unlock()
 		length := 3 * this _stride * size height / 2
-		byteBuffer := ByteBuffer new(ptr, length, func (buffer: ByteBuffer))
-		super(byteBuffer, size, _stride, _uvOffset)
+		super(ByteBuffer new(ptr, length), size, _stride, _uvOffset)
```
sebastianbaginski: Nice. This change alone can be the reason why it should be in the master branch.

sebastianbaginski:
@davidhesselbom @simonmika what is the policy about reviewing / testing ? should I now fetch Emil's repo and test it ?

sebastianbaginski:
@marcusnaslund I think you should give this one a priority, maybe it will fix the bugs you are getting for https://github.com/cogneco/ooc-kean/pull/844
if I recall correctly the crashes were related to `forceFree`

marcusnaslund:
@sebastianbaginski Agreed.
I will point out though that until you get answers to your questions (and are happy with them) this is still in the `peer review` state.

simonmika:
@sebastianbaginski When you do peer review you don't have to test run it.

source/base/ByteBuffer.ooc
emilwestergren: No, this was just a suggestion so feel free to suggest anything else.

source/base/ByteBuffer.ooc
emilwestergren: That probably makes more sense yes :)

source/base/ByteBuffer.ooc
sebastianbaginski: I mean if it should be `true` or `false` by default. I'm just asking, not in this PR of course.

source/base/ByteBuffer.ooc
simonmika: You are right, the logic should be inverted. If `recover` succeeded it should return `true` and that should stop the freeing. @emilwestergren can you fix this before it goes of to merging?

source/draw/gpu/opengl/AndroidContext.ooc
emilwestergren: Doesn't need to be there, I added it now for clarity. Shouldn't really be part of this PR so I can remove it if it's important.

sebastianbaginski: Do not remove it, I was just curious.

source/base/ByteBuffer.ooc
emilwestergren: Oh sorry, haven't had my coffee yet. Yes I think we agreed to default it to `false` since this is a public constructor.

source/base/ByteBuffer.ooc
emilwestergren: Already on it.

emilwestergren:
Updated.

sebastianbaginski:
Great. I'm done with review.

PR #851, opened 2015-11-20
Tools cleanup

davidhesselbom:
Cleaned up tools scripts, leaving fewer distracting files in the main folder.

### DIFF
--- a/install_rock.sh
+++ b/install_rock.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
 if [ "$(uname -o)" == "Msys" ]; then
-	tmp="$(wget -O - https://github.com/cogneco/rock/releases/latest | grep -o '\<download/rock_.*zip\>')"
-	wget https://github.com/cogneco/rock/releases/$tmp
-	echo ${tmp##*/}
-	7z e ${tmp##*/}
+	latest="$(wget -O - https://github.com/cogneco/rock/releases/latest | grep -o '\<download/rock_.*zip\>')"
+	wget https://github.com/cogneco/rock/releases/$latest
+	echo ${latest##*/}
+	7z e ${latest##*/}
 else
-	tmp="$(wget -O - https://github.com/cogneco/rock/releases/latest | grep -o '\<download/rock_.*deb\>')"
-	wget https://github.com/cogneco/rock/releases/$tmp
-	echo ${tmp##*/}
-	dpkg-deb -x ${tmp##*/} .
+	latest="$(wget -O - https://github.com/cogneco/rock/releases/latest | grep -o '\<download/rock_.*deb\>')"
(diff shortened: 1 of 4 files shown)


davidhesselbom:
Looks like I made some incorrect assumptions "cleaning up" `install_rock.sh`. Let me try again...

davidhesselbom:
@jonathanudd Peer review?

jonathanudd:
We are not running appveyor any more right? If we don't do we need to keep the windows part in for installing rock?

davidhesselbom:
I guess not. I can remove it.

jonathanudd:
Other than that I think it looks good.

install_rock.sh
```diff
@@ -1,14 +0,0 @@
-#!/bin/bash
-if [ "$(uname -o)" == "Msys" ]; then
-	tmp="$(wget -O - https://github.com/cogneco/rock/releases/latest | grep -o '\<download/rock_.*zip\>')"
-	wget https://github.com/cogneco/rock/releases/$tmp
-	echo ${tmp##*/}
-	7z e ${tmp##*/}
-else
```
marcusnaslund: I don't know about this. @simonmika - do we want to get rid of everything related to Windows?

davidhesselbom: As far as the script goes, it's all versioned, so we can revive if need be (although if the file is moved or renamed again, it'll be harder to find this part of it).

davidhesselbom: If we're dropping Windows support though, should we still keep support for running without a GPU?

simonmika: We should keep support for Windows as long as the impact on our speed forward is low. We might need it soon again.

Support for running without GPU is probably even more interesting, long term.

davidhesselbom: Okay, I'll back up my branch and remove the last commit from the PR.

marcusnaslund: Good good.

davidhesselbom: And so I did, a long time ago :)

marcusnaslund: > And so I did, a long time ago :)

Oh, right, so you did. I was expecting some news below this, but I realize that was incorrect of me.

marcusnaslund:
What else is supposed to be in `tools`, btw? Does this need a separate `ci` folder?

davidhesselbom:
Not sure, I just wanted some consistency across projects. I can move it elsewhere, if you want.

davidhesselbom:
Memtest could go in `tools`, I suppose.

marcusnaslund:
Right, it's the same in vidproc. OK, that's fine by me.

davidhesselbom:
What is this "vidproc" you speak of?

PR #852, opened 2015-11-20
Changed names of fields in size.

simonmika:
Renamed  `width`, `height`, `depth` in `FloatSize2D`, `IntSize2D`, `FloatSize3D`, `IntSize3D` to `x`, `y`, `z` in accordance with #783.


simonmika:
@sebastianbaginski can you peer review?

sebastianbaginski:
Quite a big one, is it possible to split it up somehow ? Like separate PRs for each Size type ? Could be processed by 4 people simultaneously.

sebastianbaginski:
OK nevermind I can take this one alone.

simonmika:
That would also be a lot of work. If we want to split up the review work it would be better to split it up on parts of the code base. Multiple PR means you would need to always know the kind of size you are reviewing.

source/draw/RasterBgr.ooc
```diff
@@ -110,7 +110,7 @@ RasterBgr: class extends RasterPacked {
 						result += (distance) sqrt() / 3
 					}
 				}
-			result /= ((this size width squared() + this size height squared()) as Float sqrt())
+			result /= this size length
```
niklasborwell: Don't you think this kind of changes should be done separately? This pull request is quite huge as it is, and this kind of changes probably adds a lot of time to the reviewing.

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
```diff
@@ -30,12 +30,12 @@ GraphicBufferYuv420Semiplanar: class extends RasterYuv420Semiplanar {
 	stride ::= this _stride
 	_uvOffset: Int
 	uvOffset ::= this _uvOffset
-	uvPadding ::= (this _uvOffset - this _stride * this _size height)
+	uvPadding ::= (this _uvOffset - this _stride * this _size y)
 	init: func ~fromBuffer (=_buffer, size: IntSize2D, =_stride, =_uvOffset) {
-		ptr := _buffer lock()
+		Pointer := _buffer lock()
```
niklasborwell: Same here. It seems unrelated to this pull request.

sebastianbaginski: Why a local variable starts with upper case ?

source/draw/RasterMonochrome.ooc
```diff
@@ -151,7 +151,7 @@ RasterMonochrome: class extends RasterPacked {
 						result += distance sqrt()
 					}
 				}
-			result /= ((this size width squared() + this size height squared()) as Float sqrt())
+			result /= ((this size x squared() + this size y squared()) as Float sqrt())
```
sebastianbaginski: Why not `this size length` as in other files ?

source/draw/RasterUv.ooc
```diff
@@ -100,7 +100,7 @@ RasterUv: class extends RasterPacked {
 						result += (distance) sqrt() / 3
 					}
 				}
-			result /= ((this size width squared() + this size height squared()) as Float sqrt())
+			result /= ((this size x squared() + this size y squared()) as Float sqrt())
```
sebastianbaginski: Same here , `length`

source/math/FloatSize3D.ooc
```diff
@@ -21,59 +21,59 @@ import IntSize3D
...
 FloatSize3D: cover {
-	width, height, depth: Float
-	volume ::= this width * this height * this depth
+	x, y, z: Float
+	volume ::= this x * this y * this z
 	length ::= this norm
-	empty ::= this width == 0 || this height == 0 || this depth == 0
-	norm ::= (this width squared() + this height squared() + this depth squared()) sqrt()
-	azimuth ::= this height atan2(this width)
+	empty ::= this x == 0 || this y == 0 || this z == 0
```
sebastianbaginski: I know that it was like that before, but how does it work in ooc, is the Int converted to Float or Float to Int for comparison ? Shouldn't we use a kind of `floatIsZero(this x)` anyway ? Not this PR just asking.

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
emilwestergren: Should at least be `pointer` and I agree, why is this included here?

source/math/FloatSize3D.ooc
emilwestergren: That's a good question, these things appear in many places so we need some clarity here.

sebastianbaginski: This test fails
```
almostZero := (0.1+0.1+0.1) - 0.3
empty = FloatSize2D new( almostZero, 0.1f)
expect(empty empty, is true)
```

sebastianbaginski:
conflicts @simonmika 
rebase and I will go through this again

source/draw/RasterMonochrome.ooc
simonmika: Missed this one.

simonmika:
This change breaks the API so please take it easy.

source/draw/RasterBgr.ooc
marcusnaslund: Yes, definitely, don't make it larger than it has to be.

sebastianbaginski:
`test.sh` works on this one.

sebastianbaginski:
This compiles and runs, tests passed. Review done.

source/draw/RasterBgr.ooc
simonmika: I agree that this one is more then big enough and that I should have avoided it this time. But in general I think we should allow for smaller improvements to the code quality while doing other changes. Else we end up with not getting the other once done.

source/math/FloatTransform3D.ooc
```diff
@@ -142,11 +142,11 @@ FloatTransform3D: cover {
...
 	createTranslation: static func (xDelta, yDelta, zDelta: Float) -> This { This new(1.0f, 0.0f, 0.0f, 0.0f, 1.0f, 0.0f, 0.0f, 0.0f, 1.0f, xDelta, yDelta, zDelta) }
 	createTranslation: static func ~float (delta: Float) -> This { This createTranslation(delta, delta, delta) }
-	createTranslation: static func ~size (delta: FloatSize3D) -> This { This createTranslation(delta width, delta height, delta depth) }
+	createTranslation: static func ~size (delta: FloatSize3D) -> This { This createTranslation(delta x, delta y, delta z) }
 	createTranslation: static func ~point (delta: FloatPoint3D) -> This { This createTranslation(delta x, delta y, delta z) }
 	createScaling: static func (xFactor, yFactor, zFactor: Float) -> This { This new(xFactor, 0.0f, 0.0f, 0.0f, yFactor, 0.0f, 0.0f, 0.0f, zFactor, 0.0f, 0.0f, 0.0f) }
 	createScaling: static func ~float (factor: Float) -> This { This createScaling(factor, factor, factor) }
-	createScaling: static func ~size (factor: FloatSize3D) -> This { This createScaling(factor width, factor height, factor width) }
+	createScaling: static func ~size (factor: FloatSize3D) -> This { This createScaling(factor x, factor x, factor z) }
```
marcusnaslund: YES! YES! I DID NOT READ ALL THESE FILES IN VAIN! :D Should be `factor y` as the second argument there.

PR #853, opened 2015-11-20
Added memcheck script for kean tests. No leak check for now, only inv…

davidhesselbom:
…alid memory access. There are so many leaks from strings and other things in the tests that Atom crashes when scrolling around in the log (which has nearly 100,000 lines in it). It's a good thing to be able to test for invalid reads, though (and there were only a few of those, last time I checked).

The script will run `Tests` built by running `test.sh`. That means if you want to run memcheck for e.g. only the `Text` tests, you first call `test.sh base/Text`, then `memcheck.sh`. It would perhaps be nice to integrate with `test.sh` instead, so that you call `test.sh memcheck base/Text` or something.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ rock_tmp
 /test/*/output
 tests.use
 Tests
+*.valgrindlog
--- a/tools/memcheck.sh
+++ b/tools/memcheck.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+valgrind --log-file=memcheck.valgrindlog --suppressions=./kean.supp --num-callers=500 --time-stamp=yes ./Tests -n $@


davidhesselbom:

marcusnaslund:
Good start, integrating the two so that `./tools/test.sh base` is a valid call would be a nice feature also, as you mentioned. If a lot of the memory leaks are caused by similar things we should start looking into if we can make a huge difference in the memory leaks in some easy/automated way, so that at least we can read the complete log whenever we want.

marcusnaslund:
After fixing some memory-related issues in `Fixture` and some tests I can now open the full log in Atom. Should we change or do we not care enough?

davidhesselbom:
Add `--leak-check=full` after the word `valgrind` in the script and see for yourself. Your call.

I noticed now that there's a `-n` near the end that shouldn't have been there...

marcusnaslund:
> Add --leak-check=full after the word valgrind

That's what I did, you know, otherwise I wouldn't have known ;)

davidhesselbom:
Okay, well, you might have added it somewhere else, or replaced something, and I wouldn't know...

PR #854, opened 2015-11-20
Added squared distance between two FloatPoint2D's

tomasmahlberg:
For use in e.g. optimized algorithms since it's cheaper than `distance`.

### DIFF
--- a/source/math/FloatPoint2D.ooc
+++ b/source/math/FloatPoint2D.ooc
@@ -32,6 +32,10 @@ FloatPoint2D: cover {
 		(this scalarProduct(other) / (this norm * other norm)) clamp(-1, 1) acos() * (this x * other y - this y * other x < 0 ? -1 : 1)
 	}
 	distance: func (other: This) -> Float { (this - other) norm }
+	distanceSquared: func (other: This) -> Float {
+		distanceVector := this - other
+		distanceVector x * distanceVector x + distanceVector y * distanceVector y
+	}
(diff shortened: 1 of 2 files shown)


tomasmahlberg:
@marcusnaslund Review please.

source/math/FloatPoint2D.ooc
```diff
@@ -32,6 +32,10 @@ FloatPoint2D: cover {
 		(this scalarProduct(other) / (this norm * other norm)) clamp(-1, 1) acos() * (this x * other y - this y * other x < 0 ? -1 : 1)
 	}
 	distance: func (other: This) -> Float { (this - other) norm }
+	distanceSquared: func (other: This) -> Float {
+		point := this - other
```
marcusnaslund: This name bothers me, but I'm not sure what to call it. The subtraction of a point with another is really a vector. Maybe `distanceVector`? What do you think?

tomasmahlberg: Yes, I though about this too, but in the end I chose `point` mostly because it is a `FloatPoint2D` in terms of data representation. But I will have to agree that `distanceVector` maybe is more representative. Will change.

marcusnaslund: Maybe the recent `Size`/`Point` discussions will sort this out.

tomasmahlberg:
Updated. Forgot to add a test though I remembered now... Will fix immediately.

tomasmahlberg:
Updated again.

PR #856, opened 2015-11-20
Allow CLI arguments without short identifier

jonathanudd:
I have added `add` functions without the short identifier parameter in `CLIArgumentParser`.


jonathanudd:
This is ready for peer review @marcusnaslund.

PR #863, opened 2015-11-23
Merged win32 and x11 modules into ui

sebastianbaginski:
Next step to remove widgets module. 
`X11` and `Win32` modules merged with `ui`
depends on https://github.com/cogneco/ooc-kean/pull/817

### DIFF
--- a/ooc-ui.use
+++ b/ooc-ui.use
@@ -2,3 +2,9 @@ Name: ooc UI
 Description: User interface stuff
 SourcePath: source/ui
 Imports: NativeWindow
+Imports: x11/X11Window
+Imports: win32/Win32Window
+version(windows)
+	Libs: -lgdi32
+version(!windows)
+	Libs: -lX11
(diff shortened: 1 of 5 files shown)


marcusnaslund:
#817 is merged.

sebastianbaginski:
@marcusnaslund : thanks
@jonathanudd : can you review

ooc-ui.use
```diff
@@ -1,4 +1,12 @@
 Name: ooc UI
 Description: User interface stuff
 SourcePath: source/ui
+SourcePath: source/ui/x11
+SourcePath: source/ui/win32
```
jonathanudd: I don't think you need three `SourcePath`. `SourcePath: source/ui` would be enough.

jonathanudd:
Apart from the three `SourcePath` lines I think it looks good.

ooc-ui.use
sebastianbaginski: You're right, it works. Updated.

ooc-ui.use
```diff
@@ -2,3 +2,9 @@ Name: ooc UI
 Description: User interface stuff
```
marcusnaslund: Terrible, terrible, but not this PR. :)

PR #864, opened 2015-11-23
Fixed bug in RasterYuv422Semipacked pixel access operator

sebastianbaginski:
Fixed write to invalid memory address in pixel `set` operator in `RasterYuv422Semipacked`.
Fixes https://github.com/cogneco/ooc-kean/issues/811
@marcusnaslund can you review

### DIFF
--- a/source/draw/RasterYuv422Semipacked.ooc
+++ b/source/draw/RasterYuv422Semipacked.ooc
@@ -82,7 +82,7 @@ RasterYuv422Semipacked: class extends RasterPacked {
 			index := (this buffer pointer + y * this stride + x * this bytesPerPixel) as ColorMonochrome* // U or V value
 			(index + 1)@ = ColorMonochrome new(value y)
 			(index - 2*(x % 2))@ = ColorMonochrome new(value u)
-			(index - 2*((x + 1) % 2))@ = ColorMonochrome new(value v)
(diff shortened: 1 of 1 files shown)


source/draw/RasterYuv422Semipacked.ooc
```diff
@@ -82,7 +82,7 @@ RasterYuv422Semipacked: class extends RasterPacked {
 			index := (this buffer pointer + y * this stride + x * this bytesPerPixel) as ColorMonochrome* // U or V value
 			(index + 1)@ = ColorMonochrome new(value y)
 			(index - 2*(x % 2))@ = ColorMonochrome new(value u)
-			(index - 2*((x + 1) % 2))@ = ColorMonochrome new(value v)
+			(index + 2*((x + 1) % 2))@ = ColorMonochrome new(value v)
```
sebastianbaginski: According to the `operator[]` above it should be `+`. If you print memory addresses when settings pixel at `(0,0)` you can see that this writes to memory 2 bytes before the buffer base address.

marcusnaslund:
If you explain how the data is actually supposed to be stored. :)

source/draw/RasterYuv422Semipacked.ooc
marcusnaslund: That makes sense, thanks for the explanation.

PR #866, opened 2015-11-23
Fixed double free in text test

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/865
@marcusnaslund can you review

### DIFF
--- a/test/base/TextBuilderTest.ooc
+++ b/test/base/TextBuilderTest.ooc
@@ -12,7 +12,7 @@ TextBuilderTest: class extends Fixture {
 			expect(text take() count, is equal to(7))
 			expect(text == t"vidflow")
 			tb = TextBuilder new(t"1,2,3" split(','))
-			text = tb join(';')
+			text = tb join(';') take()
 			expect(text == t"1;2;3")
 			tb free()
 			text free()


sebastianbaginski:
Actually maybe @tomasmahlberg could review this ?

marcusnaslund:
Maybe removing line 17 is the better solution. No use in `take()` if you just `free()` after.

sebastianbaginski:
If I remove the `free` as I wanted to do before, then TextTest generates invalid memory reads. I think we should follow @tomasmahlberg 's suggestion:
>  I think the general recommendation is to take() when you assign to a variable

This one works and does not generate any invalid memory access anymore.

tomasmahlberg:
Yes, I think we should do it like this, especially here in the test where people might look for examples on how to use things.

marcusnaslund:
Alright, but @tomasmahlberg please explain why I need to take and free. Feels like that negates the whole purpose of the `OwnedBuffer`.

tomasmahlberg:
@marcusnaslund I think the motivation for using this back-end was more to reduce the need for temporary variables. For example, there is actually no need to assign the result of the `join` here at all. One could just do it on one line without memory leaks: `expect(tb join(';') == t"1;2;3")`. I think it is situations similar to this that we want to simplify. When explicitly assigning something to a variable the general recommendation is to call `take` and `free` explicitly but in the end any way that does not leak memory should be fine I guess.

marcusnaslund:
Alright, thanks.

PR #867, opened 2015-11-23
FloatMatrix * FloatMatrix multiplication optimization

davidhesselbom:
This greatly reduces the number of calls to `OwnedBuffer free`.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -250,7 +250,7 @@ FloatMatrix : cover {
 					accumulator -= lowerElements[x2 + y * lower take() width] * resultElements[x + x2 * result take() width]
 				value := lowerElements[y + y * lower take() width]
 				if (value != 0)
-					resultElements[x + y * result take()width] = accumulator / value
+					resultElements[x + y * result take() width] = accumulator / value
 				else
 					raise("Division by zero in FloatMatrix forwardSubstitution")
 			}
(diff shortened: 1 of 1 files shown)


davidhesselbom:
@tomasmahlberg Peer review?

source/math/FloatMatrix.ooc
```diff
@@ -348,15 +348,16 @@ FloatMatrix : cover {
 				raise("Invalid dimensions in FloatMatrix * operator: left width must match right height!")
 		}
 		otherWidth := other take() width
+		(thisWidth, thisHeight) := (t width, t height)
 		result := This new(otherWidth, t height)
```
tomasmahlberg: `t height`=> `thisHeight` here also maybe.

davidhesselbom: Sure, I'll get that fixed, too.

tomasmahlberg:
Did you try doing this for `upper` and `lower` `width` in `_forwardSubstitution` and `_backwardSubstitution` also? Could maybe save some additional calls to `free`.

davidhesselbom:
I did try that, and it did remove even more `OwnedBuffer` calls. However, those two functions look very, very similar, and I think it would be better to somehow merge them before applying identical optimizations in both.

tomasmahlberg:
Alright, this looks good now.

source/math/FloatMatrix.ooc
```diff
@@ -250,7 +250,7 @@ FloatMatrix : cover {
 					accumulator -= lowerElements[x2 + y * lower take() width] * resultElements[x + x2 * result take() width]
 				value := lowerElements[y + y * lower take() width]
 				if (value != 0)
-					resultElements[x + y * result take()width] = accumulator / value
```
marcusnaslund: What? This works?

davidhesselbom: Sadly...

PR #868, opened 2015-11-23
Removed widgets module

sebastianbaginski:
`widgets` module merged into `ui`
@davidhesselbom can you review

### DIFF
--- a/source/ui/win32/Win32DisplayWindow.ooc
+++ b/source/ui/win32/Win32DisplayWindow.ooc
@@ -18,7 +18,7 @@
 import DisplayWindow
 use ooc-math
 use ooc-draw
-use ooc-ui
+import Win32Window
 
 version(windows) {
 Win32DisplayWindow: class extends DisplayWindow {
--- a/source/ui/x11/UnixWindow.ooc
+++ b/source/ui/x11/UnixWindow.ooc
@@ -15,12 +15,12 @@
  * along with this software. If not, see <http://www.gnu.org/licenses/>.
  */
 
-import DisplayWindow
 use ooc-draw-gpu
 use ooc-math
 use ooc-draw
 use ooc-opengl
-use ooc-ui
+import DisplayWindow
+import X11Window
 
 version(unix || apple) {
 UnixWindowBase: class extends DisplayWindow {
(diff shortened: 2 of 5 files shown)


davidhesselbom:
Does this remove the ability to run in windows, or run without openGL?

sebastianbaginski:
No, you can run without gpu as before. It just moves some files around.

davidhesselbom:
Alright, so how come we needed `ooc-widgets` in the first place? Did you come up with something here you didn't think of earlier, or...?

sebastianbaginski:
`widgets` module was a quick fix I needed for the IR-Tracker. I had a lot of other stuff to fix back then in a short amount of time.

davidhesselbom:
Alright.

source/ui/x11/UnixWindow.ooc
```diff
@@ -15,12 +15,12 @@
...
 use ooc-draw-gpu
 use ooc-math
 use ooc-draw
 use ooc-opengl
-use ooc-ui
+import DisplayWindow
+import X11Window
 
 version(unix || apple) {
 UnixWindowBase: class extends DisplayWindow {
```
davidhesselbom: I know it wasn't introduced here, but it would be nice to name this `UnixWindow` and rename `UnixWindow` `UnixOpenGLWindow` or something...

source/ui/win32/Win32DisplayWindow.ooc
```diff
@@ -18,7 +18,7 @@
 import DisplayWindow
 use ooc-math
 use ooc-draw
-use ooc-ui
+import Win32Window
 
 version(windows) {
 Win32DisplayWindow: class extends DisplayWindow {
```
davidhesselbom: And this one has `DisplayWindow` while the Unix equivalent has no `Display` in its name. But no, it shouldn't be fixed here and now.

source/ui/x11/UnixWindow.ooc
sebastianbaginski: I agree. There will be more changes in this module, I will fix this one too.

davidhesselbom:
Didn't test it, but it looks good to me.

PR #869, opened 2015-11-23
Added ResampleLinear() to FloatPoint2DVectorList.

niklasborwell:
Added a function to re-sample a data set with linear interpolation.


niklasborwell:
@bettinaselig Can you peer review? If you're not too busy

bettinaselig:
Sure.

bettinaselig:
What shall this function do?

niklasborwell:
You mean what it will be used for? Or what the function does?

bettinaselig:
Both.

niklasborwell:
The function is just a simple re-sampling with linear interpolation, no anti-aliasing, but we might extend it later. 
We need this when we try to estimate the sync between the gyro data and the video stream. The problem is that the gyro data has a higher sampling rate, and neither of the sample rates are constant, which means we can't calculate the cross-correlation directly. 
So I will use this to re-sample the gyro and video stream data so that we can compute the cross-correlation.

bettinaselig:
Looks good.

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -107,4 +108,30 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
...
 		result
 	}
+	resampleLinear: func (start, end, interval: Float) -> This {
+		// Assumes list is sorted by x values.
+		resultCount := ((end - start) / interval) ceil() as Int + 1
+		result := This new()
+		previousIndex := 0
+		for (i in 0 .. resultCount) {
+			point := FloatPoint2D new(start + i * interval, 0.0f)
+			if (point x <= this[0] x)
```
marcusnaslund: To avoid all the `memcpy`, maybe do the `thisPointer` trick here as well? If you care about performance.

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -107,4 +108,30 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
...
+				point y = this[this count - 1] y
+			else {
+				leftPoint, rightPoint: FloatPoint2D
+				for (j in previousIndex .. this count)
+					if (this[j] x <= point x) {
+						leftPoint = this[j]
+						rightPoint = this[Int minimum(j + 1, this count - 1)]
+						previousIndex = j
+					} else break
+				weight := (Float absolute(point x - leftPoint x) / Float absolute(rightPoint x - leftPoint x)) clamp(0.0f, 1.0f)
```
marcusnaslund: Do you really need to `clamp`?

marcusnaslund:
Done reviewing.

source/math/FloatPoint2DVectorList.ooc
bettinaselig: If `point x > rightPoint x`?

niklasborwell: No, you're right, it's not necessary here.

bettinaselig: If `point x > rightPoint x`?

source/math/FloatPoint2DVectorList.ooc
niklasborwell: Well, I thought about it, but I think the difference would be negligible compared to the ugliness of the code, since I'm indexing at so many different indices. Also performance is not critical, so I think we can tackle it if we need it.

source/math/FloatPoint2DVectorList.ooc
niklasborwell: No, it's covered by the previous if cases.

niklasborwell:
Updated.

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -107,4 +108,30 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
...
+				point y = this[this count - 1] y
+			else {
+				leftPoint, rightPoint: FloatPoint2D
+				for (j in previousIndex .. this count)
+					if (this[j] x <= point x) {
+						leftPoint = this[j]
+						rightPoint = this[Int minimum(j + 1, this count - 1)]
+						previousIndex = j
+					} else break
+				weight := Float absolute(point x - leftPoint x) / Float absolute(rightPoint x - leftPoint x)
```
bettinaselig: You never check if `point x` is smaller than `rightPoint x`, so you need to `clamp`. Don't ignore me! I bite you in the ass otherwise!

source/math/FloatPoint2DVectorList.ooc
bettinaselig: If you say so...

source/math/FloatPoint2DVectorList.ooc
bettinaselig: See above.

niklasborwell: Ok, I double checked this pretty carefully to prove that I'm right! 
It is never checked explicitly, but since we are assuming that the list is sorted by x values, this case won't happen. `rightPoint` is always next in the list after `leftPoint`, so `rightPoint x > leftPoint x`. And the only way `leftPoint x > point x` is if `point x <= this[0] x` which is checked earlier. Also, `rightPoint x == leftPoint x` is never true even if there are duplicate points, because it will skip directly to the second point in the j loop. I will add a test with duplicate points though, that would be useful.

niklasborwell:
Updated.

source/math/FloatPoint2DVectorList.ooc
bettinaselig: Ok. I am convinced :+1:

PR #871, opened 2015-11-23
Added methods to test floating point equality within tolerance

sebastianbaginski:
Updated `FloatVectorTest` with test that would previously fail due to floating point rounding errors.
Problem mentioned in https://github.com/cogneco/ooc-kean/issues/870 is not fixed here, it deserves its own PR.
@marcusnaslund can you review

### DIFF
--- a/source/math/FloatVector2D.ooc
+++ b/source/math/FloatVector2D.ooc
@@ -22,7 +22,7 @@ FloatVector2D: cover {
 	x, y: Float
 	area ::= this x * this y
 	length ::= this norm
-	empty ::= !(this x > 0 && this y > 0)
+	empty ::= this x equals(0.0f) || this y equals(0.0f) || this x < 0.0f || this y < 0.0f
 	norm ::= (this x squared() + this y squared()) sqrt()
 	azimuth ::= this y atan2(this x)
 	absolute ::= This new(Float absolute(this x), Float absolute(this y))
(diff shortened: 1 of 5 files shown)


source/math/FloatVector2D.ooc
```diff
@@ -22,7 +22,7 @@ FloatVector2D: cover {
 	x, y: Float
 	area ::= this x * this y
 	length ::= this norm
-	empty ::= !(this x > 0 && this y > 0)
+	empty ::= this y equals(0.0f) || this x equals(0.0f) || this x < 0.0f || this y < 0.0f
```
marcusnaslund: Uhm, why not `<=`?

marcusnaslund:
I think `tolerance` is a better name than `within`.

source/math/FloatVector2D.ooc
sebastianbaginski: `<=` does not use the new `equals` methods and the test would fail as before.

sebastianbaginski:
OK, renamed to `tolerance`.

source/math/FloatVector2D.ooc
marcusnaslund: Alright. At least change the position of `x` and `y` in the first two statements and I'm happy.

sebastianbaginski: Done.

PR #872, opened 2015-11-23
Fixed bug in TextBuilder prepend and append methods

sebastianbaginski:
You can crash `TextBuilderTest` when compiled and run with `./test.sh base/TextBuilder -Dsafe` (not only text builder...).
This PR fixes references to invalid vector index in TextBuilder.


sebastianbaginski:
@marcusnaslund can you review ?

PR #876, opened 2015-11-23
LinkedList now automatically wraps data in a Cell

### LINKED ISSUE #628: collections/LinkedList class is suspicious
When working on https://github.com/cogneco/ooc-kean/pull/627 I got crashes related to LinkedList. Exception backtraces were kept in a LinkedList, it crashed every time when I tried to iterate over the list and call `free` on the elements inside.
The loop looked like this
```ooc
list := LinkedList<Something>
size := list size
while (size > 0) {
    element := list head
    if (element)
        element free()
    list removeAt(0)
    --size
}
```
I didn't felt like digging into it, I just changed the container type to Stack and it worked.
Could be nice to write unit tests.
Btw. is there anything wrong with the code above ?

marcusnaslund:
@davidhesselbom or @sebastianbaginski - can one of you take this?

I managed to fix the invalid memory read in `LinkedListTest`, it was due to the data being removed (old line 105), *before* reading it and returning. That also made it possible to use `Cell` instead of the pointer solution.

The user has to take care of the data being put in the `LinkedList`, but that should be nothing new.

Fixes #628

### DIFF
--- a/source/collections/LinkedList.ooc
+++ b/source/collections/LinkedList.ooc
@@ -39,7 +39,7 @@ LinkedList: class <T> {
 			raise("Index out of bounds in LinkedList add~withIndex")
 	}
 	get: func (index: SSizeT) -> T {
-		this getNode(index) data@
+		this getNode(index) data
 	}
 	getNode: func (index: SSizeT) -> Node<T> {
 		if (index < 0 || index >= this _size)
@@ -89,21 +89,20 @@ LinkedList: class <T> {
 	first: func -> T {
 		data: T
 		if (this _head next != this _head)
-			data = this _head next data@
+			data = this _head next data
 		data
 	}
 	last: func -> T {
 		data: T
 		if (this _head prev != this _head)
-			data = this _head prev data@
+			data = this _head prev data
 		data
 	}
 	removeAt: func (index: SSizeT) -> T {
-		item := null
+		item: T = null
 		if (this _head next != this _head && 0 <= index && index < this _size) {
 			toRemove := this getNode(index)
-			this removeNode(toRemove)
-			item = toRemove data
+			item = this removeNode(toRemove)
 		} else
 			raise("Index out of bounds in LinkedList removeAt")
 		item
@@ -117,13 +116,15 @@ LinkedList: class <T> {
 		}
 		result
 	}
-	removeNode: func (toRemove: Node<T>) {
+	removeNode: func (toRemove: Node<T>) -> T {
 		toRemove prev next = toRemove next
 		toRemove next prev = toRemove prev
 		toRemove prev = null
 		toRemove next = null
+		data := toRemove data
 		toRemove free()
 		this _size -= 1
+		data
 	}
 	removeLast: func -> Bool {
 		result := false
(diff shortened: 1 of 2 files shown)


source/collections/LinkedList.ooc
```diff
@@ -117,13 +116,15 @@ LinkedList: class <T> {
...
-	removeNode: func (toRemove: Node<T>) {
+	removeNode: func (toRemove: Node<T>) -> T {
 		toRemove prev next = toRemove next
 		toRemove next prev = toRemove prev
 		toRemove prev = null
 		toRemove next = null
+		data := toRemove data
 		toRemove free()
 		this _size -= 1
+		data
```
sebastianbaginski: What happens here if you put an object into node and call `removeNode()` ? `Node` calls `free` on the `Cell` object, and `Cell free()` object calls `gc_free` on the data it contains. If you intend to return and keep using the data from `Node`, shouldn't this be more like
```ooc
removeNode: func (toRemove: Node<T>) -> T {
 		toRemove prev next = toRemove next
 		toRemove next prev = toRemove prev
 		toRemove prev = null
 		toRemove next = null
		data := toRemove data
		toRemove data = null  // <-- no `free` will be called on `data` now
 		toRemove free()
 		this _size -= 1
		data
}
```
?

source/collections/LinkedList.ooc
```diff
@@ -153,12 +154,20 @@ operator []= <T>(list: LinkedList<T>, index: Int, value: T) { list set(index, va
...
+	}
 	init: func
-	init: func ~withParams (=prev, =next, =data)
+	init: func ~withParams (=prev, =next, item: T) {
+		this _data = Cell new(item)
+	}
 	free: override func {
-		if (T inheritsFrom?(Object))
-			gc_free(data)
+		this _data free()
```
sebastianbaginski: With above comment, shouldn't this be
```ooc
if (this _data)
    this _data free()
```
?

source/collections/LinkedList.ooc
sebastianbaginski: Not really like that, but you get the point. It should just take the value away from the Cell object so it does not try to free it.

marcusnaslund: Ah, very true, thanks.

source/collections/LinkedList.ooc
```diff
@@ -153,12 +154,20 @@ operator []= <T>(list: LinkedList<T>, index: Int, value: T) { list set(index, va
 Node: class <T> {
 	prev: This<T>
 	next: This<T>
-	data: T* //TODO: Auto-wrap covers in Cell<T>
+	_data: Cell<T>
```
davidhesselbom: So a `Node` now always contains a `Cell`, even if `T` isn't a cover?

davidhesselbom: I really wouldn't recommend putting anything but covers inside Cells. Anywhere.

marcusnaslund: Any reason other than that it's unneccessary?

davidhesselbom: Well, it's nice to be able to assume that if you have a `Cell`, you will never find an object inside. But I don't understand why you are wrapping everything in a `Cell` in the first place. I don't think we're doing that anywhere else...? `HashDictionary` does it, but only if `T` is a cover, not otherwise. And I don't see why any one `LinkedList` should be able to contain more than one type of `T` at the same time.

marcusnaslund: No, that makes sense, I was just lazy. Will fix.

source/collections/LinkedList.ooc
marcusnaslund: I think `free`ing the node should not `free` its data.

marcusnaslund:
OK, I'm open to suggestions here. 

Now I only wrap covers in `Cell`, as it should be. One has to manually free that `Cell` and not just the data, though. Another solution would be to have the `data` property of `Node` return the whole `Cell` and not just its contents, to make it more obvious.

davidhesselbom:
Okay, so, why do we need a `Cell` in the first place` We do fine without them in the `VectorList` classes.

marcusnaslund:
I do remember what a hassle this was with `Promise` and `Future` - it did not work without them there. I'm not exactly sure why, though.

Issue #620, opened 2015-10-27
Some tests are for manual inspection - and should not always be run?


marcusnaslund:
We have some tests such as `ProfilingTest`, `CLIParserTest`, `DebugPrintTest` etc. which print out stuff like:
```
ProfilingTest [TODO: Not implemented as a fixture!]
```
This text was put there to remind us that these tests need to be updated. Some of these, for example `CLIParserTest`, can be automated and rewritten as a `Fixture` instance. (which has now been done for `ArgumentParser`)

Others, like `DebugPrintTest`, `SvgPlotTest` etc. are tests which rely on manual inspection and **cannot** be written as automated tests. For those tests, I propose we introduce a flag (or rather use the `debugTests` which already exists for tests that do printouts/files - meaning they require manual inspection), and do the above printout "not implemented as a fixture", or something else, inside the version block.

Input is welcome.

marcusnaslund:
@davidhesselbom @simonmika - input please.

davidhesselbom:
Some of the things that "require manual inspection" (like `SvgPlotTest`) can probably be made automatable by comparing the output to a pre-made, committed reference output.

simonmika:
Remove the TODO outputs, they just clog up the output. And while you are at it remove the timing printouts. They should be behind some version flag.

marcusnaslund:
I'm putting this in cleanup then.

PR #877, opened 2015-11-23
Removed "TODO: Not yet implemented as fixture" printouts

marcusnaslund:
In accordance with #620 
I will instead create issues for turning these tests into something using `Fixture`.
Anything to say @davidhesselbom ?

### DIFF
--- a/test/base/SynchronizedTest.ooc
+++ b/test/base/SynchronizedTest.ooc
@@ -1,39 +1,39 @@
 use ooc-base
 import threading/Thread, os/Time
 
-TestClass: class {
-	refCounter: ReferenceCounter
-	init: func {
-		refCounter = ReferenceCounter new(this)
+version(debugTests) {
+	TestClass: class {
+		refCounter: ReferenceCounter
+		init: func {
+			refCounter = ReferenceCounter new(this)
+		}
+		dispose: func {
+			refCounter decrease()
+		}
 	}
-	dispose: func {
-		refCounter decrease()
+	
+	obj := TestClass new()
+	
+	thread1 := Thread new(func1)
+	thread2 := Thread new(func2)
+	
+	thread1 start()
+	thread2 start()
+	
+	func1: func {
+		for (i in 0 .. 10) {
+			obj1 := obj
+			Time sleepMilli(100)
+		}
 	}
+	
+	func2: func {
+		for (i in 0 .. 10) {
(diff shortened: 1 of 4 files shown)


test/base/SynchronizedTest.ooc
```diff
@@ -1,39 +1,39 @@
...
-	refCounter: ReferenceCounter
-	init: func {
-		refCounter = ReferenceCounter new(this)
+version(debugTests) {
+	TestClass: class {
+		refCounter: ReferenceCounter
+		init: func {
+			refCounter = ReferenceCounter new(this)
+		}
+		dispose: func {
```
davidhesselbom: `free`?

davidhesselbom: Is this function even called from anywhere?

marcusnaslund: This whole test is weird. I don't know why the diff is so crazy in this file. Do `?w=1` and you'll see it's not this PR.

test/base/SynchronizedTest.ooc
```diff
@@ -1,39 +1,39 @@
...
-func1: func {
-	for (i in 0 .. 10) {
-		obj1 := obj
-		Time sleepMilli(100)
+	
+	func2: func {
+		for (i in 0 .. 10) {
+			obj2 := obj
+			Time sleepMilli(76)
+		}
```
davidhesselbom: And isn't it a little strange to define these functions after they were used? Consider moving these lines up to before the other things happen.

test/base/SynchronizedTest.ooc
davidhesselbom: I know, I just don't know if we should even keep the test.

marcusnaslund: I will create an issue.

marcusnaslund: #886

davidhesselbom: Very well.

davidhesselbom: So you should probably open issues for the other tests you just removed the printouts from, too.

marcusnaslund: Right. Coming up.

marcusnaslund: #887

PR #878, opened 2015-11-23
Split up Thread.ooc into multiple files

### LINKED ISSUE #703: Split Thread.ooc into separate files
Right now, `Thread.ooc` contains the abstract classes for `Thread`, `Mutex`, `RecursiveMutex`, `ThreadLocal`, and `WaitCondition`. These should be cleaned up and broken up into at least four (maybe five) separate files.

This will also help move these files out of the sdk - see #659

marcusnaslund:
This will break the threading API.

Fixes #703


marcusnaslund:
@sebastianbaginski - peer review? :)

source/sdk/threading/native/MutexUnix.ooc
```diff
@@ -1,4 +1,4 @@
-import ../Thread
+import ../Mutex
 
 version(unix || apple) {
 	include pthread | (_XOPEN_SOURCE=500)
```
sebastianbaginski: `_XOPEN_SOURCE=500` is this how you test for conditions in version blocks in ooc ? Certainly not valid syntax for C preprocessor.

sebastianbaginski:
Looks good. Does it compile on windows ?

marcusnaslund:
> Does it compile on windows ?

Yes. Just discovered `FloatVectorListTest` fails on Windows though (expected 10.00000 but was 10.0000 or something stupid like that), but I don't think we care that much about Windows right now.

PR #884, opened 2015-11-24
Added first derivative of pixel in RasterMonochrome.

tomasmahlberg:


tomasmahlberg:
@sebastianbaginski Can you look?

sebastianbaginski:
It looks good. The only thing bothering me is the fact that the expressions
```ooc
(8 * source[x + step] - 8 * source[x - step] + source[x - 2 * step] - source[x + 2 * step]) as Float / (12.0f * step)
```
and
```ooc
(8 * source[x + sourceWidth * step] - 8 * source[x - sourceWidth * step] + source[x - sourceWidth * 2 * step] - source[x + sourceWidth * 2 * step]) as Float / (12.0f * step)
```
are now repeated twice in this file.
Since ooc does not support macros, I'm not sure if it's worth to introduce performance penalty with a helper function. I would just leave it as it is now.
Which means I'm done reviewing.

marcusnaslund:
What is going on here? :) Derivative of what?

sebastianbaginski:
Of pixel :> A handy shortcut for "*convolution of pixel neighbourhood with derivative kernel*" :P

PR #894, opened 2015-11-25
Fixed typo in Quaternion cover

tomasmahlberg:
Couldn't find any explicit references to this (in any project), so this should be enough.


tomasmahlberg:
@marcusnaslund

marcusnaslund:
Thank you.

PR #895, opened 2015-11-25
Reduced memory leaks in Fixture

marcusnaslund:
An example of valgrind output for `test/math` before:
```
==00:00:00:20.094 29312==    definitely lost: 16,078,012 bytes in 2,339 blocks
==00:00:00:20.094 29312==    indirectly lost: 19,915,488 bytes in 611,227 blocks
==00:00:00:20.094 29312==      possibly lost: 108 bytes in 3 blocks
```

and after:
```
==00:00:00:20.443 28643==    definitely lost: 20,888,344 bytes in 203,616 blocks
==00:00:00:20.443 28643==    indirectly lost: 1,402,236 bytes in 207,626 blocks
==00:00:00:20.443 28643==      possibly lost: 0 bytes in 0 blocks
```

And this should be even better once we add `. free()` to all tests. I will get to work on that after this.

@sebastianbaginski - can you peer review?

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -24,7 +24,7 @@ Fixture: abstract class {
 	exitHandlerRegistered: static Bool
 	failureNames: static VectorList<String>
 	name: String
-	tests := VectorList<Test> new()
+	tests := VectorList<Test> new(32, false)
 	printFailures: static func {
 		if (This failureNames && (This failureNames count > 0)) {
 			"Failed tests: %i [" printf(This failureNames count)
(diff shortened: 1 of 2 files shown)


sebastianbaginski:
:+1:

sebastianbaginski:
Looks good. We should also fix the leaks from `DateTime toString` (or rewrite with `toText` ?), they are bloating the valgrind log in kean.

marcusnaslund:
> We should also fix the leaks from `DateTime toString`

That would be great. Do you have any ideas? We'll be moving to `Text` *eventually*, so it's not worth spending too much time on.

marcusnaslund:
Also, thanks for the memcheck script @davidhesselbom

PR #896, opened 2015-11-25
Made OpenGLTransform default OpenGLMap.

emilwestergren:
A transform vertex shader is now always used to make sure all coordinate systems are correct when drawing. OpenGLMapTransform will eventually be merged into OpenGLMap, but this change doesn't break the API so it will do as a start.

@davidpiuva peer

### DIFF
--- a/source/draw/gpu/opengl/OpenGLMap.ooc
+++ b/source/draw/gpu/opengl/OpenGLMap.ooc
@@ -49,7 +49,6 @@ OpenGLMap: class extends GpuMap {
 		if (vertexSource == null || fragmentSource == null)
 			Debug raise("Vertex or fragment shader source not set")
 	}
-	init: func ~defaultVertex (fragmentSource: String, context: OpenGLContext) { this init(slurp("shaders/default.vert"), fragmentSource, context) }
 	free: override func {
 		for (i in 0 .. this _context getMaxContexts()) {
 			if (this _program[i] != null)
@@ -108,6 +107,7 @@ OpenGLMap: class extends GpuMap {
 		this apply(action)
 		(action as Closure) free()
 	}
+	new: static func ~default (fragmentSource: String, context: OpenGLContext) -> This { OpenGLMapTransform new(fragmentSource, context) }
 }
 OpenGLMapMesh: class extends OpenGLMap {
 	init: func (context: OpenGLContext) { super(This vertexSource, This fragmentSource, context) }


davidpiuva:
ok

davidpiuva:
It breaks InMotion for some reason.

emilwestergren:
Haha, there is probably some code that depends on the untransformed vertex shader. I guess it needs to be carefully adapted to this change.

davidpiuva:
Does this new version scale the image by default to preserve pixels instead of keeping the same size?

emilwestergren:
No, with OpenGLTransform the size of the image passed as parameter will be the size of the quad that is drawn and in local coordinates (i.e. a smaller image will appear in the top left corner).

davidpiuva:
When would I ever need that?

emilwestergren:
Need what?

davidpiuva:
Applying a translation and scaling to preserve pixel size. I always keep things where they are when I change resolution because I want the whole image unless I choose a transform.

emilwestergren:
That's how Simon wanted it.

davidpiuva:
But it is useless for me.

emilwestergren:
So you're saying you always want the texture to automatically scale/shrink over the entire surface?

davidpiuva:
Yes, I only preserve pixel size when I do table programming or GUI drawing. Otherwise it is just a complicated way of generating the unit transform for OpenGL.

davidpiuva:
@simonmika Can we have both pixel preserving and scaling default transforms?

simonmika:
@davidpiuva Please suggest an API for that. Maybe it should be a property of the canvas?

davidpiuva:
canvas scaleMode = ScaleMode Reference
Good for drawing GUI and debugging.

canvas scaleMode = ScaleMode Local
Good for exact GPGPU math operations where a pixel, row or column represents an item using unsigned integers.

canvas scaleMode = ScaleMode Normalized
Good when you want the cropping to be preserved but change DPI.

emilwestergren:
Why `ScaleMode` ? Should it only apply to scaling?

davidpiuva:
Just hard to name it. What would you call it?

emilwestergren:
`CoordinateSystem`?

davidpiuva:
System?

emilwestergren:
And if it's changed, it should apply to all the functions in the draw API right?

emilwestergren:
`CoordinateType`?

davidpiuva:
Applying to all makes sense.
The name should be compact enough to not bloated the impression of the code but still be specific.

emilwestergren:
I guess it would be nice, gonna take some work to get everything working though.

marcusnaslund:
What's the status of this, seven days later?

emilwestergren:
It's idle until we decide how to fix the API.

marcusnaslund:
You can reopen this when it's active again.

simonmika:
@sebastianbaginski: Can you come up with a decent name for this?

sebastianbaginski:
I don't know, what exactly needs renaming ?

simonmika:
We want to have a property of a `Mapping` that controls how the input data and the output data is scaled relative to each other.

sebastianbaginski:
`ScaleMode` sounds good but is bound to scaling only.
Some others `PixelMappingMode`, `PixelScaleMode` `PixelMapMode` `PixelScalingMode` `MappingMode` `PixelTransformMode`
Quite hard to come up with something obvoius.

davidpiuva:
`Unit`?

PR #899, opened 2015-11-25
Fix accidental crash of Fixture when tests failed

marcusnaslund:
The memory optimizations I did for Fixture earlier in #895 accidentally caused some crashes when tests failed (oops). This PR corrects for that and also removes the leaking iterator on line 84.

@sebastianbaginski - thoughts?

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -67,7 +67,7 @@ Fixture: abstract class {
 				e message = test name
 				result = r = false
 				failures add(e)
-			}			
+			}
 			This _print(r ? "." : "f")
 		}
 		if (!result) {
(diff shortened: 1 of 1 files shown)


source/unit/Fixture.ooc
```diff
@@ -52,7 +52,7 @@ Fixture: abstract class {
 		this add(test)
 	}
 	run: func -> Bool {
-		failures := VectorList<TestFailedException> new()
+		failures := VectorList<TestFailedException> new(32, false)
```
sebastianbaginski: `32` just because you really want to set the second argument to `false`, right ?

marcusnaslund: Exactly.

PR #901, opened 2015-11-25
Use stride instead of width in RasterMonochrome getDerivative

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/900
`getFirstDerivative` now uses image stride to calculate offsets in pixel buffer
`stride` is signed now to avoid potential pitfalls when using it in subtraction
@tomasmahlberg can you take a look, as you worked on this code recently ?

### DIFF
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -173,17 +173,17 @@ RasterMonochrome: class extends RasterPacked {
 	}
 	getFirstDerivative: func (x, y: Int) -> (Float, Float) {
 		step := 2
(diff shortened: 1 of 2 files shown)


tomasmahlberg:
Looks good. I guess it could be argued that the variable should be called `sourceStride` now instead, but I don't think it matters either way.

sebastianbaginski:
Oh yeah, you're right :) I will change that.

tomasmahlberg:
OK :)

sebastianbaginski:
@tomasmahlberg  ok, you can check again

PR #902, opened 2015-11-25
Moved Process from SDK to Concurrent namespace

marcusnaslund:
What the title says.

### DIFF
--- a/ooc-concurrent.use
+++ b/ooc-concurrent.use
@@ -4,3 +4,4 @@ SourcePath: source/concurrent
 Imports: Promise
 Imports: PromiseCollector
 Imports: ThreadPool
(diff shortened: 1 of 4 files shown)


sebastianbaginski:
Are you sure you want to keep `Process` in the same unit with `ThreadPool` and `Promise` ? To me seems more natural to have `Process` along with `File`, `Pipe` etc. somewhere in `io` module. In fact it is possible to use the same base abstraction for `File` and `Process` - both can be read from and written to.

marcusnaslund:
This was actually @simonmika 's idea. I will talk to him again.

marcusnaslund:
We're going with io.

PR #903, opened 2015-11-25
General cleanup in math.ooc

marcusnaslund:
Small fixes. Moved some stuff around so that properties are at the top.


marcusnaslund:
Thoughts on this? @davidhesselbom

source/sdk/math.ooc
```diff
@@ -74,22 +70,8 @@ extend Int {
 		result % divisor
 	}
 	clamp: func (floor: This, ceiling: This) -> This {
-		if (this > ceiling)
-			ceiling
-		else if (this < floor)
-			floor
-		else
-			this
+		this > ceiling ? ceiling : (this < floor ? floor : this)
```
davidhesselbom: I'm sure this wasn't possible at some point. Either way, I don't think it reads very clearly...

source/sdk/math.ooc
```diff
@@ -152,6 +135,9 @@ operator == (left, right: Range) -> Bool { left min == right min && left max ==
 operator != (left, right: Range) -> Bool { !(left == right) }
 
 extend Double {
+	pi ::= static 3.14159_26535_89793_23846_26433_83279
+	e ::= static 2.718281828459045235360287471352662497757247093699959574966
```
davidhesselbom: Funny how these don't have nearly the same number of digits...

source/sdk/math.ooc
```diff
@@ -358,6 +318,9 @@ extend Float {
 }
 
 extend LDouble {
+	pi ::= static 3.14159_26535_89793_23846_26433_83279
```
davidhesselbom: Is the value here a double or a long double? How can you tell?

source/sdk/math.ooc
```diff
@@ -74,22 +70,8 @@ extend Int {
 		result % divisor
 	}
 	clamp: func (floor: This, ceiling: This) -> This {
```
davidhesselbom: floor, ceiling: This

source/sdk/math.ooc
marcusnaslund: I don't think it's too bad, and having two `?:`-operators is nothing new. Do a regex search for `.*\?.*: \(.*\?.*:.*\)`.

davidhesselbom: No, it's not "too bad", it just takes a bit of getting used to. For a function that does so little work, I guess it's fine.

davidhesselbom: It would be slower, I guess, but `this maximum(floor) minimum(ceiling)` should also work (if `maximum` and `minimum` are methods and not static functions).

marcusnaslund:
Updated.

marcusnaslund:
Any more comments @davidhesselbom ?

source/sdk/math.ooc
```diff
@@ -358,6 +318,9 @@ extend Float {
 }
 
 extend LDouble {
+	pi ::= static 3.14159_26535_89793_23846_26433_83279 as This
+	e ::= static 2.718281828459045235360287471352662497757247093699959574966 as This
```
davidhesselbom: Okay, but does this actually have LDouble precision, or is it truncated to a Double before being cast to an LDouble? (Not that it matters, but I'm curious.)

davidhesselbom:
Nothing important, no.

source/sdk/math.ooc
marcusnaslund: I find nothing in the documentation about how to specify an `LDouble`, so I don't know.

sebastianbaginski: Try with capital `L`: value := 3.14L`

davidhesselbom: Are you just guessing now, or do you have reason to believe it'll work? You wouldn't happen to know how to declare a UInt8, would you?

sebastianbaginski: Adding L suffix to float literal works in C, to specify a long double.
I'm not aware of any suffix for unsigned short in standard C. There is a suffix for microsoft compilers though, but I don't remember what was it, maybe `10u` or something like that.

marcusnaslund: Seems to work with an `L`, thanks.

PR #905, opened 2015-11-26
Made FloatMatrix independent of Geometry.

simonmika:
In preparation of creating a geometry namespace FloatMatrix is made independent on all future geometry classes.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -16,58 +16,56 @@
 
 use ooc-base
 import math
-import IntVector2D
-import FloatPoint3D
 
-FloatMatrix : cover {
+FloatMatrix: cover {
 	// x = column
 	// y = row
-	_dimensions: IntVector2D
-	dimensions: IntVector2D { get {
-		result := this _dimensions
-		this free(Owner Receiver)
-		result
-	}}
+	_width: Int
 	width: Int { get {
-		result := this _dimensions x
+		result := this _width
 		this free(Owner Receiver)
 		result
 	}}
+	_height: Int
 	height: Int { get {
-		result := this _dimensions y
+		result := this _height
 		this free(Owner Receiver)
 		result
 	}}
 	isNull: Bool { get { // TODO: Better name?
-		result := this _dimensions empty
+		result := this _width == 0 || this _height == 0
 		this free(Owner Receiver)
 		result
 	}}
 	isSquare: Bool { get {
-		result := this _dimensions square
+		result := this _width == this _height
 		this free(Owner Receiver)
 		result
 	}}
 	isVector: Bool { get {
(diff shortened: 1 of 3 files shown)


simonmika:
@tomasmahlberg Please review.

tomasmahlberg:
Looks right to me.

marcusnaslund:
Why?

simonmika:
Because `FloatMatrix` is going into `ooc-math` but `FloatPoint2D` etc is going into `ooc-geometry` which will depend on `ooc-math`.

marcusnaslund:
What is going into `geometry`? I thought it was just the `Point` and `Vector` covers. They will not depend on `ooc-math`.

source/math/FloatMatrix.ooc
```diff
@@ -16,58 +16,56 @@
...
+	_height: Int
 	height: Int { get {
-		result := this _dimensions y
+		result := this _height
 		this free(Owner Receiver)
 		result
 	}}
 	isNull: Bool { get { // TODO: Better name?
-		result := this _dimensions empty
+		result := this _width == 0 || this _height == 0
```
sebastianbaginski: Replacing this line could potentially give different results:
```ooc
IntVector2D {
	empty ::= !(this x > 0 && this y > 0)
}
```
https://github.com/cogneco/ooc-kean/issues/870

marcusnaslund: Furthermore, we might not want to use `==`?

sebastianbaginski: I think they are integers.

marcusnaslund: Oh, right, nvm.

marcusnaslund: But a matrix with negative width or height is just crazy.

sebastianbaginski: I think it would crash in constructor when trying to allocate enormous amount of memory, as `malloc` accepts unsigned values :)

marcusnaslund: Haha yes. Then we're all good here.

simonmika:
The following classes will go into `geometry`, and some of them will depend on `math`:
```ooc
FloatVector2D
FloatVector3D
FloatPoint2D
FloatPoint3D
FloatBox2D
FloatShell2D
FloatTransform2D
FloatTransform3D
IntVector2D
IntVector3D
IntPoint2D
IntPoint3D
IntBox2D
IntShell2D
IntTransform2D
FloatEuclidTransform
Quaternion
FloatRotation3D
FloatPoint2DVectorList
FloatEuclidTransformVectorList
FloatConvexHull2D
```

marcusnaslund:
Validating this now, so should be in soon.

marcusnaslund:
This gets rid of a 64 byte memory leak... somehow.

PR #906, opened 2015-11-26
Simplified and generalized bilinear interpolation in RasterMonochrome

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/732
Bilinear interpolation method is moved from `RasterMonochrome` and can work with images with more than 1 bpp.
Would be easier if we could call methods on generic types. I could simply call `blend` depending on which type of `Color` is `T`. But now it seems like the only thing you can do with a generic `T` is to cast it to pointer... Not the best language for metaprogramming :/

### DIFF
--- a/source/draw/RasterCanvas.ooc
+++ b/source/draw/RasterCanvas.ooc
@@ -88,4 +88,36 @@ RasterCanvas: abstract class extends Canvas {
 			}
 		}
 	}
+	resizeBilinear: static func <T> (sourceBuffer, resultBuffer: T*, sourceBox, resultBox: IntBox2D, sourceStride, resultStride, bytesPerPixel: Int) {
+		(resultWidth, resultHeight) := (resultBox size x, resultBox size y)
+		(sourceWidth, sourceHeight) := (sourceBox size x, sourceBox size y)
+		(sourceStartColumn, sourceStartRow) := (sourceBox leftTop x, sourceBox leftTop y)
+		(resultStartColumn, resultStartRow) := (resultBox leftTop x, resultBox leftTop y)
+		for (row in 0 .. resultHeight) {
+			sourceRow := ((sourceHeight as Float) * row) / resultHeight + sourceStartRow
+			sourceRowUp := sourceRow floor() as Int
+			weightDown := sourceRow - sourceRowUp as Float
+			sourceRowDown := (sourceRow - weightDown) as Int + 1
+			if (sourceRowDown >= sourceHeight)
+				weightDown = 0.0f
+			for (column in 0 .. resultWidth) {
+				sourceColumn := ((sourceWidth as Float) * column) / resultWidth + sourceStartColumn
+				sourceColumnLeft := sourceColumn floor() as Int
+				weightRight := sourceColumn - sourceColumnLeft as Float
+				if (sourceColumnLeft + 1 >= sourceWidth)
+					weightRight = 0.0f
+				This _blendSquare(sourceBuffer as UInt8*, resultBuffer as UInt8*, sourceStride, resultStride, sourceRowUp, sourceColumnLeft, row + resultStartRow, column + resultStartColumn, (1.0f - weightDown) * (1.0f - weightRight), (1.0f - weightDown) * weightRight, weightDown * (1.0f - weightRight), weightDown * weightRight, bytesPerPixel)
+			}
+		}
+	}
+	_blendSquare: static func (sourceBuffer, resultBuffer: UInt8*, sourceStride, resultStride, sourceRow, sourceColumn, row, column: Int, weightTopLeft, weightTopRight, weightBottomLeft, weightBottomRight: Float, bytesPerPixel: Int) {
+		finalValue: UInt8 = 0
+		for (i in 0 .. bytesPerPixel) {
+			finalValue = weightTopLeft > 0.0f ? weightTopLeft * sourceBuffer[sourceColumn * bytesPerPixel + sourceRow * sourceStride + i] : 0
+			finalValue += weightTopRight > 0.0f ? weightTopRight * sourceBuffer[(sourceColumn + 1) * bytesPerPixel + sourceRow * sourceStride + i] : 0
+			finalValue += weightBottomLeft > 0.0f ? weightBottomLeft * sourceBuffer[sourceColumn * bytesPerPixel + (sourceRow + 1) * sourceStride + i] : 0
+			finalValue += weightBottomRight > 0.0f ? weightBottomRight * sourceBuffer[(sourceColumn + 1) * bytesPerPixel + (sourceRow + 1) * sourceStride + i] : 0
+			resultBuffer[column * bytesPerPixel + row * resultStride + i] = finalValue
+		}
+	}
 }
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -67,43 +67,11 @@ RasterMonochrome: class extends RasterPacked {
 		else {
 			result = This new(size)
 			match (method) {
-				case InterpolationMode Smooth => This _resizeBilinear(this, result)
+				case InterpolationMode Smooth => RasterCanvas resizeBilinear(this buffer pointer as ColorMonochrome*, result buffer pointer as ColorMonochrome*, IntBox2D new(this size), IntBox2D new(result size), this stride, result stride, this bytesPerPixel)
 				case => RasterCanvas resizeNearestNeighbour(this buffer pointer as ColorMonochrome*, result buffer pointer as ColorMonochrome*, IntBox2D new(this size), IntBox2D new(result size), this stride, result stride, this bytesPerPixel)
 			}
 		}
 		result
-	}
-	_resizeBilinear: static func (source, result: This) {
-		resultBuffer := result buffer pointer
-		sourceBuffer := source buffer pointer
-		(resultWidth, resultHeight, resultStride) := (result size x, result size y, result stride)
-		(sourceWidth, sourceHeight, sourceStride) := (source size x, source size y, source stride)
-		for (row in 0 .. resultHeight) {
-			sourceRow := ((sourceHeight as Float) * row) / resultHeight
-			sourceRowUp := sourceRow floor() as Int
-			weightDown := sourceRow - sourceRowUp as Float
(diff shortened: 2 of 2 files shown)


source/draw/RasterMonochrome.ooc
```diff
@@ -67,44 +67,12 @@ RasterMonochrome: class extends RasterPacked {
...
-				if (columnRightValid)
-					valueRowUp += weightRight * sourceBuffer[sourceColumnRight + sourceRowUp * sourceStride]
-				valueRowDown := 0.0f
-				if (rowDownValid)
-					valueRowDown += (1.0f - weightRight) * sourceBuffer[sourceColumnLeft + sourceRowDown * sourceStride] + (columnRightValid ? (weightRight * sourceBuffer[sourceColumnRight + sourceRowDown * sourceStride]) : 0.0f)
-				pixelValue := weightDown * valueRowDown + (1.0f - weightDown) * valueRowUp
-				resultBuffer[column + row * resultStride] = pixelValue as UInt8
-			}
-		}
-	}
```
sebastianbaginski: This code is moved to RasterCanvas.

source/draw/RasterCanvas.ooc
```diff
@@ -88,4 +88,42 @@ RasterCanvas: abstract class extends Canvas {
...
+				sourceColumn := ((sourceWidth as Float) * column) / resultWidth + sourceStartColumn
+				sourceColumnLeft := sourceColumn floor() as Int
+				weightRight := sourceColumn - sourceColumnLeft as Float
+				if (sourceColumnLeft + 1 >= sourceWidth)
+					weightRight = 0.0f
+				This _blendSquare(sourceBuffer as UInt8*, resultBuffer as UInt8*, sourceStride, resultStride, sourceRowUp, sourceColumnLeft, row + resultStartRow, column + resultStartColumn, (1.0f - weightDown) * (1.0f - weightRight), (1.0f - weightDown) * weightRight, weightDown * (1.0f - weightRight), weightDown * weightRight, bytesPerPixel)
+			}
+		}
+	}
+	_blendSquare: static func (sourceBuffer, resultBuffer: UInt8*, sourceStride, resultStride, sourceRow, sourceColumn, row, column: Int, weightTopLeft, weightTopRight, weightBottomLeft, weightBottomRight: Float, bytesPerPixel: Int) {
```
sebastianbaginski: This method is needed because in ooc you can't call methods on generic types, so you can't use already existing `blend` methods from `Color` classes.
It takes source and result buffer and blends four pixels from source into one pixel in result according to given weights.

source/draw/RasterCanvas.ooc
```diff
@@ -88,4 +88,42 @@ RasterCanvas: abstract class extends Canvas {
...
+		finalValue: UInt8 = 0
+		for (i in 0 .. bytesPerPixel) {
```
sebastianbaginski: It loops for every channel in pixel. `bytesPerPixel` is a compile time constant, this loop should be unrolled by optimizer.

source/draw/RasterCanvas.ooc
```diff
@@ -88,4 +88,42 @@ RasterCanvas: abstract class extends Canvas {
...
+			if (weightTopLeft > 0.0f)
+				finalValue = weightTopLeft * sourceBuffer[sourceColumn * bytesPerPixel + sourceRow * sourceStride + i]
+			if (weightTopRight > 0.0f)
+				finalValue += weightTopRight * sourceBuffer[(sourceColumn + 1) * bytesPerPixel + sourceRow * sourceStride + i]
+			if (weightBottomLeft > 0.0f)
+				finalValue += weightBottomLeft * sourceBuffer[sourceColumn * bytesPerPixel + (sourceRow + 1) * sourceStride + i]
+			if (weightBottomRight > 0.0f)
+				finalValue += weightBottomRight * sourceBuffer[(sourceColumn + 1) * bytesPerPixel + (sourceRow + 1) * sourceStride + i]
```
sebastianbaginski: Here I need to multiply by `bytesPerPixel` because `sourceBuffer` is a pointer to `uint8_t` and the `operator[]` will not have the correct offset.

source/draw/RasterCanvas.ooc
```diff
@@ -88,4 +88,42 @@ RasterCanvas: abstract class extends Canvas {
 			}
 		}
 	}
+	resizeBilinear: static func <T> (sourceBuffer, resultBuffer: T*, sourceBox, resultBox: IntBox2D, sourceStride, resultStride, bytesPerPixel: Int) {
+		(resultWidth, resultHeight) := (resultBox size x, resultBox size y)
+		(sourceWidth, sourceHeight) := (sourceBox size x, sourceBox size y)
+		sourceStartColumn := sourceBox leftTop x
+		sourceStartRow := sourceBox leftTop y
+		resultStartColumn := resultBox leftTop x
+		resultStartRow := resultBox leftTop y
```
marcusnaslund: Maybe write lines 94-97 on two rows using tuples as well.

source/draw/RasterCanvas.ooc
```diff
@@ -88,4 +88,42 @@ RasterCanvas: abstract class extends Canvas {
...
+		for (row in 0 .. resultHeight) {
+			sourceRow := ((sourceHeight as Float) * row) / resultHeight + sourceStartRow
+			sourceRowUp := sourceRow floor() as Int
+			weightDown := sourceRow - sourceRowUp as Float
```
marcusnaslund: Subtraction or cast happens first?

source/draw/RasterCanvas.ooc
marcusnaslund: That is a looooooong list of arguments.

source/draw/RasterCanvas.ooc
marcusnaslund: Have you checked the C code?

sebastianbaginski: It won't happen on the C level. I meant the optimizer from gcc. Does rock even have one ? :P

source/draw/RasterCanvas.ooc
marcusnaslund: Can the weights be negative?

source/draw/RasterCanvas.ooc
marcusnaslund: What I meant was, rock might obfuscate this into something that gcc cannot optimize.

source/draw/RasterCanvas.ooc
sebastianbaginski: Cast : `lang_Numbers__Float weightDown = sourceRow - ((lang_Numbers__Float) (sourceRowUp));`

marcusnaslund: Thanks. That's not what you want though?

source/draw/RasterCanvas.ooc
sebastianbaginski: It's just a loop in C too.

source/draw/RasterCanvas.ooc
sebastianbaginski: Good idea, updated.

source/draw/RasterCanvas.ooc
sebastianbaginski: I hope not.

source/draw/RasterCanvas.ooc
marcusnaslund: Thank you rock

source/draw/RasterCanvas.ooc
marcusnaslund: Then these if-statements kind of go against the programming style we have here.

source/draw/RasterCanvas.ooc
sebastianbaginski: `sourceRow` is already a float, I think I was protecting myself from rock.

source/draw/RasterCanvas.ooc
sebastianbaginski: Statements below the ifs are valid if and only if weights are greater than zero. I don't care what someone will pass into this function, I simply do not want to do the calculations if weights are not positive.

source/draw/RasterCanvas.ooc
sebastianbaginski: I was thinking about using a Float[4] to pass weights. But then I think `weightRight := sourceColumn - sourceColumnLeft as Float` is easier to understand than `weights[3] = sourceColumn - sourceColumnLeft as Float`

source/draw/RasterCanvas.ooc
marcusnaslund: Yes,  but this function is only used by `resizeBilinear` where you can already show that they're in the `[0.0f, 1.0f]` range.

sebastianbaginski: `weight != 0.0f` is simply wrong condition if you are adding up unsigned values with weights. I refuse to write incorrect code :P

marcusnaslund: But if any weight is zero, you are doing `finalValue += 0.0f`, and you'll be happy.

sebastianbaginski: Corresponding `weight` is set to zero if the pixel is outside of the image. This expression `weightBottomLeft * sourceBuffer[sourceColumn * bytesPerPixel + (sourceRow + 1) * sourceStride + i]` would still access the pixel buffer at incorrect offset even with `weightBottomLeft` being zero, so we can't simply get rid of the `if` statements.

marcusnaslund: Yargh, I guess you win this round.

marcusnaslund: Short-circuiting arithmetic statements would be nice though.

sebastianbaginski: You mean `finalValue += weightBottomRight > 0.0f ? weightBottomRight * sourceBuffer[(sourceColumn + 1) * bytesPerPixel + (sourceRow + 1) * sourceStride + i] : 0` ?

marcusnaslund:
> Bilinear interpolation method is moved from RasterMonochrome and can work with images with more than 1 bpp.

Would be nice with tests for the other classes. Not for this PR though.

sebastianbaginski:
Sure, I have more PR stacked on top of this :)

sebastianbaginski:
Updated

PR #907, opened 2015-11-26
Removed async parameter from toRaster.

emilwestergren:
@davidpiuva is this ok with you? From now on `toRaster` is strictly synchronous so you have to use `toRasterAsync` for asynchronus reads.

* Breaks API


davidpiuva:
It is okay as long as the change is applied correctly. It would improve readability of my code.

emilwestergren:
@davidpiuva peer

davidpiuva:
ok

davidpiuva:
All affected features runs okay.

PR #909, opened 2015-11-26
Added offset and sort function to FloatPoint2DVectorList.

niklasborwell:


source/math/FloatPoint2DVectorList.ooc
```diff
@@ -87,6 +87,54 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
...
+		else if (start < end) {
+			pivot := This _partition(array, start, end, This _medianOfThree(array, start, end))
+			if (pivot > start)
+				This _quicksort(array, start, pivot - 1)
+			if (pivot < end)
+				This _quicksort(array, pivot + 1, end)
+		}
+	}
+	sortByX: func {
+		This _quicksort(this _vector _backend as FloatPoint2D*, 0, this count - 1)
```
niklasborwell: I'm not sure we want to add the sort in this way. I basically just copied the code from FloatVectorList and modified it to use FloatPoint2D. I didn't add a sortByY because I think sortByX would be the most common usage.

niklasborwell:
@tomasmahlberg Peer review?

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -87,6 +87,54 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
 		}
 		result
 	}
+	offset: func (offset: FloatPoint2D) -> This {
```
tomasmahlberg: This function name is a bit ambiguous but from looking at the signature it should be clear what it does, so it's OK I guess.

tomasmahlberg:
Looks OK.

source/math/FloatPoint2DVectorList.ooc
niklasborwell: I was thinking about maybe calling it shift, but that is used in a different way in FloatVectorList.

tomasmahlberg: Maybe actually addOffset is better as offset might sound like you are shifting it... I don't know... let's maybe just keep it like this?

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -87,6 +87,54 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
...
+		result := This new()
+		thisPointer := this pointer as FloatPoint2D*
+		for (i in 0 .. this count)
+			result add(thisPointer[i] + offset)
+		result
+	}
+	_swap: static func (array: FloatPoint2D*, i, j: Int) {
+		t := array[i]
```
marcusnaslund: `temporary`.

source/math/FloatPoint2DVectorList.ooc
marcusnaslund: I prefer `addOffset`.

marcusnaslund: Or `offsetAll`.

marcusnaslund: Wait, the `+` operator already does the exact same thing.

niklasborwell: ..Doh.

niklasborwell:
Updated.

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -87,10 +87,52 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
...
+	}
+	sortByX: func {
+		This _quicksortX(this _vector _backend as FloatPoint2D*, 0, this count - 1)
+	}
 	operator + (value: FloatPoint2D) -> This {
 		result := This new()
-		for (i in 0 .. this _count)
-			result add(this[i] + value)
+		thisPointer := this pointer as FloatPoint2D*
+		for (i in 0 .. this count)
```
marcusnaslund: Don't remove the `_`.

source/math/FloatPoint2DVectorList.ooc
```diff
@@ -87,10 +87,52 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
 		}
 		result
 	}
+	_swap: static func (array: FloatPoint2D*, i, j: Int) {
```
marcusnaslund: Hmm, might be nice with a generic `swap` in `VectorList`. Not for this PR though.

niklasborwell: Yeah that would be a good idea.

niklasborwell:
Updated.

PR #910, opened 2015-11-26
Cleaned up properties in Quaternion

tomasmahlberg:


tomasmahlberg:
@marcusnaslund (peer) review?

source/math/Quaternion.ooc
```diff
@@ -36,15 +36,15 @@ Quaternion: cover {
...
-	isIdentity ::= (this w == 1.0f && this x == 0.0f && this y == 0.0f && this z == 0.0f)
-	isNull ::= (this w == 0.0f && this x == 0.0f && this y == 0.0f && this z == 0.0f)
-	norm ::= (this real squared() + (this imaginary norm) squared()) sqrt()
-	normalized ::= (this / this norm) as This
+	isValid ::= this real == this real && this imaginary isValid
+	isIdentity ::= this real == 1.f && this imaginary x == 0.f && this imaginary y == 0.f && this imaginary z == 0.f
+	isNull ::= this real == 0.f && this imaginary x == 0.f && this imaginary y == 0.f && this imaginary z == 0.f
+	norm ::= (this real squared() + this imaginary norm squared()) sqrt()
+	normalized ::= this / this norm
 	logarithmImaginaryNorm ::= ((this logarithm) imaginary) norm
```
tomasmahlberg: This has to look like this...

test/math/QuaternionTest.ooc
```diff
@@ -86,7 +85,7 @@ QuaternionTest: class extends Fixture {
 			expect(conjugate z == -quaternion5 z)
 		})
 		this add("normalized", func {
-			normalized := quaternion0 normalized
+			normalized := quaternion0 normalized as Quaternion
```
tomasmahlberg: Needed here because `expect` was throwing weird errors otherwise.

source/math/Quaternion.ooc
```diff
@@ -36,15 +36,15 @@ Quaternion: cover {
...
 
-	inverse ::= This new(this w / (this norm * this norm), -this x / (this norm * this norm), -this y / (this norm * this norm), -this z / (this norm * this norm))
-	isValid ::= (this w == this w && this x == this x && this y == this y && this z == this z)
-	isIdentity ::= (this w == 1.0f && this x == 0.0f && this y == 0.0f && this z == 0.0f)
-	isNull ::= (this w == 0.0f && this x == 0.0f && this y == 0.0f && this z == 0.0f)
-	norm ::= (this real squared() + (this imaginary norm) squared()) sqrt()
-	normalized ::= (this / this norm) as This
+	isValid ::= this real == this real && this imaginary isValid
+	isIdentity ::= this real == 1.f && this imaginary x == 0.f && this imaginary y == 0.f && this imaginary z == 0.f
+	isNull ::= this real == 0.f && this imaginary x == 0.f && this imaginary y == 0.f && this imaginary z == 0.f
```
tomasmahlberg: Should one use `equals` for float literal comparison in situations like these now that we have it?

source/math/Quaternion.ooc
marcusnaslund: Actually, we should probably use `equals()` when comparing Float.. or? @sebastianbaginski

source/math/Quaternion.ooc
marcusnaslund: Oh, haha, I just said the same thing higher up but it's more relevant here. @sebastianbaginski ?

source/math/Quaternion.ooc
sebastianbaginski: Yeah I strongly recommend using `equals` for floats because of https://github.com/cogneco/ooc-kean/issues/860

test/math/QuaternionTest.ooc
marcusnaslund: That's because you removed the cast from the property. Why the cast was needed at all... I have no idea.

tomasmahlberg: Yes, I know. I think the problem is in the `expect` function here. The explicit cast in the property is not needed for general usage, I think it was basically there because it fixed the test.

tomasmahlberg:
Updated.

test/math/QuaternionTest.ooc
```diff
@@ -72,7 +72,6 @@ QuaternionTest: class extends Fixture {
 		this add("inverse", func {
 			inverse := this quaternion2 inverse
 			product := this quaternion2 * inverse
-			//productReal :=
```
marcusnaslund: oops

PR #912, opened 2015-11-26
Fixed 16 MB memory leak in Random generator tests

marcusnaslund:
All the other big ones are related to `Raster` classes, so I'm hoping on you @sebastianbaginski


sebastianbaginski:
I will look into it, but the last time we fixed leaks from `open` it crashed your system :/

PR #914, opened 2015-11-26
=)

fredrikbryntesson:


=)
```diff
@@ -0,0 +1,8 @@
+=)
+=)
+=)
+=)
+=)
```
marcusnaslund: This line causes segfaults.

sebastianbaginski:
This should go upstream. We need more happiness asap.

marcusnaslund:
Will you peer review @sebastianbaginski ? All tests check out and magic has no complaints.

marcusnaslund:
Passes validation. The phone seems a lot happier.

PR #920, opened 2015-11-27
Implemented replaceAll for Text

sebastianbaginski:
Added `replaceAll` so we can use `Text` instead of `String` in `DateTime`.
@tomasmahlberg can you also check that I got the `ownership` part right ?

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -149,6 +149,31 @@ Text: cover {
 			--rightPosition
 		this slice(leftPosition, rightPosition - leftPosition + 1)
 	}
+	replaceAll: func (toReplace, newValue: This) -> This {
+		t := this take()
+		result: This
+		position := t find(toReplace)
+		if (position == -1)
+			result = t copy()
+		else {
+			previous := 0
+			result = This new()
+			toReplaceTaken := toReplace take()
+			newValueTaken := newValue take()
+			toReplaceCount := toReplaceTaken count
+			while (position > -1) {
+				result += t slice(previous, position - previous) + newValueTaken
+				previous = position + toReplaceCount
+				position = t find(toReplaceTaken, previous)
+			}
+			if (previous < t count)
+				result += t slice(previous, t count - previous)
+		}
+		toReplace free(Owner Receiver)
+		newValue free(Owner Receiver)
+		this free(Owner Receiver)
+		result
+	}
 	toString: func -> String { this _buffer toString() }
 	toInt: func -> Int { this toLLong() as Int }
 	toLong: func -> Long { this toLLong() as Long }
--- a/test/base/TextTest.ooc
+++ b/test/base/TextTest.ooc
@@ -8,11 +8,15 @@ TextTest: class extends Fixture {
 		super("Text")
 		this add("constructors", func {
 			text := Text new(c"test string", 5)
-			expect(text give() toString() == "test ")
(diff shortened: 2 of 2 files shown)


source/base/Text.ooc
```diff
@@ -149,6 +149,30 @@ Text: cover {
...
+		t := this take()
+		result: This
+		position := t find(toReplace)
+		if (position == -1)
+			result = t copy()
+		else {
+			previous := 0
+			result = This new()
+			x := toReplace take()
+			y := newValue take()
```
sebastianbaginski: I saw that variables with one letter name are used in this file in other methods.

source/base/Text.ooc
```diff
@@ -149,6 +149,30 @@ Text: cover {
...
+			while (position > -1) {
+				result += t slice(previous, position - previous) + y
```
marcusnaslund: We have a `TextBuilder`, you know, wouldn't that be more efficient?

source/base/Text.ooc
marcusnaslund: Yeah, I can accept `t`, but `x` and `y` are pretty non-explanatory. `toReplaceTaken` and `newValueTaken`, is that too much?

test/base/TextTest.ooc
```diff
@@ -177,17 +192,32 @@ this add("text literal toString", func {
...
 		})
+		this add("replace all", func {
+			text := t"aa00aa00"
+			notReplaced := text replaceAll(t"xx", t"yy")
+			expect(text == notReplaced)
+			replaced := text replaceAll(t"00", t"bb")
+			expect(replaced == t"aabbaabb")
+			replaced = t"xYxYxY" replaceAll(t"x", t"y")
+			expect(replaced == t"yYyYyY")
+			replaced = t"ooc is great language" replaceAll(t"great", t"the best") replaceAll(t"ooc", t"C++")
```
marcusnaslund: Haha :)

source/base/Text.ooc
sebastianbaginski: I thought about it too, but would the code be more readable ? It's not that we will run `replaceAll` million times per frame

test/base/TextTest.ooc
sebastianbaginski: What ? It's just a random test string.

tomasmahlberg: I LOL'd. :)

tomasmahlberg:
Ownership looks about right in `replaceAll`. Maybe we can come up with a standard for naming taken variables in situations like this? Like adding a short suffix, maybe `newValue_` for `newValue` or something similar? Don't really like the look of `newValueTaken`.

source/base/Text.ooc
marcusnaslund: You're right.

marcusnaslund:
@tomasmahlberg - will you make a cleanup issue about that? We should discuss `t` while we're at it.

tomasmahlberg:
Issue #922.

marcusnaslund:
Thanks.
I hope you'll be implementing this in `DateTime` (and `TimeSpan`) soon @sebastianbaginski :)

sebastianbaginski:
Yes, I have next small PR stacked.
I also think we should start working on replacing `String` with `Text` in `File`. Memory leaks from the `File` class are quite annoying.

PR #924, opened 2015-11-27
Added print and println to Text

sebastianbaginski:
Adds basic printing functions to `Text` class.

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -174,6 +174,14 @@ Text: cover {
 		this free(Owner Receiver)
 		result
 	}
+	print: func {
+		printf("%s", this _buffer _backend pointer as Char*)
+		this free(Owner Receiver)
+	}
+	println: func {
+		printf("%s\n", this _buffer _backend pointer as Char*)
+		this free(Owner Receiver)
+	}
 	toString: func -> String { this _buffer toString() }
 	toInt: func -> Int { this toLLong() as Int }
 	toLong: func -> Long { this toLLong() as Long }
--- a/source/base/TextBuilder.ooc
+++ b/source/base/TextBuilder.ooc
@@ -91,7 +91,7 @@ TextBuilder: class {
 			result = TextBuffer empty
 		Text new(result give())
 	}
-	println: func { this toString() println().free() }
+	println: func { this toText() println() }
 	operator [] (index: Int) -> Char {
 		i := index
 		position := 0


source/base/Text.ooc
```diff
@@ -174,6 +174,14 @@ Text: cover {
 		this free(Owner Receiver)
 		result
 	}
+	print: func {
+		printf("%s", this _buffer _backend pointer as Char*)
+		this free(Owner Receiver)
+	}
+	println: func {
+		printf("%s\n", this _buffer _backend pointer as Char*)
```
marcusnaslund: Aren't these strings going to leak?

sebastianbaginski: I think these are literals:
```C
        __Text_strLit2 = (void*) lang_String__makeStringLiteral("%s\n", 3);
        __Text_strLit1 = (void*) lang_String__makeStringLiteral("%s", 2);
```

marcusnaslund: OK then.

source/base/Text.ooc
```diff
@@ -174,6 +174,14 @@ Text: cover {
 		this free(Owner Receiver)
 		result
 	}
+	print: func {
+		printf("%s", this _buffer _backend pointer as Char*)
```
emilwestergren: Please use `Debug print`, `printf` doesn't work on Android.

source/base/Text.ooc
sebastianbaginski: Yeah you never know without looking at C code. There was a bug related to this in plotting module not so long ago.

source/base/Text.ooc
marcusnaslund: Oh, that's good to know.

sebastianbaginski: That's interesting. Does the process on android system have a `stdout` stream attached ? I mean, can you do `fwrite(stdout, "my string\n")` instead of `println` ?

emilwestergren: Of course it *works* but by registering a callback to `Debug print` we can redirect any prints through their own logging system.

emilwestergren: `printf` works but Android defaults `stdout` to `/dev/null`. I think you can change that in a settings file somewhere, but currently we use the registered callback in `Debug print` which uses Android's logging functions.

sebastianbaginski: One can do that without callbacks by reopening `stdout` with `freopen`, you can redirect anything that goes to stdout anywhere you like. Are you sure we should bind the Debug module into `Text` ? `String` class didn't use Debug in `print`.

emilwestergren: http://developer.android.com/tools/debugging/debugging-log.html#viewingStd

emilwestergren: Yeah I changed my mind, don't change it.

emilwestergren: Just be aware that `printf` won't show in the Android log by default.

sebastianbaginski: OK, I think we should extend `Debug` class to support `Text`.

emilwestergren: Yes, sounds much better.

PR #927, opened 2015-11-27
Added a status printout in the testscript

jonathanudd:
This is to support presentation of statuses on print.


jonathanudd:
@marcusnaslund who do you want to do the peer review on this?

marcusnaslund:
I believe I'm competent enough myself. :)

marcusnaslund:
There is already a printout if something fails, like `Failed tests: 2 [FloatBox2D, FloatVector2D]`. Couldn't you read that, and assume it's OK if there is no such line?

PR #928, opened 2015-11-27
Added format support to Text

sebastianbaginski:
Will be needed in `DateTime`.
I tried to move `Format.ooc` out of SDK but it's not that easy...
@marcusnaslund

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -182,6 +182,15 @@ Text: cover {
 		printf("%s\n", this _buffer _backend pointer as Char*)
 		this free(Owner Receiver)
 	}
+	format: func (args: ...) -> This {
+		string := this take() toString()
+		resultString := string format(args)
+		string free()
+		result := This new(resultString) copy()
+		resultString free()
+		this free(Owner Receiver)
+		result
+	}
 	toString: func -> String { this _buffer toString() }
 	toInt: func -> Int { this toLLong() as Int }
 	toLong: func -> Long { this toLLong() as Long }
--- a/test/base/TextTest.ooc
+++ b/test/base/TextTest.ooc
@@ -218,6 +218,14 @@ TextTest: class extends Fixture {
 			replaced = t"a complicated sentence with multiple replaced words" replaceAll(t"a complicated", t"not so complicated") replaceAll(t"multiple", t"very few") replaceAll(t"words", t"letters")
 			expect(replaced == "not so complicated sentence with very few replaced letters")
 		})
+		this add("format test", func {
+			text := t"%i x %i" format(10, 15)
+			expect(text == t"10 x 15")
+			text = t"%02d" format(4)
+			expect(text == t"04")
+			text = t"%s %s %i %i" format("test", "number", 0, 1)
+			expect(text == t"test number 0 1")
+		})
 	}
 }
 


source/base/Text.ooc
```diff
@@ -182,6 +182,16 @@ Text: cover {
 		printf("%s\n", this _buffer _backend pointer as Char*)
 		this free(Owner Receiver)
 	}
+	format: func (args: ...) -> This {
+		t := this take()
+		string := t toString()
```
sebastianbaginski: It depends on `String` for now, but at least `DateTime` and `TimeSpan` can be implemented without `String` now.

marcusnaslund:
> I tried to move Format.ooc out of SDK but it's not that easy...

Actually, @simonmika had some idea about moving `Text` into the sdk instead. Nothing's decided there yet though. We should talk about that.

source/base/Text.ooc
marcusnaslund: Does that mean we will be able to eliminate those leaks?

marcusnaslund:
@tomasmahlberg can you please check the `OwnedBuffer` usage here?

source/base/Text.ooc
```diff
@@ -182,6 +182,16 @@ Text: cover {
 		printf("%s\n", this _buffer _backend pointer as Char*)
 		this free(Owner Receiver)
 	}
+	format: func (args: ...) -> This {
+		t := this take()
```
tomasmahlberg: This is maybe a bit unnecessary since you only use `t` in one place, better just use `this take()` explicitly perhaps?

sebastianbaginski: OK

sebastianbaginski:
Alright, but since you had that issue about removing SDK I thought I can help you a bit ;)

source/base/Text.ooc
```diff
@@ -182,6 +182,16 @@ Text: cover {
...
+		string := t toString()
+		resultString := string format(args)
+		string free()
+		result := This new(resultString) copy()
+		resultString free()
+		this free(Owner Receiver)
+		result give()
```
tomasmahlberg: This `give()` could be moved to the end of the line where you assign `result`, but it's not important.

source/base/Text.ooc
sebastianbaginski: That's my goal.

marcusnaslund:
Right, there's still a lot of stuff that should be moved out, but maybe some stuff that should be left. Let's discuss this soon.

source/base/Text.ooc
sebastianbaginski: So the `give` is needed, it is not enough to call `copy` ?

marcusnaslund: Let's do it, it saves us a line.

source/base/Text.ooc
sebastianbaginski: @tomasmahlberg can you check again ?

source/base/Text.ooc
tomasmahlberg: Yes, I have no idea actually, you'd have to trace this function to figure it out...

sebastianbaginski: `copy` already sets owner to receiver, so I think `give` is not needed here.

sebastianbaginski:
Updated.

PR #933, opened 2015-11-27
Moved Process from the SDK to source/io

marcusnaslund:
@sebastianbaginski - we're moving these to io instead of concurrent as you said. A future PR will clean these files up.


marcusnaslund:
Do you think `Pipe` belongs there too?

sebastianbaginski:
> Do you think Pipe belongs there too?

Sure, a pipe is one of possible ways to do `io` between processes.

PR #936, opened 2015-11-28
Removed Coro.ooc

marcusnaslund:
We don't use it, so we're throwing it away.


sebastianbaginski:
This is interesting.
We should probably remove it anyway.

PR #939, opened 2015-11-28
Cleaned up types.ooc

marcusnaslund:
Cleaned up the file, only style fixes.
Removed the interface.

(Hint: append ?w=1 to the URL when looking at the diff)


sebastianbaginski:
> Removed the interface.

Too bad, shame that metaprogramming in ooc is so painful.

marcusnaslund:
I think there is still hope that we can revive it some day, but in the mean time, it goes.

PR #942, opened 2015-11-29
Cleaned up win32 specific files

marcusnaslund:
We're keeping these since removing them would break Win32 compatibility. 
Only style fixes. 

(Hint: append ?w=1 to the URL when looking at the diff)

### DIFF
--- a/source/sdk/native/win32/errors.ooc
+++ b/source/sdk/native/win32/errors.ooc
@@ -1,49 +1,44 @@
-
 version(windows) {
+	import native/win32/types
 
-    import native/win32/types
+	include windows
 
-    include windows
+	GetLastError: extern func -> Int
 
-    GetLastError: extern func -> Int
+	ERROR_HANDLE_EOF: extern Int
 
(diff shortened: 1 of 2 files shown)


source/sdk/native/win32/errors.ooc
```diff
@@ -1,49 +1,44 @@
...
-
+	import native/win32/types
+
+	include windows
+
+	GetLastError: extern func -> Int
+
+	ERROR_HANDLE_EOF: extern Int
+
+	FormatMessage: extern func (dwFlags: DWORD, lpSource: Pointer, dwMessageId: DWORD, dwLanguageId: DWORD, lpBuffer: LPTSTR, nSize: DWORD, Arguments: VaList*) -> DWORD
```
sebastianbaginski: This can be simplified, but nevermind.
Encoding a type in variable name, programming style straight from m$...

marcusnaslund: More could be done, but at least these files are magic compliant now.

PR #945, opened 2015-11-29
Removed ArrayList dependancy from Iterable<T>

marcusnaslund:
Removed the not used `toList` from `Iterable<T>`. I assume we will continue to use iterators, but we certainly don't want `ArrayList`. 

Also made some quick fixes to make the file magic compliant.

Since you touched this file last time, I'm asking you for peer review @thomasfanell

### DIFF
--- a/source/sdk/lang/Iterators.ooc
+++ b/source/sdk/lang/Iterators.ooc
@@ -1,36 +1,24 @@
-import structs/[List, ArrayList] /* for Iterable<T> toArrayList */
-
-/**
- * iterators
- */
 Iterable: abstract class <T> {
 	iterator: abstract func -> Iterator<T>
-	/** Return the contents of the iterable as an ArrayList. */
-	toList: func -> ArrayList<T> {
-		result := ArrayList<T> new()
-		for(elem: T in this)
-			result add(elem)
-		result
-	}
 	reduce: func (f: Func (T, T) -> T) -> T {
 		iter := iterator()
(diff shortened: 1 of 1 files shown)


source/sdk/lang/Iterators.ooc
```diff
@@ -1,36 +1,24 @@
...
-	}
 	reduce: func (f: Func (T, T) -> T) -> T {
 		iter := iterator()
 		acc := f(iter next(), iter next())
-		while(iter hasNext?())
+		while (iter hasNext?())
 			acc = f(acc, iter next())
 		acc
 	}
 	each: func (f: Func (T)) {
```
thomasfanell: @marcusnaslund iterator leakage?

source/sdk/lang/Iterators.ooc
```diff
@@ -1,36 +1,24 @@
...
-		for(elem in this)
+		for (elem in this)
 			f(elem)
 	}
 	// Return false to break
 	eachUntil: func (f: Func (T) -> Bool) {
```
thomasfanell: @marcusnaslund same here?

source/sdk/lang/Iterators.ooc
marcusnaslund: Oh definitely, but that's not a part of this PR.

marcusnaslund:
Thanks for the comments @thomasfanell - but they have nothing to do with what this PR does. :)

PR #950, opened 2015-11-30
Implemented solver for tridiagonal systems

### LINKED ISSUE #858: Optimize interpolateCubicSpline in FloatVectorList
`interpolateCubicSpline`, introduced in #840, uses inefficient ways both for storing data by having a large square tridiagonal matrix as a `FloatMatrix` and also for solving it, using LUP decomposition. There are much faster ways of solving these types of systems and we might want to do it piece-wise instead of finding all parameters at once.

tomasmahlberg:
Fixes #858.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -236,6 +236,39 @@ FloatMatrix: cover {
 		this free(Owner Receiver)
 		result
 	}
+	// Thomas algorithm for tridiagonal systems
+	// https://en.wikipedia.org/wiki/Tridiagonal_matrix_algorithm
+	solveTridiagonal: func (y: This) -> This {
+		diagonal := this take() getDiagonal() take()
+		upperDiagonal := this take() getDiagonal(1, 0) take()
+		lowerDiagonal := this take() getDiagonal(0, 1) take()
+		rightHandSide := y take() copy() take()
+		d := diagonal elements
+		u := upperDiagonal elements
+		l := lowerDiagonal elements
+		rhs := rightHandSide elements
+		result := diagonal create()
+		length := result take() height
+		u[0] /= d[0]
+		rhs[0] /= d[0]
+		for (index in 1 .. length) {
+			if (index < length - 1)
+				u[index] /= d[index] - l[index - 1] * u[index - 1]
+			rhs[index] = (rhs[index] - l[index - 1] * rhs[index - 1]) / (d[index] - l[index - 1] * u[index - 1])
+		}
+		result[0, length - 1] = rhs[length - 1]
+		for (iteration in 0 .. length - 1) {
+			index := length - 2 - iteration
+			result[0, index] = rhs[index] - u[index] * result take()[0, index + 1]
+		}
+		diagonal free()
+		upperDiagonal free()
+		lowerDiagonal free()
+		rightHandSide free()
+		this free(Owner Receiver)
+		y free(Owner Receiver)
+		result
+	}
 	getDiagonal: func (x := 0, y := 0) -> This {
 		t := this take()
 		length := Int minimum(t width - x, t height - y)
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -236,6 +236,15 @@ FloatMatrix: cover {
 		this free(Owner Receiver)
 		result
 	}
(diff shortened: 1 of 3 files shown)


tomasmahlberg:
@marcusnaslund review

source/math/FloatMatrix.ooc
```diff
@@ -236,6 +236,48 @@ FloatMatrix: cover {
 		this free(Owner Receiver)
 		result
 	}
+	// Thomas algorithm for tridiagonal systems
```
marcusnaslund: What an appropriate name. :)

source/math/FloatMatrix.ooc
```diff
@@ -236,6 +236,48 @@ FloatMatrix: cover {
...
+	// https://en.wikipedia.org/wiki/Tridiagonal_matrix_algorithm
+	solveTridiagonal: func (y: This) -> This {
+		diagonal := this take() getDiagonal() take()
```
marcusnaslund: I was asked by @simonmika to write a short description of `OwnedBuffer` for people, but I honestly think you're the better person for that. Do you have time? It can (and should be) pretty short, just explaining what goes on in calls like this, what `take()`, `give()` etc. mean and when they're needed.

source/math/FloatMatrix.ooc
```diff
@@ -236,6 +236,48 @@ FloatMatrix: cover {
...
+		upperDiagonal := this take() getDiagonal(1, 0) take()
+		lowerDiagonal := this take() getDiagonal(0, 1) take()
+		rightHandSide := y take() copy() take()
+		d := diagonal elements
+		u := upperDiagonal elements
+		l := lowerDiagonal elements
```
marcusnaslund: How about naming these `diagonal`, `upperDiagonal` and `lowerDiagonal`, and naming the matrix instances something like `diagonalVector`?

source/math/FloatMatrix.ooc
```diff
@@ -236,6 +236,48 @@ FloatMatrix: cover {
...
+		}
+		diagonal free()
+		upperDiagonal free()
+		lowerDiagonal free()
+		rightHandSide free()
+		this free(Owner Receiver)
+		y free(Owner Receiver)
+		result
+	}
+	getDiagonal: func (x := 0, y := 0) -> This {
```
marcusnaslund: `startX` and `startY`, perhaps

source/math/FloatMatrix.ooc
tomasmahlberg: Sure, I will try to come up with something at some point today. I can send it to you as a draft in any case.

marcusnaslund: Thanks!

source/math/FloatMatrix.ooc
tomasmahlberg: The problem with long variable names here is that it will bloat the algorithm calculations below enormously.

source/math/FloatMatrix.ooc
tomasmahlberg: Agreed.

source/math/FloatMatrix.ooc
marcusnaslund: `diagonal`, `upper`, `lower`, then.

marcusnaslund: It would be more OK to have `i` instead of `index` if it still gets too bloated, though.

tomasmahlberg: That could work, let's see how it looks.

tomasmahlberg:
Updated.

source/math/FloatMatrix.ooc
```diff
@@ -236,6 +236,48 @@ FloatMatrix: cover {
...
+		lower := lowerDiagonalVector elements
+		rhs := rightHandSide elements
+		result := diagonalVector create()
+		length := result take() height
+		upper[0] /= diagonal[0]
+		rhs[0] /= diagonal[0]
+		for (index in 1 .. length) {
+			if (index < length - 1)
+				upper[index] /= diagonal[index] - lower[index - 1] * upper[index - 1]
+			rhs[index] = (rhs[index] - lower[index - 1] * rhs[index - 1]) / (diagonal[index] - lower[index - 1] * upper[index - 1])
```
marcusnaslund: This looks better.

PR #951, opened 2015-11-30
fixed some memory leaks for Constraint classes

sebastianbaginski:
Base class for Constraints now calls `free` on its child.
Fixed memory issues related to cell ownership.
@marcusnaslund

### DIFF
--- a/source/unit/Constraints.ooc
+++ b/source/unit/Constraints.ooc
@@ -31,6 +31,19 @@ Modifier: abstract class {
 	child: This = null
 	init: func
 	init: func ~parent (=parent)
+	free: override func {
+		if (this child) {
+			if (this child parent == this)
+				this child parent = null
+			this child free()
+		}
+		if (this parent) {
+			if (this parent child == this)
+				this parent child = null
+			this parent free()
+		}
+		super()
+	}
 	verify: func ~parent (value: Object, =child) -> Bool {
 		this parent != null ? this parent verify(value, this): this test(value)
 	}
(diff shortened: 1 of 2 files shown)


source/unit/Constraints.ooc
```diff
@@ -31,6 +31,19 @@ Modifier: abstract class {
...
 	init: func ~parent (=parent)
+	free: override func {
+		if (this child) {
+			if (this child parent == this)
+				this child parent = null
+			this child free()
+		}
+		if (this parent) {
+			if (this parent child == this)
+				this parent child = null
```
sebastianbaginski: This and similar line above is to prevent circular `free` calls, and allow to clear entire hierarchy of objects from any point in the parent-child chain.

source/unit/Constraints.ooc
```diff
@@ -274,18 +287,18 @@ CompareWithinConstraint: class extends CompareConstraint {
 		this precision = precision as LDouble
 		this comparer = func (value, correct: Object) -> Bool { this testChild(value) }
 		f := func (value, correct: Object) -> Bool { (value as Cell<Float> get() - correct as Cell<Float> get()) abs() < precision }
-		CompareConstraint new(this, this correct, f, ComparisonType Within)
+		CompareConstraint new(this, Cell<Float> new(this correct as Cell<Float> get()), f, ComparisonType Within)
```
sebastianbaginski: `this correct` was shared between parent and several children, it was not clear who should delete it and was causing invalid delete / free after implementing `free` for `Modifier`.

marcusnaslund:
`FloatComplexVectorList` fails on CircleCI. There are some invalid read/writes in that test, IIRC.

sebastianbaginski:
Rebased. On my machine it behaves like on the master branch now, it has one uninitialized value read on master too.

marcusnaslund:
Same problem. The test on upstream fails on Windows as well, so there's something about that uninitialized value read that goes unnoticed on our machines for some reason.

sebastianbaginski:
Still fails on Circle, weird. Hold on with this for now.

marcusnaslund:
Alright, try rebasing this now.

sebastianbaginski:
OK, rebased.

PR #953, opened 2015-11-30
Fixed uninitialized memory read in FloatComplexVectorList

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/952
@marcusnaslund

### DIFF
--- a/test/math/FloatComplexVectorListTest.ooc
+++ b/test/math/FloatComplexVectorListTest.ooc
@@ -34,7 +34,7 @@ FloatComplexVectorListTest: class extends Fixture {
 	complexNumberArray add(complexNumber2)
 	complexNumberArray add(complexNumber3)
 
-	tolerance := 0.00001
+	tolerance := 0.00001f
 
 	init: func {
 		super("FloatComplexVectorList")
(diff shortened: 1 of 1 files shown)


test/math/FloatComplexVectorListTest.ooc
```diff
@@ -34,7 +34,7 @@ FloatComplexVectorListTest: class extends Fixture {
 	complexNumberArray add(complexNumber2)
 	complexNumberArray add(complexNumber3)
 
-	tolerance := 0.00001
+	tolerance := 0.00001f
```
marcusnaslund: Aaaaaaah. I've fallen for this so many times. Nice catch.

sebastianbaginski: Looks like Constraints can't handle different parameter types for tolerance and float arguments.
This change fixed the uninitialized memory read.

test/math/FloatComplexVectorListTest.ooc
```diff
@@ -70,10 +70,10 @@ FloatComplexVectorListTest: class extends Fixture {
...
 			sum := list sum
 			mean := list mean
-			expect(sum real == 6.0f)
-			expect(sum imaginary == 4.0f)
-			expect(mean real == 1.5f)
-			expect(mean imaginary == 1.0f) //FIXME getting errors using "is equal to" here because... ooc (expected 1.00000000 got 1.00000000)
+			expect(sum real, is equal to(6.0f) within(tolerance))
+			expect(sum imaginary, is equal to(4.0f) within(tolerance))
+			expect(mean real, is equal to(1.5f) within(tolerance))
+			expect(mean imaginary, is equal to(1.0f) within(tolerance))
```
sebastianbaginski: These lines are only to get rid of TODO.

test/math/FloatComplexVectorListTest.ooc
marcusnaslund: That's strange since IIRC they're all cast to `LDouble` anyway.

sebastianbaginski: That's right. I don't feel like digging deeper into this though :)

marcusnaslund:
Tests now work on Windows as well. :)

PR #954, opened 2015-11-30
Fixed array indexing in safe mode

sebastianbaginski:
Other project can now be compiled with -Dsafe flag.
@marcusnaslund ready

### DIFF
--- a/source/sdk/lang/Array.h
+++ b/source/sdk/lang/Array.h
@@ -1,6 +1,5 @@
-#pragma once
-#ifndef ___lang_array___
-#define ___lang_array___
+#ifndef OOC_LANG_ARRAY_H
+#define OOC_LANG_ARRAY_H
 
 #ifdef __OOC_USE_GC__
 #define array_malloc GC_malloc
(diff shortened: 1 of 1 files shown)


source/sdk/lang/Array.h
```diff
@@ -12,20 +11,13 @@
...
-
 #define _lang_array__Array_new(type, size) ((_lang_array__Array) { size, array_malloc((size) * sizeof(type)) });
 
 #if defined(safe)
 #define _lang_array__Array_get(array, index, type) ( \
-    (index < 0 || index >= array.length) ? \
-    lang_Exception__Exception_throw((lang_Exception__Exception *) lang_Exception__OutOfBoundsException_new_noOrigin(index, array.length)), \
-    *((type*) NULL) : \
-    ((type*) array.data)[index])
+    ((type*) array.data)[(index < 0 || index >= array.length) ? lang_Exception__Exception_throw((lang_Exception__Exception *) lang_Exception__OutOfBoundsException_new_noOrigin(index, array.length)), -1 : index])
```
marcusnaslund: So the change is from NULL to `-1`? What does it mean?

sebastianbaginski: I removed the conditional expression and replaced it with simple array access, but the index calculation can throw exception now.
`-1` is there to indicate invalid index, array at `-1` is never accessed because exception is thrown. It uses the comma operator to first throw exception and then return correct type for array index.
```C
x = a, b;
```
x has the value and type of b but a is evaluated first

marcusnaslund: The exception was thrown first in the previous code as well, so why does this work better?

sebastianbaginski: Previously the `_lang_array__Array_set` macro was causing confusion when rock was trying to resolve the types after macro expansion. Both versions should do the same thing, in C at least. But the error was thrown at the ooc stage, so I think rock did macro expansion too and failed.

marcusnaslund: Ah, poor little rock.
Thanks!

PR #955, opened 2015-11-30
Fixed remaining memory leaks in Constraints

sebastianbaginski:
Added `IsConstraint` to the constraint hierarchy.
Removed suppressions for Constraints from kean.supp. Only suppression still used is the one for string literals.
@marcusnaslund ready


kean.supp
```diff
@@ -1,18 +1 @@
...
-   Memcheck:Leak
-   ...
-   fun:Fixture__Fixture_expect_*
-}
-{
-   fixture_getequal
-   Memcheck:Leak
-   ...
-   fun:Constraints__IsConstraints___getequal__
-}
```
sebastianbaginski: This file is empty now, but maybe we should keep it for future suppressions ?

marcusnaslund:
Woah, hang on, `Fixture` doesn't leak at all anymore?

sebastianbaginski:
No more leaks from Fixture now. These were the last ones.

kean.supp
marcusnaslund: No point in keeping an empty file.

sebastianbaginski: Updated

marcusnaslund:
Nice!

PR #956, opened 2015-11-30
Reduced memory leaks from loops in CreateAlphaByteString

marcusnaslund:

### DIFF
--- a/source/draw/CreateAlphaByteString.ooc
+++ b/source/draw/CreateAlphaByteString.ooc
@@ -32,9 +32,12 @@ CreateAlphaByteString: class {
 			ipbuffer pointer[i] = buf pointer[(i * 4) + 3]
 		ip = ipbuffer pointer as Int*
 		if (ipbuffer size > 3)
-			imageArray += ip[0] toString()
-		for (i in 1 .. ipbuffer size / 4)
-			imageArray += "," + (i % 32 == 0 ? "\n" : " ") + ip[i] toString()
+			imageArray = imageArray >> ip[0] toString()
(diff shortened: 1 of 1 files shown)


source/draw/CreateAlphaByteString.ooc
```diff
@@ -32,9 +32,12 @@ CreateAlphaByteString: class {
 			ipbuffer pointer[i] = buf pointer[(i * 4) + 3]
 		ip = ipbuffer pointer as Int*
 		if (ipbuffer size > 3)
-			imageArray += ip[0] toString()
-		for (i in 1 .. ipbuffer size / 4)
-			imageArray += "," + (i % 32 == 0 ? "\n" : " ") + ip[i] toString()
+			imageArray = imageArray >> ip[0] toString()
+		for (i in 1 .. ipbuffer size / 4) {
+			intString := ip[i] toString()
+			imageArray = imageArray >> "," >> (i % 32 == 0 ? "\n" : " ") >> intString
```
sebastianbaginski: Does this part `imageArray >> "," >> (i % 32 == 0 ? "\n" : " ")` looks OK with valgrind ? I never know when a string like that is a static literal and I try to free it.

sebastianbaginski:
Looks good except the one part I'm not sure of.

source/draw/CreateAlphaByteString.ooc
marcusnaslund: > I never know when a string like that is a static literal

You and me both. Thanks for the comment, I will check it.

marcusnaslund: No invalid read/writes.

PR #958, opened 2015-11-30
Cleaned up Process* classes

marcusnaslund:
There are some possible leaks from iterators remaining, and the style isn't perfectly ours, but at least we get rid of 90 lines.
@sebastianbaginski ?

### DIFF
--- a/source/io/native/ProcessUnix.ooc
+++ b/source/io/native/ProcessUnix.ooc
@@ -20,7 +20,6 @@ signal: extern func (Int, Pointer)
 ProcessUnix: class extends Process {
 	init: func ~unix (=args)
 
-	/** terminate my child pid! */
 	terminate: func {
 		if (pid)
 			kill(pid, SIGTERM)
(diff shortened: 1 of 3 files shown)


source/io/native/ProcessUnix.ooc
```diff
@@ -71,40 +66,22 @@ ProcessUnix: class extends Process {
 			}
 
 			message = message + "\n"
-			if (stdErr) {
+			if (stdErr)
 				stdErr write(message)
-			} else {
+			else
 				stderr write(message)
```
sebastianbaginski: Could it work with the `??` operator:
```ooc
(stdErr ?? stderr) write(message)
```
?

sebastianbaginski: I'm just curious, you don't have to change it.

sebastianbaginski:
Looks ok.

source/io/native/ProcessUnix.ooc
marcusnaslund: Cool, it does! I won't change it though.

sebastianbaginski: Great, thanks for testing

PR #959, opened 2015-11-30
Removed duplicated code in Profiler

sebastianbaginski:
Code for creating output string was repeated twice.
Fixes some minor memory leaks.
@marcusnaslund

### DIFF
--- a/source/base/Profiler.ooc
+++ b/source/base/Profiler.ooc
@@ -33,11 +33,7 @@ Profiler: class {
 		this _timer stop()
 	}
 	printResults: static func {
-		This _profilers apply(func (profiler: This) {
-			outputString := profiler _name + " Time: " & ("%.3f" formatFloat(profiler _timer _result / 1000.0f)) >> " Average: " & ("%.3f" formatFloat(profiler _timer _average / 1000.0f))
-			Debug print(outputString)
-			outputString free()
-		})
+		This _logResults(func (s: String) { Debug print(s) })
 	}
 	free: override func {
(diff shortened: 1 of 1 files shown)


source/base/Profiler.ooc
```diff
@@ -67,4 +60,13 @@ Profiler: class {
 		This _profilers free()
 		This _profilers = null
 	}
+	_logResults: static func (logMethod: Func (String)) {
+		for (i in 0 .. This _profilers count) {
```
sebastianbaginski: I needed to use a loop because I got segmentation fault when trying to free the closure here.

source/base/Profiler.ooc
```diff
@@ -33,11 +33,7 @@ Profiler: class {
 		this _timer stop()
 	}
 	printResults: static func {
-		This _profilers apply(func (profiler: This) {
-			outputString := profiler _name + " Time: " & ("%.3f" formatFloat(profiler _timer _result / 1000.0f)) >> " Average: " & ("%.3f" formatFloat(profiler _timer _average / 1000.0f))
-			Debug print(outputString)
-			outputString free()
-		})
+		This _logResults(func (s: String) { Debug print(s) })
```
marcusnaslund: This closure will leak, right?

sebastianbaginski: It is `free`d in `_logResults`.

marcusnaslund: Right, thanks.

PR #968, opened 2015-11-30
Cleaned up mangling.ooc

marcusnaslund:
Reversed the if statement on new line 7 to avoid the return call.
We should discuss where this file should be placed, having a folder for only one file only used by one other file seems absurd to me.

Comments @sebastianbaginski ?


sebastianbaginski:
We should probably inline it in the file where it is used.

PR #969, opened 2015-12-01
Fixed bug in TextBuilder append method

sebastianbaginski:
Fixed bug in `append` method (data of other string builder was freed when appending). Fixed memory leaks and invalid reads in `TextBuilderTest`

Fixes https://github.com/cogneco/ooc-kean/issues/948

@marcusnaslund


test/base/TextBuilderTest.ooc
```diff
@@ -67,6 +74,9 @@ TextBuilderTest: class extends Fixture {
 			s := t take() toString()
 			expect(s == t)
 			s free()
+			sb free()
+			sb2 free()
+			sb3 free()
```
marcusnaslund: Something like `(s, sb, sb2, sb3) free()` would be so nice, but it doesn't work.

PR #970, opened 2015-12-01
Fixed memory leak in sdk/io/Writer.ooc

sebastianbaginski:
@marcusnaslund


marcusnaslund:
I don't know how much we care about "fixing" the io classes because I believe @simonmika thinks we should rewrite them from scratch later. But this doesn't hurt of course.

PR #976, opened 2015-12-01
Removed Bag.ooc and cleaned up other Hash classes

marcusnaslund:
We don't need `Bag.ooc`.

A future PR will hopefully be able to merge `HashBag.ooc` and `HashDictionary` so we only have `HashMap` left as a backend for our `HashDictionary`. I will also try to eliminate memory leaks.

Hint: Append `?w=1` when viewing the diff.


sebastianbaginski:
looks ok

PR #979, opened 2015-12-01
Corrected buffer ownership in RasterBgr open

sebastianbaginski:
@marcusnaslund this is the last one


marcusnaslund:
Will memcheck for the tests be a lot nicer now? I will try it later.

sebastianbaginski:
Yes, it should be better.

PR #980, opened 2015-12-01
Moved SynchronizedQueue to `concurrent` module

sebastianbaginski:
I think this releases the `collections -> sdk/threading` dependency.
@marcusnaslund


marcusnaslund:
Creates a `concurrent -> collections` dependancy, but maybe we have that already. We should do this anyhow.

marcusnaslund:
Oh right, of course we already have that.

PR #981, opened 2015-12-01
Moved Vector.ooc into SDK/structs

sebastianbaginski:
@marcusnaslund


test/math/FloatComplexTest.ooc
```diff
@@ -22,6 +22,7 @@ import math
 import structs/ArrayList
 import FloatComplex
 import lang/IO
+import structs/Vector
```
marcusnaslund: Does this need to be here at all?

sebastianbaginski: Yes. It won't compile otherwise.

marcusnaslund: Ah, the test uses a `HeapVector` for some reason.

PR #983, opened 2015-12-01
Removed HashBag.ooc

marcusnaslund:
Another one bites the dust.
@sebastianbaginski ?

Later, I want to rework then with `Text` as key instead of `String` (and remove related + other memory leaks). Right now, that's not possible.


test/collections/HashDictionaryTest.ooc
```diff
@@ -159,16 +159,16 @@ HashDictionaryTest: class extends Fixture {
...
 			expect(dictionary contains("Second"), is true)
 			dictionary remove("Second")
-			expect(dictionary count, is equal to(2))
+			expect(dictionary count as Int, is equal to(2))
 			expect(dictionary empty, is false)
 			expect(dictionary contains("Second"), is false)
 			dictionary remove("Third")
 			dictionary remove("First")
-			expect(dictionary count, is equal to(0))
+			expect(dictionary count as Int, is equal to(0))
```
sebastianbaginski: What happened here ? :)

marcusnaslund: `count` is now of type `SizeT`, which cannot be compared with `Int` easily since `rock` would rather cast it to `Float`........

sebastianbaginski: Nice, I need to remember this one.

marcusnaslund: Note: It worked well locally, but crashed on Circle.

PR #984, opened 2015-12-01
Changed name to CharBuffer and CharBufferIterator.

simonmika:
This makes it possible to use the name `Buffer` for other purposes.


simonmika:
@sebastianbaginski Can you peer review this?

PR #986, opened 2015-12-01
Improved Memory Management.

simonmika:
Created new API for working with memory.

### DIFF
--- a/source/sdk/Buffer.ooc
+++ b/source/sdk/Buffer.ooc
@@ -1,0 +1,84 @@
+/*
+* Copyright (C) 2015 - Simon Mika <simon@mika.se>
+*
+* This sofware is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License as published by the Free Software Foundation; either
+* version 2.1 of the License, or (at your option) any later version.
+*
+* This software is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


simonmika:
@marcusnaslund Can you peer rieview this?

source/sdk/Buffer.ooc
```diff
@@ -0,0 +1,84 @@
...
+	operator []= (start: Int, data: This) {
+		data copyTo(this[start])
+	}
+	operator []= (range: Range, data: This) {
+		data copyTo(this[range])
+	}
+	operator == (other: This) -> Bool {
+		(this _size == other _size) && (memcmp(this _pointer, other _pointer, this _size) == 0)
+	}
+	empty: static This { get { This new() } }
```
marcusnaslund: We generally keep all properties at the top of the class.

marcusnaslund:
Looks good. We might also want something to resize the `Buffer` though, as a replacement for `realloc`?

source/sdk/Buffer.ooc
simonmika: And all static stuff at the bottom.

simonmika:
Feel free to add a `resize` method.

simonmika:
@marcusnaslund Just because this can be better there are no need to stop this PR from merging. Don't let the perfect solution become the enemy of the good as long as the good leads to the perfect.

marcusnaslund:
No, but I *am* stopping this PR because it is failing tests.

simonmika:
Good.

marcusnaslund:
I will take it from here in a new PR.

PR #987, opened 2015-12-01
Fixes Image coordinateSystem

ziquanyu:
Fix "Image coordinateSystem", it's fine to use it now. We can change RasterImage or GpuImage and change will affect image after it's rendered.

Auto-test is passed on my computer. It's tested to change RasterImage and GpuImage, both ways work now.


ziquanyu:
@sebastianbaginski Can you review this when you have time, thanks!

source/draw/gpu/GpuSurface.ooc
```diff
@@ -62,9 +62,16 @@ GpuSurface: abstract class extends Canvas {
 		translation := this _toLocal * FloatTransform3D createTranslation(box leftTop x, box leftTop y, this focalLength) * this _toLocal
 		translation * toReference * FloatTransform3D createScaling(box size x / 2.0f, box size y / 2.0f, 1.0f)
 	}
-	_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D) -> FloatTransform3D {
+	_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D, flipVector: FloatVector3D) -> FloatTransform3D {
```
sebastianbaginski: Can't we pass the `image transform` here instead of `flipVector` ? Maybe it could be useful later ?

source/draw/gpu/GpuSurface.ooc
```diff
@@ -62,9 +62,16 @@ GpuSurface: abstract class extends Canvas {
...
 		scaling := FloatTransform3D createScaling(box size x as Float / imageSize x, box size y as Float / imageSize y, 1.0f)
 		translation := FloatTransform3D createTranslation(box leftTop x as Float / imageSize x, box leftTop y as Float / imageSize y, 0.0f)
+		if (!(flipVector x + 1.0f)) {
+			translation a = -1.0f
```
sebastianbaginski: What if `translation a` was 2.0f, 0.5f, etc... previously ?

source/draw/gpu/GpuSurface.ooc
```diff
@@ -62,9 +62,16 @@ GpuSurface: abstract class extends Canvas {
...
+			translation m = 1.0f
+		} else if (!(flipVector y + 1.0f)) {
```
sebastianbaginski: Do you mean `flipVector y equals(-1.0f)` ?

source/geometry/IntTransform2D.ooc
```diff
@@ -57,6 +58,7 @@ IntTransform2D: cover {
 	}
 	determinant ::= this a * this e * this i + this d * this h * this c + this g * this b * this f - this g * this e * this c - this d * this b * this i - this a * this h * this f
 	translation ::= IntVector2D new(this g, this h)
+	diagonal ::= IntVector3D new(this a, this e, this i)
```
sebastianbaginski: I don't know if this should be here, the class is 2D and returns something 3D. @marcusnaslund

ziquanyu: I don't know either. If we just pass the transform, we don't really need this.

source/draw/gpu/GpuSurface.ooc
ziquanyu: yes

source/draw/gpu/GpuSurface.ooc
ziquanyu: pass the transform is an alternative.

source/geometry/IntTransform2D.ooc
marcusnaslund: If we don't need it, let's not introduce this.

source/draw/gpu/GpuSurface.ooc
ziquanyu: I'm not really sure about the details of this part. But as I've tested, this translation has always identity matrix. Also, I change size and position. I got correct result. the translation is more like initial transform and be applied to vertex shader. Do you get what I mean?

marcusnaslund: I see @sebastianbaginski 's point. `!(flipVector x + 1.0f)` does not look like a nice boolean statement.

ziquanyu: if we change it to transform, we don't have "equals"

ziquanyu: we can still use it, i will change it anyway.

source/draw/gpu/GpuSurface.ooc
ziquanyu: we can still use it, i will change it anyway.

source/draw/gpu/GpuSurface.ooc
sebastianbaginski: OK, so @ziquanyu explained to me that it is always an identity anyway.
@ziquanyu : can you do it like `translation a = - translation a` ? I think it will be more clear that you want to flip things.

ziquanyu: yes of couse.

source/geometry/IntTransform2D.ooc
```diff
@@ -16,6 +16,7 @@
 
 import math
 import IntVector2D
+import IntVector3D
```
sebastianbaginski: I think you can remove this `import` now.

test/draw/gpu/ImageCoordinateSystemTest.ooc
```diff
@@ -0,0 +1,77 @@
...
+		this add("change Raster coordinateSystem YDownward + XRightward + Default", func {
+			correctImage := RasterBgra open("test/draw/gpu/input/Hercules.png")
+			sourceImage coordinateSystem = CoordinateSystem YDownward
+			gpuImage := gpuContext createBgra(sourceImage size)
+			gpuImage canvas clear()
+			gpuImage canvas draw(sourceImage)
+			rasterFromGpu := gpuImage toRaster()
+			expect(rasterFromGpu distance(correctImage), is equal to(0.0f))
+		})
+
```
marcusnaslund: While you're at it, skip this empty line.

ziquanyu:
@sebastianbaginski everything else is okay now?

source/draw/gpu/GpuSurface.ooc
```diff
@@ -62,9 +62,16 @@ GpuSurface: abstract class extends Canvas {
...
 		translation * toReference * FloatTransform3D createScaling(box size x / 2.0f, box size y / 2.0f, 1.0f)
 	}
-	_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D) -> FloatTransform3D {
+	_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D, flipVector: FloatVector2D) -> FloatTransform3D {
 		scaling := FloatTransform3D createScaling(box size x as Float / imageSize x, box size y as Float / imageSize y, 1.0f)
 		translation := FloatTransform3D createTranslation(box leftTop x as Float / imageSize x, box leftTop y as Float / imageSize y, 0.0f)
+		if (flipVector x equals(-1.0f)) {
+			translation a = - translation a
+			translation m = 1.0f
+		} else if (flipVector y equals(-1.0f)) {
```
sebastianbaginski: Can both `flipVector x` and `flipVector y` be `-1.0f` at the same time ?

source/draw/gpu/GpuSurface.ooc
```diff
@@ -62,9 +62,16 @@ GpuSurface: abstract class extends Canvas {
...
+			translation f = - translation f
+			translation n = 1.0f
```
sebastianbaginski: If `translation` is identity, do you have to set `translation n = 1.0f` explicitly ?

sebastianbaginski:
I think the two above are my final comments.

source/draw/gpu/GpuSurface.ooc
ziquanyu: yes, then you flip both x and y

sebastianbaginski: Ok, so I think you need to remove the `else`, because if `x == -1.0f` then the `else if` is never checked.

source/draw/gpu/GpuSurface.ooc
ziquanyu: yes, identity matrix has only diagonal elements are ones. that n is not at diagonal part.

source/draw/gpu/GpuSurface.ooc
ziquanyu: or maybe i can remove if instead.

sebastianbaginski:
OK, if this does what you want then I think it's ready.

test/draw/gpu/ImageCoordinateSystemTest.ooc
```diff
@@ -0,0 +1,76 @@
...
+		super("ImageCoordinateSystemTest")
+		sourceImage := RasterBgra open("test/draw/gpu/input/Hercules.png")
+		this add("change Raster coordinateSystem XLeftward", func {
+			correctImage := RasterBgra open("test/draw/gpu/correct/XLeftward.png")
+			sourceImage coordinateSystem = CoordinateSystem XLeftward
+			gpuImage := gpuContext createBgra(sourceImage size)
+			gpuImage canvas clear()
+			gpuImage canvas draw(sourceImage)
+			rasterFromGpu := gpuImage toRaster()
+			expect(rasterFromGpu distance(correctImage), is equal to(0.0f))
```
marcusnaslund: Add `within(tolerance)` with a very small `tolerance`.

ziquanyu: okay

ziquanyu:
@marcusnaslund Is there anything else I need to change?

source/draw/gpu/GpuSurface.ooc
```diff
@@ -62,9 +62,17 @@ GpuSurface: abstract class extends Canvas {
...
-	_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D) -> FloatTransform3D {
+	_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D, flipVector: FloatVector2D) -> FloatTransform3D {
 		scaling := FloatTransform3D createScaling(box size x as Float / imageSize x, box size y as Float / imageSize y, 1.0f)
 		translation := FloatTransform3D createTranslation(box leftTop x as Float / imageSize x, box leftTop y as Float / imageSize y, 0.0f)
+		if (flipVector x equals(-1.0f)) {
+			translation a = - translation a
+			translation m = 1.0f
+		}
+		if (flipVector y equals(-1.0f)) {
+			translation f = - translation f
```
marcusnaslund: Remove the space after `-` here and on line 69, and I'm happy.

ziquanyu:
@marcusnaslund  It's done this time.

marcusnaslund:
Oops. This breaks the Android build.

PR #990, opened 2015-12-02
Add IntVectorList init~withValue(). Compare to FloatVectorList.

bettinaselig:
Add `IntVectorList init~withValue()`, which corresponds to  `FloatVectorList init~withValue()`

### DIFF
--- a/source/math/IntVectorList.ooc
+++ b/source/math/IntVectorList.ooc
@@ -29,6 +29,11 @@ IntVectorList: class extends VectorList<Int> {
 		super(other _vector)
 		this _count = other count
 	}
+	init: func ~withValue (capacity, value: Int) {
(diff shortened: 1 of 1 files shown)


bettinaselig:
@marcusnaslund Peer review?

source/math/IntVectorList.ooc
```diff
@@ -29,6 +29,11 @@ IntVectorList: class extends VectorList<Int> {
 		super(other _vector)
 		this _count = other count
 	}
+	init: func ~withValue (capacity: Int, value: Int) {
```
marcusnaslund: `redundant type information, reduce to (capacity, value: Int)`

bettinaselig:
Updated.

PR #992, opened 2015-12-02
Cleaned up fancy backtrace loading

marcusnaslund:
I really dig that `??` operator.

The third case is not needed since we don't compile the fancy backtrace that way, and this helps us to move out `sdk/io`.

@sebastianbaginski ?


sebastianbaginski:
Alright, if you say if it's not needed.

PR #993, opened 2015-12-02
Fixed recycling of EGLBgra images.

emilwestergren:
* When `EGLBgra` images are created from the buffer of a `GraphicBufferYuv420Semiplanar`, they are kept as a member for the instance for future use. 
* When `GraphicBufferYuv420Semiplanar` is freed, the `EGLBgra` is recycled.
* When creating a new `GraphicBufferYuv420Semiplanar`, if the `GraphicBuffer` matches a recycled `EGLBgra` image, the `EGLBgra` can be reused.
* Extreme performance increase

@sebastianbaginski peer

### DIFF
--- a/source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
+++ b/source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
@@ -19,8 +19,10 @@ import math
 use ooc-draw
 use ooc-geometry
 use ooc-base
+use ooc-collections
 use ooc-draw-gpu
 import GraphicBuffer, AndroidContext, EGLBgra
+import threading/Mutex
 
 version(!gpuOff) {
 GraphicBufferYuv420Semiplanar: class extends RasterYuv420Semiplanar {
@@ -31,6 +33,9 @@ GraphicBufferYuv420Semiplanar: class extends RasterYuv420Semiplanar {
 	_uvOffset: Int
 	uvOffset ::= this _uvOffset
 	uvPadding ::= (this _uvOffset - this _stride * this _size y)
+	_rgba: EGLBgra = null
+	_bin := static VectorList<EGLBgra> new()
+	_mutex := static Mutex new()
 	init: func ~fromBuffer (=_buffer, size: IntVector2D, =_stride, =_uvOffset) {
 		pointer := _buffer lock()
 		_buffer unlock()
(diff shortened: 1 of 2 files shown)


source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
```diff
@@ -31,6 +33,9 @@ GraphicBufferYuv420Semiplanar: class extends RasterYuv420Semiplanar {
 	_uvOffset: Int
 	uvOffset ::= this _uvOffset
 	uvPadding ::= (this _uvOffset - this _stride * this _size y)
+	_rgba: EGLBgra = null
+	_bin := static VectorList<EGLBgra> new()
```
emilwestergren: Will be replaced with `SynchronizedVectorList`.

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
```diff
@@ -38,21 +43,45 @@ GraphicBufferYuv420Semiplanar: class extends RasterYuv420Semiplanar {
...
+			}
+		}
+		This _mutex unlock()
 		result
 	}
-	kean_draw_graphicBufferYuv420Semiplanar_new: unmangled static func (buffer: GraphicBuffer, size: IntVector2D, stride, uvOffset: Int) -> This {
-		This new(buffer, size, stride, uvOffset)
+	clean: static func {
+		for (i in 0 .. This _bin count)
+			This _bin remove(i) free()
```
sebastianbaginski: No need to lock here because it is called at exit ?

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
```diff
@@ -38,21 +43,45 @@ GraphicBufferYuv420Semiplanar: class extends RasterYuv420Semiplanar {
...
+		this _rgba
+	}
+	_search: static func (buffer: GraphicBuffer) -> EGLBgra {
+		This _mutex lock()
+		result: EGLBgra = null
+		for (i in 0 .. This _bin count) {
+			image := This _bin[i]
+			if (image buffer _handle == buffer _handle) {
+				result = image
+				break
```
sebastianbaginski: I've seen this code somewhere, could be nice to have a generic locked cache implementation. Not this PR of course :)

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
emilwestergren: It's not guaranteed to be thread safe so I'll add lock.

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
emilwestergren: Yes, similar logic is duplicated in many places.

sebastianbaginski:
Looks good.

source/draw/gpu/opengl/GraphicBufferYuv420Semiplanar.ooc
emilwestergren: Implementing one right now.

PR #996, opened 2015-12-02
Added generic RecycleBin

emilwestergren:
Added generic RecycleBin for recycling objects.

### DIFF
--- a/source/collections/RecycleBin.ooc
+++ b/source/collections/RecycleBin.ooc
@@ -1,0 +1,58 @@
+/*
+* Copyright (C) 2015 - Simon Mika <simon@mika.se>
+*
+* This sofware is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License as published by the Free Software Foundation; either
+* version 2.1 of the License, or (at your option) any later version.
+*
+* This software is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with This software. If not, see <http://www.gnu.org/licenses/>.
+*/
+import threading/Mutex
+
+RecycleBin: class <T> {
+	_mutex := Mutex new()
+	_free: Func (T)
+	//TODO: Replace with SynchronizedVectorList
+	_list: VectorList<T>
+	_size: Int
+	init: func (=_size, =_free) { this _list = VectorList<T> new(_size) }
+	free: override func {
+		this clear()
+		this _list free()
+		this _mutex free()
+		super()
+	}
+	add: func (object: T) {
+		this _mutex lock()
+		if (this _list count == this _size)
+			this _free(this _list remove(0))
+		this _list add(object)
+		this _mutex unlock()
+	}
+	search: func (matches: Func (T) -> Bool) -> T {
+		result: T = null
+		this _mutex lock()
+		for (i in 0 .. this _list count) {
+			if (matches(this _list[i])) {
+				result = this _list remove(i)
+				break
+			}
+		}
+		this _mutex unlock()
+		(matches as Closure) free()
+		result
+	}
+	clear: func {
+		this _mutex lock()
+		for (i in 0 .. this _list count)
+			this _free(this _list remove())
+		this _mutex unlock()
+	}
+}
--- a/source/draw/gpu/opengl/_RecycleBin.ooc
+++ b/source/draw/gpu/opengl/_RecycleBin.ooc
@@ -23,7 +23,7 @@ import OpenGLPacked, OpenGLMonochrome, OpenGLBgra, OpenGLBgr, OpenGLUv
 import threading/Mutex
 
 version(!gpuOff) {
-RecycleBin: class {
+_RecycleBin: class {
 	_monochrome := VectorList<OpenGLMonochrome> new()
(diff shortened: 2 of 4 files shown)


emilwestergren:
@sebastianbaginski something like this? (don't peer review yet)

sebastianbaginski:
Yes, something like this. I was also thinking about having a possibility to set maximum number of elements in cache, and throw out old ones automatically. But maybe in separate class, which could reuse this one.

emilwestergren:
Yes I'll add that.

emilwestergren:
One problem with that is that you usually recycle stuff inside an object's `free` function. In that case how do you free the object from inside this class?

emilwestergren:
Maybe you could to supply a `free` closure that will run on an object when the bin is full.

sebastianbaginski:
Yes, a closure should work.

source/draw/gpu/opengl/_RecycleBin.ooc
```diff
@@ -23,7 +23,7 @@ import OpenGLPacked, OpenGLMonochrome, OpenGLBgra, OpenGLBgr, OpenGLUv
 import threading/Mutex
 
 version(!gpuOff) {
-RecycleBin: class {
+_RecycleBin: class {
```
emilwestergren: This will eventually be replaced by the generic `RecycleBin` class so I temporarily renamed it to a private class.

emilwestergren:
@marcusnaslund can you write tests for this?

marcusnaslund:
No problem.

emilwestergren:
Updated.

emilwestergren:
I'm not sure if it should be located in `ooc-collections` so correct it if it's wrong.

emilwestergren:
Updated.

source/collections/RecycleBin.ooc
```diff
@@ -0,0 +1,61 @@
...
+RecycleBin: class <T> {
+	_mutex := Mutex new()
+	_free: Func (T)
+	//TODO: Replace with SynchronizedVectorList
+	_list := VectorList<T> new()
+	_maxCount: Int
+	init: func (=_maxCount, =_free)
+	free: override func {
+		this _mutex lock()
+		this clear()
```
sebastianbaginski: A deadlock :(

emilwestergren: Ops

emilwestergren:
Updated.

source/collections/RecycleBin.ooc
```diff
@@ -0,0 +1,59 @@
...
+		super()
+	}
+	add: func (object: T) {
+		this _mutex lock()
+		this _list add(object)
+		if (this _list count > this _maxCount)
+			this _free(this _list remove(0))
+		this _mutex unlock()
+	}
+	search: func (matches: Func (T) -> Bool) -> T {
```
sebastianbaginski: What's the status of `generics + covers` ? Can we keep covers in this bin too ?
Last time I checked passing covers as arguments to closures had bugs.

emilwestergren: I don't know.

sebastianbaginski: Nevermind, this class will be useful anyway.

marcusnaslund: The latest rock version fixed some of these bugs, but I don't know either.

emilwestergren: I guess we'll find out when that happens, I don't think it will be very common for us to recycle covers anyway.

sebastianbaginski:
What's with this one ? @marcusnaslund

marcusnaslund:
I don't know, noone has flagged it for review. Also I haven't had time to write tests yet. If you can, please do.

sebastianbaginski:
Sorry I have other things now, just curious.

marcusnaslund:
@emilwestergren - move this up for review if you think it's ready. I will make a separate PR with tests and merge them together.

emilwestergren:
OK, I have barely tested it though, that's why I was waiting for real tests before merge, but I guess it's up to you.

emilwestergren:
Also, this class will be much cleaner when `SynchronizedVectorList` is implemented so maybe that should be the priority before this one.

source/collections/RecycleBin.ooc
```diff
@@ -0,0 +1,59 @@
...
+* This software is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with This software. If not, see <http://www.gnu.org/licenses/>.
+*/
+import threading/Mutex
+import VectorList
```
marcusnaslund: You should rebase, see that rock won't compile, and remove this line. :)

emilwestergren:
Updated.

marcusnaslund:
I see you've added tests, nice. Can we bump this up to `final review`?

source/collections/RecycleBin.ooc
```diff
@@ -0,0 +1,58 @@
...
+	free: override func {
+		this clear()
+		this _list free()
+		this _mutex free()
+		super()
+	}
+	add: func (object: T) {
+		this _mutex lock()
+		this _list add(object)
+		if (this _list count > this _maxCount)
```
marcusnaslund: You should probably do this before you `add`, to avoid a possible unnecessary reallocation.

marcusnaslund: Related to that, you could create `_list` in the constructor with `_maxCount` as an argument.

emilwestergren:
Updated.

PR #997, opened 2015-12-02
Added implicit toText() for Int, UInt, Float, Double

### LINKED ISSUE #681: Implicit toString() and toText()
Currently, we do stuff like
```ooc
// x, y, z are Int, Float or something other than a String
println("Hello world " + x toString() + " " + y toString() +  " " + z toString())
```
Which is rather messy. 
```ooc
println("Hello world " + x + " " + y + " " + z)
```
is a lot neater, but it doesn't work.

Is it possible to override operators for `String` so that the right-hand argument is automatically converted to `String` if it isn't one already? Are there problems that could arise from such overloads?

EDIT: This applies to `toText`, too.
EDIT 2: Actually, we currently do things like this:
```ooc
// x, y, z are Int, Float or something other than a String
(xString, yString, zString) := (x toString(), y toString(), z toString())
println("Hello world " << xString >> " " & yString >>  " " & zString)
```

marcusnaslund:
Fixes #681 (no point in adding implicit `toString` anymore)

Creates invalid memory read/writes right now because of bugs in `Text print()`, so waiting on that.

### DIFF
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -53,6 +53,30 @@ Text: cover {
 	operator != (other: String) -> Bool { !(this == other) }
 	operator != (other: This) -> Bool { !(this == other) }
 	operator + (other: This) -> This { This new(this _buffer + other _buffer) }
+	operator + (other: Int) -> This {
+		converted := other toString()
+		result := this + This new(converted)
+		converted free()
+		result
+	}
+	operator + (other: UInt) -> This {
+		converted := other toString()
+		result := this + This new(converted)
+		converted free()
+		result
+	}
+	operator + (other: Float) -> This {
+		converted := other toString()
+		result := this + This new(converted)
+		converted free()
+		result
+	}
+	operator + (other: Double) -> This {
+		converted := other toString()
+		result := this + This new(converted)
+		converted free()
+		result
+	}
 	beginsWith: func (other: This) -> Bool { this slice(0, Int minimum(other count, this count)) == other }
 	beginsWith: func ~string (other: String) -> Bool { this beginsWith(This new(other)) }
 	beginsWith: func ~character (character: Char) -> Bool {
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -53,18 +53,10 @@ Text: cover {
 	operator != (other: String) -> Bool { !(this == other) }
 	operator != (other: This) -> Bool { !(this == other) }
 	operator + (other: This) -> This { This new(this _buffer + other _buffer) }
-	operator + (other: Int) -> This {
-		this + other toText() take()
-	}
-	operator + (other: UInt) -> This {
-		this + other toText()
-	}
-	operator + (other: Float) -> This {
-		this + other toText() take()
-	}
-	operator + (other: Double) -> This {
-		this + other toText()
-	}
+	operator + (other: Int) -> This { this + other toText() }
+	operator + (other: UInt) -> This { this + other toText() }
+	operator + (other: Float) -> This { this + other toText() }
+	operator + (other: Double) -> This { this + other toText() }
 	beginsWith: func (other: This) -> Bool { this slice(0, Int minimum(other count, this count)) == other }
 	beginsWith: func ~string (other: String) -> Bool { this beginsWith(This new(other)) }
 	beginsWith: func ~character (character: Char) -> Bool {
@@ -335,7 +327,10 @@ extend Int {
 
 extend UInt {
 	toText: func -> Text {
-		Text new(this toString())
+		string := this toString()
+		result := Text new(string) copy()
+		string free()
+		result
 	}
 }
 
@@ -350,6 +345,9 @@ extend Float {
 
 extend Double {
 	toText: func -> Text {
-		Text new(this toString())
+		string := this toString()
+		result := Text new(string) copy()
+		string free()
+		result
 	}
 }
--- a/source/base/Text.ooc
+++ b/source/base/Text.ooc
@@ -54,28 +54,16 @@ Text: cover {
 	operator != (other: This) -> Bool { !(this == other) }
 	operator + (other: This) -> This { This new(this _buffer + other _buffer) }
 	operator + (other: Int) -> This {
-		converted := other toString()
-		result := this + This new(converted)
-		converted free()
-		result
+		this + other toText() take()
 	}
 	operator + (other: UInt) -> This {
-		converted := other toString()
-		result := this + This new(converted)
-		converted free()
-		result
+		this + other toText()
 	}
 	operator + (other: Float) -> This {
-		converted := other toString()
-		result := this + This new(converted)
-		converted free()
-		result
+		this + other toText() take()
 	}
 	operator + (other: Double) -> This {
-		converted := other toString()
-		result := this + This new(converted)
-		converted free()
-		result
+		this + other toText()
 	}
 	beginsWith: func (other: This) -> Bool { this slice(0, Int minimum(other count, this count)) == other }
 	beginsWith: func ~string (other: String) -> Bool { this beginsWith(This new(other)) }
@@ -335,3 +323,33 @@ makeTextLiteral: func (str: CString, strLen: Int) -> Text { Text new(str, strLen
 operator == (left: String, right: Text) -> Bool { Text new(left) == right }
 operator != (left: String, right: Text) -> Bool { !(left == right) }
 operator []= (left: TextBuffer, range: Range, right: Text) { right _buffer copyTo(left slice(range min, range count)) }
+
+extend Int {
+	toText: func -> Text {
+		string := this toString()
+		result := Text new(string) copy()
+		string free()
+		result
+	}
+}
+
+extend UInt {
+	toText: func -> Text {
+		Text new(this toString())
+	}
+}
+
+extend Float {
+	toText: func -> Text {
+		string := this toString()
+		result := Text new(string) copy()
+		string free()
+		result
+	}
+}
+
+extend Double {
+	toText: func -> Text {
+		Text new(this toString())
+	}
+}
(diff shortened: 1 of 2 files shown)


marcusnaslund:
peer review @davidhesselbom ?

test/base/TextTest.ooc
```diff
@@ -226,6 +226,14 @@ TextTest: class extends Fixture {
 			text = t"%s %s %i %i" format("test", "number", 0, 1)
 			expect(text == t"test number 0 1")
 		})
+		this add("implicit toText", func {
+			one := t"123" + 456 + 7.89f
+			expect(one == t"1234567.89")
+			two := Text new("12") + 3.45f + 6.78 + 9
```
davidhesselbom: That's odd, I always thought evaluation happened from right to left (for reasons I don't remember).

source/base/Text.ooc
```diff
@@ -53,6 +53,30 @@ Text: cover {
...
+		result := this + This new(converted)
+		converted free()
+		result
+	}
+	operator + (other: Double) -> This {
+		converted := other toString()
+		result := this + This new(converted)
+		converted free()
+		result
+	}
```
davidhesselbom: Imagine how many lines an interface could have saved us in a situation like this...

source/base/Text.ooc
```diff
@@ -53,6 +53,30 @@ Text: cover {
 	operator != (other: String) -> Bool { !(this == other) }
 	operator != (other: This) -> Bool { !(this == other) }
 	operator + (other: This) -> This { This new(this _buffer + other _buffer) }
+	operator + (other: Int) -> This {
+		converted := other toString()
```
davidhesselbom: Why don't you add `toText` functions for these 4 data types, first?

marcusnaslund: Dependancy issues, but we're getting there.

marcusnaslund:
Anything else @davidhesselbom ? `toText` is coming to the numeric types, but not any day soon.

source/base/Text.ooc
davidhesselbom: ... dependency on what? It seems like a logical first thing to add so you don't have to go back and modify this code again later and change `toString` to `toText`. What am I missing?

davidhesselbom:
Other than that, no, I have nothing to add.

source/base/Text.ooc
marcusnaslund: It's the same reason we can't have `toText` in `Cell` right now. These files are in the sdk, and cannot depend on things outside the sdk. Once we've got `Text` inside the sdk, it's another story.

davidhesselbom: I added this in `Text.ooc`:
```ooc
extend Int {
	toText: func -> Text {
		Text new(this toString())
	}
}
```
Now this program works:
```ooc
use ooc-base

t := 5 toText()
t toString() println()
```

marcusnaslund: Damn, we have extensions. Thanks.

davidhesselbom: ... and maybe using extensions isn't how you should do things, but keep in mind that this what `math.ooc` does.

It's a different story for `Cell`, because it's a generic class, and extending them doesn't work very well due to bugs in rock. Extending basic covers like `Int` is the *only* way you can define new functions on them, though, and there's not necessarily any harm in doing in it several places, I don't think. For instance, you could put `toText` functions for all of the numeric types in a single file named `toTextExtensions.ooc` or something, add that file to `ooc-base.use`, and (here's the bad part) add `use ooc-base` anywhere you need to call `toText` on a numeric type.

marcusnaslund: Needing to add `ooc-base` to call `toText` is not so surprising, since we're getting a type (`Text`) that is in base anyhow. But I see your point, I will change to use extensions instead.

davidhesselbom: Right, but if `Text` was in the SDK, you wouldn't need to `use` anything, as is the case with `String`. For what that's worth.

marcusnaslund: Huh, actually, this does not work for me:
```ooc
extend Float {
	toText: func -> Text {
		converted := this toString()
		result := Text new(converted)
		converted free()
		result
	}
}

operator + (other: Float) -> This {
	this + other toText()
}
```
I get all sorts of weird characters in the resulting text instead. If I don't free `converted`, it works, but then it of course leaks memory. I don't get the same problem for `Int`.

marcusnaslund: Actually, it seems that `TextBuffer` just takes the pointer to the string. So should it actually be freed?

davidhesselbom: I think you may need to use an argument in the `free` function, but I still don't really understand `OwnedBuffer`...

marcusnaslund: There is no argument to the `String free()` method.

davidhesselbom: Maybe it would be a good idea to write a short program that converts between `String` and `Text` (importantly, both literals and non-literals) that does not leak memory and that does not result in invalid reads/writes (which I'm sure a memtest will prove is the reason for the "weird characters" in your program).

marcusnaslund: The thing is, I had no invalid read/writes when I did this without extensions..

davidhesselbom: That's... odd.

davidhesselbom: Did you actually run memtest that time, or were you just lucky, and the memory your pointer pointed to still contained the correct data, even if it was, in fact, invalid?

marcusnaslund: I was running memtest, because I had some problems earlier with `print` and `println` which @sebastianbaginski has now fixed.

davidhesselbom: Okay. I have no reason to believe the problem is caused by the extension itself (not because I trust rock, but because I have no idea how an extension could cause the problem).

marcusnaslund: I will dive deeper into the mystery when I have time.

marcusnaslund:
Ah, now I get it. When not using an extension, the concatenated `Text` is created *before* the `String` is freed. 
With the extension, I free the `String` first, and then return a `Text` with its data, which is then concatenated. I have to free the `String` object without freeing its data.

sebastianbaginski:
You can also simply copy the text `result := Text new(converted) copy()`

marcusnaslund:
I'm not sure how you mean @sebastianbaginski ?

sebastianbaginski:
```ooc
extend Float {
    toText: func -> Text {
        converted := this toString()
        result := Text new(converted) copy()
        converted free()
        result
    }
}
```
because Text constructor from string does not give the ownership of data to the text object.

marcusnaslund:
Thanks. That works... but it leaks a few bytes of memory

sebastianbaginski:
OK, I can check it out if you want.

marcusnaslund:
Updated. Seems to work well now.

marcusnaslund:

test/base/TextTest.ooc
```diff
@@ -226,6 +226,17 @@ TextTest: class extends Fixture {
 			text = t"%s %s %i %i" format("test", "number", 0, 1)
 			expect(text == t"test number 0 1")
 		})
+		this add("implicit toText", func {
+			one := t"123" + 456 + 7.89f
```
sebastianbaginski: Do you like this api ? I was never a fan of such implicit conversions.

sebastianbaginski:
Looks correct.

test/base/TextTest.ooc
marcusnaslund: Except `one := (t"123" toFloat() + 456 + 7.89f) toText()` would give you something like `one := t"586.89"`...
I do like to have the option of writing like this, instead of a myriad of `toText()`/`toString()` calls.

sebastianbaginski: That was just an example, integers and text could come from a function :)

marcusnaslund: To be fair, one could say the same thing about `value := 2.31f + 7`: Is it a `Float` or `Int`? For those situations we have the rule that `Float` plus anything is always a `Float`, if possible. What we say now is a `Text` plus anything is a `Text`, if possible.

sebastianbaginski: Yes but this is a standard in (almost) any programming language to be able to add floats to integers and `value = 2.5 + 6` won't surprise anyone. Certainly it is not a standard to add string objects and numbers with `operator +` like `value = "8" + 7`. What should be the result ? string `"87"` or  a number `15` ? There is no standard behaviour and this is a possible source of confusion for newcommers, becase I know people who would prefer the second option (and I've seen people implementing and using it, despite the ambiguity).
Of course I'm not asking you to skip this PR.

davidhesselbom: Here's one of the reasons I suggested implicit conversions in the first place:
```ooc
	line: static func ~dashed (start, end: FloatPoint2D, lineWidth: Float, opacity: Float, color: String, dashFloatPair: FloatPoint2D) -> String {
		"<line x1='" << start x toString() >> "' y1='" & start y toString() >> "' x2='" & end x toString() >> "' y2='" & end y toString() >> "' stroke-width='" & lineWidth toString() >> "' stroke='" >> color >> "' stroke-opacity='" & opacity toString() >> "' stroke-dasharray='" & dashFloatPair toString() >> "'/>\n"
	}

```
@sebastianbaginski I think the implicit conversion it's clear enough, but I also think you have a point. Another way is to use a varargs concatenation function that accepts arguments of any type and builds a longer string from them. In C#,
```cs
return String.Concat(new object[] { "<line x1='", start.x, "' y1="', start.y, '" x2="' /* etc */ });
```
Unfortunately, varargs in ooc give no compile-time guarantee that the objects passed to such a function actually have a `toString()` function, and neither would passing any collection of `T`. Working interfaces would be a godsend here.

PR #1005, opened 2015-12-02
Changed the imaginary part of quaternions from point to vector. 

niklasborwell:
Part of #783.

### DIFF
--- a/source/geometry/FloatEuclidTransform.ooc
+++ b/source/geometry/FloatEuclidTransform.ooc
@@ -17,7 +17,7 @@
 use ooc-collections
 import math
 use ooc-math
-import FloatPoint3D
+import FloatVector3D
 import FloatRotation3D
 import FloatTransform2D
 import FloatTransform3D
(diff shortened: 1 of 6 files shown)


marcusnaslund:
I really don't want to carry on this discussion of point vs vector, but, really? Have we discussed this specific thing in that issue?

niklasborwell:
Haha well, we definitely don't want to resurrect the whole discussion, so maybe we can just decide quietly. This specific case, as opposed to the translation part of Euclid transforms, was never discussed. 
Simon asked me to replace parts in the code that I think should be vectors instead of points, and according to wikipedia: 
>A number of the form a + 0i + 0j + 0k, where a is a real number, is called real, and a number of the form 0 + bi + cj + dk, where b, c, and d are real numbers, and at least one of b, c or d is nonzero, is called pure imaginary. If a + bi + cj + dk is any quaternion, then a is called its scalar part and bi + cj + dk is called its *vector part*. The scalar part of a quaternion is always real, and the vector part is always pure imaginary. Even though every quaternion can be viewed as a vector in a four-dimensional vector space, it is common to define a vector to mean a pure imaginary quaternion. With this convention, a vector is the same as an element of the vector space R3.

So it seems to me that the imaginary should be represented with a vector. What do you think?

marcusnaslund:
You're right, the imaginary part of the product of two pure imaginary quaternions is isomorphic to the cross product of two vectors with equal values. That's not a perfect argument in my opinion, as I could argue that a quaternion is simply a point in a 4D space. Then again, one usually thinks of `i`, `j` and `k` in linear algebra as vectors (and their names coincide with the imaginary parts of the quaternion for a reason) so the sum of them should also be.

...so after thinking about it I'm not opposed to this anymore. If you want to do this, let's go for it.

niklasborwell:
Well, I never really cared that much for this overall change to begin with, but since it was decided...I usually think of quaternions like it's own representation simply with 4 floats, but I thought a vector was more correct than a point in this case. 
But now I'm not so sure anymore. I mean, the point is actually also a vector, but a position vector. And the so called vector is supposed to be a displacement vector. So maybe this should actually be a point? I got confused by the names, and now I'm not sure anymore.

niklasborwell:
I've thought about it a bit. Unless you see a good reason I think we should skip this. What do you think?

marcusnaslund:
I like how we convinced each other of the other's opinion. I can see the idea but I don't see the benefit of changing stuff everywhere unless we all agree.

niklasborwell:
I will close this and keep the branch then. We can resurrect the pull reqeust if someone feels like it should be changed.

PR #1006, opened 2015-12-03
Removed hard coded exception function name from Array.h

sebastianbaginski:
This gets rid of hard-coded mangled function name from Array.h and use our friendly name instead. Now will be easier to move Exception.ooc to another folder.


sebastianbaginski:
@marcusnaslund review ?

PR #1014, opened 2015-12-03
Corrected ColorMonochrome constructor with 32 and 64 bit unsigned parameters

sebastianbaginski:
Test case added.
@marcusnaslund 

Bug was caused by rock picking the `Float` version of constructor for unsigned parameters.

### DIFF
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -25,6 +25,10 @@ ColorMonochrome: cover {
 	normalized ::= this y as Float / 255
 	init: func@ (=y)
 	init: func@ ~default { this init(0) }
+	init: func@ ~uint (i: UInt) { this init(i as UInt8) }
+	init: func@ ~uint16 (i: UInt16) { this init(i as UInt8) }
+	init: func@ ~uint32 (i: UInt32) { this init(i as UInt8) }
+	init: func@ ~uint64 (i: UInt64) { this init(i as UInt8) }
 	init: func@ ~int (i: Int) { this init(i as UInt8) }
 	init: func@ ~float (f: Float) { this init(f*255.0f clamp(0.0f, 255.0f) as UInt8) }
 	init: func@ ~double (d: Double) { this init(d*255.0f clamp(0.0f, 255.0f) as UInt8) }
--- a/test/draw/RasterMonochromeTest.ooc
+++ b/test/draw/RasterMonochromeTest.ooc
@@ -98,6 +98,19 @@ RasterMonochromeTest: class extends Fixture {
 			image referenceCount decrease()
 			image2 referenceCount decrease()
 		})
+		this add("color monochrome from unsigned types", func {
+			image := RasterMonochrome new(IntVector2D new(10, 10))
+			value: UInt = 128
+			value32: UInt32 = 128
+			value64: UInt64 = 128
+			image[0, 0] = ColorMonochrome new(value32)
+			image[0, 1] = ColorMonochrome new(value64)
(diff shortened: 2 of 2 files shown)


test/draw/RasterMonochromeTest.ooc
```diff
@@ -98,6 +98,16 @@ RasterMonochromeTest: class extends Fixture {
 			image referenceCount decrease()
 			image2 referenceCount decrease()
 		})
+		this add("color monochrome from unsigned types", func {
+			image := RasterMonochrome new(IntVector2D new(10, 10))
+			value32: UInt32 = 128
+			value64: UInt64 = 128
+			image[0, 0] = ColorMonochrome new(value32)
+			image[0, 1] = ColorMonochrome new(value64)
+			expect(image[0, 0] y as Int, is equal to(value32 as Int))
```
sebastianbaginski: Fixture does not support unsigned. Remove `as Int` and expect will fail "expected  equal to  '128', was '128.00000000'"

marcusnaslund: We should probably add that since this happens in other places. Will you or shall I?

sebastianbaginski: Do that if you have time. Otherwise I can.

marcusnaslund: I'm on it.

source/draw/Color.ooc
```diff
@@ -25,6 +25,9 @@ ColorMonochrome: cover {
 	normalized ::= this y as Float / 255
 	init: func@ (=y)
 	init: func@ ~default { this init(0) }
+	init: func@ ~uint16 (i: UInt16) { this init(i as UInt8) }
+	init: func@ ~uint32 (i: UInt32) { this init(i as UInt8) }
+	init: func@ ~uint64 (i: UInt64) { this init(i as UInt8) }
```
marcusnaslund: What happens if you have a `UInt` and call one of these? Just curious. Does rock cast to float anyway?

sebastianbaginski: UInt should be Uint32 by default. Let's see what rock does ...

sebastianbaginski: I have updated the commit so I think you know the answer.

marcusnaslund: Rock does good things sometimes :)

test/draw/RasterMonochromeTest.ooc
```diff
@@ -98,6 +98,19 @@ RasterMonochromeTest: class extends Fixture {
...
+			value64: UInt64 = 128
+			image[0, 0] = ColorMonochrome new(value32)
+			image[0, 1] = ColorMonochrome new(value64)
+			image[0, 2] = ColorMonochrome new(value)
+			expect(image[0, 0] y as Int, is equal to(value32 as Int))
+			expect(image[0, 1] y as Int, is equal to(value64 as Int))
+			expect(image[0, 2] y as Int, is equal to(value as Int))
+			image referenceCount decrease()
+		})
 		/*this add("distance, convertFrom RasterBgra", func {
```
marcusnaslund: Not this PR, but since this is stuff you've been working on, what's the status on this?

sebastianbaginski: I don't know, I think this is the first time I see it :)

marcusnaslund: Take a look at it later. :)

sebastianbaginski: OK

PR #1017, opened 2015-12-03
added auto-test file and fix for Image coordinate system.

ziquanyu:
Fixed this PR.  Tested on both. It should be fine this time.


ziquanyu:
@marcusnaslund  have fixed this time. both tests are passed.

marcusnaslund:
I will be very thorough this time. Thank you.

ziquanyu:
Do that, I think it will be fine this time. I did run test on only kean, that's the reason i didn't notice it affects on others.

ziquanyu:
you can maybe wait a bit for this PR. I will make some tests to test this again. it seems more complex than I thought. Even if it runs well now, yet it may cause problem in some situation on both pc and phone.

ziquanyu:
@marcusnaslund  I close this PR since too much were changed. I will make new PR when the tests are done on both pc and phone.

PR #1021, opened 2015-12-04
Added constructor from RasterImage to RasterPacked

sebastianbaginski:
Will be needed later.
@marcusnaslund

### DIFF
--- a/source/draw/RasterPacked.ooc
+++ b/source/draw/RasterPacked.ooc
@@ -109,6 +109,11 @@ RasterPacked: abstract class extends RasterImage {
 		this _buffer = original buffer copy()
 		this _stride = original stride
 	}
+	init: func ~fromRasterImage (original: RasterImage) {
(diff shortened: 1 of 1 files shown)


source/draw/RasterPacked.ooc
```diff
@@ -109,6 +109,11 @@ RasterPacked: abstract class extends RasterImage {
 		this _buffer = original buffer copy()
 		this _stride = original stride
 	}
+	init: func ~fromRasterImage (original: RasterImage) {
+		super(original)
+		this _stride = this bytesPerPixel * original width
+		this _buffer = ByteBuffer new(this stride * original height)
```
marcusnaslund: Do we want the buffer to own the data here as well?

sebastianbaginski: I think this creates a recyclable buffer, same as other init methods here.

PR #1024, opened 2015-12-04
UInt support in Constraints

marcusnaslund:
Also removed some casts in tests. @sebastianbaginski ?


sebastianbaginski:
Looks good. So  Uint 16, 32, 64 still left to do ?

sebastianbaginski:
ByteBuffer failed, does it work locally ? You can put the flag again yourself

marcusnaslund:
Yargh. Yes, it works locally. I will memcheck this test.

marcusnaslund:
Oh, nvm, it does not. I will need to cast the "correct" argument instead...

marcusnaslund:
> Uint 16, 32, 64 still left to do ?

`UInt64` already exists, and `UInt16` we don't use.

PR #1027, opened 2015-12-04
Tests for drawing on raster canvas with different coordinate systems

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/1022
Depends on https://github.com/cogneco/ooc-kean/pull/1020


sebastianbaginski:
@marcusnaslund rebased and ready

sebastianbaginski:
https://github.com/cogneco/ooc-kean/issues/1029
coordinate system is not meant to be written to as here, but I think for now we can leave it as it is useful to see that it works

PR #1028, opened 2015-12-04
Cleaned out some global functions from math.ooc

marcusnaslund:
See any mistakes @sebastianbaginski ?

### DIFF
--- a/source/draw/RasterBgra.ooc
+++ b/source/draw/RasterBgra.ooc
@@ -161,13 +161,13 @@ RasterBgra: class extends RasterPacked {
 	operator []= (point: IntPoint2D, value: ColorBgra) { this[point x, point y] = value }
 	operator [] (point: FloatPoint2D) -> ColorBgra { this [point x, point y] }
 	operator [] (x, y: Float) -> ColorBgra {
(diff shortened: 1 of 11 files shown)


marcusnaslund:
*Incredibly* fascinating. This makes the last `Promise` test fail. Same thing locally.

marcusnaslund:
Fixed it.

sebastianbaginski:
So what was the problem ?

marcusnaslund:
Incorrect conversion:
https://github.com/marcusnaslund/ooc-kean/commit/c00be75d0ead5a65819cd4feaac0e51fd1718bff

sebastianbaginski:
OK, but it compiled without problems ? Do you think there are more places here on in the other project where this could happen ?

marcusnaslund:
I will look at this some more.

marcusnaslund:
There we go, there's no room for unintended sideeffects anymore since all three versions were removed this time. Will require a corresponding PR in the other project though, I'm on it. Peer review again?

sebastianbaginski:
OK I think you know what you are doing :)

PR #1030, opened 2015-12-04
Moved logic from FloatRotation3D to Quaternion and fixed formula

tomasmahlberg:
So, whilst moving the `angle` method from `FloatRotation3D` to `Quaternion` I also noticed that it was calculating the wrong results. Apparently there was a factor `2.0f` missing from the formula (See http://math.stackexchange.com/questions/90081/quaternion-distance). Additionally a test was missing for this particular function so it has passed by unnoticed for a while. This is now fixed.


tomasmahlberg:
@marcusnaslund review?

PR #1032, opened 2015-12-04
Moved synchronization objects from Gles3Fence to OpenGLFence.

emilwestergren:
Moved Mutex and WaitCondition up an abstraction level. Probably shouldn't be here in the long term but it's nice to be able to begin waiting on a gpu fence even though it has not been created yet on the GPU.

@sebastianbaginski peer

### DIFF
--- a/source/draw/gpu/opengl/OpenGLFence.ooc
+++ b/source/draw/gpu/opengl/OpenGLFence.ooc
@@ -17,19 +17,37 @@
 use ooc-draw-gpu
 import backend/[GLFence, GLContext]
 import OpenGLContext
+import threading/[Mutex, WaitCondition]
 
 version(!gpuOff) {
 OpenGLFence: class extends GpuFence {
 	_backend: GLFence
-	init: func (context: OpenGLContext) {
-		this _backend = context backend as GLContext createFence()
-	}
+	_syncCondition := WaitCondition new()
+	_mutex := Mutex new()
+	_context: OpenGLContext
+	init: func (=_context)
 	free: override func {
+		this _mutex free()
+		this _syncCondition free()
 		this _backend free()
 		super()
 	}
-	wait: func { this _backend clientWait() }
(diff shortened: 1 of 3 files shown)


source/draw/gpu/opengl/OpenGLFence.ooc
```diff
@@ -17,19 +17,37 @@
...
+import threading/[Mutex, WaitCondition]
 
 version(!gpuOff) {
 OpenGLFence: class extends GpuFence {
 	_backend: GLFence
-	init: func (context: OpenGLContext) {
-		this _backend = context backend as GLContext createFence()
-	}
+	_syncCondition := WaitCondition new()
+	_mutex := Mutex new()
```
sebastianbaginski: We should maybe create a class for these two (`WaitCondition` + `Mutex`), it is a common pair to use. Any suggestions for name ?

emilwestergren: Sounds like a good idea, I think @simonmika agrees. Regarding name I'm not sure.
* `SynchronizationLock`
* `WaitLock`
* `BlockLock` (Catchy right?)

Do we want to keep `signal` `wait` and `broadcast` or create a new interface?

sebastianbaginski: I vote for keeping the interface, as everyone is already familiar with it.
`MutexCondition`, `WaitLock`, `SynchronizedLock`
yes, I guess nothing beats `BlockLock`

marcusnaslund: I like `WaitLock` but it may be too close to `WaitCondition`. `SynchronizationLock` is also good.

emilwestergren: https://github.com/cogneco/ooc-kean/issues/1043

PR #1036, opened 2015-12-06
Automated DebugPrintTest

marcusnaslund:
@sebastianbaginski  - peer review?

Not a great test, but certainly better than before.


sebastianbaginski:
Yes it is better.

PR #1037, opened 2015-12-06
Wrote a test for Synchronized

### LINKED ISSUE #886: SynchronizedTest does nothing useful and is not a Fixture instance
We should rewrite this test into a meaningful Fixture instance.

marcusnaslund:
Not a great test, but better than one that does nothing.
Fixes #886 

It's basically just a `Mutex` test, but then `Synchronized` is basically just a `Mutex` wrapper. Its existence would make more sense if we had interfaces, but, alright, it's there and it's used - so it deserves a real test.


marcusnaslund:

PR #1038, opened 2015-12-07
Added constructor from RasterImage to RasterBgr

sebastianbaginski:
Removes TODO and explicit coordinate system assignment from `convertFrom` in `RasterBgr`.
@marcusnaslund

### DIFF
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -52,7 +52,8 @@ RasterBgr: class extends RasterPacked {
 	init: func ~allocateStride (size: IntVector2D, stride: UInt) { super(size, stride) }
 	init: func ~fromByteBufferStride (buffer: ByteBuffer, size: IntVector2D, stride: UInt) { super(buffer, size, stride) }
 	init: func ~fromByteBuffer (buffer: ByteBuffer, size: IntVector2D) { this init(buffer, size, this bytesPerPixel * size x) }
-	init: func ~fromRasterImage (original: This) { super(original) }
+	init: func ~fromRasterBgr (original: This) { super(original) }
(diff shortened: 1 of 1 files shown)


source/draw/RasterBgr.ooc
```diff
@@ -52,7 +52,8 @@ RasterBgr: class extends RasterPacked {
 	init: func ~allocateStride (size: IntVector2D, stride: UInt) { super(size, stride) }
 	init: func ~fromByteBufferStride (buffer: ByteBuffer, size: IntVector2D, stride: UInt) { super(buffer, size, stride) }
 	init: func ~fromByteBuffer (buffer: ByteBuffer, size: IntVector2D) { this init(buffer, size, this bytesPerPixel * size x) }
-	init: func ~fromRasterImage (original: This) { super(original) }
+	init: func ~fromRasterBgr (original: This) { super(original) }
```
marcusnaslund: Is this necessary? There is a constructor which takes a `RasterPacked` in the base class (which is what `super(original)` will call anyway).

sebastianbaginski: `init: func ~fromRasterBgr (original: This) { super(original) } ` will call `super(RasterPacked)`
and `init: func ~fromRasterImage (original: RasterImage) { super(original) }` will call `super(RasterImage)`. It won't compile the `convertFrom` method without the `fromRasterImage` version as it uses the base RasterImage in there.

sebastianbaginski: Ah now I see what you mean, I think this one is needed for the `copy` method

marcusnaslund: Alright.

PR #1040, opened 2015-12-07
Cleanup of medianPosition

### LINKED ISSUE #621: FloatPoint2DVectorList medianPosition
`FloatPoint2DVectorList medianPosition` is a strange function:
```ooc
	medianPosition: func -> FloatPoint2D {
		result := FloatPoint2D new()
		sortedX := getX()
		sortedX sort()
		sortedY := getY()
		sortedY sort()
		result x = sortedX[sortedX count / 2]
		result y = sortedY[sortedY count / 2]
		sortedX free()
		sortedY free()
		result
	}
```
1. It will return a point with an x and y position that exists in the list, but not necessarily a point that exists in the list. For the points `(0,0)`, `(1,2)`, `(2,0)`, the result will be `(1, 0)`. How would this point be of any use to us?
1. Is there a more correct way to compute a multi-dimensional median of anything?
1. Should use `FloatVectorList median()` on `getX()` and `getY()` rather than sort and get the `[count / 2]` value, assuming we do want to compute the median this way.
    - This will also give us the correct median value for a list with an even number of elements (the average of the two medians), but if we actually care about returning a point from the list, that might be a problem 

Assuming we want the function to behave the way it currently does,
```ooc
	medianPosition: func -> FloatPoint2D {
		(xValues, yValues) := (this getX(), this getY())
		result := FloatPoint2D new(xValues median(), yValues median())
		xValues free()
		yValues free()
		result
	}
```
should do it, but I'm not sure that's what we want. Or whether we even need this function.

marcusnaslund:
Fixes #621 
Peer review @bettinaselig ?

### DIFF
--- a/source/geometry/FloatPoint2DVectorList.ooc
+++ b/source/geometry/FloatPoint2DVectorList.ooc
@@ -61,15 +61,10 @@ FloatPoint2DVectorList: class extends VectorList<FloatPoint2D> {
 		result
 	}
 	medianPosition: func -> FloatPoint2D {
-		result := FloatPoint2D new()
-		sortedX := this getX()
-		sortedX sort()
-		sortedY := this getY()
-		sortedY sort()
-		result x = sortedX[sortedX count / 2]
-		result y = sortedY[sortedY count / 2]
-		sortedX free()
-		sortedY free()
+		(xValues, yValues) := (this getX(), this getY())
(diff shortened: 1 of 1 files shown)


bettinaselig:
Sure.

bettinaselig:
Looks good.

PR #1042, opened 2015-12-07
Added missing toUv function to Color covers

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/1011
@marcusnaslund


source/draw/Color.ooc
```diff
@@ -72,6 +73,7 @@ ColorUv: cover {
 	init: func@ ~double (d: Double) { this init(d*255.0f clamp(0.0f, 255.0f) as UInt8) }
 	copy: func -> This { This new(this u, this v) }
 	toMonochrome: func -> ColorMonochrome { ColorMonochrome new() }
+	toUv: func -> This { this }
```
sebastianbaginski: This is a cover so this function is already called on copy.

marcusnaslund:
Did you just update this?

sebastianbaginski:
Yes, I think so.

PR #1047, opened 2015-12-07
Moved VectorList to sdk

sebastianbaginski:
VectorList moved to SDK and made available by default
@marcusnaslund


marcusnaslund:
Nice. Will you start replacing `ArrayList` uses?

sebastianbaginski:
I was thinking about moving Text into SDK first, as this would allow us to rewrite File to use Text instead, and get rid of those annoying memory leaks inside File class.

marcusnaslund:
Alright - you do that and I can take a look at replacing `ArrayList` if/when I have time.

PR #1051, opened 2015-12-08
`lastIndexOf` method for Text

sebastianbaginski:
Replaces https://github.com/cogneco/ooc-kean/pull/1050
@marcusnaslund


sebastianbaginski:
Actually no need to have both PRs, let's go with just this one.

sebastianbaginski:
This will be needed in the `File` class.

PR #1053, opened 2015-12-08
Allow negative size and offset in OutOfBounds exception

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/1008
@marcusnaslund

### DIFF
--- a/source/sdk/lang/Exception.ooc
+++ b/source/sdk/lang/Exception.ooc
@@ -197,13 +197,13 @@ OSException: class extends Exception {
 }
 
 OutOfBoundsException: class extends Exception {
-	init: func (=origin, accessOffset, elementLength: SizeT) {
+	init: func (=origin, accessOffset, elementLength: Int) {
 		init(accessOffset, elementLength)
 	}
-	init: func ~noOrigin (accessOffset, elementLength: SizeT) {
-		message = "Trying to access an element at offset %d, but size is only %d!" format(accessOffset, elementLength)
+	init: func ~noOrigin (accessOffset, elementLength: Int) {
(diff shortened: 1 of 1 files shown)


source/sdk/lang/Exception.ooc
```diff
@@ -197,13 +197,13 @@ OSException: class extends Exception {
...
 
 OutOfBoundsException: class extends Exception {
-	init: func (=origin, accessOffset, elementLength: SizeT) {
+	init: func (=origin, accessOffset, elementLength: Int) {
 		init(accessOffset, elementLength)
 	}
-	init: func ~noOrigin (accessOffset, elementLength: SizeT) {
-		message = "Trying to access an element at offset %d, but size is only %d!" format(accessOffset, elementLength)
+	init: func ~noOrigin (accessOffset, elementLength: Int) {
+		message = "Trying to access an element at offset %i, but size is %i!" format(accessOffset, elementLength)
```
marcusnaslund: Is there a reason for changing to `%i`?

sebastianbaginski: I think there was a difference between the two but can't remember the details now...

marcusnaslund: http://www.cplusplus.com/reference/cstdio/printf/ lists both `%d` for `%i` for signed integers. But oh well, nice with refreshing changes. :)

sebastianbaginski: Yes I think there was a difference for scanf, but not printf.

marcusnaslund: "%d scans an integer as a signed decimal number, but %i defaults to decimal but also allows hexadecimal (if preceded by "0x") and octal if preceded by "0"."
I learned something today, thanks.

Issue #475, opened 2015-09-30
"empty" as a function or property name


davidhesselbom:
It is generally considered a bad idea to use words like "empty" for function/property names, because it's not clear whether it's a verb or an adjective. There are many places in our codebase where `empty` is a property that returns a `Bool`. Should those be changed into `isEmpty`? Or should it be obvious that
1. `This empty` returns an "empty" instance of something,
2. `this empty` returns a bool stating whether an instance is "empty", and
3. `this empty()` "empties" an instance?

simonmika:
My thinking is currently:
1. use `This empty`
2. `this empty` or `this isEmpty`, not sure
3. in this case I suggest using `this clear()` instead

davidhesselbom:
That's all I need for #442, but I'm leaving this open until we've reached a decision about number 2.

marcusnaslund:
There are a lot of `empty` properties in math, so maybe before #491 it would be nice to have consensus on number 2.

davidpiuva:
I prefer isEmpty for clarity.

davidhesselbom:
Yes, I'd like some consistency about this. And I agree `isEmpty` is better for properties, but if we choose to go with `isEmpty`, should all properties that return a `Bool` start with `is` for consistency? I wouldn't mind if they did.

emilwestergren:
I agree with Dr. Hesselbom, if we go with `isEmpty` then we need to be consistent.

marcusnaslund:
To revive this and hopefully reach a decision: I've also seen the `identitiy` property which in fact did not check whether an instance was the identity, but rather created a new identity element. I think this was in `FloatRotation3D`. Not exactly clear as it is now.

davidpiuva:
`makeIdentity` and `isIdentity` would separate explicitly.

emilwestergren:
I prefer
`identity` - returns new
`isIdentity` - return bool

marcusnaslund:
Yes, I also agree with prefixing `is` for properties that return a `Bool`.

simonmika:
In a properly designed language there should not be any conflicts with having a static property and an instance property with the same name. As they are properties on two different things (object vs class) there are really no risk for misunderstandings. Some languages however are not designed well and therefore this issue arises.

I don't favor using `is` prefixes always but it might be the least bad thing to do in same cases.

niklasborwell:
Agree with 
`identity` - returns new
`isIdentity` - return bool
[...]

PR #1054, opened 2015-12-08
Remove `empty` from Vector.

bettinaselig:
A suggestion. Please comment.

### DIFF
--- a/source/geometry/FloatBox2D.ooc
+++ b/source/geometry/FloatBox2D.ooc
@@ -40,7 +40,7 @@ FloatBox2D: cover {
 	rightCenter ::= FloatPoint2D new(this right, this center y)
 	topCenter ::= FloatPoint2D new(this center x, this top)
 	bottomCenter ::= FloatPoint2D new(this center x, this bottom)
-	empty ::= this size empty
+	empty ::= this size x <= 0.0f || this size y <= 0.0f
 	area ::= this size x * this size y
 	init: func@ (=leftTop, =size)
 	init: func@ ~fromIntBox2D (box: IntBox2D) { this init(box left, box top, box width, box height) }
--- a/source/geometry/IntVector2D.ooc
+++ b/source/geometry/IntVector2D.ooc
@@ -23,7 +23,7 @@ IntVector2D: cover {
 	x, y: Int
 	area ::= this x * this y
 	square ::= this x == this y
-	isZeroVector ::= this size x <= 0.0f || this size y <= 0.0f
+	isZero ::= this area <= 0
 	length ::= ((this x squared() + this y squared()) as Float sqrt())
 	basisX: static This { get { This new(1, 0) } }
 	basisY: static This { get { This new(0, 1) } }
--- a/source/geometry/IntVector2D.ooc
+++ b/source/geometry/IntVector2D.ooc
@@ -23,7 +23,7 @@ IntVector2D: cover {
 	x, y: Int
 	area ::= this x * this y
 	square ::= this x == this y
-	empty ::= !(this x > 0 && this y > 0)
+	isZeroVector ::= this size x <= 0.0f || this size y <= 0.0f
 	length ::= ((this x squared() + this y squared()) as Float sqrt())
 	basisX: static This { get { This new(1, 0) } }
 	basisY: static This { get { This new(0, 1) } }
--- a/source/geometry/FloatVector2D.ooc
+++ b/source/geometry/FloatVector2D.ooc
@@ -24,7 +24,6 @@ FloatVector2D: cover {
 	x, y: Float
 	area ::= this x * this y
 	length ::= this norm
-	empty ::= this x equals(0.0f) || this y equals(0.0f) || this x < 0.0f || this y < 0.0f
 	norm ::= (this x squared() + this y squared()) sqrt()
 	azimuth ::= this y atan2(this x)
 	absolute ::= This new(Float absolute(this x), Float absolute(this y))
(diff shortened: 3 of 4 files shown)


source/geometry/IntVector2D.ooc
```diff
@@ -23,7 +23,7 @@ IntVector2D: cover {
 	x, y: Int
 	area ::= this x * this y
 	square ::= this x == this y
-	empty ::= !(this x > 0 && this y > 0)
+	isZeroVector ::= this size x <= 0.0f || this size y <= 0.0f
```
emilwestergren: I think `isZero` is enough here.

source/geometry/FloatBox2D.ooc
```diff
@@ -40,7 +40,7 @@ FloatBox2D: cover {
 	rightCenter ::= FloatPoint2D new(this right, this center y)
 	topCenter ::= FloatPoint2D new(this center x, this top)
 	bottomCenter ::= FloatPoint2D new(this center x, this bottom)
-	empty ::= this size empty
```
marcusnaslund: I would suggest `area ::= this area equals(0.f)` or `area ::= this area <= 0.f`

source/geometry/FloatBox2D.ooc
```diff
@@ -40,7 +40,7 @@ FloatBox2D: cover {
...
+	empty ::= this size x <= 0.0f || this size y <= 0.0f
```
marcusnaslund: I would suggest `area ::= this area equals(0.f)` or `area ::= this area <= 0.f`

marcusnaslund: And similarly for `IntBox`

source/geometry/IntVector2D.ooc
marcusnaslund: The name ties in to #475 - we need a decision there as well.

marcusnaslund: Of course, this property should be in all `Vector` covers.

source/geometry/FloatBox2D.ooc
sebastianbaginski: We need to add more functions to Float extension, as this will fail for my favourite float point case when `size x, size y = (0.1 + 0.1 + 0.1) - 0.3`.

marcusnaslund: I'm sorry, I don't see how that applies here @sebastianbaginski ?

sebastianbaginski: `this size x <= 0.0f || this size y <= 0.0f` will be false.

source/geometry/IntVector2D.ooc
niklasborwell: A zero vector has all elements zero, no? http://mathworld.wolfram.com/ZeroVector.html
`isZeroVector ::= this size x == 0.0f && this size y == 0.0f`
Having a property that describes whether one element is zero is still strange to me, but maybe there is another name that makes sense.

source/geometry/FloatBox2D.ooc
marcusnaslund: Right, but the box won't be empty either. :)

source/geometry/IntVector2D.ooc
marcusnaslund: Right, thanks @niklasborwell , it should be `&&`. Do we ever need the `||` case?

niklasborwell: Nowhere in the current code at least. Except for boxes, but they should have their own properties.

marcusnaslund: Good.

source/geometry/FloatBox2D.ooc
bettinaselig: We haven't defined that yet... unfortunately... 
Can we solve one problem after the other?

source/geometry/IntVector2D.ooc
```diff
@@ -23,7 +23,7 @@ IntVector2D: cover {
 	x, y: Int
 	area ::= this x * this y
 	square ::= this x == this y
-	empty ::= !(this x > 0 && this y > 0)
+	isZero ::= this area <= 0
```
emilwestergren:
```ooc
isZero ::= abs(this x) < tolerance && abs(this y) < tolerance
```

sebastianbaginski: This will go into Float extension like `equals`.

emilwestergren: OK, `this area <= 0` is just plain wrong.

bettinaselig: What is `tolerance`? I think it is getting a bit too specific here.

sebastianbaginski: Exactly (also see the discussion above).

sebastianbaginski: No, don't add it here, we're gonna update the Float class to compare with tolerance.

sebastianbaginski: @bettinaselig see https://github.com/cogneco/ooc-kean/issues/860

niklasborwell: I think the float comparison is OK here, but it should be
`isZero ::= this area == 0`
or preferably
`isZero ::= this x == 0 && this y == 0`

sebastianbaginski: OK just noticed we're commenting on the `IntBox` about floating point rounding errors ...

bettinaselig:
Discussion IRL needed.

PR #1057, opened 2015-12-08
Started removing ArrayList uses

marcusnaslund:
"Talk to your kids about `ArrayList`, before someone else does."

@sebastianbaginski ?

### DIFF
--- a/source/io/Process.ooc
+++ b/source/io/Process.ooc
@@ -23,7 +23,7 @@ Process: abstract class {
 	pid = 0: Long
 
 	new: static func ~fromArray (args: String[]) -> This {
-		p := VectorList<String> new()
+		p := VectorList<String> new(args length, false)
 		for (i in 0 .. args length) {
 			arg := args[i]
 			p add(arg)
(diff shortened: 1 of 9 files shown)


source/io/Process.ooc
```diff
@@ -108,7 +108,11 @@ Process: abstract class {
 
 	// Returns a representation of the command, escaped to some point.
 	getCommandLine: func -> String {
-		args join(" ") replaceAll("\\", "\\\\")
+		result := ""
+		for (i in 0 .. args count - 1)
+			result = result << args[i] << " "
+		result = result << args[args count - 1]
```
sebastianbaginski: Previous code did nothing when `args` was empty. I think this will crash now.
I don't know if it is possible, though, but I wanted to point it out anyway.

marcusnaslund: Hm, not sure if we should have a check for that.
`// The first argument should be the path to the executable.`

sebastianbaginski: OK. I like how they were not sure about it (first argument "should" be the path) :)

marcusnaslund:
This breaks something in the other project, will look at it again in a little while.

marcusnaslund:
There we go. @sebastianbaginski , mind checking this again?

sebastianbaginski:
sure

sebastianbaginski:
It looks fine, I think. Are you sure nothing leaks after you took the ownership away from vector ?

marcusnaslund:
> Are you sure nothing leaks after you took the ownership away from vector ?

The old implementation with `ArrayList` didn't free anything either. :) The other project explicitly frees all the `String`s that go in here. We should not do it that way in the long run, but then we shouldn't be doing it with `String` in the long run either, so I guess it's OK to leave it like this for now?

sebastianbaginski:
Sure, it will be rewritten anyway.

PR #1060, opened 2015-12-08
Added nullTerminated method to Text, fixed raw method in TextBuffer

sebastianbaginski:
Fixes https://github.com/cogneco/ooc-kean/issues/1059
Fixes https://github.com/cogneco/ooc-kean/issues/1056
@marcusnaslund

### DIFF
--- a/source/sdk/Text.ooc
+++ b/source/sdk/Text.ooc
@@ -190,6 +190,17 @@ Text: cover {
 		this free(Owner Receiver)
 		result
 	}
+	nullTerminated: func -> This {
+		t := this take()
+		result: This
+		if (t endsWith('\0'))
+			result = this
+		else {
+			result = t + t"\0"
+			this free(Owner Receiver)
+		}
+		result
+	}
 	print: func {
 		fwrite(this _buffer _backend pointer as Char*, 1, this _buffer _backend size, stdout)
 		this free(Owner Receiver)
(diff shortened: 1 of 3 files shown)


source/sdk/Text.ooc
```diff
@@ -190,6 +190,17 @@ Text: cover {
 		this free(Owner Receiver)
 		result
 	}
+	nullTerminated: func -> This {
```
simonmika: Remove! There is no such thing as a null terminated text.

source/sdk/TextBuffer.ooc
```diff
@@ -20,11 +20,7 @@ import OwnedBuffer
 
 TextBuffer: cover {
 	_backend: OwnedBuffer
-	raw: Char* { get {
-		result := this _backend pointer as Char*
-		this free(Owner Receiver)
-		result
-	}}
+	raw ::= this _backend pointer as Char*
```
simonmika: Wrong, call to property should always free if called with owner receiver.

source/sdk/Text.ooc
sebastianbaginski: So how do you plan to pass a content of Text to the system calls which expects `\0` at the end of char buffer ?

source/sdk/TextBuffer.ooc
sebastianbaginski: This is not valid for property which should return a pointer to internal buffer. Please provide a single use case where this is correct to free a buffer and then return pointer to it.

simonmika:
When you want to no which memory it pointed to. I think it is better to remove `raw` then to make exceptions.

PR #1061, opened 2015-12-08
Empty

bettinaselig:
According to the discussion today, I changed `Vector` and `Box`:
- Vectors have now properties `hasZeroArea` or `hasZeroVolume`
- Boxes have now properties `isEmpty` instead of `empty`

### DIFF
--- a/source/geometry/FloatBox2D.ooc
+++ b/source/geometry/FloatBox2D.ooc
@@ -40,8 +40,8 @@ FloatBox2D: cover {
 	rightCenter ::= FloatPoint2D new(this right, this center y)
 	topCenter ::= FloatPoint2D new(this center x, this top)
 	bottomCenter ::= FloatPoint2D new(this center x, this bottom)
-	empty ::= this size empty
-	area ::= this size x * this size y
+	hasZeroArea ::= this size hasZeroArea
+	area ::= this size area
 	init: func@ (=leftTop, =size)
 	init: func@ ~fromIntBox2D (box: IntBox2D) { this init(box left, box top, box width, box height) }
 	init: func@ ~fromPoints (first, second: FloatPoint2D) {
@@ -118,15 +118,15 @@ FloatBox2D: cover {
 	ceiling: func -> This { This new(this leftTop ceiling(), this size ceiling()) }
 	floor: func -> This { This new(this leftTop floor(), this size floor()) }
 	operator + (other: This) -> This {
-		if (this empty)
+		if (this hasZeroArea)
 			other
-		else if (other empty)
+		else if (other hasZeroArea)
 			this
 		else
 			this union(other)
 	}
 	operator - (other: This) -> This {
-		if (this empty || other empty)
+		if (this hasZeroArea || other hasZeroArea)
 			This new()
 		else
 			this intersection(other)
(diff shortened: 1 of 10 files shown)


bettinaselig:
@niklasborwell @emilwestergren @sebastianbaginski @marcusnaslund @davidpiuva @simonmika 
Any comments?

sebastianbaginski:
Nice. I like this `isEmpty ::= this size hasZeroArea`, we're getting pretty close to Shakespeare Language - http://shakespearelang.sourceforge.net/report/shakespeare/#SECTION00091000000000000000

davidpiuva:
`isEmpty` for a box is weird since we don't place items in the boxes.

source/geometry/IntPoint3D.ooc
```diff
@@ -22,6 +22,8 @@ use ooc-base
 
 IntPoint3D: cover {
 	x, y, z: Int
+	volume ::= this x * this y * this z
+	hasZeroVolume ::= this volume == 0
```
marcusnaslund: This belongs in `IntVector3D`

source/geometry/FloatBox2D.ooc
```diff
@@ -40,8 +40,8 @@ FloatBox2D: cover {
 	rightCenter ::= FloatPoint2D new(this right, this center y)
 	topCenter ::= FloatPoint2D new(this center x, this top)
 	bottomCenter ::= FloatPoint2D new(this center x, this bottom)
-	empty ::= this size empty
-	area ::= this size x * this size y
+	isEmpty ::= this size hasZeroArea
```
davidpiuva: A large box can still be empty so this will mess up the box metaphor.

marcusnaslund: I think that's a matter for philosophical debate. Is the box really large if it has zero area? :)

sebastianbaginski: I think David meant a large box, like for keeping stuff inside, which is empty because there is no stuff inside but is still large at the same time.

davidpiuva: It can have zero volume but still contain a singularity and therefore not be empty.

bettinaselig:
@marcusnaslund can decide if `Box` should also have `hasZeroArea` instead of `isEmpty`.

bettinaselig:
Need to update the tests.

source/geometry/FloatBox2D.ooc
marcusnaslund: Ah, got it. In our 3D world, it's quite hard to store stuff in a "large" 2D box anyhow. :) 
But I see the problem. I'm OK with `hasZeroArea` for box as well. What do you think?

davidpiuva: `hasZeroArea` is a better metaphor.

bettinaselig:
Wohooo! Done.

bettinaselig:

sebastianbaginski:
ok, set the flag yourself when you are done with adapting the code in other project

simonmika:
@davidpiuva We have a volume or an area in a box. Please try to understand the naming not misunderstand it on purpose.

simonmika:
I suggest at least going for `hasNoArea` and `hasNoVolume` instead although it still a unnecessarily long property.

davidpiuva:
I do not missunderstand anything. I just feel that `empty` belongs to container classes.

bettinaselig:
`empty` is away now anyway.

simonmika:
The beauty of object oriented programming is that the names methods and properties is to be understood in the relation to the class they operate on.

PR #1063, opened 2015-12-08
Replaced String with Text in FStream

sebastianbaginski:
Depends on https://github.com/cogneco/ooc-kean/pull/1060
Not for review yet


sebastianbaginski:
It is impossible without having a way to create Text with null character at the end of buffer.

PR #1065, opened 2015-12-08
added missing floating point comparison functions

sebastianbaginski:
I think we can close https://github.com/cogneco/ooc-kean/issues/860 after including these functions.

### DIFF
--- a/source/sdk/math.ooc
+++ b/source/sdk/math.ooc
@@ -39,7 +39,7 @@ extend Int {
 	maximumValue ::= static INT_MAX
 	pi ::= static 3
 	e ::= static 2
-	
+
 	modulo: func (divisor: This) -> This {
 		result := this
 		if (result < 0)
@@ -114,7 +114,8 @@ operator != (left, right: Range) -> Bool { !(left == right) }
 extend Double {
 	pi ::= static 3.14159_26535_89793_23846_26433_83279
 	e ::= static 2.718281828459045235360287471352662497757247093699959574966
-	
+	defaultTolerance ::= static 0.0001
+
 	cos: extern (cos) func -> This
 	sin: extern (sin) func -> This
 	tan: extern (tan) func -> This
@@ -157,7 +158,11 @@ extend Double {
 	clamp: func (floor, ceiling: This) -> This {
 		this > ceiling ? ceiling : (this < floor ? floor : this)
 	}
-	equals: func (other: This, tolerance := 0.0001) -> Bool { (this - other) abs() < tolerance }
+	equals: func (other: This, tolerance := This defaultTolerance) -> Bool { (this - other) abs() < tolerance }
+	lessThan: func (other: This, tolerance := This defaultTolerance) -> Bool { this - other < tolerance }
+	greaterThan: func (other: This, tolerance := This defaultTolerance) -> Bool { this - other > tolerance }
+	lessOrEqual: func (other: This, tolerance := This defaultTolerance) -> Bool { this equals(other) || this lessThan(other) }
+	greaterOrEqual: func (other: This, tolerance := This defaultTolerance) -> Bool { this equals(other) || this greaterThan(other) }
 }
 
 extend Float {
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
Do we want `isZero ::= this equals(0.0)` property ?

source/sdk/math.ooc
```diff
@@ -39,7 +39,7 @@ extend Int {
 	maximumValue ::= static INT_MAX
 	pi ::= static 3
 	e ::= static 2
-	
+
```
marcusnaslund: What editor do you use?

sebastianbaginski: Atom. I think there was a tab in there.

marcusnaslund: VSCode keeps the tab, it seems. Can I fix this? @thomasfanell

thomasfanell: File -> User Settings: ```"editor.insertSpaces": "auto"```

marcusnaslund: Should probably be the default setting.

source/sdk/math.ooc
```diff
@@ -158,6 +158,10 @@ extend Double {
 		this > ceiling ? ceiling : (this < floor ? floor : this)
 	}
 	equals: func (other: This, tolerance := 0.0001) -> Bool { (this - other) abs() < tolerance }
+	lessThan: func (other: This, tolerance := 0.0001) -> Bool { this - other < tolerance }
+	greaterThan: func (other: This, tolerance := 0.0001) -> Bool { this - other > tolerance }
+	lessOrEqual: func (other: This, tolerance := 0.0001) -> Bool { this equals(other) || this lessThan(other) }
+	greaterOrEqual: func (other: This, tolerance := 0.0001) -> Bool { this equals(other) || this greaterThan(other) }
```
marcusnaslund: I think I'd prefer some static `defaultTolerance` property.

sebastianbaginski: updated

source/sdk/math.ooc
```diff
@@ -291,13 +297,18 @@ extend Float {
...
+	greaterThan: func (other: This, tolerance := This defaultTolerance) -> Bool { this - other > tolerance }
+	lessOrEqual: func (other: This, tolerance := This defaultTolerance) -> Bool { this equals(other) || this lessThan(other) }
+	greaterOrEqual: func (other: This, tolerance := This defaultTolerance) -> Bool { this equals(other) || this greaterThan(other) }
 }
 
 extend LDouble {
 	pi ::= static 3.14159_26535_89793_23846_26433_83279L
 	e ::= static 2.718281828459045235360287471352662497757247093699959574966L
-	
+	defaultTolerance ::= static 0.0001L
```
sebastianbaginski: Can't reuse the value from Double class because rock:
> error Property tries to define a field in type extension.
>
>    defaultTolerance ::= static Double defaultTolerance as This

marcusnaslund: Right, `Double` already extends `LDouble`, so that makes sense. Too bad.
Guess we can't do it this way then, sorry.

marcusnaslund:
These methods should be tested in `test/sdk/MathTest.ooc`. Now or later?

sebastianbaginski:
They are not used anywhere now. You can merge it and tomorrow before 12.00 you will have PR with tests

marcusnaslund:
> tomorrow before 12.00

Haha I'm not here until 13.00 tomorrow, so you have one more hour. ;)

source/sdk/math.ooc
```diff
@@ -157,7 +158,11 @@ extend Double {
 	clamp: func (floor, ceiling: This) -> This {
 		this > ceiling ? ceiling : (this < floor ? floor : this)
 	}
-	equals: func (other: This, tolerance := 0.0001) -> Bool { (this - other) abs() < tolerance }
+	equals: func (other: This, tolerance := This defaultTolerance) -> Bool { (this - other) abs() < tolerance }
```
emilwestergren: We probably should use a relative error in the long term for these comparisons.

sebastianbaginski: Sure absolute error can't cover all possible use cases, but I guess that's better than nothing

emilwestergren: Yes and you can supply your own tolerance if you want to anyway.

PR #1066, opened 2015-12-08
Added functions to create rotations from euler angles.

niklasborwell:
Also added note to clarify the order which rotations should be applied.
I tried making the 
`createFromEulerAngles: static func (...)`
functions like
`init: func@ ~fromEulerAngles (...)`
but it seems ooc doesn't like calling static functions from constructors.

### DIFF
--- a/test/geometry/FloatRotation3DTest.ooc
+++ b/test/geometry/FloatRotation3DTest.ooc
@@ -19,8 +19,6 @@ use ooc-geometry
 use ooc-unit
 import math
 
-//TODO: Missing properties: x, y, z. Missing methods: createRotationX, createRotationY, createRotationZ
-
 FloatRotation3DTest: class extends Fixture {
 	quaternion0 := Quaternion new(33.0f, 10.0f, -12.0f, 54.5f)
 	quaternion1 := Quaternion new(10.0f, 17.0f, -10.0f, 14.5f)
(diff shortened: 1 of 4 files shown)


niklasborwell:
@tomasmahlberg peer review?

test/geometry/FloatRotation3DTest.ooc
```diff
@@ -86,6 +86,15 @@ FloatRotation3DTest: class extends Fixture {
...
 		})
+		this add("euler angles conversion", func {
+			x := 0.1f
+			y := 0.23f
+			z := 0.04f
+			rotation := FloatRotation3D createFromEulerAngles(x, y, z)
+			expect(x, is equal to(rotation x) within(tolerance))
+			expect(y, is equal to(rotation y) within(tolerance))
+			expect(z, is equal to(rotation z) within(tolerance))
+		})
```
marcusnaslund: This might take care of the `TODO` on line 22, what do you think?

niklasborwell: Yeah. Btw, if you want to review this directly please go ahead. Tomas has already left for the day.

marcusnaslund: No I'm just casually reading through. Will wait until tomorrow.

tomasmahlberg:
Too bad about the constructors... Otherwise, this looks good to me. Nice to have the multiplication order well-defined here, I am going to put these creators to use immediately.

PR #1068, opened 2015-12-08
Made Receiver default owner for Closure.

emilwestergren:
Fixes https://github.com/cogneco/ooc-kean/issues/1062
@sebastianbaginski @simonmika peer review

### DIFF
--- a/test/sdk/OwnedClosureTest.ooc
+++ b/test/sdk/OwnedClosureTest.ooc
@@ -1,0 +1,57 @@
+use ooc-base
+use ooc-unit
+import math
+
+MyClass: class {
+	content: Int { get set }
+	init: func { this content = 0 }
+	apply: func (action: Func (This)) {
+		action(this)
+		(action as Closure) free(Owner Receiver)
+	}
+}
+
+OwnedClosureTest: class extends Fixture {
+	init: func {
+		super("OwnedClosure")
+		adder1 := 1
+		adder2 := 2
+		adder3 := 3
+		function1 := func (instance: MyClass) {
+			instance content += adder1
+			instance content += adder2
+			instance content += adder3
+		}
+		function2 := func (instance: MyClass) {
+			instance content += adder1
+			instance content += adder2
+			instance content += adder3
+		}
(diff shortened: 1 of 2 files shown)


simonmika:
I think this doesn't work as it requires a default constructor for covers. We probably need to change the order of the `Owner` enum.

marcusnaslund:
@simonmika is right, this will not work, unfortunately.

emilwestergren:
Oh that's right. So I change the order of the enum instead?

emilwestergren:
Is this how it's supposed to work:
```ooc
closure := (function as Closure)
closure = closure take()
closure free(Owner Sender)
```

emilwestergren:
It gets pretty messy to use this in general since you need to cast back and forth from `Closure` and `Func`.
```ooc
f := func {} as Closure
f = f take()
(f as Func)()
f free(Owner Sender)
```

emilwestergren:
Updated.

simonmika:
This looks right. I agree with the messiness. maybe there is a way to extend `Func` although I'm doubting. Let us go ahead anyhow as this is better then what we are having today.

simonmika:
Peer review done.

marcusnaslund:
Won't other things be `Owner Receiver` instead of `Owner Unknown` by default? Will that cause any sideeffects? I will have to test this rather closely.

test/sdk/OwnedClosureTest.ooc
```diff
@@ -0,0 +1,57 @@
...
+			f := function2 as Closure
+			instance1 := MyClass new()
+			expect(f owner == Owner Receiver)
+			f = f give()
+			f free(Owner Sender)
+			expect(f owner == Owner Receiver)
+			(f as Func(MyClass))(instance1)
+			instance1 apply(f as Func(MyClass))
+			expect(instance1 content, is equal to(12))
+			instance1 free()
```
marcusnaslund: I assume you split these up afterwards? Anyway, just call them `instance` without the `1`.

emilwestergren:
Updated.

PR #1072, opened 2015-12-09
Fixed incorrect offset in bilinear rescaling in RasterPacked

sebastianbaginski:
Offset used incorrect stride.
Updated the tests to use both interpolation modes in drawing.
@marcusnaslund ready


source/draw/RasterPacked.ooc
```diff
@@ -70,7 +70,7 @@ RasterPackedCanvas: abstract class extends RasterCanvas {
 		(sourceWidth, sourceHeight) := (sourceBox size x, sourceBox size y)
 		(sourceStartColumn, sourceStartRow) := (sourceBox leftTop x, sourceBox leftTop y)
 		(resultStartColumn, resultStartRow) := (resultBox leftTop x, resultBox leftTop y)
-		(sourceStride, resultStride) := (source stride / bytesPerPixel, target stride / bytesPerPixel)
```
sebastianbaginski: We didn't notice it before as bilinear resampling was tested only on monochrome images.

PR #1073, opened 2015-12-09
Started removing static members in math.ooc

marcusnaslund:
Will require changes in other files, so this is just for example. Is this what we want @simonmika ?

### DIFF
--- a/source/geometry/FloatPoint3D.ooc
+++ b/source/geometry/FloatPoint3D.ooc
@@ -22,7 +22,7 @@ use ooc-base
 
 FloatPoint3D: cover {
 	x, y, z: Float
-	norm ::= (this x squared() + this y squared() + this z squared()) sqrt()
+	norm ::= (this x squared + this y squared + this z squared) sqrt()
 	azimuth ::= this y atan2(this x)
 	isValid ::= (this x == this x && this y == this y && this z == this z)
 	elevation: Float {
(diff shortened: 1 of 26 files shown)


simonmika:
It looks like the right way to go. Long term we will want unit tests as well.

marcusnaslund:
I have already written a lot of unit tests for this, which have to be updated now also.

marcusnaslund:
Care to take over peer review @sebastianbaginski ?

source/geometry/FloatPoint3D.ooc
```diff
@@ -45,7 +45,7 @@ FloatPoint3D: cover {
 		This new(radius * (azimuth cos()) * (elevation sin()), radius * (azimuth sin()) * (elevation sin()), radius * (elevation cos()))
 	}
 	angles: static func (rx, ry, n: Float) -> This {
-		z := n*n sqrt() / (1 + ry tan() squared() + rx tan() squared())
+		z := n*n sqrt() / (1 + ry tan() squared + rx tan() squared)
```
sebastianbaginski: Same here, `x^2` is a property, but `x^(1/2)` is not ?

source/math/FloatVectorList.ooc
```diff
@@ -228,7 +228,7 @@ FloatVectorList: class extends VectorList<Float> {
 		result := This new(size)
 		factor := 1.0f / (sqrt(2.0f * Float pi) * sigma)
```
sebastianbaginski: `1.0f / (2.0f * Float pi) sqrt() * sigma` ?

source/math/FloatVectorList.ooc
```diff
@@ -228,7 +228,7 @@ FloatVectorList: class extends VectorList<Float> {
...
 		for (i in 0 .. size)
-			result add((factor * pow(Float e, -0.5f * ((i - (size - 1.0f) / 2.0f) squared()) / (sigma squared()))) as Float)
+			result add((factor * pow(Float e, -0.5f * ((i - (size - 1.0f) / 2.0f) squared) / (sigma squared))) as Float)
```
sebastianbaginski: Why the `pow` global function call is left here ?

sebastianbaginski:
There are some global function calls left. Do you want to take care of them here or another PR ? Otherwise looks good.

marcusnaslund:
I don't want to make this PR too huge, already had one where I fixed some, I will make a third that takes care of the rest.

Issue #966, opened 2015-11-30
Replace FileDescriptor with File in PipeUnix


marcusnaslund:
`FileDescriptor` has been deprecated and the only remaining use is in `PipeUnix`, after that it can be removed.

PR #1074, opened 2015-12-09
Removed a lot of unnecessary and old stuff from the SDK

marcusnaslund:
Mostly comments and old `TODO`s and `FIXME`s we either don't care about, have no idea what they're for, or will not work (like extending `Cell<T>`)

@sebastianbaginski ?

### DIFF
--- a/source/sdk/os/FileDescriptor.ooc
+++ b/source/sdk/os/FileDescriptor.ooc
@@ -19,7 +19,7 @@ FileDescriptor: cover from Int {
 
 	read: func ~evilAlloc (len: Int) -> Pointer {
 		buf := gc_malloc(len)
-		read(buf, len) // todo: check errors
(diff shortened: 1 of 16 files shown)


source/sdk/os/FileDescriptor.ooc
```diff
@@ -19,7 +19,7 @@ FileDescriptor: cover from Int {
 
 	read: func ~evilAlloc (len: Int) -> Pointer {
```
sebastianbaginski: I'm pretty sure I saw this one in the valgrind logs the other day...

marcusnaslund: Oh really?

marcusnaslund: #966

PR #1077, opened 2015-12-10
Base Application class

sebastianbaginski:


sebastianbaginski:
I hope this will be extended by event processing in the future. It will do for now.

marcusnaslund:
I guess there's no way to write a meaningful test.

sebastianbaginski:
We could create the Application object and feed it with artificial data and test if the for loop is correct, but I don't think it's worth it now.

marcusnaslund:
No, let's leave it.

PR #1078, opened 2015-12-10
GuiApplication subclass and basic event processing system

sebastianbaginski:
Depends on https://github.com/cogneco/ooc-kean/pull/1077
not for review yet

### DIFF
--- a/source/ui/Application.ooc
+++ b/source/ui/Application.ooc
@@ -15,4 +15,5 @@ Application: class {
 		this _arguments free()
 		super()
 	}
+	processEvents: virtual func
 }
--- a/source/ui/DisplayWindow.ooc
+++ b/source/ui/DisplayWindow.ooc
@@ -15,15 +15,60 @@
  * along with this software. If not, see <http://www.gnu.org/licenses/>.
  */
 
+use ooc-base
 use ooc-draw
 use ooc-geometry
 import x11/UnixWindow
 import win32/Win32DisplayWindow
+import GuiEvent
 
 DisplayWindow: abstract class {
+	_mousePressHandler: Event1<MouseEvent>
+	_mouseReleaseHandler: Event1<MouseEvent>
+	_keyPressHandler: Event1<KeyboardEvent>
+	_keyReleaseHandler: Event1<KeyboardEvent>
 	init: func (size: IntVector2D, title: String)
+	free: override func {
+		if (this _mousePressHandler)
+			this _mousePressHandler free()
+		if (this _mouseReleaseHandler)
+			this _mouseReleaseHandler free()
+		if (this _keyPressHandler)
+			this _keyPressHandler free()
+		if (this _keyReleaseHandler)
+			this _keyReleaseHandler free()
+		super()
+	}
 	draw: abstract func (image: Image)
 	refresh: virtual func
+	addMousePressHandler: func (handler: Event1<MouseEvent>) {
+		this _mousePressHandler = this _mousePressHandler ? this _mousePressHandler add(handler) : handler
+	}
+	addMouseReleaseHandler: func (handler: Event1<MouseEvent>) {
+		this _mouseReleaseHandler = this _mouseReleaseHandler ? this _mouseReleaseHandler add(handler) : handler
+	}
+	addKeyPressHandler: func (handler: Event1<KeyboardEvent>) {
+		this _keyPressHandler = this _keyPressHandler ? this _keyPressHandler add(handler) : handler
+	}
+	addKeyReleaseHandler: func (handler: Event1<KeyboardEvent>) {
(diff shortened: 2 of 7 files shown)


sebastianbaginski:
@marcusnaslund ready
it is not used anywhere yet, and most of these classes are not exposed in the `use` file, but I'd prefer to commit in small chunks

source/ui/Application.ooc
```diff
@@ -15,4 +15,5 @@ Application: class {
 		this _arguments free()
 		super()
 	}
+	processEvents: virtual func
```
sebastianbaginski: It is here because in the future it could be possible to process other types of events even when gui is off (like Timer events)

source/ui/DisplayWindow.ooc
```diff
@@ -15,15 +15,60 @@
  * along with this software. If not, see <http://www.gnu.org/licenses/>.
  */
 
+use ooc-base
```
sebastianbaginski: Needed for `Event1`

source/ui/DisplayWindow.ooc
```diff
@@ -15,15 +15,60 @@
...
+	addMouseReleaseHandler: func (handler: Event1<MouseEvent>) {
+		this _mouseReleaseHandler = this _mouseReleaseHandler ? this _mouseReleaseHandler add(handler) : handler
+	}
+	addKeyPressHandler: func (handler: Event1<KeyboardEvent>) {
+		this _keyPressHandler = this _keyPressHandler ? this _keyPressHandler add(handler) : handler
+	}
+	addKeyReleaseHandler: func (handler: Event1<KeyboardEvent>) {
+		this _keyReleaseHandler = this _keyReleaseHandler ? this _keyReleaseHandler add(handler) : handler
+	}
+	_onMousePress: func (event: MouseEvent) {
```
sebastianbaginski: These private methods will be used to pass the event to the window (for now from inside of event loop but it should be the work of yet non-existent `EventDispatcher`).

source/ui/EventLoop.ooc
```diff
@@ -0,0 +1,16 @@
+use ooc-base
+import DisplayWindow
+import x11/X11EventLoop
+
+EventLoop: abstract class {
+	init: func
+	processEvents: abstract func (receiver: DisplayWindow)
```
sebastianbaginski: For now we can have `receiver` as parameter, but a real event processing system should keep a global list of created widgets and use it to dispatch events.

marcusnaslund:
Have someone else peer review this please.

source/ui/GuiApplication.ooc
```diff
@@ -0,0 +1,24 @@
+use ooc-base
+import DisplayWindow
+import EventLoop
+
+GuiApplication: class extends Application {
```
sebastianbaginski: This will be used to simplify the other project even more.

source/ui/GuiApplication.ooc
```diff
@@ -0,0 +1,24 @@
...
+	_window: DisplayWindow
```
sebastianbaginski: It's funny to have a `GuiApplication` with just one widget, but I guess we have to start somewhere...

source/ui/GuiEvent.ooc
```diff
@@ -0,0 +1,23 @@
+use ooc-base
+use ooc-geometry
+
+GuiEvent: abstract class {
+	init: func
+}
+
+MouseEvent: class extends GuiEvent {
```
sebastianbaginski: Can be extended later with mouse button etc. 
Will do for a start

sebastianbaginski:
@emilwestergren review if you have time, otherwise @jonathanudd please

source/ui/DisplayWindow.ooc
```diff
@@ -15,15 +15,60 @@
...
+		if (this _keyPressHandler)
+			this _keyPressHandler free()
+		if (this _keyReleaseHandler)
+			this _keyReleaseHandler free()
+		super()
+	}
 	draw: abstract func (image: Image)
 	refresh: virtual func
+	addMousePressHandler: func (handler: Event1<MouseEvent>) {
+		this _mousePressHandler = this _mousePressHandler ? this _mousePressHandler add(handler) : handler
```
emilwestergren: Can't you implement +/- operator for this?

sebastianbaginski: We removed it some time ago as it was not working in some cases, and we wanted to have single interface in all cases.

sebastianbaginski: https://github.com/cogneco/ooc-kean/pull/758

emilwestergren: Ok, shame

marcusnaslund:
> not for review yet

if it has no flag, it won't be reviewed. :)

PR #1079, opened 2015-12-10
Added Text support in Cell toString

### LINKED ISSUE #689: Cell toString() has no overload for Text

marcusnaslund:
Fixes #689 

We can do this now. Allows us to use `is equal to` with `Text` without crashing the whole test.

@thomasfanell - peer review?

### DIFF
--- a/source/sdk/lang/types.ooc
+++ b/source/sdk/lang/types.ooc
@@ -127,6 +127,7 @@ Cell: class <T> {
 	}
 	toString: func -> String {
 		match (this val) {
+			case value: Text => value toString()
 			case value: Bool => value toString()
 			case value: Char => value toString()
 			case value: Int => value toString()


source/sdk/lang/types.ooc
```diff
@@ -135,6 +135,7 @@ Cell: class <T> {
 			case value: UInt => value toString()
 			case value: ULong => value toString()
 			case value: UInt64 => value toString()
+			case value: Text => value toString()
```
thomasfanell: I would rather see this at the top or bottom, not hidden between numeric types.

marcusnaslund: Haha yes, I don't know why I placed it there.

PR #1080, opened 2015-12-11
Fixed invalid read/writes in LinkedList

marcusnaslund:
Replaces #876 

Fixes some bugs in `LinkedList`, including the invalid read/writes that showed up during testing. Also changed the use of `SSizeT` to `Int`.

Next step is to have auto-wrapping in `Cell`, which is still causing some issues. 

@sebastianbaginski ?

### DIFF
--- a/source/collections/LinkedList.ooc
+++ b/source/collections/LinkedList.ooc
@@ -1,8 +1,8 @@
 import structs/List
 
 LinkedList: class <T> {
-	_size = 0 : SizeT
-	size ::= this _size as Int
+	_size := 0
+	size ::= this _size
 	_head: Node<T>
 	head ::= this _head
 	init: func {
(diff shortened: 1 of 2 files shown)


source/collections/LinkedList.ooc
```diff
@@ -98,12 +98,11 @@ LinkedList: class <T> {
 			data = this _head prev data@
 		data
 	}
-	removeAt: func (index: SSizeT) -> T {
-		item := null
+	removeAt: func (index: Int) -> T {
+		item: T = null
```
sebastianbaginski: Aha, so what was the type of `item` before ?

sebastianbaginski:
Looks good. What was causing the errors ?

marcusnaslund:
`removeNode` was called *before* that node's data was read and returned.

sebastianbaginski:
Ok thanks

source/collections/LinkedList.ooc
marcusnaslund: I don't even want to guess. Maybe it worked fine, but I've had to cast `null` sometimes in arguments, so might as well be explicit here.

PR #1081, opened 2015-12-11
added default VS Code settings

thomasfanell:


thomasfanell:
I think by excluding the `.libs` and `rock_tmp` folders from search, one can not search for generated C files. Is that a good thing?

marcusnaslund:
I guess that means the default search (if you specify no folder) will not go there. But if you specifically want to search in `rock_tmp`, VScode will still allow it? No?

sebastianbaginski:
> one can not search for generated C files

For me this is a major reason to stay with Atom, I peek into the C sources at least few times a day :)

thomasfanell:
@sebastianbaginski as I suspected, I should probably remove that filter then.

sebastianbaginski:
@thomasfanell but if if can be done manually like @marcusnaslund just said, then it's fine

marcusnaslund:
@thomasfanell no extra labels on PRs. :)

thomasfanell:
@marcusnaslund sorry, old habit.

marcusnaslund:
Of course, @sebastianbaginski , you can always override this in your own settings.

sebastianbaginski:
Then OK, don't mind my comment.

thomasfanell:
Well ok then... :+1:

PR #1082, opened 2015-12-11
Possibility to create GuiApplication without default window

sebastianbaginski:
Will be needed in other project.
This basically adds one constructor.
@marcusnaslund do you want me to ask someone else ?

### DIFF
--- a/source/ui/GuiApplication.ooc
+++ b/source/ui/GuiApplication.ooc
@@ -1,15 +1,20 @@
 use ooc-base
+use ooc-geometry
 import DisplayWindow
 import EventLoop
+import Application
 
 GuiApplication: class extends Application {
 	_window: DisplayWindow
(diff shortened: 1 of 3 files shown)


marcusnaslund:
Nah, this is small enough.

source/ui/GuiApplication.ooc
```diff
@@ -18,7 +23,13 @@ GuiApplication: class extends Application {
 			this _eventLoop free()
 		super()
 	}
+	createWindow: virtual func (windowSize: IntVector2D, windowName: String) {
```
marcusnaslund: Should this be public?

sebastianbaginski: I think so, anyone who has an application object can now create the window anytime he wants, not only in constructor.

PR #1084, opened 2015-12-11
Ownership of GraphicBuffer instance can now be passed when freeing it

emilwestergren:
When creating a `GraphicBuffer` in Vidhance, you can specify who owns the backend instance with the parameter `_ownsBackend`. So when freeing the `GraphicBuffer`, you need to specify the owner to the user through `passOwnership: Bool`.

`_free: func (backend: Pointer, passOwnership: Bool)`

`passOwnership == true`
 The user can do whatever he wants with it (i.e. recycle, free etc)
`passOwnership == false`
 The user must free it.

@sebastianbaginski @simonmika opinions?

* Breaks Vidhance API

### DIFF
--- a/source/draw/gpu/opengl/GraphicBuffer.ooc
+++ b/source/draw/gpu/opengl/GraphicBuffer.ooc
@@ -39,7 +39,7 @@ GraphicBufferUsage: enum (*2) {
 GraphicBuffer: class {
 	_allocate: static Func (Int, Int, Int, Int, Pointer*, Pointer*, Int*)
 	_createFromHandle: static Func (Int, Int, Int, Int, Int, Pointer, Bool, Pointer*, Pointer*)
-	_free: static Func (Pointer)
+	_free: static Func (Pointer, Bool)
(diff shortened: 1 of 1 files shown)


simonmika:
We probably never want to recycle something allocated by somebody else. At least not something with a specific size.

emilwestergren:
No, but the user (i.e. the wrapper) perhaps wants to recycle stuff that we are finished using.

emilwestergren:
The real issue here is that some `GraphicBuffer` instances should simply be freed when Vidhance doesn't need them anymore, but some needs to be returned to the user. Either this is managed through the `free` callback, like here, or if you have a better suggestion then please feel free to help me out.

emilwestergren:
After discussion with @sebastianbaginski we decided to keep this logic in the integration code. The backend to a `GraphicBuffer` will consist of the pointer to the instance and a flag defining the action to execute when it is freed. This way Vidhance is isolated from any recycling/freeing logic needed by the user.

emilwestergren:
See https://github.com/cogneco/ooc-kean/pull/1096

PR #1087, opened 2015-12-11
GpuFence wait now returns bool.

emilwestergren:
A status bool is returned to make sure the waiting did not stop because of error.

@marcusnaslund peer

### DIFF
--- a/source/draw/gpu/opengl/backend/gles3/Gles3Fence.ooc
+++ b/source/draw/gpu/opengl/backend/gles3/Gles3Fence.ooc
@@ -29,11 +29,11 @@ Gles3Fence: class extends GLFence {
 		version(debugGL) { validateEnd("Fence free") }
 		super()
 	}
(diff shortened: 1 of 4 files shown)


source/draw/gpu/opengl/backend/gles3/Gles3Fence.ooc
```diff
@@ -29,11 +29,11 @@ Gles3Fence: class extends GLFence {
 		version(debugGL) { validateEnd("Fence free") }
 		super()
 	}
-	clientWait: func (timeout: UInt64 = ULLONG_MAX) {
+	clientWait: func (timeout: UInt64 = ULLONG_MAX) -> Bool {
```
marcusnaslund: Not strictly necessary, so you choose, but you used `override` for the other functions.

emilwestergren:
Updated.

PR #1090, opened 2015-12-13
Removed the remaining global math functions

marcusnaslund:
Breaks API.

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -315,8 +315,8 @@ Quaternion: cover {
 		constantD := (vectorQuantityZ transpose() * matrixQuantityS * matrixQuantityS * vectorQuantityZ)[0, 0]
 
 		for (_ in 0 .. maximumIterationCount) {
-			functionValue := pow(initialGuess, 4) - (constantA + constantB) * initialGuess squared - constantC * initialGuess + (constantA * constantB + constantC * sigma - constantD)
-			derivativeValue := 4 * pow(initialGuess, 3) + 2 * (constantA + constantB) * initialGuess - constantC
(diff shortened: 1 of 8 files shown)


marcusnaslund:
Don't know why the `Text` and `TextLiteral` tests suddenly fail. Have to check that.

davidpiuva:
What is the purpose? This makes math code completely unreadable.

sebastianbaginski:
I think the only reason is to remove any global functions. But I agree, `cos(x * pow(Math pi, 2))` looks better than `(x * Math pi pow(2)) cos()`.

marcusnaslund:
Should we have some exceptions to the no global functions-rule here?

davidpiuva:
If type matching is implemented correctly with overloading then they are not really global since the same type must be mentioned.

sebastianbaginski:
@marcusnaslund I'm not sure really, but now when I think about it, it seems like a change that only makes complicated code less readable.

davidpiuva:
We might not want to write the type explicitly as `Float pow(A, B)` but it would be better to work towards `A ^ B`.

sebastianbaginski:
`^` is XOR operator in C, it could be quite confusing to redefine it. I saw some languages using `a ** b` for `pow(a, b)`, but maybe this is even less readable than `a pow(b)`.

davidpiuva:
`A pow B` then?

davidpiuva:
There should be an exception to "global" function as long as they are defined within the same class and the first argument has the `This` type for safe overloading just like the mangled code generated in C99 for every class method.

niklasborwell:
I would also vote for keeping the global math functions, to avoid situations similar to something I wrote just a moment ago:
`factor * tan(value1 + value2)` would need to be written `factor * ((value1 + value2) tan())`

marcusnaslund:
Only `pow` and `atan2` will go. OK? Peer review @niklasborwell

PR #1091, opened 2015-12-14
VS Code integration

thomasfanell:
After this functionality is implemented in magic, the following tasks will be supported directly from VS Code:
* build project
* compile current file
* test project
* test current file
* analyze project
* analyze current file

### DIFF
--- a/.configuration/build.json
+++ b/.configuration/build.json
@@ -1,5 +1,6 @@
 {
 	"name": "ooc-kean",
+	"description": "ooc-kean build configurations",
 	"rootDirectory": "project/root/directory/here",
 	"targets": [
 		{
@@ -18,10 +19,10 @@
 			"arguments": ["-x"]
 		},
 		{
-			"name": "analyze",
+			"name": "verify",
 			"description": "Style analysis",
 			"command": "magic",
-			"arguments": ["-s", "-t", "analyze"]
+			"arguments": ["-s", "-t", "verify"]
 		},
 		{
 			"name": "test",
--- a/.configuration/build.json
+++ b/.configuration/build.json
@@ -1,44 +1,41 @@
-[
-	{
-		"name": "ooc-kean release build",
-		"rootDirectory": ".",
-		"command": {
-			"name": "Build ooc-kean",
+{
+	"name": "ooc-kean",
+	"rootDirectory": "project/root/directory/here",
+	"targets": [
+		{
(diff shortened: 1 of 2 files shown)


marcusnaslund:
> After this functionality is implemented in magic

When will that be?

thomasfanell:
@marcusnaslund not too long. The basic functionality will probably be in place during this week, if no unforeseen problems occur.

thomasfanell:
@marcusnaslund is there a task you would like to see in this list?

sebastianbaginski:
@thomasfanell what is `analyze project / file` ?

thomasfanell:
@sebastianbaginski code analysis, project wide or the currently opened file

sebastianbaginski:
Alright, so something like `count lines, methods, % of comments etc.` or more like `static code analysis for possible logic / syntax errors` ? It'd be nice to have either of them :)

thomasfanell:
Code metrics analysis is a nice idea, but initially it will only be style analysis.

marcusnaslund:
Should this wait until `magic` is ready for it? Otherwise I guess this breaks the ability to compile tests from `VSCode`? (Although I don't know if anyone does it right now...?)

thomasfanell:
Yes, this should wait. One could execute the tasks, but magic would not recognize the commands.

thomasfanell:
I'll flag for a peer review when I have basic functionality up and running

marcusnaslund:
Excellent.

marcusnaslund:
> The basic functionality will probably be in place during this week

I guess we jinxed it. :)

thomasfanell:
It needs to work with PRInt too, which delayed it slightly.

sebastianbaginski:
@thomasfanell do you think you could integrate CppCheck (http://cppcheck.sourceforge.net/) at some point ? It could be nice to run even basic static analysis on the rock output.

thomasfanell:
@sebastianbaginski I will add it to the list, good idea!

marcusnaslund:
Sorry, I'm closing this so as to avoid having `PRInt` go through it every time I restart it (which has been a lot recently). Once you're making progress again, feel free to reopen it.

sebastianbaginski:
@marcusnaslund @jonathanudd Could we have a `ignore` label which would cause PRInt skip the labelled PR ?

PR #1096, opened 2015-12-14
Removed ownsMemory flag.

emilwestergren:
* No recycle/free logic is now part of the `GraphicBuffer` wrapper inside Vidhance.
* Breaks Vidhance API

@sebastianbaginski might as well take a look at this since you know why it needs changing


sebastianbaginski:
Looks good.

Issue #541, opened 2015-10-13
Remove dependancy on math in ooc-collections


marcusnaslund:
The only obstacle to moving `math.ooc` from the sdk into `ooc-math` is the use of `Range` in `ooc-collections`. Because `ooc-math` has a dependancy on `ooc-collections`, we would otherwise get a circular dependancy. The `Range` cover is extended in `math.ooc`.

If we could accomplish this, we "never" need to have `import math` in any file, only `use ooc-math`.

davidhesselbom:
The `Range` extension could be moved into `lang/Types` where `Range` is defined, but it uses `Int clamp`, which is also defined in `math.ooc`...

sebastianbaginski:
Maybe you could hard-code the uses of `Int clamp` this one time ?:) By hard-code I mean inline it manually

davidhesselbom:
Might have to, yeah. It just bothers me, this whole dependency thing...

marcusnaslund:
Well this can be done, because I just did it. :)
All the kean tests build and run, **but** I get a thousand or so warnings of the type `incompatible implicit declaration of built-in function ‘sqrt’` because rock throws in stuff like
```c
#define Extensions__sqrt sqrt
```
after moving all the extensions to a file in `ooc-math` called `Extensions.ooc`. It seems like rock does something special for definitions in the sdk..?

marcusnaslund:
This is pretty easy to do now. Question is, do we really want `math.ooc` in `source/math`, or should it stay in the `sdk`? @simonmika ?

simonmika:
I think that basic operations on `float`, `int`, etc could stay in the sdk but the other option is also viable. Any suggestions? @fredrikbryntesson @thomasfanell @sebastianbaginski @marcusnaslund ?

sebastianbaginski:
For me everything that solves the weird
```ooc
use ooc-math
import math
```
is good enough.

thomasfanell:
I would like to see the `sdk` folder disappear from `ooc-kean`, since `ooc-kean` is our real SDK. But, if that is not feasible in the foreseeable future, then basic operations that act on "built-in" types should stay in `sdk`.

simonmika:
I suggest the following long term vision:

- [ ] Remove `ooc-` from use file names in kean.
- [ ] Rename `sdk` to `system`.
- [ ] Stop importing the `sdk` but instead having `use system` explicitly in like every source file.
[...]

PR #1098, opened 2015-12-14
Automatically import math to all files

marcusnaslund:
Added `Import: math` to `sdk.use` according to the discussion in #541 - now we no longer need `import math` anywhere.

@sebastianbaginski @simonmika ?


sebastianbaginski:
Looks good.

simonmika:
Good.

PR #1099, opened 2015-12-15
fixed throwing exception in Array in safe mode

sebastianbaginski:
Previously the exception was only created without throwing it...
@marcusnaslund


marcusnaslund:
Haha, oops.

PR #1100, opened 2015-12-15
Added missing override for some functions

marcusnaslund:
Following the new rule that all functions that override an `abstract` method must have the `override` keyword.

@sebastianbaginski ?


sebastianbaginski:
Is it an error to add `override` to a function not declared in base class ?

marcusnaslund:
Yes. `error cannot override inherited member because it is not marked virtual 'add'`

sebastianbaginski:
ok thanks

PR #1103, opened 2015-12-16
fixed version flags in ui module

sebastianbaginski:
Fixes android build of other project

### DIFF
--- a/source/ui/x11/X11EventLoop.ooc
+++ b/source/ui/x11/X11EventLoop.ooc
@@ -4,6 +4,7 @@ import include/x11
 import ../[EventLoop, DisplayWindow, GuiEvent]
 import UnixWindow
 
+version((unix || apple) && !android) {
 X11EventLoop: class extends EventLoop {
 	init: func
 	processEvents: override func (receiver: DisplayWindow) {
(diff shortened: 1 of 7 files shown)


source/ui/x11/X11Window.ooc
```diff
@@ -19,7 +19,7 @@ use ooc-ui
 import include/x11
 use ooc-draw
 
-version(unix || apple) {
+version((unix || apple) && !android) {
```
marcusnaslund: Isn't there a parsing problem with `version` still?

sebastianbaginski: I think it is like that in Exception class too.

source/ui/x11/X11EventLoop.ooc
```diff
@@ -4,6 +4,7 @@ import include/x11
 import ../[EventLoop, DisplayWindow, GuiEvent]
 import UnixWindow
 
+version(!android && (unix || apple)) {
```
marcusnaslund: Or does this not parse correctly?

sebastianbaginski: I have both this and the other version in this PR and it seems to compile for android.

sebastianbaginski:
Updated.

PR #1107, opened 2015-12-16
Size of boxes must now be nonnegative

### LINKED ISSUE #1106: Do not allow negative sizes for FloatBox, IntBox
When `FloatBox` or `IntBox` is initialised with a negative vector as size, the member variables `leftTop` and `size` need to be recalculated. 

Something like:
```
init: func@ (point: FloatPoint2D, vector: FloatVector2D) {
    if (vector x < 0) {
        point x += vector x
        vector x = -vector x
    }
    if (vector y < 0) {
        point y += vector y
        vector y = -vector y
    }
    this leftTop = point
    this size = vector
}
```

Decided in IRL discussion after closing https://github.com/cogneco/ooc-kean/pull/1054.

marcusnaslund:
Fixes #1106

### DIFF
--- a/source/geometry/FloatBox2D.ooc
+++ b/source/geometry/FloatBox2D.ooc
@@ -14,13 +14,13 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use ooc-base
+use ooc-collections
 use ooc-math
 import FloatPoint2D
 import FloatVector2D
 import IntBox2D
 import FloatPoint2DVectorList
-use ooc-base
-use ooc-collections
 
 FloatBox2D: cover {
 	leftTop: FloatPoint2D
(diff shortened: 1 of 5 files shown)


marcusnaslund:
Peer review @bettinaselig ?

source/geometry/FloatBox2D.ooc
```diff
@@ -41,7 +41,18 @@ FloatBox2D: cover {
...
+		if (size x < 0) {
+			leftTop x += size x
+			size x = -size x
+		}
+		if (size y < 0) {
+			leftTop y += size y
+			size y = -size y
+		}
+		this leftTop = leftTop
+		this size = size
```
bettinaselig: According to https://github.com/cogneco/ooc-kean/issues/814 we want to avoid that local variables are called like class variables. I guess that is why we usually do not use the `.` for the input variables of the functions.

test/geometry/FloatShell2DTest.ooc
```diff
@@ -86,9 +86,9 @@ FloatShell2DTest: class extends Fixture {
 			expect(increased top, is equal to(-1.0f) within(tolerance))
 			expect(increased right, is equal to(2.0f) within(tolerance))
 			expect(increased bottom, is equal to(3.0f) within(tolerance))
-			expect(decreased left, is equal to(1.0f) within(tolerance))
+			expect(decreased left, is equal to(0.0f) within(tolerance))
 			expect(decreased top, is equal to(1.0f) within(tolerance))
-			expect(decreased right, is equal to(0.0f) within(tolerance))
+			expect(decreased right, is equal to(1.0f) within(tolerance))
```
bettinaselig: Why this change?

marcusnaslund: Because, with the nonnegative size, `left` cannot be to the right of `right`.

bettinaselig: Ok.

marcusnaslund: Which makes sense, so this is good. :)

source/geometry/FloatBox2D.ooc
marcusnaslund: Right you are. I will fix.

marcusnaslund:
Updated.

bettinaselig:
Now I am happy :-)

PR #1109, opened 2015-12-17
Changed from conjugate to inverse in FloatRotation3D

tomasmahlberg:
Also fixed `*` operator in `Quaternion` and removed duplicated functionality.


tomasmahlberg:
@marcusnaslund

marcusnaslund:
Line 94 of `Quaternion.ooc` as well.

tomasmahlberg:
Line 221 as well. This should also be inverse. It seems like we have duplicate functionality here as well. Both `apply` and the `*` operator for `Quaternion` and `FloatPoint3D` essentially does the same thing.

marcusnaslund:
Fix those as well then. :)

tomasmahlberg:
Updated.

PR #1110, opened 2015-12-18
Changed global math functions to Float versions

marcusnaslund:
@sebastianbaginski ?

### DIFF
--- a/source/sdk/math.ooc
+++ b/source/sdk/math.ooc
@@ -3,14 +3,14 @@ use math
 
 FLT_EPSILON: extern Float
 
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
So now there is no way of doing `sqrt` with double precision ? Not that I need it, though

marcusnaslund:
You can do it through the `sqrt` method of `Double`. I don't think we should have both versions global (at least not now) since we kind of only use single precision.

sebastianbaginski:
Right, thanks

Issue #763, opened 2015-11-13
Wrap C memory routines inside the Buffer class


marcusnaslund:
~~Even though we don't use GC, we still use `gc_free()` (as defined in `sdk/lang/Memory.ooc`) to not confuse the function with the regular `free()` we use on classes.~~

~~The names starting with `gc_` are a bit misleading, but more than just renaming them to something else, we should wrap `malloc`, `calloc` and `free` in some object-oriented way. Either by extending the `Pointer` class or by making a new `Buffer` implementation.~~

---
**Update (2016-01-28):**

- [x] Add a `moveTo` function wrapping `memmove`
- [x] `malloc`, `calloc`, `free` calls shall later utilize the `Buffer` class in `sdk/Buffer.ooc`. 
- [x] Verify that `malloc` etc. actually return something under the `safe` flag.

marcusnaslund:
Making a note here: There should also be a `moveTo` function wrapping `memmove`.

marcusnaslund:
With #1112 we now have `sdk/Buffer`. Let's make sure we also use it.

davidpiuva:
With our own memory manager, we could make assertions that pointers are valid before use in the `-Dsafe` version.

marcusnaslund:
Can we do something like this? We introduce a version flag `-DdebugMemory`:
```ooc
version (!debugMemory) {
    malloc: extern func (SizeT) -> Pointer
    free: extern func (Pointer)
    // etc. etc.
} else {
    _malloc: extern (malloc) func (SizeT) -> Pointer
    malloc: func (size: SizeT) -> Pointer {
        result := _malloc(size)
        "allocated %p" printfln(result)
        result
    }
    _free: extern (free) func (Pointer)
    free: func (pointer: Pointer) {
        "freeing %p" printfln(result)
        _free(pointer)
    }
```
That way, we don't always get an unneccessary function call for each allocation/free.

PR #1112, opened 2015-12-18
Added a new Buffer cover for memory management

marcusnaslund:
Adds an object-oriented approach to memory management.
Adding stuff to `Pointer` doesn't seem to work.

Replaces #986 

Part of the work on #763


marcusnaslund:

marcusnaslund:
Also @simonmika ?

source/sdk/Buffer.ooc
```diff
@@ -0,0 +1,70 @@
...
+	slice: func ~untilEnd (start: Int) -> This {
+		this slice(start, this _size - start)
+	}
+	slice: func (start, distance: Int) -> This { // call by value -> modifies copy of cover
+		size := distance absolute
+		if (start < 0)
+			start = this _size + start
+		if (distance < 0)
+			start -= size
+		start < this size ? This new(this _pointer + start, Int minimum(size, this _size - start)) : This empty
```
sebastianbaginski: I think this function is incorrect. `This new(pointer + start, size)` will create a `Buffer` which will try to call `free()` on a pointer which was not allocated by `malloc`. This may crash or even worse, not crash immediately.
Another problem is when someone calls `buffer2 := buffer slice(0, ...)` and attempts to free both buffers (double free ofc).

source/sdk/Buffer.ooc
```diff
@@ -0,0 +1,70 @@
...
+			this copyTo(result)
+		} else
+			result = this
+		result
+	}
+	copyTo: func (destination: This) -> Int {
+		result := 0
+		if (this _pointer != null && this _size != 0 && destination _pointer != null && destination _size != 0) {
+			result = Int minimum(this _size, destination _size)
+			memcpy(destination _pointer, this _pointer, result)
```
sebastianbaginski: What about overlapping memory regions ? :) If we want to never ever use plain C memory management functions, we need to consider this case as well. Maybe for now add a flag to this method `copyTo: func (destination: This, mayOverlay := false)` and if the flag is true call `memmove`. Or measure the difference between `memcpy` and `memmove` and if it is not significant, then always call `memmove` ?

sebastianbaginski:
There are some problems which needs to be solved before final review. This class could use a `realloc` wrapper method too.

source/sdk/Buffer.ooc
marcusnaslund: I don't see how overlap is possible with two independently created `Buffer` instances.

marcusnaslund:
> This class could use a realloc wrapper method too.

Heh, that's what I said before, and yet I forgot about it myself. Will add.

source/sdk/Buffer.ooc
marcusnaslund: Hmm, should we handle this in some sophisticated way or just allocate new space for the new `Buffer` and copy data to it?

source/sdk/Buffer.ooc
sebastianbaginski: It all depends on how we deal with `slice` method, if it's like now with added some `ownership` then this question is still valid.

source/sdk/Buffer.ooc
sebastianbaginski: `Owner Sender / Receiver` ? :) I don't know, a simple boolean `ownsMemory` or something could also work.

sebastianbaginski: Or just allocate and copy the slice, then we don't have to deal with overlapping memory blocks and the code will be simpler. What do you think ?

simonmika: `slice` should never copy. If a copy is wanted one should slice and then copy. Else we need to have partial compares and a lot of other complex methods. I suggest leaving it like it is. Calling free on a buffer that is the result is an error as is using a slice after its "parent" was freed. One has to know what to do when using a buffer directly. Buffer is just a object oriented approach to memory handling. Everything can still be screwed up. For more fail safe ways to handle memory one should always use something else like `OwnedBuffer` or even `ByteBuffer`.

source/sdk/Buffer.ooc
simonmika: Buffers can overlap. That is the responsibility of the caller to this and other methods to ensure it does not overlap. This constraint is no different than calling `memcpy` directly.

marcusnaslund:
With @simonmika 's comments I guess this goes back to peer review. @sebastianbaginski ?

source/sdk/Buffer.ooc
sebastianbaginski: Yes, and since we already wrap the `memcpy` functionality we can as well wrap `memmove`. I mean to avoid code like
```ooc
if (buffersOverlap(buffer1, buffer2) == false)
    buffer1 copyTo(buffer2)
else
   memmove(buffer2 _pointer, buffer1 _pointer, buffer1 size)
```
on the client side.
Anyway it can be added later.

source/sdk/Buffer.ooc
```diff
@@ -0,0 +1,77 @@
...
+	free: func@ -> Bool {
+		result := this _pointer != null && this _size != 0
+		if (result) {
+			free(this _pointer)
+			this _pointer = null
+			this _size = 0
+		}
+		result
+	}
+	extend: func (size: Int) {
```
sebastianbaginski: I think this should call by reference `func@`.

source/sdk/Buffer.ooc
```diff
@@ -0,0 +1,77 @@
...
+		newPointer: Pointer = null
+		if (size > this _size)
+			newPointer = realloc(this _pointer, size)
+		if (newPointer != null)
+			this _pointer = newPointer
```
sebastianbaginski: ... because of this line.

source/sdk/Buffer.ooc
marcusnaslund: Thanks

source/sdk/Buffer.ooc
marcusnaslund: So a `moveTo` as well?

marcusnaslund:
Updated

source/sdk/Buffer.ooc
sebastianbaginski: Sure, why not.

source/sdk/Buffer.ooc
```diff
@@ -0,0 +1,78 @@
...
+		if (result) {
+			free(this _pointer)
+			this _pointer = null
+			this _size = 0
+		}
+		result
+	}
+	extend: func@ (size: Int) {
+		newPointer := this _pointer
+		if (size > this _size) {
```
sebastianbaginski: How about changing this method name to `resize` and this condition to `if (size != this _size)` ?

marcusnaslund: Right, I don't know why I limited myself in this way.

source/sdk/Buffer.ooc
marcusnaslund: Let's hold off on that until a future PR.

sebastianbaginski:
I think it looks ok now.

Issue #694, opened 2015-11-05
Implement efficient HashMap/HashDictionary with no memory leaks


emilwestergren:

marcusnaslund:
Related: `HashBag` (which `HashDictionary` currently uses) is more or less a wrapper for `HashMap`, which makes `HashDictionary` a wrapper of a wrapper, sort of. We can probably rearrange some stuff and get rid of both `Bag` and `HashBag`.

marcusnaslund:
As an update, `Bag` and `HashBag` are now gone.

marcusnaslund:
Here is what should be done:

- [x] Clean up `HashMap` (#1346)
- [ ] Add support for `Text` as key
- [ ] Change default key to `Text`, rather than `String`, in `HashDictionary`. Update all tests accordingly.
- [ ] Fix remaining memory leaks.

marcusnaslund:
`SynchronizedResource` uses `HashMap`, and it does not always perform well in stress tests...
```ooc
(SIGSEGV) segmentation fault
[fancy backtrace]
0     BacktraceHandler backtrace()                                                                           in Backtrace                                 (at /home/mnaslund/versioned/ooc/ooc-kean/source/system/Backtrace.ooc:26)                 
1     Exception getCurrentBacktrace()                                                                        in Exception                                 (at /home/mnaslund/versioned/ooc/ooc-kean/source/system/Exception.ooc:162)                
2     _signalHandler()                                                                                       in Exception                                 (at /home/mnaslund/versioned/ooc/ooc-kean/source/system/Exception.ooc:226)                
3     killpg()                                                                                               in                                           (at (null):0)                                                                             
4     nss_hosts_lookup()                                                                                     in                                           (at /build/eglibc-3GlaMS/eglibc-2.19/string/../sysdeps/x86_64/multiarch/memcmp-sse4.S:798)
5     HashMap _keyEquals()                                                                                   in structs/HashMap                           (at /home/mnaslund/versioned/ooc/ooc-kean/source/system/structs/HashMap.ooc:328)          
6     HashMap getEntry()                                                                                     in structs/HashMap                           (at /home/mnaslund/versioned/ooc/ooc-kean/source/system/structs/HashMap.ooc:137)          
7     HashMap get()                                                                                          in structs/HashMap                           (at /home/mnaslund/versioned/ooc/ooc-kean/source/system/structs/HashMap.ooc:229)          
[...]

PR #1114, opened 2015-12-18
Cleaned up Hash classes

marcusnaslund:
Cleans up memcheck output, removed some unneeded stuff, general cleanup of code.
Part of the work on #694 

Next step is to change the key from `String` to `Text`.


sebastianbaginski:
this improves the code, so why not :)

PR #1119, opened 2015-12-20
Cleaned up test output

marcusnaslund:
Some people thought the output got a bit clogged up with all the timings. Now test time is only printed if it took more than 0.01 s, and total time is only printed at the end. @sebastianbaginski ?

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -26,6 +26,7 @@ Fixture: abstract class {
 	name: String
 	tests := VectorList<Test> new(32, false)
 	printFailures: static func {
+		"Total time: %.2f s" printfln(This totalTime)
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Looks like this:
```
11:49:51 FloatRandomGenerator ........ done in 0.22s
11:49:52 Range .... done
11:49:52 IntVectorList ... done
11:49:52 BoolVectorList ....... done
11:49:52 IntRandomGenerator ...... done in 0.22s
11:49:52 FloatExtension ... done
11:49:52 FloatMatrix ...................... done
11:49:52 FloatComplexVectorList ...... done
Total time: 28.14 s
```

sebastianbaginski:
I think it looks better now.

Issue #774, opened 2015-11-16
Start replacing toString() with toText()


marcusnaslund:
With the `parse` functions now taking `Text` instead of `String` arguments (in source/math) it is time to deprecate `toString()` and implement `toText()` throughout the library.

- [x] source/math
- [x] source/geometry

marcusnaslund:
Actually, this should be done in `source/geometry` as well, so I'm reopening this.

thomasfanell:
@marcusnaslund issue description modified

PR #1120, opened 2015-12-21
added toText() to source/math classes

thomasfanell:
Reference: #774 

**Progress:**
- [x] BoolVectorList
- [x] FloatComplex
- [x] FloatComplexVectorList
- [x] FloatMatrix
- [x] FloatVectorList
- [x] IntVectorList

Closing #774


source/sdk/Text.ooc
```diff
@@ -326,6 +326,10 @@ operator == (left: String, right: Text) -> Bool { Text new(left) == right }
 operator != (left: String, right: Text) -> Bool { !(left == right) }
 operator []= (left: TextBuffer, range: Range, right: Text) { right _buffer copyTo(left slice(range min, range count)) }
 
+extend Bool {
+	toText: func -> Text { this ? t"true" : t"false" }
+}
+
```
marcusnaslund: The reason all `toText` were put in this file was because, at the time, `Text.ooc` resided in `source/base`. Now that `Text` is a part of the sdk, we could put all `toText` functions in `Numbers.ooc` instead, alongside `toString`. What do you think?

thomasfanell: Well since it's a SDK type, why not

marcusnaslund: Great. You can make it a part of this PR as well.

thomasfanell:
@marcusnaslund peer review?

marcusnaslund:
Looks good.

PR #1121, opened 2015-12-22
Adapted to class structure style policy - Part 1

thomasfanell:
This PR covers:
* source/base
* source/cli
* source/collections
* source/concurrent

This PR concern only class structure, there are no general style fixes (that comes later).

### DIFF
--- a/source/base/Event.ooc
+++ b/source/base/Event.ooc
@@ -17,14 +17,20 @@
 
 Event: class {
 	_head: Func
-	head ::= this _head
 	_tail: This
+	head ::= this _head
 	tail ::= this _tail
 	init: func ~nil {
 		this init(func) // TODO: this is really a stupid way to create null pointer although no null text is required
 	}
 	init: func (=_head)
 	init: func ~add (=_head, =_tail)
+	free: override func {
+		(this _head as Closure) free()
+		if (this _tail != null)
+			this _tail free()
+		super()
+	}
 	add: func (action: Func) -> This {
 		this add(This new(action, null))
 	}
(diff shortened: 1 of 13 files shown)


source/base/ReferenceCounter.ooc
```diff
@@ -22,6 +21,12 @@ ReferenceCounter: class {
 		this _lock = Mutex new(mutexType)
 //		"  RC, #{This count}, #{this _target class name}" println()
 	}
+	free: override func {
+		This count -= 1
+//		"--RC, #{This count}, #{this _target class name}" println()
```
marcusnaslund: Will you remove this comment and the one above while you're at it? They shouldn't be here

source/base/Event.ooc
```diff
@@ -17,14 +17,20 @@
 
 Event: class {
 	_head: Func
-	head ::= this _head
 	_tail: This
+	head ::= this _head
```
marcusnaslund: Hmm I'm not sure about this. I think it could be good to have the corresponding property right next to its member. Will need to think about this.

source/base/ReferenceCounter.ooc
thomasfanell: I was actively avoiding that and other style fixes, since I did not want to clutter the commits. Should I clean up as much as I can, then?

source/base/Event.ooc
thomasfanell: I don't agree. Fields should always grouped at the top. I we decide otherwise, it should be done very, very soon. Since there are *a lot* of files to adapt. @simonmika, do chime in.

source/base/ReferenceCounter.ooc
marcusnaslund: No that sounds like a good plan. I've just wanted to remove these comments for a long time, but keep forgetting it every time.

source/base/Event.ooc
marcusnaslund: You may be right. We seem to be splitting them up a lot more here than in the other project though, so there will be more to change there.

source/base/ReferenceCounter.ooc
thomasfanell: Fixed. I might as well do some minor cleanup while i'm at it. I'll only correct the most obvious and bothersome style violations though. We'll have to make another pass at a later date.

marcusnaslund: Agreed.

source/collections/LinkedList.ooc
```diff
@@ -2,9 +2,9 @@ import structs/List
 
 LinkedList: class <T> {
 	_size := 0
-	size ::= this _size
 	_head: Node<T>
 	head ::= this _head
+	size ::= this _size
```
marcusnaslund: `_size` is above `_head`, so `size` should also be above `head`

thomasfanell: fixed

source/base/Event.ooc
simonmika: Properties should follow their corresponding fields.

PR #1124, opened 2015-12-22
Adapted to class structure style policy - Part 4

thomasfanell:
This PR covers:
* source/io
* source/math

### DIFF
--- a/source/math/FloatComplexVectorList.ooc
+++ b/source/math/FloatComplexVectorList.ooc
@@ -21,6 +21,35 @@ use ooc-collections
 import FloatVectorList
 
 FloatComplexVectorList: class extends VectorList<FloatComplex> {
+	sum: FloatComplex {
+		get {
+			result := FloatComplex new()
+			for (i in 0 .. this _count)
+				result = result + this[i]
+			result
+		}
+	}
+	mean ::= this sum / this _count
+	real: FloatVectorList {
+		get {
+			result := FloatVectorList new()
+			for (i in 0 .. this _count) {
+				currentPoint := this[i]
+				result add(currentPoint real)
+			}
+			result
+		}
+	}
+	imaginary: FloatVectorList {
+		get {
+			result := FloatVectorList new()
+			for (i in 0 .. this _count) {
+				currentPoint := this[i]
+				result add(currentPoint imaginary)
+			}
+			result
+		}
+	}
 	init: func ~default {
 		super()
 	}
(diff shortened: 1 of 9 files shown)


source/math/FloatComplexVectorList.ooc
```diff
@@ -77,6 +50,33 @@ FloatComplexVectorList: class extends VectorList<FloatComplex> {
...
+		result _vector = this _vector
+		result _count = this _count
+		result
+	}
+	copy: func -> This {
+		result := This new(this _count)
+		for (i in 0 .. this _count)
+			result add(this[i])
+		result
+	}
```
marcusnaslund: Do we need this? Or is it just so that the return type is `XVectorList` instead of `VectorList<X>`?

source/math/FloatMatrix.ooc
```diff
@@ -20,17 +20,18 @@ FloatMatrix: cover {
...
 	height: Int { get {
 		result := this _height
 		this free(Owner Receiver)
 		result
 	}}
+	width: Int { get {
+		result := this _width
+		this free(Owner Receiver)
+		result
+	}}
```
marcusnaslund: Note: `_width` comes before `_height`.

source/math/FloatMatrix.ooc
```diff
@@ -395,6 +365,28 @@ FloatMatrix: cover {
...
+		version (safe) {
+			if (x < 0 || y < 0 || x >= t width || y >= t height)
+				raise("Accessing matrix element out of range in get operator")
+		}
+		result := t elements[x + y * t width]
+		this free(Owner Receiver)
+		result
+	}
+	// NOTE: Because rock doesn't understand the concept of inline functions,
+	// this function has been inlined manually in many places in this file for performance reasons.
```
marcusnaslund: We do this in a lot of places these days, so you can probably remove these comments too.

source/math/FloatVectorList.ooc
```diff
@@ -80,6 +59,27 @@ FloatVectorList: class extends VectorList<Float> {
...
+		super(capacity)
+		for (i in 0 .. capacity)
+			this add(value)
+	}
+	toVectorList: func -> VectorList<Float> {
+		result := VectorList<Float> new(this count)
+		result _vector = this _vector
+		result _count = this count
+		result
+	}
```
marcusnaslund: Also, why this function? `myVectorList as VectorList<Float>` would do the exact same thing and wouldn't force me to copy everything. Do we ever use it?

source/math/FloatComplexVectorList.ooc
thomasfanell: Hmm?

marcusnaslund: There's a `copy` method in `VectorList`.

thomasfanell: I think maybe you are missing the point of these PRs... :-)

thomasfanell: Any changes to the code, syntax-wise, should not be in this PR, as it will clutter them up and also make them huge.

thomasfanell: I'll note this down though, and go through them all once these are done.

marcusnaslund: I'm not suggesting you fix this now, all these changes are just bringing it to my attention. I'm just asking you since you're mr. Rock.

marcusnaslund: Thanks.

source/math/FloatVectorList.ooc
```diff
@@ -500,4 +310,194 @@ FloatVectorList: class extends VectorList<Float> {
...
+		result[0, 1] = 1.f
+		for (index in 1 .. size - 1) {
+			result[index, index - 1] = 1.f
+			result[index, index] = 4.f
+			result[index, index + 1] = 1.f
+		}
+		result[size - 1, size - 2] = 1.f
+		result[size - 1, size - 1] = 2.f
+		result
+	}
```
marcusnaslund: There's some cleanup that can be done in the above functions. Don't think `quicksort` should be static and the comments aren't very useful. `_nthElement` should be `_nthSmallestElement`. Writing this here so you will note it down and one of us will fix this after this PR.

PR #1125, opened 2015-12-22
Adapted to class structure style policy - Part 5

thomasfanell:
This PR covers:
* source/net
* source/plot

### DIFF
--- a/source/plot/SvgWriter2D.ooc
+++ b/source/plot/SvgWriter2D.ooc
@@ -34,10 +34,10 @@ SvgWriter2D: class {
 	init: func ~svgPlotWithFilename (filename: String, args: ...) {
 		this init(File new(filename), args)
 	}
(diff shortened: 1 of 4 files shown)


source/plot/SvgWriter2D.ooc
```diff
@@ -10,11 +10,11 @@ import LinePlotData2D
 import svg/Shapes
 
 SvgWriter2D: class {
+	numberOfPlotsHorizontally: Int
```
marcusnaslund: Hmm we should never have public members in classes. This must have slipped through. Can you fix this now (as this PR is pretty small anyway)

thomasfanell: Sure thing.

thomasfanell: Fixed. I had to add some properties in order to keep the API intact.

Issue #1007, opened 2015-12-03
rock: class names from sdk/lang/types hard coded in source code


sebastianbaginski:
If we want to move / rename classes from these files we first need to deal with hard coded paths and class names in rock sources.

marcusnaslund:
I would prefer to move most (or all) files if possible at once, otherwise we have to force people to update rock every other day.

marcusnaslund:
As an update: Only four files remain. They cannot be moved right now.

marcusnaslund:
- References to `lang_array` in `source/rock/backend/cnaughty/CGenerator.ooc` and `source/rock/middle/Type.ooc`
- One reference to `lang_String` in `source/rock/backend/cnaughty/Skeleton.ooc`
- References to `lang_types` in `source/rock/backend/cnaughty/CastWriter.ooc`, `source/rock/backend/cnaughty/CGenerator.ooc`, `source/rock/backend/cnaughty/ClassDeclWriter.ooc`, `source/rock/backend/cnaughty/FunctionCallWriter.ooc`, `source/rock/middle/FuncType.ooc`,

marcusnaslund:
Moved to #1278

PR #1128, opened 2015-12-22
Moved what is possible to the sdk root directory

marcusnaslund:
Doing what is possible on #1007 - remaining stuff will need changes in rock.


marcusnaslund:
@sebastianbaginski ?

PR #1137, opened 2015-12-28
Operator formatting part 3

thomasfanell:
Formatted operators in
* source/geometry

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -197,36 +197,22 @@ Quaternion: cover {
 		"Real: " << "%8f" formatFloat(this real) >>
 		" Imaginary: " & "%8f" formatFloat(this imaginary x) >> " " & "%8f" formatFloat(this imaginary y) >> " " & "%8f" formatFloat(this imaginary z)
 	}
-	operator == (other: This) -> Bool {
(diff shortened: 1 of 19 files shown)


marcusnaslund:
I guess you *could* put `<` and `>` operators above `==` and `!=`, then the latter would align up nicely with `>=` and `<=`. This is true for a lot of covers in this PR.

thomasfanell:
ok i'm on it

source/geometry/Quaternion.ooc
```diff
@@ -346,6 +332,4 @@ Quaternion: cover {
 	}
 	kean_math_quaternion_new: unmangled static func (w, x, y, z: Float) -> This { This new(w, x, y, z) }
 }
-operator * (value: Float, other: Quaternion) -> Quaternion {
-	other * value
-}
+operator * (value: Float, other: Quaternion) -> Quaternion { other * value }
```
marcusnaslund: How about an empty line after classes/covers?

thomasfanell: I considered that, but since it's nothing below the operators, I let them be. Should I insert a new line?

marcusnaslund: Nah I'm ok with this.

PR #1138, opened 2015-12-28
General style fix - first iteration

thomasfanell:
* General style fixes (correct spacing between block types etc)
* Remove unnecessary comments
* Fix property syntax `foo: Int { get { this _foo } }` to `foo ::= this _foo`

**Comments in the SDK (or in files that were previously in the SDK) are largely left intact for now**

### DIFF
--- a/source/base/TimeSpan.ooc
+++ b/source/base/TimeSpan.ooc
@@ -20,6 +20,7 @@ use ooc-base
 TimeSpan: cover {
 	_ticks: Int64 = 0
 	ticks ::= this _ticks
+
 	init: func@ (=_ticks)
 	init: func@ ~fromHourMinuteSec (hour, minute, second, millisecond: Int) {
 		this _ticks = DateTime timeToTicks(hour, minute, second, millisecond)
--- a/source/base/TimeSpan.ooc
+++ b/source/base/TimeSpan.ooc
@@ -31,7 +31,7 @@ TimeSpan: cover {
 	elapsedDays: func -> Int64 { this ticks / DateTime ticksPerDay }
 	elapsedWeeks: func -> Int64 { this ticks / DateTime ticksPerWeek }
 
-	defaultFormat: static const Text = t"%w weeks, %d days, %h hours, %m minutes, %s seconds, %z milliseconds"
+	defaultFormat: static Text = t"%w weeks, %d days, %h hours, %m minutes, %s seconds, %z milliseconds"
 	// supported formatting expressions:
 	//  %w - weeks (rounded down)
 	//  %d - days (<7)
--- a/source/base/TimeSpan.ooc
+++ b/source/base/TimeSpan.ooc
@@ -22,28 +22,15 @@ TimeSpan: cover {
 	ticks ::= this _ticks
 
 	init: func@ (=_ticks)
-	init: func@ ~fromHourMinuteSec (hour, minute, second, millisecond: Int) {
-		this _ticks = DateTime timeToTicks(hour, minute, second, millisecond)
-	}
+	init: func@ ~fromHourMinuteSec (hour, minute, second, millisecond: Int) { this _ticks = DateTime timeToTicks(hour, minute, second, millisecond) }
 	negate: func -> This { This new(-1 * this ticks) }
-	elapsedMilliseconds: func -> Int64 {
-		this ticks / DateTime ticksPerMillisecond
-	}
-	elapsedSeconds: func -> Int64 {
-		this ticks / DateTime ticksPerSecond
-	}
-	elapsedMinutes: func -> Int64 {
-		this ticks / DateTime ticksPerMinute
-	}
-	elapsedHours: func -> Int64 {
-		this ticks / DateTime ticksPerHour
-	}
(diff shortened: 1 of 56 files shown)


source/base/TimeSpan.ooc
```diff
@@ -109,42 +97,19 @@ TimeSpan: cover {
...
+	second: static func -> This { This seconds(1) }
+	seconds: static func (count: Double) -> This { This new(DateTime ticksPerSecond * count) }
+	minute: static func -> This { This minutes(1) }
+	minutes: static func (count: Double) -> This { This new(DateTime ticksPerMinute * count) }
+	hour: static func -> This { This hours(1) }
+	hours: static func (count: Double) -> This { This new(DateTime ticksPerHour * count) }
+	day: static func -> This { This days(1) }
+	days: static func (count: Double) -> This { This new(DateTime ticksPerDay * count) }
+	week: static func -> This { This weeks(1) }
+	weeks: static func (count: Double) -> This { This new(DateTime ticksPerWeek * count) }
```
marcusnaslund: Maybe split these up so that `millisecond`, `second`, `minute` etc. come first, then `milliseconds`, `seconds`, `minutes`, etc.

source/geometry/FloatPoint2D.ooc
```diff
@@ -21,8 +21,10 @@ use ooc-base
 
 FloatPoint2D: cover {
 	x, y: Float
```
marcusnaslund: Another interesting style issue. `x, y: Float`, or
```ooc
x: Float
y: Float
```
This way is obviously better here, but is it always?

source/geometry/FloatPoint3D.ooc
```diff
@@ -29,7 +30,8 @@ FloatPoint3D: cover {
 		if (r != 0.0f)
 			r = (this z / r) clamp(-1.0f, 1.0f) acos()
 		r
-	} }
+	}}
```
marcusnaslund: I think magic dislikes this.

source/math/FloatRandomGenerator.ooc
```diff
@@ -17,7 +17,7 @@
 import os/Time
 
 FloatRandomGenerator: abstract class {
-	permanentSeed: static UInt = 0
+	permanentSeed ::= static 0
```
marcusnaslund: This should be allowed to change, it is for debugging.

source/math/IntRandomGenerator.ooc
```diff
@@ -18,7 +18,7 @@ use ooc-math
 import os/Time
 
 IntRandomGenerator: abstract class {
-	permanentSeed: static Int = 0
+	permanentSeed ::= static 0
```
marcusnaslund: Same here as for the Float version

source/geometry/FloatPoint2D.ooc
thomasfanell: I think it's always better to put one declaration per line.

source/base/TimeSpan.ooc
thomasfanell: Agreed.

source/geometry/FloatPoint2D.ooc
marcusnaslund: > I think it's always better to put one declaration per line.

Even in `FloatRotation3D` which as 16 `Float`s?

source/geometry/FloatPoint3D.ooc
thomasfanell: It does? I seem to recall implementing this rule it so it would allow this particular case, but I could be wrong. Either way, I think this looks better, don't you?

marcusnaslund: Did that change recently? We decided on one some weeks ago because magic allowed one (the one which was already there) and disallowed the other.

source/geometry/FloatPoint2D.ooc
thomasfanell: You mean `FloatTransform3D`? I see your point. I guess this is one of those judgement calls you have to make when writing the class/cover.

marcusnaslund: Right.

source/geometry/FloatPoint3D.ooc
thomasfanell: No that was quite a long time ago. I'd say two or so months ago.

thomasfanell: No complaints from magic here. So what is the policy?

marcusnaslund: Ah, we have both in the code. I see. Anyway, since we write `empty: static This { get { This new() } }`, there's a point in always having `} }` instead of `}}`. I'm not saying `} }` looks better though.

marcusnaslund: `}}` is best.

thomasfanell:
I am closing this for now since it's too big to handle now that everyone is back from leave. I'll split this up into manageable pieces at a later date.

marcusnaslund:
Understandable. A lot of good stuff here though, so looking forward to it.

PR #1140, opened 2015-12-29
Moved even more files from sdk/lang

marcusnaslund:
It turns out you have to be extra careful about which order these lines are added. Alphabetic order works most of the time.

@sebastianbaginski ?


sebastianbaginski:
So it won't compile if you include them in wrong order ? Nice, just like windows headers.

marcusnaslund:
Correct, and you get the most obscure and weirdest errors possible.

Issue #706, opened 2015-11-06
Clean and improve sdk/math.ooc


marcusnaslund:
There are a number of things to do.

- [x] Fix indentation and make it magic compliant (not necessary now, but will be later)
- [x] Remove global functions, instead make sure we use methods defined as extensions on `Int`, `Float` and so on.
- [x] Remove the global variables `PI`, `constantPi` and so on, instead use them as extensions.
- [x] ~~Move the `Range` extension and make it a part of `Range` in `sdk/lang/Numbers.ooc`~~
- [x] Fix inconsistencies and errors with `%` and `modulo` (see #508)
- [x] Change static methods into regular methods. Some can probably be properties.

Eventually, we want to make this a part of `ooc-math` and thus get rid of the `import math` line altogether. The above will help a great deal and will need to be done first.

marcusnaslund:
Let's keep `sin`, `cos`, `tan`, `asin`, `acos`, `atan` as global functions? How about `sqrt`? Also, should we change them to their `Float` counterparts since that's what we mainly use? @niklasborwell @sebastianbaginski

sebastianbaginski:
I think you know what is my opinion
```ooc
sin(x * pow(Math pi, 2)) * atan(pow(y, 3))
```
vs.
```ooc
(x * Math pi pow(2)) sin() * y pow(3) atan()
```
first one is much easier to read, at least for me. It is just the natural format, which we all know from math classes.

marcusnaslund:
`y pow(3)` looks good to me though. But I see your point about `sin`, `cos` etc. and I'm OK with keeping them. Right now they take and return `Double`, though.

niklasborwell:
Yeah it would make sense to have float functions instead.

sebastianbaginski:
... not mentioning the fact that everyone will need to remember to write `(-5) pow(2)` instead of `-5 pow(2)` :)

marcusnaslund:
Fixed by #1172

PR #1141, opened 2015-12-29
Changed multiple static functions in math.ooc to nonstatic

marcusnaslund:
Fixes the last part of #706 

This will need quite a lot of changes, mostly because of `minimum` and `maximum`, so let's all agree on this before we proceed. 

@simonmika says there should be no static methods on the number covers. That means instead of `Int maximum(2,3)` we do `2 maximum(3)`. Instead of `Float linearInterpolation(0, 1, 0.5f)` we do `0.5f linearInterpolation(0, 1)`, and so on. I can see a benefit of that.

Thoughts? @sebastianbaginski @thomasfanell @niklasborwell


marcusnaslund:
I will do everything except `minimum` and `maximum` now. Those two are big enough changes for their own PR.

marcusnaslund:
Peer review @thomasfanell ?

thomasfanell:
@marcusnaslund looks good (and this is much better than before). Good job.

PR #1145, opened 2015-12-29
Moved Backtrace and Exception out of sdk/lang

marcusnaslund:
Didn't think this would work but turns out I had to update some stuff in the other project.

@sebastianbaginski ?


sebastianbaginski:
There is a unmangled function `kean_lang_exception_outOfBoundsException_throw` in Exception.ooc, maybe you need to update it too. It is referenced in Array.h : https://github.com/cogneco/ooc-kean/blob/master/source/sdk/lang/Array.h#L18 in safe mode.

marcusnaslund:
I guess I don't need to, but I should

PR #1149, opened 2016-01-04
Added support for combined single char cli argument flags

jonathanudd:
Before this you could not write `-abc` to set the flags `a`, `b` and `c`.


jonathanudd:
@sebastianbaginski could you review this?

sebastianbaginski:
Looks ok.

PR #1150, opened 2016-01-04
Changed multiplication operator in Quaternion

tomasmahlberg:
Before #1109 we had two different implementations of multiplication between a `Quaternion` and a `FloatPoint3D`. Apparently the one that was not used and that was removed is faster so this PR changes `*` to use this method instead.

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -206,7 +206,7 @@ Quaternion: cover {
 	operator - -> This { This new(-this real, -this imaginary) }
 	operator / (value: Float) -> This { This new(this w / value, this x / value, this y / value, this z / value) }
 	operator * (value: Float) -> This { This new(this w * value, this x * value, this y * value, this z * value) }
-	operator * (value: FloatPoint3D) -> FloatPoint3D { (this * This new(0.0f, value) * this inverse) imaginary }
+	operator * (value: FloatPoint3D) -> FloatPoint3D { This hamiltonProduct(This hamiltonProduct(this, This new(0.0f, value)), this inverse) imaginary }
 	operator as -> String { this toString() }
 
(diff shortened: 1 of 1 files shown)


tomasmahlberg:
@marcusnaslund

source/geometry/Quaternion.ooc
```diff
@@ -210,7 +210,7 @@ Quaternion: cover {
 	operator - -> This { This new(-this real, -this imaginary) }
 	operator / (value: Float) -> This { This new(this w / value, this x / value, this y / value, this z / value) }
 	operator * (value: Float) -> This { This new(this w * value, this x * value, this y * value, this z * value) }
-	operator * (value: FloatPoint3D) -> FloatPoint3D { (this * This new(0.0f, value) * this inverse) imaginary }
+	operator * (value: FloatPoint3D) -> FloatPoint3D { This hamiltonProduct(This hamiltonProduct(this, This new(0.0f, value)), this inverse) imaginary }
```
marcusnaslund: What? This is faster? It just looks a lot more complicated.

tomasmahlberg: It's about 25% faster over 1 million operations with identical results using random input quaternions.
```
apply Time: 935.624 Average: 935.624
* Time: 1248.205 Average: 1248.205
```

sebastianbaginski: Nice results. What optimization level have you tested ?

tomasmahlberg: No optimization at first. Tried now with different levels and seem to get approximately the same ratio for all of them.

marcusnaslund:
Sorry, didn't think this would conflict with another PR, but it did.

tomasmahlberg:
Rebased.

PR #1151, opened 2016-01-04
Quaternion fix

emilwestergren:
Renamed function for creating a quaternion from axis angle vector https://en.wikipedia.org/wiki/Axis%E2%80%93angle_representation

* Breaks API

@niklasborwell @marcusnaslund peer review

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -18,6 +18,7 @@ use ooc-collections
 use ooc-math
 import FloatPoint3D
 import FloatTransform3D
+import FloatVector3D
 
 Quaternion: cover {
 	real: Float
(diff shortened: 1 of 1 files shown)


source/geometry/Quaternion.ooc
```diff
@@ -218,35 +215,29 @@ Quaternion: cover {
...
-				angularVelocity y * sin(angle / 2.0f) / angle,
-				angularVelocity z * sin(angle / 2.0f) / angle
+				rotation x * sin(angle / 2.0f) / angle,
+				rotation y * sin(angle / 2.0f) / angle,
+				rotation z * sin(angle / 2.0f) / angle
 				)
 		}
 		result
 	}
+	createRotation: static func (angle: Float, direction: FloatPoint3D) -> This {
```
niklasborwell: It seems this function does the same as createFromAxisAngle. Can we confirm and remove one of them?

PR #1154, opened 2016-01-05
Added normalized property for FloatPoint3D and FloatVector3D.

emilwestergren:
@marcusnaslund peer review


emilwestergren:
Not sure if the 0 check should actually be included.

marcusnaslund:
> Not sure if the 0 check should actually be included.

I think not. In the case you get something close to a zero vector you will avoid crash because of zero division, but instead return something that isn't normalized anyway and might crash something else.

emilwestergren:
Updated.

PR #1155, opened 2016-01-05
fixed small leaks from Fixture

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -48,7 +48,6 @@ Fixture: abstract class {
 		result := true
 		dateString := DateTime now toText(t"%hh:%mm:%ss") toString() & " " << this name + " "
 		This _print(dateString)
-		dateString free()
 		timer := ClockTimer new() . start()
 		for (i in 0 .. this tests count) {
 			test := tests[i]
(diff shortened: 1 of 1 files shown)


source/unit/Fixture.ooc
```diff
@@ -122,7 +122,9 @@ Fixture: abstract class {
 	is ::= static IsConstraints new()
 	testsFailed: static Bool { get { This _testsFailed } }
 	printFailures: static func {
-		"Total time: %.2f s" printfln(This totalTime)
+		timeString := "Total time: %.2f s" format(This totalTime)
+		This _print(timeString)
```
marcusnaslund: If `_print` would free its incoming argument we could get rid of a lot of `free()` calls to strings in this file.

source/unit/Fixture.ooc
```diff
@@ -122,7 +122,9 @@ Fixture: abstract class {
...
+		timeString free()
 		if (This failureNames && (This failureNames count > 0)) {
 			"Failed tests: %i [" printf(This failureNames count)
```
marcusnaslund: This should use `_print` in the same way I guess

source/unit/Fixture.ooc
sebastianbaginski: That's true, I will update it.

sebastianbaginski:
Updated.

source/unit/Fixture.ooc
```diff
@@ -122,9 +120,9 @@ Fixture: abstract class {
 	is ::= static IsConstraints new()
 	testsFailed: static Bool { get { This _testsFailed } }
 	printFailures: static func {
-		"Total time: %.2f s" printfln(This totalTime)
+		This _print("Total time: %.2f s" format(This totalTime))
```
marcusnaslund: You're missing a newline here now.

marcusnaslund: `Total time: 36.76 sFailed tests: 3 [FloatVector3D, Quaternion, FloatVector2D]`

sebastianbaginski: Thanks, fixed

PR #1157, opened 2016-01-05
Quaternion cleanup

emilwestergren:
Replaced `FloatPoint3D` with `FloatVector3D` where appropriate and some other minor improvements.

* Breaks API

@niklasborwell peer review

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -28,12 +28,6 @@ Quaternion: cover {
 	y ::= this imaginary y
 	z ::= this imaginary z
 
-	// NOTE: The coordinates are represented differently in C# Kean:
-	// x = this w
-	// y = this x
-	// z = this y
-	// w = this z
-
 	isValid ::= this real isNumber && this imaginary isValid
 	isIdentity ::= this real equals(1.f) && this imaginary x equals(0.f) && this imaginary y equals(0.f) && this imaginary z equals(0.f)
 	isZero ::= this real equals(0.f) && this imaginary x equals(0.f) && this imaginary y equals(0.f) && this imaginary z equals(0.f)
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -219,6 +219,7 @@ Quaternion: cover {
 	createRotationX: static func (angle: Float) -> This { This createRotation(angle, FloatVector3D new(1.0f, 0.0f, 0.0f)) }
 	createRotationY: static func (angle: Float) -> This { This createRotation(angle, FloatVector3D new(0.0f, 1.0f, 0.0f)) }
 	createRotationZ: static func (angle: Float) -> This { This createRotation(angle, FloatVector3D new(0.0f, 0.0f, 1.0f)) }
+	createFromAxisAngle: static func (vector: FloatVector3D) -> This { This createRotation(vector norm, vector normalized) }
 	hamiltonProduct: static func (left, right: This) -> This {
 		(a1, b1, c1, d1) := (left w, left x, left y, left z)
 		(a2, b2, c2, d2) := (right w, right x, right y, right z)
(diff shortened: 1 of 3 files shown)


source/geometry/Quaternion.ooc
```diff
@@ -28,15 +29,9 @@ Quaternion: cover {
...
-	// NOTE: The coordinates are represented differently in C# Kean:
-	// x = this w
-	// y = this x
-	// z = this y
-	// w = this z
-
 	isValid ::= this real isNumber && this imaginary isValid
 	isIdentity ::= this real equals(1.f) && this imaginary x equals(0.f) && this imaginary y equals(0.f) && this imaginary z equals(0.f)
-	isNull ::= this real equals(0.f) && this imaginary x equals(0.f) && this imaginary y equals(0.f) && this imaginary z equals(0.f)
+	isZero ::= this real equals(0.f) && this imaginary x equals(0.f) && this imaginary y equals(0.f) && this imaginary z equals(0.f)
```
niklasborwell: This could be mistaken for a quaternion representing zero rotation which is the same as `isIdentity`. Do we even use this property? I don't think we should ever create a (0,0,0,0) quaternion. Right now we are doing so in the FloatRotation3D default constructor, but I would probably remap that to identity later.

emilwestergren: I think the most common name for a quaternion with all 0's is "Zero quaternion", but I may be wrong. I think it's pretty consistent with some of the other geometry types:
Zero vector - (0, 0, 0)
Zero matrix - (0, 0, 0, 0, ...)
Zero quaternion - (0, 0, 0, 0)

source/geometry/Quaternion.ooc
```diff
@@ -214,16 +209,17 @@ Quaternion: cover {
...
 		This createRotationZ(rotationZ) * This createRotationY(rotationY) * This createRotationX(rotationX)
 	}
-	createRotation: static func (angle: Float, direction: FloatPoint3D) -> This {
+	createRotation: static func (angle: Float, direction: FloatVector3D) -> This {
 		halfAngle := angle / 2.0f
 		point3DNorm := direction norm
 		if (point3DNorm != 0.0f)
 			direction /= point3DNorm
-		This new(0.0f, halfAngle * direction) exponential
+		This new(0.0f, (halfAngle * direction) toFloatPoint3D()) exponential
```
niklasborwell: I thought we were changing the imaginary to a vector? Or is that too much for this pull request. It kind of seems related to the direction change in this function.

emilwestergren: That is my suggestion and it is included in this issue https://github.com/cogneco/ooc-kean/issues/1156, but I left it out in this PR.

source/geometry/Quaternion.ooc
niklasborwell: Can we remove this property entirely?

source/geometry/Quaternion.ooc
niklasborwell: Ok, fine.

source/geometry/Quaternion.ooc
emilwestergren: Let it stay. https://msdn.microsoft.com/en-us/library/windows/desktop/bb281613%28v=vs.85%29.aspx

niklasborwell: It shouldn't compare floats directly either.

niklasborwell: But fine, it zero quaternion seems to be well defined, so I yield.

emilwestergren:
Is there a way to restart the PRInt test?

marcusnaslund:
> Is there a way to restart the PRInt test?

There's a `Run tests` button if you look at the details, but maybe it's just visible to me? @jonathanudd

marcusnaslund:
Will merge this as soon as PRInt says ok.

jonathanudd:
Yes the `Run tests` button in PRInt is only visible to the the one running the server. Should I make it visible for everyone?

PR #1158, opened 2016-01-05
Removed - operator and changed + to * in FloatEuclidTransform

tomasmahlberg:
Since some of the operations carried out by the `+` operator are non-commutative it makes more sense to change this to a `*` operator for clarity. If we want to do what `-` was doing previously we should have to do `euclidTransform1 * euclidTransform2 inverse` instead of `euclidTransform1 - euclidTransform2`.
@niklasborwell @emilwestergren Thoughts?


niklasborwell:
Yes, this is what I would do too.

emilwestergren:
Looks good

tomasmahlberg:
@marcusnaslund Ready for review, tests probably need to be restarted.

tomasmahlberg:
Rebased.

PR #1160, opened 2016-01-05
Added maximum/minimum for Int64/UInt64

emilwestergren:
@marcusnaslund peer review


marcusnaslund:
See #1141 - @simonmika doesn't think there should be any static methods in `math.ooc`.

emilwestergren:
Ok, but that's how it's currently done for `Int`. Either I change it for `UInt64` now and it's inconsistent, or we change everything later.

marcusnaslund:
Fair enough. Let's fix it later once we've discussed it thoroughly.

PR #1161, opened 2016-01-05
Removed FileDescriptor and modified PipeUnix to match

### LINKED ISSUE #966: Replace FileDescriptor with File in PipeUnix

marcusnaslund:
Fixes #966


marcusnaslund:
@sebastianbaginski - do you know what `close` does? It doesn't even take an argument. Also, we don't call it where we use this in the other project.

source/io/native/PipeUnix.ooc
```diff
@@ -1,77 +1,72 @@
...
-		fd close()
+		0
+		//fd := _getFD(end)
+		//fd == 0 ? 0 : fd close()
 	}
 	close: func ~both {
-		readFD close()
-		writeFD close()
+		//readFD close()
+		//writeFD close()
```
marcusnaslund: @sebastianbaginski - do you know what `close` does? It doesn't even take an argument. Also, we don't call it where we use this in the other project.

sebastianbaginski: I think it calls `close(fd)` on a FileDescriptor object. Basically it just closes the file. See `man close`.

sebastianbaginski: So why did you remove it from here ? Also I think we should call it in the other project. Always clean up your stuff, you know. But the pipe will be closed by the OS when the process exits anyway, so not really a big problem.

marcusnaslund: > So why did you remove it from here ?

Oh, I'm not finished, hence no review flag. :) Just wanted your input. My point is that there is no `close(fd)`, just `close()`, so what does it really do? Or does rock throw in the argument automagically?

sebastianbaginski: I don't know, it should be more like `close(this)` in `FileDescriptor`

marcusnaslund:
Updated and ready for review.

PR #1165, opened 2016-01-07
added toText() to geometry classes

thomasfanell:
* added `toText()` to geometry classes

### DIFF
--- a/source/geometry/FloatBox2D.ooc
+++ b/source/geometry/FloatBox2D.ooc
@@ -129,6 +129,7 @@ FloatBox2D: cover {
 		This createAround(newCenter, newSize)
 	}
 	toString: func -> String { "#{this leftTop toString()}, #{this size toString()}" }
+	toText: func -> Text { this leftTop toText() + t", " + this size toText() }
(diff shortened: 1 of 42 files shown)


marcusnaslund:
These should be accompanied by tests as well to make sure they output the same as their string counterparts.

thomasfanell:
Some of them are not equal. Some of the toString's are sloppy and ugly, such as ending with trailing separators. Furthermore, some of them are formatted to print 8 decimal points and Text are only formatted to two decimal points. Should we create corresponding formatting options for Text then? (see `sdk/Format.ooc`)

marcusnaslund:
> Should we create corresponding formatting options for Text then? (see sdk/Format.ooc)

We should, but not in this PR. Eight decimals is probably in general unneccessary anyway. We should have matching tests for these functions anyhow, but it's alright if you want to do it later.

thomasfanell:
Well I can make tests that tests expected output, no problem... but now you know why I didn't include tests in this PR. :-)

thomasfanell:
I'll create some basic tests, hang on...

marcusnaslund:
Perfect. :) Readd the flag when you're done.

PR #1166, opened 2016-01-07
Fixed bug in Quaternion.

emilwestergren:
Creating a Quaternion from a zero vector will cause undefined behaviour so I added a check for this.

@niklasborwell peer review


marcusnaslund:
Aren't we doing something wrong if we're doing that?

niklasborwell:
Initially that's what we were thinking  too, but actually, in axis angle representation an identity rotation is represented by a zero vector.

emilwestergren:
What @niklasborwell said

marcusnaslund:
Alrighty then.

PR #1167, opened 2016-01-08
Renamed axis-angle to Euler vector in Quaternion

tomasmahlberg:
The Euler vector is the product of the axis and angle in the axis-angle representation and what we are using here.


tomasmahlberg:
@marcusnaslund

PR #1168, opened 2016-01-08
Removed more question marks

### LINKED ISSUE #1162: Remove question marks from (SDK) functions
There are also a lot in `source/net`, and in `CsvReader`, `Pipe`.

marcusnaslund:
Fixes #1162 

Remaining uses of `?` are `instanceOf?`, `inheritsFrom?`, `hasNext?` and `hasPrev?` which are hardcoded to be like that in rock... :( Don't know if we can do much about it @thomasfanell 


sebastianbaginski:
Looks ok.

PR #1170, opened 2016-01-08
Changed even more static methods in math.ooc

marcusnaslund:
These were a lot easier than I thought. The only static methods that now remain are `minimum` and `maximum`, they will be dealt with after the weekend.

Peer review @thomasfanell ?

### DIFF
--- a/source/plot/Axis.ooc
+++ b/source/plot/Axis.ooc
@@ -12,7 +12,7 @@ Axis: class {
 	min: Float { get set }
 	max: Float { get set }
 	tick: Float { get set (length) {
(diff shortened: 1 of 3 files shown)


thomasfanell:
@marcusnaslund Looks good. Although I think both `minimum` and `maximum` should be static. It makes more sense in my opinion to do `Int maximum` than say, `someInt maximum`. I think most people would expect them to be static.

marcusnaslund:
The `geometry` and `math` classes don't have static `maximum`/`minimum` functions, so changing it for numeric types would be more consistent.

PR #1171, opened 2016-01-08
Fixed VS Code filter

thomasfanell:
Fixed so every `.valgrindlog` file is blocked from search, not just the ones in the project root.


marcusnaslund:
How can print fail this?

marcusnaslund:
Shouldn't we also have `"**/*~": true` here as for the other project?

thomasfanell:
Probably, yes... i'll fix

PR #1172, opened 2016-01-11
Changed minimum and maximum of numeric types to non-static methods

### LINKED ISSUE #706: Clean and improve sdk/math.ooc

marcusnaslund:
Finally fixes #706

### DIFF
--- a/source/draw/Image.ooc
+++ b/source/draw/Image.ooc
@@ -67,7 +67,8 @@ Image: abstract class {
 		super()
 	}
 	resizeWithin: func (restriction: IntVector2D) -> This {
-		this resizeTo(((this size toFloatVector2D()) * Float minimum(restriction x as Float / this size x as Float, restriction y as Float / this size y as Float)) toIntVector2D())
+		restrictionFraction := (restriction x as Float / this size x as Float) minimum(restriction y as Float / this size y as Float)
+		this resizeTo((this size toFloatVector2D() * restrictionFraction) toIntVector2D())
 	}
 	resizeTo: abstract func (size: IntVector2D) -> This
 	resizeTo: virtual func ~withMethod (size: IntVector2D, method: InterpolationMode) -> This {
--- a/source/draw/RasterBgr.ooc
+++ b/source/draw/RasterBgr.ooc
@@ -94,8 +94,8 @@ RasterBgr: class extends RasterPacked {
 					if (c distance(o) > 0) {
 						maximum := o
 						minimum := o
-						for (otherY in Int maximum(0, y - this distanceRadius) .. Int minimum(y + 1 + this distanceRadius, this size y))
-							for (otherX in Int maximum(0, x - this distanceRadius) .. Int minimum(x + 1 + this distanceRadius, this size x))
+						for (otherY in 0 maximum(y - this distanceRadius) .. (y + 1 + this distanceRadius) minimum(this size y))
+							for (otherX in 0 maximum(x - this distanceRadius) .. (x + 1 + this distanceRadius) minimum(this size x))
 								if (otherX != x || otherY != y) {
 									pixel := (other as This)[otherX, otherY]
 									if (maximum blue < pixel blue)
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -97,8 +97,8 @@ RasterMonochrome: class extends RasterPacked {
 					if (c distance(o) > 0) {
 						maximum := o
 						minimum := o
-						for (otherY in Int maximum(0, y - 2) .. Int minimum(y + 3, this size y))
-							for (otherX in Int maximum(0, x - 2) .. Int minimum(x + 3, this size x))
+						for (otherY in 0 maximum(y - 2) .. (y + 3) minimum(this size y))
+							for (otherX in 0 maximum(x - 2) .. (x + 3) minimum(this size x))
 								if (otherX != x || otherY != y) {
 									pixel := (other as This)[otherX, otherY]
 									if (maximum y < pixel y)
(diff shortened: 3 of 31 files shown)


marcusnaslund:
Not quite ready but this is what we want right @simonmika ?

marcusnaslund:
Maybe @thomasfanell feels like doing some more peer reviewing?

simonmika:
Looks good

marcusnaslund:
@sebastianbaginski - do you have time to peer review?

sebastianbaginski:
Sure, I'll take a look.

source/draw/Image.ooc
```diff
@@ -69,7 +69,7 @@ Image: abstract class {
 		super()
 	}
 	resizeWithin: func (restriction: IntVector2D) -> This {
-		this resizeTo(((this size toFloatVector2D()) * Float minimum(restriction x as Float / this size x as Float, restriction y as Float / this size y as Float)) toIntVector2D())
+		this resizeTo(((this size toFloatVector2D()) * (restriction x as Float / this size x as Float) minimum(restriction y as Float / this size y as Float)) toIntVector2D())
```
sebastianbaginski: To be honest one can argue that previously this line was slightly more readable. What do you think ?

marcusnaslund: Hm, yes, but I think it was pretty terrible to begin with. I will split this up on two lines instead.

source/draw/RasterBgr.ooc
```diff
@@ -94,8 +94,8 @@ RasterBgr: class extends RasterPacked {
 					if (c distance(o) > 0) {
 						maximum := o
 						minimum := o
-						for (otherY in Int maximum(0, y - this distanceRadius) .. Int minimum(y + 1 + this distanceRadius, this size y))
-							for (otherX in Int maximum(0, x - this distanceRadius) .. Int minimum(x + 1 + this distanceRadius, this size x))
+						for (otherY in 0 maximum(y - this distanceRadius) .. (y + 1 + this distanceRadius) minimum(this size y))
```
sebastianbaginski: Maybe there is no problem with readability, but `for (x in 0 maximum(...) .. y)` looks quite uncommon. Maybe it will take some time until I get used to it.

source/draw/Image.ooc
```diff
@@ -69,7 +69,8 @@ Image: abstract class {
 		super()
 	}
 	resizeWithin: func (restriction: IntVector2D) -> This {
-		this resizeTo(((this size toFloatVector2D()) * Float minimum(restriction x as Float / this size x as Float, restriction y as Float / this size y as Float)) toIntVector2D())
+		restrictionFraction := (restriction x as Float / this size x as Float) minimum(restriction y as Float / this size y as Float)
+		this resizeTo((this size toFloatVector2D() * restrictionFraction) toIntVector2D())
```
marcusnaslund: There we go. @sebastianbaginski

source/draw/RasterMonochrome.ooc
```diff
@@ -97,8 +97,8 @@ RasterMonochrome: class extends RasterPacked {
 					if (c distance(o) > 0) {
 						maximum := o
 						minimum := o
-						for (otherY in Int maximum(0, y - 2) .. Int minimum(y + 3, this size y))
-							for (otherX in Int maximum(0, x - 2) .. Int minimum(x + 3, this size x))
+						for (otherY in 0 maximum(y - 2) .. (y + 3) minimum(this size y))
+							for (otherX in 0 maximum(x - 2) .. (x + 3) minimum(this size x))
```
sebastianbaginski:
```C++
--this->readability;
```

source/draw/RasterBgr.ooc
marcusnaslund: Yeah. We might not want `minimum` or `maximum` calls in `for`-calls to begin with.

sebastianbaginski:
Code in PR looks OK. Code in general, at least few examples when not so much :)

marcusnaslund:
> Code in general, at least few examples when not so much :)

Say again?

sebastianbaginski:
At least few examples in this PR where the code does not look so good.

source/draw/Image.ooc
sebastianbaginski: OK looks slightly better

sebastianbaginski:
Also, are you sure that there is no `-x func()` vs. `(-x) func()` bug anywhere now ?

marcusnaslund:
Those for-loops (as in `RasterMonochrome` and some other `Raster` classes) you mentioned are worse, I agree. But I'd rather see we calculate the loop limits before and assign them to variables. For example
```ooc
for (otherY in 0 maximum(y - this distanceRadius) .. (y + 1 + this distanceRadius) minimum(this size y))
```
should probably be
```ooc
yLoopMin := 0 maximum(y - this distanceRadius
yLoopMax := this size y minimum(y + 1 + this distanceRadius)
for (otherY in yLoopMin .. yLoopMax)
```

marcusnaslund:
> Also, are you sure that there is no -x func() vs. (-x) func() bug anywhere now ?

I've tried to pretty careful. I will go through it again and look specifically for that, thanks.

sebastianbaginski:
> I've tried to pretty careful. I will go through it again and look specifically for that, thanks.

OK, thanks. Review done.

PR #1173, opened 2016-01-11
Removed outdated TODOs

marcusnaslund:
Cleaned out TODOs that aren't useful, outdated, or simply won't be fixed.

### DIFF
--- a/source/base/Event.ooc
+++ b/source/base/Event.ooc
@@ -83,7 +83,7 @@ Event1: class <T> {
 	}
 }
 
-Event2: class <T0, T1> { // TODO: Write tests and fix this
+Event2: class <T0, T1> {
 	_head: Func(T0, T1)
 	_tail: This<T0, T1>
 	head ::= this _head
--- a/source/io/CsvReader.ooc
+++ b/source/io/CsvReader.ooc
@@ -2,8 +2,6 @@ use ooc-base
 use ooc-collections
 import io/[File, FileReader]
 
-// TODO: Handle multi-line string literals
-
 CsvReader: class extends Iterator<VectorList<Text>> {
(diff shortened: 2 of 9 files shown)


source/base/Event.ooc
```diff
@@ -83,7 +83,7 @@ Event1: class <T> {
 	}
 }
 
-Event2: class <T0, T1> { // TODO: Write tests and fix this
```
marcusnaslund: I just don't know what's left to fix.

source/io/CsvReader.ooc
```diff
@@ -2,8 +2,6 @@ use ooc-base
 use ooc-collections
 import io/[File, FileReader]
 
-// TODO: Handle multi-line string literals
```
marcusnaslund: Discussed in October that we have no use for this atm https://github.com/cogneco/ooc-kean/issues/566

test/math/FloatComplexTest.ooc
```diff
@@ -61,7 +61,6 @@ FloatComplexTest: class extends Fixture {
 			expect((complexNumber2 / complexNumber0) real, is equal to(2.6f) within(tolerance))
 			expect((complexNumber2 / complexNumber0) imaginary, is equal to(0.2f) within(tolerance))
 		})
-		// TODO: Add tests for compound assignment operators
```
marcusnaslund: We don't test operators for most classes/covers, which is a problem, but it's not specific to `FloatComplex`. I will create a more general issue.

marcusnaslund:
Comments @sebastianbaginski ?

sebastianbaginski:
No, I have nothing to add

PR #1174, opened 2016-01-11
Quaternion cleanup

emilwestergren:
@marcusnaslund @tomasmahlberg peer

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -37,65 +37,54 @@ Quaternion: cover {
 	conjugate ::= This new(this real, -this imaginary)
 	inverse ::= this conjugate / (this real squared + this imaginary norm squared)
 	transform ::= this toFloatTransform3D()
-	// NOTE: Separation into parts assumes order of application X -> Y -> Z
-	rotationX: Float {
-		get {
-			result: Float
-			value := this w * this y - this z * this x
-			if ((value abs() - 0.5f) abs() < This precision)
-				result = 0.0f
-			else
-				result = (2.0f * (this w * this x + this y * this z)) atan2(1.0f - 2.0f * (this x squared + this y squared))
-			result
-		}
-	}
-	rotationY: Float {
-		get {
-			result: Float
-			value := this w * this y - this z * this x
-			if ((value abs() - 0.5f) abs() < This precision)
-				result = value sign * (Float pi / 2.0f)
-			else
-				result = ((2.0f * value) clamp(-1, 1)) asin()
-			result
-		}
-	}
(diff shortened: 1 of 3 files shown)


source/geometry/Quaternion.ooc
```diff
@@ -94,12 +86,12 @@ Quaternion: cover {
 	exponential: This {
 		get {
```
marcusnaslund: `exponential: This { get { ` while you're at it

source/geometry/Quaternion.ooc
```diff
@@ -146,8 +138,7 @@ Quaternion: cover {
 	distance: func (other: This) -> Float { (this - other) norm }
 	angle: func (other: This) -> Float {
 		result := 2.0f * (this dotProduct(other) absolute) acos()
-		result = result == result ? result : 0.0f
-		result
+		result == result ? result : 0.0f
```
marcusnaslund: I believe `result isNumber` works instead of `result == result`

emilwestergren:
Hold on, I think I broke something, test is failing

source/geometry/Quaternion.ooc
emilwestergren: `result isNumber ? result : 0.0f` always returns `0.0f`

emilwestergren: It works if I add `isNumber ::= this == this` to `Float: cover...`. Must be messing up in the implicit cast or something.

marcusnaslund: Oh. We probably should, then.

emilwestergren:
Updated.

marcusnaslund:
@tomasmahlberg - any comments?

tomasmahlberg:
Nope, go ahead.

source/geometry/Quaternion.ooc
```diff
@@ -34,75 +34,57 @@ Quaternion: cover {
...
+	rotationZ: Float { get {
+		result: Float
+		value := this w * this y - this z * this x
+		if ((value abs() - 0.5f) abs() < This precision)
+			result = 2.0f * (this z atan2(this w))
+		else
+			result = (2.0f * (this w * this z + this x * this y)) atan2(1.0f - 2.0f * (this y squared + this z squared))
+		result
+	}}
+	direction ::= (this logarithm imaginary) normalized
```
marcusnaslund: Since this changed to a one-liner, it's probably better suited above line 40 with the others. After that, I'm happy.

emilwestergren:
Updated.

marcusnaslund:
Thank you.

PR #1175, opened 2016-01-11
Added script for test coverage analysis

### LINKED ISSUE #446: Code coverage in tests
E.g. `FloatBox2DTest` doesn't test any of the functions `union`, `intersection`, `bounds`, `createAround`, `adaptTo`... 

Is there any way we can detect this kind of thing automatically? A list of functions and properties in cover/class should be easy to extract from an `ooc` file, and at the very least, a tool for code coverage should be able to detect when any of those functions/properties are not used a single once in a test for that cover/class, especially when the test file is named after the cover/class itself. Would this be a suitable task for magic?

marcusnaslund:
Yes, bash is evil itself, but it's how we do things at the moment.

This script generates the directory `coverage` with coverage information on every file. We might want to do something similar for the other project?

(Finally) fixes #446

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ rock_tmp
 tests.use
 Tests
 *.valgrindlog
+/coverage/
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,6 +9,7 @@
 		".libs": true,
 		"rock_tmp": true,
 		".vscode/": true,
+		"coverage/": true,
 		"source/draw/stb_image": true,
 		"**/*.valgrindlog": true,
 		"**/*~": true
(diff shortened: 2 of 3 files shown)


marcusnaslund:
What do you think @sebastianbaginski ?

sebastianbaginski:
Do you have to build the binaries with `-fprofile-arcs -ftest-coverage` ?
I can't see why we shouldn't have this, nice.

marcusnaslund:
> Do you have to build the binaries with -fprofile-arcs -ftest-coverage ?

`--coverage` is an alias for that, and is the argument sent to `test.sh`.

sebastianbaginski:
OK.

marcusnaslund:
Right now we get coverage information for all `.c` files as well. Would be nice if we could hide that, but haven't figured out how yet.

PR #1176, opened 2016-01-11
Renamed createRotation in Quaternion

tomasmahlberg:
Renamed `createRotation` to `createFromAxisAngle` and removed usage of the `exponential` property as this just creates unnecessary overhead and makes the code harder to follow.


tomasmahlberg:
@emilwestergren review?

tomasmahlberg:
Hold on, need to fix the test.

tomasmahlberg:
Updated.

marcusnaslund:
So, why was `exponential` there to begin with if removing it affects nothing?

marcusnaslund:
Looks good.

PR #1177, opened 2016-01-11
Optimized weightedMean in Quaternion

### LINKED ISSUE #617: Evaluate performance of Quaternion weightedMean
The performance of the static function `weightedMean` in `Quaternion` needs to be evaluated and eventual optimizations should be made. Right now it uses `FloatMatrix` operations which could potentially be replaced by nested for loops. It should be evaluated if it would be worth sacrificing readability of the code if the performance is noticeably better.

tomasmahlberg:
Fixes #617. This speeds things up by about 40%. Saves some lines but looks just about as messy as the old code.

@marcusnaslund Do you think this is better?

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -240,42 +240,39 @@ Quaternion: cover {
 		// B - attitudeProfile
 		// q - result
 		// S - matrixQuantityS
-		// V - referenceVectors
 		// W - observationVectors
 		// Y - gibbsVector
 		// Z - vectorQuantityZ
 		// TODO: Fix singularity problem when the angle is close to PI, using sequential rotations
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Certainly not worse. 40% faster sounds very nice.

marcusnaslund:
How does this relate to https://github.com/cogneco/ooc-kean/issues/617 ?

tomasmahlberg:
@marcusnaslund  What do you mean? It seems quite obvious to me.

marcusnaslund:
> What do you mean? It seems quite obvious to me.

Oh, haha, sorry. I was wondering if this PR would close that issue, but didn't notice you already had written that. Let's blame it on the fact that it's monday. :)

tomasmahlberg:
Hehe, no problem. :)

PR #1179, opened 2016-01-11
Minor increase in test coverage + general cleanup

marcusnaslund:
While testing the new `lcov` tool provided in #1175 I'm updating some general test coverage and cleanup here and there. There's a lot of small mixed changes in this PR, but mostly just added tests.


source/sdk/threading/Mutex.ooc
```diff
@@ -59,15 +58,10 @@ MutexGlobal: class extends Mutex {
...
 RecursiveMutex: abstract class extends Mutex {
 	lock: abstract func
-
 	unlock: abstract func
 
-	with: func (f: Func) {
-		lock()
-		f()
-		unlock()
-	}
```
marcusnaslund: Already exists in `Mutex`

source/base/Debug.ooc
```diff
@@ -30,13 +30,13 @@ DebugLevel: enum {
 
 Debug: class {
 	_level: static DebugLevel = DebugLevel Everything
-	printFunction: static Func (String) = func (s: String) { println(s) }
+	_printFunction: static Func (String) = func (s: String) { println(s) }
```
marcusnaslund: No public fields in classes. If you want to change it, use `initialize`.

codecoverage.sh
```diff
@@ -11,3 +11,4 @@ fi
 lcov --quiet -t 'OOC Test coverage' -o $INFO_FILE -c --directory ./.libs/ooc/ --base-directory .
 genhtml --quiet -o coverage $INFO_FILE --num-spaces 4
 rm $INFO_FILE
+xdg-open 'coverage/index.html'
```
marcusnaslund: This seems to work well. Is there a better, more standardized way to open a webpage? @fredrikbryntesson ?

marcusnaslund:
Ready for peer review @thomasfanell or @sebastianbaginski , whoever is up for it.

sebastianbaginski:
Looks ok, @thomasfanell can you set a flag after you review (or without review)

codecoverage.sh
fredrikbryntesson: xdg-open is very standardized but maybe I check like this would be good?
if which xdg-open > /dev/null
then
  xdg-open 'coverage/index.html'
elif which gnome-open > /dev/null
then
  gnome-open 'coverage/index.html'
else
  echo "Could not detect the web browser to use."
fi

marcusnaslund: Sure, I'll do that, thanks.

PR #1180, opened 2016-01-11
Reinstated test for ThreadPool

marcusnaslund:
With PRInt, instead of Circle, this seems to run just fine.


marcusnaslund:
Well, no.

PR #1181, opened 2016-01-12
Removed as String operators from geometry covers

marcusnaslund:
As discussed yesterday, @thomasfanell - we don't do this anywhere else, the SDK hardly even does it, and we don't use it.

Peer review?


thomasfanell:
@marcusnaslund Agreed, looks good to me.

PR #1185, opened 2016-01-12
Uniform name of static free functions

### LINKED ISSUE #727: cleanup, dispose, clean...
Now that we've finally gotten rid of `__destroy__`, the next thing to fix is the naming of the various other functions we use for cleanup, among them `dispose`, `cleanup`, and `clean`. There may be reasons why they shouldn't all be called the same thing (e.g. `free`), but generally, you shouldn't have to guess which one is used for a particular class.

marcusnaslund:
Fixes #727


marcusnaslund:
Peer review @thomasfanell ?

thomasfanell:
@marcusnaslund good riddance

PR #1191, opened 2016-01-13
Modified static uses of non-static methods

marcusnaslund:
In ooc, it appears you can use non-static methods as static methods, which is either a bug or a feature. Anyhow, this is more in line with other `maximum`/`minimum` calls of our geometry covers and with #1172 

@sebastianbaginski ?


sebastianbaginski:
> which is either a bug or a feature. 

I think it just generates the same code, passing `this` as first parameter of function call. All methods are kind of static in C code anyway :)

PR #1192, opened 2016-01-13
Updated pure getter properties to style guidelines

marcusnaslund:
Peer review @sebastianbaginski or @thomasfanell ?

### DIFF
--- a/source/sdk/io/File.ooc
+++ b/source/sdk/io/File.ooc
@@ -25,11 +25,9 @@ File: abstract class {
 	extension: String { get { getExtension() } }
 	nameWithoutExtension: String { get { getNameWithoutExtension() } }
 	pathWithoutExtension: String { get { getPathWithoutExtension() } }
-	children: VectorList<This> {
-		get {
-			getChildren()
-		}
-	}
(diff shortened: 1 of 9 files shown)


sebastianbaginski:
It looks OK.

source/sdk/io/File.ooc
```diff
@@ -25,11 +25,9 @@ File: abstract class {
 	extension: String { get { getExtension() } }
 	nameWithoutExtension: String { get { getNameWithoutExtension() } }
 	pathWithoutExtension: String { get { getPathWithoutExtension() } }
-	children: VectorList<This> {
-		get {
-			getChildren()
-		}
-	}
+	children: VectorList<This> { get {
+		getChildren()
```
sebastianbaginski: `This` or `this` ?

marcusnaslund: Yes, but that's missing in 1000 other places in this file. We want to enforce `this`, so I will be adding it in some future PR. For now, this does exactly what it used to do.

sebastianbaginski: OK

Issue #747, opened 2015-11-12
Is it possible to be smarter in Text slice()?


marcusnaslund:
Is there really a smarter way, which this TODO comment is looking for in `Text.ooc`?
```ooc
if (this _buffer owner == Owner Receiver)
	result = result copy() // TODO: Could we be smarter here?
```
Creating this issue so it's not forgotten. If not, the comment should be removed.

simonmika:
It requires the introduction of a reference counter. It is possible but still unclear whether it is something we want.

PR #1196, opened 2016-01-13
We don't have to be smarter (in Text) right now

marcusnaslund:
Removing this `TODO` and in the process closing #747 

We don't seem to need this now, and if we ever want to make this better, by introducing a reference counter (see the issue), we will create an issue *then*, not have one that just lingers.

Does that seem OK @sebastianbaginski ?


sebastianbaginski:
Yes, seems ok.

PR #1198, opened 2016-01-14
Removed inline and const keywords

### LINKED ISSUE #1197: Evaluate if the inline keyword makes any difference
It seems like `inline`, like `const`, makes little or no difference in generated code. This should be determined for sure. 

If that is the case, `inline` keywords in the source (only in `sdk`) should be removed.

marcusnaslund:
Rock just rips out the `const` keyword anyway, and `inline` won't really help `gcc` with the code that's generated anyway.

Fixes #1197 

@sebastianbaginski ?

### DIFF
--- a/source/sdk/structs/HashMap.ooc
+++ b/source/sdk/structs/HashMap.ooc
@@ -44,8 +44,8 @@ murmurHash: func <K> (keyTagazok: K) -> SizeT {
 	seed: SizeT = 1
 
 	len := K size
-	m = 0x5bd1e995 : const SizeT
-	r = 24 : const SSizeT
+	m = 0x5bd1e995: SizeT
+	r = 24: SSizeT
 	l := len
 
 	h : SizeT = seed ^ len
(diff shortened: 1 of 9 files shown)


source/sdk/structs/HashMap.ooc
```diff
@@ -44,8 +44,8 @@ murmurHash: func <K> (keyTagazok: K) -> SizeT {
 	seed: SizeT = 1
 
 	len := K size
-	m = 0x5bd1e995 : const SizeT
-	r = 24 : const SSizeT
```
sebastianbaginski: So it does completely nothing ? Can you even reassign the value after initialization ?

marcusnaslund: Exactly.

sebastianbaginski: What a shame it's so broken...

sebastianbaginski:
Well, I guess there is no point in keeping this code then. I can understand that `const` is not propagated to C code, but at least rock could refuse to compile code which reassigns a `const` variable...

PR #1199, opened 2016-01-14
Fixed bug.

emilwestergren:
* Increased size of bin - I realized it is not uncommon to hold around 100 images when recording 120/240 fps
* Fixed bug when modifying a VectorList while iterating over it

@marcusnaslund peer review


marcusnaslund:
There's no penalty when using videos of lower framerate?

emilwestergren:
No, if everything works we wouldn't even need a limit.

PR #1201, opened 2016-01-14
Removed gc_ prefix from C memory routines

marcusnaslund:
The `gc_` prefix makes no sense.

`gc_malloc(...) -> calloc(1, ...)` (or if there's a better way to write it)
`gc_calloc(..., ...) -> calloc(..., ...)`
`gc_malloc_atomic(...) -> malloc(...)`

`gc_free` has been renamed to `memfree` to make all these calls easier to find. If you have another suggestion, let me know.

After this, these calls should be replaced by the use of our `Buffer` cover which will handle memory management in an OOP manner.

### DIFF
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -38,7 +38,7 @@ ByteBuffer: class {
 			this _referenceCount free()
 		this _referenceCount = null
 		if (this _ownsMemory)
-			gc_free(this _pointer)
+			memfree(this _pointer)
 		this _pointer = null
 		super()
 	}
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -127,7 +127,7 @@ _RecyclableByteBuffer: class extends ByteBuffer {
 		}
 		This _lock unlock()
 		version(debugByteBuffer) { if (buffer == null) Debug print("No RecyclableByteBuffer available in the bin; allocating a new one") }
-		buffer == null ? This new(gc_malloc_atomic(size), size) : buffer
+		buffer == null ? This new(malloc(size), size) : buffer
 	}
 	_getBin: static func (size: Int) -> VectorList<This> {
 		size < 10000 ? This _smallRecycleBin : (size < 100000 ? This _mediumRecycleBin : This _largeRecycleBin)
--- a/source/sdk/Memory.ooc
+++ b/source/sdk/Memory.ooc
@@ -11,8 +11,7 @@ memset: extern func (Pointer, Int, SizeT) -> Pointer
 memcmp: extern func (Pointer, Pointer, SizeT) -> Int
 memmove: extern func (Pointer, Pointer, SizeT)
 memcpy: extern func (Pointer, Pointer, SizeT)
-free: extern func (Pointer)
-gc_free: extern (free) func (Pointer)
+memfree: extern (free) func (Pointer)
 alloca: extern func (SizeT) -> Pointer
 
 // note: sizeof is intentionally not here. sizeof(Int) will be translated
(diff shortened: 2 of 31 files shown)


source/sdk/Memory.ooc
```diff
@@ -2,17 +2,8 @@ include string // Memory routines in C are actually in string.h
 version(windows) { include malloc }
 else { include alloca }
 
-// GC_MALLOC zeroes the memory, so in the non-gc version, we prefer to use calloc
-// to the expense of some performance. Don't use malloc instead - existing ooc code
-// is written assuming that gc_malloc zeroes the memory.
-gc_malloc: func (size: SizeT) -> Pointer { gc_calloc(1, size) }
```
sebastianbaginski: Why do we do this ? Previously it was possible to redefine `gc_calloc` and `gc_free` like
```ooc
gc_calloc: func (nmemb, size: SizeT) -> Pointer {
    result := calloc(nmemb, size)
    "%i bytes allocated at address %i" printfln(nmemb, result)
    result
}

gc_free: func (address: Pointer) {
    "free memory at address %i" printfln(address)
    free(address)
}
```
or add another form of tracking / logging memory allocations. I did that at least once.
I think it won't be possible after this change.

davidpiuva: I also need the ability to track allocations when debugging. I would rather expand the functionality with debug calls each time memory is allocated or freed so that I can find out if an object exists when crashing.

marcusnaslund: The `gc_` prefix is outdated and in the long run misleading. The `ooc_` prefix doesn't make for a great name either, especially considering we might not call this language `ooc` for much longer.

I don't see why you can't do your memory debugging. Just change `malloc: extern func (SizeT) -> Pointer` to `_malloc: extern (malloc) func (SizeT) -> Pointer` and then add your own 

```ooc
malloc: func (size: SizeT) -> Pointer { // do whatever }
```

davidpiuva: If we only have one name then how do I use dynamic memory within the function that allocates memory for the main uses?

marcusnaslund: Sorry, I don't understand your question.

davidpiuva: When I try to use println from gc_malloc, I always get a crash. There should be a way to prevent infinite recursion when using dynamic memory from within the function that gives dynamic memory to the other callers.

marcusnaslund: Gotcha. I've never tried to do what you're suggesting, but I don't see how it would be different with `_malloc` as extern and `malloc` instead of `malloc` as extern and `gc_malloc`.

davidpiuva: We would need some kind of global flag that is set to true when entering the inner functions so that recursive calls from built in allocation is not debug printing again from the string operations. There might be other things preventing the use of println. It would be nice to have 2 debug functions where any code can be used to track memory errors. For classes, it would be very useful to get the class name and count of each leak.

marcusnaslund:
@fredrikbryntesson @sebastianbaginski - is `gc_free` -> `memfree` a good name or do you have something better in mind?

sebastianbaginski:
Sounds good enough to me. I'll let you and @fredrikbryntesson decide.

fredrikbryntesson:
Seems fine.

marcusnaslund:

source/base/ByteBuffer.ooc
```diff
@@ -127,7 +127,7 @@ _RecyclableByteBuffer: class extends ByteBuffer {
 		}
 		This _lock unlock()
 		version(debugByteBuffer) { if (buffer == null) Debug print("No RecyclableByteBuffer available in the bin; allocating a new one") }
-		buffer == null ? This new(gc_malloc_atomic(size), size) : buffer
+		buffer == null ? This new(malloc(size), size) : buffer
```
sebastianbaginski: what happened with atomic ?

marcusnaslund: See the PR description

PR #1211, opened 2016-01-18
Removed unused variables and most gcc warnings

marcusnaslund:
In preparation for enforcing `-Wall` and later `-Werror` in tests and builds.

### DIFF
--- a/test/draw/gpu/GpuSurfaceTest.ooc
+++ b/test/draw/gpu/GpuSurfaceTest.ooc
@@ -86,9 +86,6 @@ GpuSurfaceTest: class extends Fixture {
 			gpuImage canvas clear()
 			viewport := gpuImage canvas viewport
 			redBox := IntBox2D new(viewport left, viewport top, viewport width / 2, viewport height / 2)
-			yellowBox := IntBox2D new(viewport width / 2, viewport top, viewport width / 2, viewport height / 2)
-			blueBox := IntBox2D new(viewport left, viewport height / 2, viewport width / 2, viewport height / 2)
-			greenBox := IntBox2D new(viewport width / 2, viewport height / 2, viewport width / 2, viewport height / 2)
 			gpuImage canvas draw(sourceImage, redBox, viewport)
 			rasterFromGpu := gpuImage toRaster()
 			expect(rasterFromGpu distance(correctImage), is equal to(0.0f))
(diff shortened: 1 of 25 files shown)


test/draw/gpu/GpuSurfaceTest.ooc
```diff
@@ -86,9 +86,6 @@ GpuSurfaceTest: class extends Fixture {
 			gpuImage canvas clear()
 			viewport := gpuImage canvas viewport
 			redBox := IntBox2D new(viewport left, viewport top, viewport width / 2, viewport height / 2)
-			yellowBox := IntBox2D new(viewport width / 2, viewport top, viewport width / 2, viewport height / 2)
-			blueBox := IntBox2D new(viewport left, viewport height / 2, viewport width / 2, viewport height / 2)
-			greenBox := IntBox2D new(viewport width / 2, viewport height / 2, viewport width / 2, viewport height / 2)
```
marcusnaslund: @emilwestergren Maybe you know if these should be included in some test or if it's safe to remove them. They're currently not used.

marcusnaslund:

test/draw/gpu/GpuSurfaceTest.ooc
emilwestergren: I think you can remove them if they're not used.

marcusnaslund:
One remaining warning is:
```
ooc-kean/source/sdk/Exception.ooc: In function ‘Exception___popStackFrame’:
ooc-kean/source/sdk/Exception.ooc:80:1: warning: control reaches end of non-void function [-Wreturn-type]
 _getException: func -> Exception { exceptionContext exception }
```
Which doesn't make sense to me.

sebastianbaginski:
Nicely done

sebastianbaginski:
I don't know about this warning. But maybe we could disable it for this file.

Can you check if gcc diagnostic pragmas works in ooc ? Try to add something like
```ooc
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wreturn-type"

_getException: func -> Exception { exceptionContext exception }

#pragma GCC diagnostic pop
```
around `__getException` or even around the whole Exception file.

marcusnaslund:
Thanks, I will try it tomorrow.

sebastianbaginski:
OK. I'm not sure if there is a way to pass a C code "as-is" to the C backend ?

Issue #1212, opened 2016-01-18
Remove lingering gcc warnings


marcusnaslund:
`PRInt` is now running with `-Wall`, but before we can move on to `-Werror`, we need to fix these issues:

- [x] `source/sdk/Exception.ooc:80:1: warning: control reaches end of non-void function [-Wreturn-type]`
` _getException: func -> Exception { exceptionContext exception }`
Might be fixable by setting `#pragma GCC diagnostic` around that function or the entire file.

- [x] `source/draw/gpu/GpuSurface.ooc:93:1: warning: control reaches end of non-void function [-Wreturn-type]`
`_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D) -> FloatTransform3D`
This one is also mysterious, but haven't looked into it too much.

- [x] `test/io/CsvReaderTest.ooc:47:37: warning: unused variable ‘row’ [-Wunused-variable]`
`reader := CsvReader open(filename) map(VectorList<Float>, |row| This toFloats)`
Is there an ooc trick for this or should we simply not use closures in this way since they leak? @thomasfanell 

- [ ] Warnings in `draw` related to the enforcement of the `override` keyword. Something like:
```
source/draw/gpu/GpuImage.ooc: In function ‘GpuImage__GpuImage_distance’:
/home/mnaslund/versioned/ooc/ooc-kean/source/draw/gpu/GpuImage.ooc:61:5: warning: passing argument 2 of ‘((struct Image__ImageClass *)((struct lang_types__Object *)this)->class)->distance’ from incompatible pointer type [enabled by default]
source/draw/gpu/GpuImage.ooc:61:5: note: expected ‘struct Image__Image *’ but argument is of type ‘struct GpuImage__GpuImage *’
source/draw/gpu/GpuContext.ooc: In function ‘GpuContext__GpuContext_createYuv420Semiplanar_fromImages’:
/home/mnaslund/versioned/ooc/ooc-kean/source/draw/gpu/GpuContext.ooc:61:5: warning: passing argument 2 of ‘((struct AbstractContext__AbstractContextClass *)((struct lang_types__Object *)this)->class)->createYuv420Semiplanar_fromImages’ from incompatible pointer type [enabled by default]
source/draw/gpu/GpuContext.ooc:61:5: note: expected ‘struct Image__Image *’ but argument is of type ‘struct GpuImage__GpuImage *’
[...]

PR #1214, opened 2016-01-18
Added -Wall to test script

marcusnaslund:
Also fixes some of the points of #1212 

`PRInt` already runs with this flag (as of today). Once we've fixed these warnings, `PRInt` will switch to `-Werror` and not allow any new warnings.

Since this is a library, it would have been nice with `-Wall -Wextra`, but that just fills the screen with rock-related nonsense we will never fix.

### DIFF
--- a/source/sdk/Exception.ooc
+++ b/source/sdk/Exception.ooc
@@ -76,7 +76,7 @@ ExceptionContext: class {
 exceptionContext := ExceptionContext new()
 
 _pushStackFrame: func -> StackFrame { exceptionContext pushStackFrame() }
-_popStackFrame: func -> StackFrame { exceptionContext popStackFrame() }
+_popStackFrame: func { exceptionContext popStackFrame() }
 _getException: func -> Exception { exceptionContext exception }
 
 version(windows) {
--- a/test/io/CsvReaderTest.ooc
+++ b/test/io/CsvReaderTest.ooc
@@ -42,15 +42,6 @@ CsvReaderTest: class extends Fixture {
 			filename free()
 			reader free()
 		})
-		this add("map", func {
-			filename := Text new(c"test/io/input/3x3.csv", 21)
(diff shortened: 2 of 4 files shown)


source/sdk/Exception.ooc
```diff
@@ -76,7 +76,7 @@ ExceptionContext: class {
 exceptionContext := ExceptionContext new()
 
 _pushStackFrame: func -> StackFrame { exceptionContext pushStackFrame() }
-_popStackFrame: func -> StackFrame { exceptionContext popStackFrame() }
+_popStackFrame: func { exceptionContext popStackFrame() }
```
marcusnaslund: @sebastianbaginski This was the problem. Why does `push` return, but not `pop`?

test/io/CsvReaderTest.ooc
```diff
@@ -42,15 +42,6 @@ CsvReaderTest: class extends Fixture {
...
 		})
-		this add("map", func {
-			filename := Text new(c"test/io/input/3x3.csv", 21)
-			reader := CsvReader open(filename) map(VectorList<Float>, |row| This toFloats)
-			filename free()
-			reader free()
-		})
-	}
-	toFloats: static func (row: VectorList<Text>) -> VectorList<Float> {
-		row map(|value| value toFloat())
```
marcusnaslund: @thomasfanell I just removed the entire test, is that OK?

thomasfanell: @marcusnaslund yes

source/sdk/Exception.ooc
sebastianbaginski: I think these functions are used only in rock, it was like that even before I worked on this.
`pop` here means simply `take first and delete it`. Internally rock uses the last pushed stack frame (`source/rock/middle/Try.ooc`), so I think `push` returns only for convenience.

PR #1217, opened 2016-01-19
Added code coverage extension functionality to VSCode

marcusnaslund:
Unfortunately, the json file must be located in the root. :/
I don't know how you want to integrate this with magic etc. later, but this works for now. @thomasfanell - peer review?


thomasfanell:
@marcusnaslund nice :)

PR #1218, opened 2016-01-19
Moved all data from `Canvas` to a new `Brush` class.

davidpiuva:
An experiment in changing the Draw API without breaking backward compability just to explore which options are possible.

Goals:
* To not have unwanted side effects between filters when color and projection changes before a draw call.
* To only need two draw methods where one is a simple wrapper on the other (Default settings or given from another Brush). Another canvas can be used as a brush by inheritance if you just gave the same settings there. If no Brush is given to the new draw call, the called canvas itself will be the brush.

Challenges:
* One settings object for both CPU and GPU.
* Abstraction while still working with OpenGL and CPU.

### DIFF
--- a/ooc-draw.use
+++ b/ooc-draw.use
@@ -22,6 +22,7 @@ Imports: RasterYuv420Semiplanar
 Imports: RasterYuv422Semipacked
 Imports: FloatImage
 Imports: Canvas
+Imports: Brush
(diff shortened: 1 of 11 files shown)


marcusnaslund:
Sorry, conflicts due to the removal of `ooc` in all use files.

PR #1225, opened 2016-01-19
Include limits header when targeting OS X

thomasfanell:
* The header `limits` was not included when targeting OS X, causing the build to fail.

@marcusnaslund peer review?


marcusnaslund:
This might be the first OS X-related bug report here.

PR #1226, opened 2016-01-20
Draw API

davidpiuva:
An alternative draw API that can be used next to the old API for backward compability.

The only usability problem so far is that you can forget to call `draw` after creating the `DrawState`.

Remaining steps for other pull requests:
* Implement drawing of points, lines and meshes using the Pen cover, source viewport and other new settings.
* Support drawing without GPU.
* Make lots of test cases.
* Slowly get everyone to use the new API while testing everything properly.
* Clean up `pen`, `transform`, `opacity`, `blend`, `viewport`,... from canvas where they do not belong.
* Give matrices in normalized coordinates directly to the vertex shader to reduce the double, tripple, quadruple,... conversions between reference and normalized coordinates.

### DIFF
--- a/source/draw/Canvas.ooc
+++ b/source/draw/Canvas.ooc
@@ -17,7 +17,9 @@
 use base
 use geometry
 use collections
-import Image, Pen, DrawState
+import Image
+import Pen
+import DrawState
 
 InterpolationMode: enum {
 	Fast // nearest neighbour
--- a/source/draw/Canvas.ooc
+++ b/source/draw/Canvas.ooc
@@ -17,7 +17,7 @@
 use base
 use geometry
 use collections
-import Image, Pen
+import Image, Pen, DrawState
 
 InterpolationMode: enum {
 	Fast // nearest neighbour
@@ -68,6 +68,7 @@ Canvas: abstract class {
 		positions free()
 	}
 	fill: abstract func
+	draw: virtual func ~DrawState (drawState: DrawState) { Debug raise("draw~DrawState unimplemented!") }
 	draw: abstract func ~ImageSourceDestination (image: Image, source, destination: IntBox2D)
 	draw: func ~ImageDestination (image: Image, destination: IntBox2D) { this draw(image, IntBox2D new(image size), destination) }
 	draw: func ~Image (image: Image) { this draw(image, IntBox2D new(image size)) }
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -15,7 +15,9 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use geometry
-import Pen, Image, Map
+import Pen
+import Image
+import Map
 
 // Example
 // DrawState new(myImage) setMap(myShader) draw()
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -1,0 +1,64 @@
+//
+// Copyright (c) 2011-2014 Simon Mika <simon@mika.se>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use geometry
+import Pen, Image, Map
+
+// Example
+// DrawState new(myImage) setMap(myShader) draw()
+
+DrawState: cover {
+	target: Image = null
+	inputImage: Image = null
+	map: Map = null
+	opacity := 1.0f
+	_transformNormalized := FloatTransform3D identity
+	viewport := IntBox2D new(0, 0, 0, 0)
+	init: func@ ~default
+	init: func@ ~target (=target)
+	setTarget: func (target: Image) -> This {
+		this target = target
+		this
+	}
+	setMap: func (map: Map) -> This {
+		this map = map
+		this
+	}
+	setOpacity: func (opacity: Float) -> This {
+		this opacity = opacity
+		this
+	}
+	setViewport: func (viewport: IntBox2D) -> This {
+		this viewport = viewport
+		this
+	}
+	setInputImage: func (inputImage: Image) -> This {
+		this inputImage = inputImage
+		this
+	}
+	setTransformReference: func ~targetSize (transform: FloatTransform3D) -> This {
+		this setTransformNormalized(transform referenceToNormalized(this target size))
+	}
+	setTransformReference: func ~explicit (transform: FloatTransform3D, imageSize: FloatVector2D) -> This {
+		this setTransformNormalized(transform referenceToNormalized(imageSize))
+	}
+	setTransformNormalized: func (transform: FloatTransform3D) -> This {
+		this _transformNormalized = transform
+		this
+	}
+	getTransformNormalized: func -> FloatTransform3D { this _transformNormalized }
+	draw: func { this target canvas draw(this) }
+}
--- a/source/draw/gpu/opengl/OpenGLCanvas.ooc
+++ b/source/draw/gpu/opengl/OpenGLCanvas.ooc
@@ -29,14 +29,10 @@ OpenGLCanvas: class extends OpenGLSurface {
 	_renderTarget: GLFramebufferObject
 	context ::= this _context as OpenGLContext
 	draw: override func ~DrawState (drawState: DrawState) {
-		// Set shader
-		gpuMap: GpuMap = (drawState map == null) ? this context defaultMap : drawState map as GpuMap
-		// Set viewport
+		gpuMap: GpuMap = drawState map as GpuMap ?? this context defaultMap
 		viewport := (drawState viewport hasZeroArea) ? IntBox2D new(this size) : drawState viewport
 		this context backend setViewport(viewport)
-		// Set transform
 		gpuMap view = _toLocal * drawState getTransformNormalized() normalizedToReference(this size) * _toLocal
-		// Set model and projection matrices
 		if (this _focalLength > 0.0f) {
 			a := 2.0f * this _focalLength / this size x
 			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
@@ -46,15 +42,12 @@ OpenGLCanvas: class extends OpenGLSurface {
 		} else
 			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
 		gpuMap model = this _createModelTransform(IntBox2D new(this size))
-		// Set opacity
 		if (drawState opacity < 1.0f)
 			this context backend blend(drawState opacity)
 		else
 			this context backend enableBlend(false)
-		// Set texture
 		if (drawState inputImage)
 			gpuMap add("texture0", drawState inputImage)
-		// Draw
 		gpuMap use()
 		this _bind()
 		this context drawQuad()
--- a/source/draw/gpu/opengl/OpenGLCanvas.ooc
+++ b/source/draw/gpu/opengl/OpenGLCanvas.ooc
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
 	_target: OpenGLPacked
 	_renderTarget: GLFramebufferObject
 	context ::= this _context as OpenGLContext
+	draw: override func ~DrawState (drawState: DrawState) {
+		// Set shader
+		gpuMap: GpuMap = (drawState map == null) ? this context defaultMap : drawState map as GpuMap
+		// Set viewport
+		viewport := (drawState viewport hasZeroArea) ? IntBox2D new(this size) : drawState viewport
+		this context backend setViewport(viewport)
+		// Set transform
+		gpuMap view = _toLocal * drawState getTransformNormalized() normalizedToReference(this size) * _toLocal
+		// Set model and projection matrices
+		if (this _focalLength > 0.0f) {
+			a := 2.0f * this _focalLength / this size x
+			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
+			k := (this _farPlane + this _nearPlane) / (this _farPlane - this _nearPlane)
+			o := 2.0f * this _farPlane * this _nearPlane / (this _farPlane - this _nearPlane)
+			gpuMap projection = FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
+		} else
+			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
+		gpuMap model = this _createModelTransform(IntBox2D new(this size))
+		// Set opacity
+		if (drawState opacity < 1.0f)
+			this context backend blend(drawState opacity)
+		else
+			this context backend enableBlend(false)
+		// Set texture
+		if (drawState inputImage)
+			gpuMap add("texture0", drawState inputImage)
+		// Draw
+		gpuMap use()
+		this _bind()
+		this context drawQuad()
+		this _unbind()
+	}
 	init: func (=_target, context: OpenGLContext) {
 		super(this _target size, context, context defaultMap, IntTransform2D identity)
 		this _renderTarget = context _backend createFramebufferObject(this _target _backend as GLTexture, this _target size)
--- a/draw.use
+++ b/draw.use
@@ -24,6 +24,7 @@ Imports: FloatImage
 Imports: Canvas
 Imports: Pen
 Imports: Map
+Imports: DrawState
 Additionals: source/draw/stb_image/stb_image.h
 Additionals: source/draw/stb_image/stb_image_write.h
 Additionals: source/draw/stb_image/stb_image_impl.c
--- a/source/draw/RasterCanvas.ooc
+++ b/source/draw/RasterCanvas.ooc
@@ -20,6 +20,7 @@ use collections
 import Image
 import Canvas
 import Pen
+import DrawState
 import RasterImage
 
 RasterCanvas: abstract class extends Canvas {


marcusnaslund:
Sorry, conflicts due to the removal of `ooc` in all use files.

davidpiuva:
I think this prototype is ready for a first evaluation to get early feedback. @emilwestergren @sebastianbaginski @simonmika

The first commit will be removed before the real PR since the caller always knows the target resolution.

source/draw/Map.ooc
```diff
@@ -0,0 +1,24 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use geometry
+
+Map: abstract class {
```
davidpiuva: This is to allow passing shader arguments in GPU independent code. Since no methods are made virtual, it should have no performance impact.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,89 @@ OpenGLCanvas: class extends OpenGLSurface {
 	_target: OpenGLPacked
 	_renderTarget: GLFramebufferObject
 	context ::= this _context as OpenGLContext
+	// TODO: Remove once rendering is done in normalized coordinates
```
davidpiuva: Converting to reference again in just to decode back to normalized in the vertex shader is messy but first it should start working on top of existing solutions since transform is harder to test.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,89 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		// Set texture
+		if (drawState intputImage)
+			gpuMap add("texture0", drawState intputImage)
+		// Draw
+		gpuMap use()
+		this _bind()
+		this context drawQuad()
+		this _unbind()
+	}
+	draw: override func (action: Func) {
```
davidpiuva: The below methods are just moved from another class to clean up. This might become a separate PR.

source/draw/gpu/GpuMap.ooc
```diff
@@ -61,12 +62,8 @@ _Map: class {
 	}
 }
 version(!gpuOff) {
-GpuMap: abstract class {
+GpuMap: abstract class extends Map {
 	_bindings := _Map new()
-	model: FloatTransform3D { get set }
-	view: FloatTransform3D { get set }
-	projection: FloatTransform3D { get set }
-	textureTransform: FloatTransform3D { get set }
```
davidpiuva: Moved to `Map`.

davidpiuva:
@bettinaselig Peer-review usability heuristics?
@emilwestergren Peer-review code quality?

bettinaselig:
What's the difference for the user? Example?

source/draw/DrawState.ooc
```diff
@@ -0,0 +1,83 @@
...
+DrawState: cover {
+	target: Image = null
+	intputImage: Image = null
+	map: Map = null
+	opacity := 1.0f
+	_transformNormalized := FloatTransform3D identity
+	viewportLocal := IntBox2D new(0, 0, 0, 0)
+	init: func@ ~default
+	init: func@ ~target (=target)
+	setTarget: func (target: Image) -> This {
```
emilwestergren: It gets kind of bloated with all these setters and it's just so you can stack them on a single line right?

davidpiuva:
You are one of the users of the API.

source/draw/DrawState.ooc
```diff
@@ -0,0 +1,83 @@
...
+
+use geometry
+import Pen, Image, Map
+
+// Example
+// DrawState new(myImage) setMap(myShader) draw()
+
+DrawState: cover {
+	target: Image = null
+	intputImage: Image = null
```
emilwestergren: `inputImage`

bettinaselig:
Yes, I know. But what is the difference for me? Could you make an example?

source/draw/DrawState.ooc
davidpiuva: Setters per reference is unpredictable on value types since you might modify a clone and a class would have to be freed.

davidpiuva:
See https://github.com/vidhance/ooc-vidproc/pull/1403/files

In the old API, you change settings in a canvas that is shared between different filters along the graph.
`gpuOriginal uv canvas draw(view, viewportHalf, This _fillUVMap)`

In the new API, you create a `DrawState`, modify settings and call `draw` on the resulting `DrawState`.
`DrawState new(gpuOriginal uv) setViewport(viewportHalf) setMap(This _fillUVMap) draw()`

source/draw/DrawState.ooc
davidpiuva: Updated

source/draw/DrawState.ooc
```diff
@@ -0,0 +1,83 @@
...
+		result := this
+		result viewportLocal = viewport
+		result
+	}
+	setInputImage: func (inputImage: Image) -> This {
+		result := this
+		result inputImage = inputImage
+		result
+	}
+	_toNormalized: static func (imageSize: FloatVector2D, transform: FloatTransform3D) -> FloatTransform3D {
```
davidpiuva: Is there a place for these conversion functions that I have everywhere in my code?

davidpiuva:
Updated with more compact setters for reference transforms. I still have the other setters to compress.

bettinaselig:
Thanks. Ok with me.

sebastianbaginski:
I think you can write setters for covers in more compact way like
```ooc
setInputImage: func (inputImage: Image) -> This {
	this inputImage = inputImage
	this
}
```
If we allow semicolons, then you can even compact it to a single line.

source/draw/Map.ooc
sebastianbaginski: Do you think you could make a separate PR with this abstract `Map` ?

source/draw/gpu/GpuMap.ooc
sebastianbaginski: This is a good change, I think it deserves separate PR.

source/draw/Map.ooc
davidpiuva: It would take some time to remake commits but I could.

davidpiuva:
I start breaking out the `Map` change.

sebastianbaginski:
OK, thanks

davidpiuva:
@emilwestergren You can continue reviewing.

davidpiuva:
@sebastianbaginski Emil seems to be busy. Can you continue reviewing?

source/draw/DrawState.ooc
```diff
@@ -0,0 +1,79 @@
...
+	opacity := 1.0f
+	_transformNormalized := FloatTransform3D identity
+	viewport := IntBox2D new(0, 0, 0, 0)
+	init: func@ ~default
+	init: func@ ~target (=target)
+	setTarget: func (target: Image) -> This {
+		result := this
+		result target = target
+		result
+	}
```
sebastianbaginski: Maybe compact them to 
```
set* : func( x ) -> This {
    this x = x
    this
}
```
or 
```
set* : func( x ) -> This { this x = x ; this }
```

davidpiuva: Then we might get a performance problem instead. The current solution is made to work with function inlining and constant propagation.

davidpiuva: Do you mean that I should just call that directly without using function pointer?

sebastianbaginski: I mean the `result := this` part is not neccessary, `this` is already passed as a copy because the call is made on a cover (and the function signature does not have a reference indicator `@`).

davidpiuva: It is against coding convention but okay.

sebastianbaginski: You mean the semicolon ? Yeah I'm not sure about this, @marcusnaslund ?
Otherwise modifying `this` it is widely used, for example in `Text` cover.

davidpiuva: Simon said that return values in functions should be named result unless they can be written as a single line expression but that was a year ago.

source/draw/DrawState.ooc
sebastianbaginski: Good question. Static functions in `FloatTransform3D` ?

source/draw/DrawState.ooc
sebastianbaginski: OK. I think he wrote some parts of `Text` cover (or OwnedBuffer) like that himself.

davidpiuva: Okay then.

source/draw/DrawState.ooc
```diff
@@ -0,0 +1,73 @@
...
+	setViewport: func (viewport: IntBox2D) -> This {
+		this viewport = viewport
+		this
+	}
+	setInputImage: func (inputImage: Image) -> This {
+		this inputImage = inputImage
+		this
+	}
+	_toNormalized: static func (imageSize: FloatVector2D, transform: FloatTransform3D) -> FloatTransform3D {
```
davidpiuva: I tried getting this function to a reusable place around a year ago but it was later removed for not belonging to the draw API. Now I have 5 instances of the same duplicated code in different places.

Maybe a new module for managing coordinate systems using static functions?

sebastianbaginski: Sure, I guess anything is better than copying the same code in 5 different places

sebastianbaginski: Will you do that in separate PR ?

davidpiuva: Yes, I am working on a new PR.

source/draw/DrawState.ooc
marcusnaslund: No semicolons.
The return value may be (and often is) called `result`, but it's not a requirement.

davidpiuva:
I should complete the stub code in GpuCanvas.ooc, squash and rebase on the other pull requests.

davidpiuva:
Rebased

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		this context backend setViewport(viewport)
+		// Set transform
+		gpuMap view = _toLocal * drawState getTransformNormalized() normalizedToReference(this size) * _toLocal
+		// Set model and projection matrices
+		if (this _focalLength > 0.0f) {
+			a := 2.0f * this _focalLength / this size x
+			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
+			k := (this _farPlane + this _nearPlane) / (this _farPlane - this _nearPlane)
+			o := 2.0f * this _farPlane * this _nearPlane / (this _farPlane - this _nearPlane)
+			gpuMap projection = FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
```
sebastianbaginski: Do you think there is a better place for this math somewhere ? In `FloatTransform3D` maybe ?

davidpiuva: I plan to replace all the math that I copied from other draw calls once I have figured out what is a feature and what is a bug.

sebastianbaginski: OK :)

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		} else
+			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
```
davidpiuva: This looks like the place where reference coordinates are converted to normalized coordinates and flipped along Y depending on image format in the orthogonal projection matrix.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		gpuMap model = this _createModelTransform(IntBox2D new(this size))
```
davidpiuva: I don't know what to do with the model transform in normalized coordinates since I don't know if the field of view is vertical or horizontal.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		viewport := (drawState viewport hasZeroArea) ? IntBox2D new(this size) : drawState viewport
+		this context backend setViewport(viewport)
+		// Set transform
+		gpuMap view = _toLocal * drawState getTransformNormalized() normalizedToReference(this size) * _toLocal
+		// Set model and projection matrices
+		if (this _focalLength > 0.0f) {
+			a := 2.0f * this _focalLength / this size x
+			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
+			k := (this _farPlane + this _nearPlane) / (this _farPlane - this _nearPlane)
+			o := 2.0f * this _farPlane * this _nearPlane / (this _farPlane - this _nearPlane)
```
davidpiuva: I have no idea of who named these variables.

sebastianbaginski: Maybe that person wanted to have the same names as in
```ooc
FloatTransform3D: cover {
	a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p: Float
...
```

davidpiuva:
Rebased

source/draw/gpu/opengl/OpenGLCanvas.ooc
emilwestergren: Correct

source/draw/gpu/opengl/OpenGLCanvas.ooc
emilwestergren: This is the current model:
**Model matrix** - the world coordinates of the image plane, i.e. the left top corner of the quad is in (-width / 2, height / 2, focalLength)
**View matrix** - the inverse camera matrix, this is where the stabilization transform is applied
**Projection matrix** - The original image plane is reprojected to the new image plane and also the coordinates are translated to normalized coordinates. This should probably be separated into two transforms, one projection and one conversion to normalized coordinates, although you often combine these in OpenGL.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+			gpuMap projection = FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
+		} else
+			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
+		gpuMap model = this _createModelTransform(IntBox2D new(this size))
+		// Set opacity
+		if (drawState opacity < 1.0f)
```
emilwestergren: This is extremely ugly and should be fixed in a separate PR.

source/draw/gpu/opengl/OpenGLCanvas.ooc
davidpiuva: https://github.com/magic-lang/ooc-kean/pull/1264 has been created to clean this up.

source/draw/gpu/opengl/OpenGLCanvas.ooc
davidpiuva: Double character names (x1, y1, z1, w1, x2, y2, ...) in transforms would make more sense when switching between 3x3 and 4x4 transforms.

emilwestergren: @sebastianbaginski is right, it is because of `FloatTransform3D`.

davidpiuva: Should we fix `FloatTransform3D` for consistency then? I had to change from 2D to 3D before but luckily I did not use these one dimensional element names.

sebastianbaginski:
If @emilwestergren is happy with this I'll let him set the final review flag.

davidpiuva:
@sebastianbaginski Emil wanted an alternative name for `set` to be used by a new convention when the side effect is returned by return value only.

davidpiuva:
Maybe `pipe` since it looks like piping in unix commands?
`DrawState new(target) pipeMap(shader) draw()`

sebastianbaginski:
I don't know. Maybe `DrawState new(target) updateMap(shader) draw()` ?
`update` is not so short though

davidpiuva:
`give` works together with `get` but marks a difference.

sebastianbaginski:
You mean `DrawState new(target) giveMap(shader) draw()` ?

davidpiuva:
yes

sebastianbaginski:
I think I could live with this or `update`. @emilwestergren ?

sebastianbaginski:
Also @marcusnaslund - any opinions about the name ?

davidpiuva:
Updated with `give`.

emilwestergren:
I think `give` is too confusing. You give the map to the state? Like you pass ownership? The problem here is we want something that makes it more obvious that the modified state is the returned one, not the one you are calling the function on.

emilwestergren:
If we don't have any better suggestions I guess we will go with `set`. However, should we make it a general rule that we don't modify the object for any cover? Any function altering the state of the cover should return a copy.

davidpiuva:
By convention, the address of an argument passed by value should not affect the behaviour since this depends on the current implementation of the compiler. There are cases where the compiler implicitly has to clone the value to a new address causing side effects to be thrown away.

davidpiuva:
Changed back to `set`.

source/draw/DrawState.ooc
```diff
@@ -0,0 +1,68 @@
...
+	}
+	setTransformReference: func ~explicit (transform: FloatTransform3D, imageSize: FloatVector2D) -> This {
+		this setTransformNormalized(transform referenceToNormalized(imageSize))
+	}
+	setTransformNormalized: func (transform: FloatTransform3D) -> This {
+		this _transformNormalized = transform
+		this
+	}
+	getTransformNormalized: func -> FloatTransform3D {
+		this _transformNormalized
```
emilwestergren: If you want to save lines you can put these on a single one.

davidpiuva: Updated

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
 	_target: OpenGLPacked
 	_renderTarget: GLFramebufferObject
 	context ::= this _context as OpenGLContext
+	draw: override func ~DrawState (drawState: DrawState) {
+		// Set shader
```
marcusnaslund: If you want these comments, change this one to `//Set shader, viewport, and transform` and remove those on line 34 and 37.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		if (this _focalLength > 0.0f) {
+			a := 2.0f * this _focalLength / this size x
+			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
+			k := (this _farPlane + this _nearPlane) / (this _farPlane - this _nearPlane)
+			o := 2.0f * this _farPlane * this _nearPlane / (this _farPlane - this _nearPlane)
+			gpuMap projection = FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
+		} else
+			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
+		gpuMap model = this _createModelTransform(IntBox2D new(this size))
+		// Set opacity
```
marcusnaslund: The fact that we're dealing with opacity is obvious.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		if (drawState opacity < 1.0f)
+			this context backend blend(drawState opacity)
+		else
+			this context backend enableBlend(false)
+		// Set texture
```
marcusnaslund: The fact that we're dealing with texture is obvious.

source/draw/Canvas.ooc
```diff
@@ -17,7 +17,7 @@
 use base
 use geometry
 use collections
-import Image, Pen
+import Image, Pen, DrawState
```
marcusnaslund: Probably not your fault originally, but please put these on three separate lines.

source/draw/DrawState.ooc
```diff
@@ -0,0 +1,64 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use geometry
+import Pen, Image, Map
```
marcusnaslund: Also here, three lines.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -28,6 +28,38 @@ OpenGLCanvas: class extends OpenGLSurface {
 	_target: OpenGLPacked
 	_renderTarget: GLFramebufferObject
 	context ::= this _context as OpenGLContext
+	draw: override func ~DrawState (drawState: DrawState) {
+		// Set shader
+		gpuMap: GpuMap = (drawState map == null) ? this context defaultMap : drawState map as GpuMap
```
marcusnaslund: `gpuMap: GpuMap = drawState map as GpuMap ?? this context defaultMap`

davidpiuva:
Updated

davidpiuva:
PRInt is done testing.

PR #1227, opened 2016-01-20
Added final override to silence rock warning

thomasfanell:
Adding `final` when overriding an abstract function seems to eliminate the warning, and it seems to work as expected.


sebastianbaginski:
Nice. Hmm but what if we would like to override in `Class1` and then again in `Class2: extends Class1` ? 
Or do I have wrong assumptions about the `final` keyword ?

thomasfanell:
@sebastianbaginski then you're most likely screwed. We currently don't seem to have more than one level of inheritence that would require this particular fix. So this is merely buying some time.

thomasfanell:
It's not a viable long-term solution.

sebastianbaginski:
OK, I understand :) This is better than before anyway

thomasfanell:
@marcusnaslund peer?

PR #1228, opened 2016-01-20
fixed cppcheck warnings

sebastianbaginski:
Warnings when analyzing generated C code with cppcheck. Variables reassigned before old value was used.
@marcusnaslund

### DIFF
--- a/source/sdk/os/Env.ooc
+++ b/source/sdk/os/Env.ooc
@@ -29,7 +29,7 @@ Env: class {
 		set(key, value, true)
 	}
 	unset: static func (key: String) -> Int {
-		result := -1
+		result: Int
 		version(windows) {
 			// under mingw, this unsets the key
 			result = putenv((key + "=") toCString())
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Nice, I just wrote about this in the style guide.

source/sdk/os/Env.ooc
```diff
@@ -29,7 +29,7 @@ Env: class {
 		set(key, value, true)
 	}
 	unset: static func (key: String) -> Int {
-		result := -1
+		result: Int
```
marcusnaslund: Default value should be 0 instead.
```ooc
result := 0
version (windows)
    result = putenv((key + "=") toCString())
else
    unsetenv(key toCString())
result
```

sebastianbaginski: But then you have the same problem on windows again, assigning 0 and then reassign without using it.

marcusnaslund: Yeah, alright then.

PR #1229, opened 2016-01-20
removed redundant comparison in EGLImage

sebastianbaginski:
`type == TextureType Rgb` was checked twice
@marcusnaslund


marcusnaslund:
One can never be too careful.

PR #1230, opened 2016-01-20
Minor cleanup in FloatTransform covers

tomasmahlberg:
In some places:
`this` -> `This`
`~size` -> `~vector`


tomasmahlberg:
@marcusnaslund

marcusnaslund:
@fredrikbryntesson It would be nice to enforce `This` when calling static methods.

PR #1231, opened 2016-01-20
Added classes for uri parsing

jonathanudd:
The classes added in this PR are built to parse an uri string in this format `abc://username:password@example.com:123/path/data?key=value#fragid1` described [here](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier).
After the parsing is done different parts of the uri can be accessed in form of members.
This could be used to pass uri parameters from a terminal containing information about settings, if it's a local or remote file and a few more possible uses.


jonathanudd:
@sebastianbaginski could you peer reivew this?

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,53 @@
+Endpoint: class {
+	_host: VectorList<Text>
+	_port: Int
+	host: VectorList<Text> {
+		get {
+			result := VectorList<Text> new()
+			for (i in 0 .. this _host count) {
+				result add(this _host[i] take())
+			}
```
sebastianbaginski: you can removed these brackets

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,53 @@
...
+	init: func (=_host, =_port)
+	free: override func {
+		for (i in 0 .. this _host count)
+			this _host[i] free(Owner Receiver)
+		this _host free()
+		super()
+	}
+	toText: func -> Text {
+		result := Text new()
+		host := this host
```
sebastianbaginski: Can't you use `_host` member directly without copying ?

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,53 @@
...
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t":")
+			port := 0
+			if (splitted count == 2)
+				port = splitted remove() toInt()
+			host := splitted remove() split(t".")
```
sebastianbaginski: I'd also check if `splitted` is empty, I think this will crash if it is.

source/uri/Locator.ooc
```diff
@@ -0,0 +1,99 @@
...
+		get { this _authority }
+	}
+	path: VectorList<Text> {
+		get {
+			result: VectorList<Text>
+			if (this _path != null) {
+				result = VectorList<Text> new()
+				for (i in 0 .. this _path count) {
+					result add(this _path[i] take())
+				}
```
sebastianbaginski: brackets can be removed

source/uri/Locator.ooc
```diff
@@ -0,0 +1,99 @@
...
+			result += t"#" + this fragment
+		result
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t"://")
+			scheme := splitted count > 1 ? splitted remove(0) : Text empty
+			d = splitted remove()
```
sebastianbaginski: Empty check on `splitted`

source/uri/Query.ooc
```diff
@@ -0,0 +1,99 @@
+Query: class {
+	_attributes: VectorList<Text>
+	_values: VectorList<Text>
+	attributes: VectorList<Text> {
+		get {
+			result := VectorList<Text> new()
+			for (i in 0 .. this _attributes count) {
+				result add(this _attributes[i] take())
+			}
```
sebastianbaginski: brackets to remove

source/uri/Query.ooc
```diff
@@ -0,0 +1,99 @@
...
+			result
+		}
+	}
+	values: VectorList<Text> {
+		get {
+			result := VectorList<Text> new()
+			for (i in 0 .. this _values count) {
+				result add(this _values[i] take())
+			}
```
sebastianbaginski: same here

source/uri/Query.ooc
```diff
@@ -0,0 +1,99 @@
...
+		attributes := this attributes
+		values := this values
+		for (i in 0 .. attributes count) {
+			argument := attributes[i]
+			if (values[i] != Text empty)
+				argument += t"=" + values[i]
+			result += i == 0 ? argument : t";" + argument
+		}
+		if (attributes != null)
+			attributes free()
```
sebastianbaginski: haha github syntax highlighting gone mad :D

source/uri/Query.ooc
```diff
@@ -0,0 +1,99 @@
...
+		result
+	}
+	getValue: func(attribute: Text) -> Text {
+		result := Text empty
+		attributes := this attributes
+		values := this values
+		for (i in 0 .. attributes count) {
+			if (attributes[i] == attribute) {
+				result = values[i] take()
```
sebastianbaginski: maybe check if `attributes` and `values` have the same count somewhere, and throw exception if they dont

source/uri/User.ooc
```diff
@@ -0,0 +1,39 @@
...
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t":")
+			password := Text empty
+			if (splitted count == 2)
+				password = splitted remove()
+			name := splitted remove()
```
sebastianbaginski: empty check on `splitted`

sebastianbaginski:
Looks great. Review done.

source/uri/Authority.ooc
```diff
@@ -0,0 +1,43 @@
+use uri
+
+Authority: class {
+	_user: User
+	_endpoint: Endpoint
+	user: User {
+		get { this _user }
+	}
```
marcusnaslund: Why not `user ::= this _user`?

marcusnaslund: And likewise for many properties in this PR.

source/uri/Endpoint.ooc
jonathanudd: I think the check `if (!d isEmpty)` together with `if (splitted count == 2)` makes this safe.

sebastianbaginski: Ah ok, if d is not empty and does not contain `:` split will return the whole string

source/uri/Query.ooc
jonathanudd: The elements in the vectors are added in a way that should make it impossible for them to be different lengths and elements are not removed after they are added.
But I add a check and throw an exception of the lengths are not equal in the parse function.

sebastianbaginski: OK

jonathanudd:
Updated based on comments.

source/uri/Authority.ooc
```diff
@@ -0,0 +1,38 @@
...
+		if (this _endpoint != null)
+			result += this _endpoint toText()
+		result
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t"@")
+			user := splitted count > 1 ? User parse(splitted remove(0)) : null
```
marcusnaslund: Probably shouldn't have a local variable `user` and a property called `user`

source/uri/Authority.ooc
```diff
@@ -0,0 +1,38 @@
...
+			endpoint := Endpoint parse(splitted remove())
```
marcusnaslund: Same with endpoint

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,47 @@
+Endpoint: class {
+	_host: VectorList<Text>
+	_port: Int
+	host: VectorList<Text> {
```
marcusnaslund: The correct style is 
```ooc
host: VectorList<Text> { get {
...
}}
```

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,47 @@
...
+	free: override func {
+		for (i in 0 .. this _host count)
+			this _host[i] free(Owner Receiver)
+		this _host free()
+		super()
+	}
+	toText: func -> Text {
+		result := Text new()
+		for (i in 0 .. this _host count)
+			result += i == 0 ? this _host[i] take() : t"." + this _host[i] take()
```
marcusnaslund: How big is `this _host`? Consider using `TextBuilder` instead.

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,47 @@
...
+		}
+		result
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t":")
+			port := splitted count > 1 ? splitted remove() toInt() : 0
+			host := splitted remove() split(t".")
```
marcusnaslund: Same as above with `port` and `host` here

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,47 @@
...
+			for (i in 0 .. host count)
+				host[i] = host[i] claim()
```
marcusnaslund: Why `claim` instead of `take`? Is this what you asked me about before? @jonathanudd 
Can you shine some light on this @tomasmahlberg ?

source/uri/Locator.ooc
```diff
@@ -0,0 +1,89 @@
...
+
+Locator: class {
+	_scheme: Text
+	_authority: Authority
+	_path: VectorList<Text>
+	_query: Query
+	_fragment: Text
+	scheme ::= this _scheme take()
+	authority ::= this _authority
+	path: VectorList<Text> {
```
marcusnaslund: `path: VectorList<Text> { get {`

source/uri/Locator.ooc
```diff
@@ -0,0 +1,89 @@
...
+	toText: func -> Text {
+		result := Text new()
+		if (!this scheme isEmpty)
+			result += this scheme + t"://"
+		if (this authority != null)
+			result += this authority toText()
+		if (this _path != null) {
+			for (i in 0 .. this _path count)
+				result += t"/" + this _path[i] take()
+		}
```
marcusnaslund: You don't need these brackets

source/uri/Locator.ooc
```diff
@@ -0,0 +1,89 @@
...
+		if (this query != null)
+			result += t"?" + this query toText()
+		if (!this fragment isEmpty)
+			result += t"#" + this fragment
+		result
```
marcusnaslund: Maybe `base/TextBuilder` should be used here too.

source/uri/Locator.ooc
marcusnaslund: We generally don't do that level of defensive programming. Maybe in a version block.

sebastianbaginski: OK. I'm just trying to predict static analysis complaints.

source/uri/Query.ooc
```diff
@@ -0,0 +1,84 @@
+Query: class {
+	_attributes: VectorList<Text>
+	_values: VectorList<Text>
+	attributes: VectorList<Text> {
```
marcusnaslund: This property and the one below also need style fix

source/uri/Query.ooc
```diff
@@ -0,0 +1,84 @@
...
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t";")
+			attributes := VectorList<Text> new()
+			values := VectorList<Text> new()
+			for (i in 0 .. splitted count) {
+				argument := splitted[i] split(t"=")
+				if (argument count == 2)
```
marcusnaslund: `argument count == 2 ? values add(argument remove() claim() : values add(Text empty)`

source/uri/User.ooc
```diff
@@ -0,0 +1,34 @@
...
+			result += t":" + this password
+		result
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t":")
+			password := splitted count > 1 ? splitted remove() : Text empty
+			name := splitted remove()
```
marcusnaslund: Same with `name` and `password` here as above

uri.use
```diff
@@ -0,0 +1,8 @@
+Name: ooc uri
+Description: uri parser
+SourcePath: source/uri
+Imports: Locator
+Imports: Authority
+Imports: User
+Imports: Endpoint
+Imports: Query
```
marcusnaslund: Thank you VScode
I thought this was fixed? @thomasfanell

marcusnaslund:
Very nice and thorough tests. :)

source/uri/Endpoint.ooc
jonathanudd: Because we only have a getter? Would the syntax be correct if there was a setter below also?

uri.use
thomasfanell: What was fixed?

source/uri/Endpoint.ooc
marcusnaslund: This is because it's only a getter.
Otherwise it's like this
```
property: Type {
   get { ... }
   set { ... }
}
```

source/uri/Endpoint.ooc
jonathanudd: If I don't use `claim` the `Text` object owner was `Unknown` and then if `take` is called the results owner is still `Unknown` and then when you free with `Owner Receiver` the object is not freed. Maybe I have misunderstood something.

uri.use
marcusnaslund: The last line is missing newline.

thomasfanell: No it wasn't. There is an extension for this, but I would not recommend installing it at this point.

marcusnaslund: Oh OK.
Anyway, missing newline. @jonathanudd :)

source/uri/Endpoint.ooc
jonathanudd: `this _host` will probably not be very large (maybe around 3 items) and this code will probably only run once or a few times when used. I don't think much is gained by using it here or in other places in this PR.

marcusnaslund: You're probably right. Then forget the `TextBuilder` comments.

source/uri/Query.ooc
marcusnaslund: `values add(argument count == 2 ? argument remove() claim() : Text empty)` also works, but is probably less readable

jonathanudd: Can you do this without setting the value of a variable? I'm not getting this to work.

test/uri/EndpointTest.ooc
```diff
@@ -0,0 +1,44 @@
...
+		this add("parse", func {
+			one := t"one"
+			two := t"two"
+			three := t"three"
+			host := one + t"." + two + t"." + three
+			port := t"1232"
+			host_port := host + t":" + port
+			endpoint := Endpoint parse(host_port take())
+			list := endpoint host
+			expect(list[0] == one)
```
marcusnaslund: Sorry to complain a lot, but `expect` calls with `Text` can (and should) nowadays use `is equal to()` like other types.

source/uri/Endpoint.ooc
```diff
@@ -0,0 +1,45 @@
...
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t":")
+			newPort := splitted count > 1 ? splitted remove() toInt() : 0
+			newHost := splitted remove() split(t".")
+			for (i in 0 .. newHost count)
+				newHost[i] = newHost[i] claim()
```
marcusnaslund: I'm still not sure about this. Since the `Text`s you have have `Owner Unknown`, `claim()` will return a copy of the underlying `Buffer`. The original one still remains. So, there should be leaks.

marcusnaslund:
You may readd the green flag once above comments are fixed.

test/uri/EndpointTest.ooc
jonathanudd: No need to say sorry :)

marcusnaslund: Things change fast in this repo. :)

jonathanudd: I had some problems with `is equal to()` but not on all tests. I changed to `is equal to()` where it worked.

source/uri/Endpoint.ooc
jonathanudd: I changed to `take()`.

jonathanudd:
Made some updates and the tests don't have any leaks now.

marcusnaslund:
Looks great now.

source/uri/Authority.ooc
```diff
@@ -0,0 +1,46 @@
...
+			result += this _user toText() + t"@"
+		if (this _endpoint != null)
+			result += this _endpoint toText()
+		result
+	}
+	parse: static func(data: Text) -> This {
+		result: This
+		d := data take()
+		if (!d isEmpty) {
+			splitted := d split(t"@")
```
tomasmahlberg: `splitted` -> `split` if we want to be grammatically correct? (Just happened to drop by here...)

marcusnaslund: Interestingly, there are 600 000+ hits on a Google search of "splitted", and Wikitionary lists it, albeit as `(nonstandard or archaic)`. Still, you're correct.

Issue #781, opened 2015-11-16
Cleaning up Color


davidpiuva:
- [x] Comparing equality with colors of other types only makes sense with classes.
- [x] `Set` can not be safely applied to covers since it is not a reference type.
- [x] Colors have more constructors than needed.
- [ ] The `Distance` function might be incorrect for BGRA since the BGR distance is not squared and applies some weird scaling instead.
- [x] Default values for UV can not be 0 since it gives negative intensity on some wavelengths.

davidhesselbom:
Also, this:
> // FIXME: yuvToBgr2[512..end] and yuvToBgr2[256..511] are all 0, so there's some minor room for optimization here.
	// Also, since multidimensional arrays work now, the arrays below can be reorganized if so desired.

marcusnaslund:
> Comparing equality with colors of other types only makes sense with classes.

How do you mean?

> Set can not be safely applied to covers

We do similar things in many covers, especially in `source/math` and `source/geometry`. What's wrong here?

>  Colors have more constructors than needed.

We should at least be consistent with which arguments may be used to create a color. But having one for `Float` and one for `Double` is unnecessary, I believe, as rock should implicitly cast from one to the other.

> The Distance function might be incorrect

Looks strange to me too, but I'm no expert here. Do you know who wrote it?

> Default values for UV can not be 0

What should they be then?

davidpiuva:
Who would call a non virtual function without sideeffects that always returns false?

Set methods change by reference to this but covers are value types so that any side effect is thrown away on a copy.

I dont know who wrote the distance function.

Default uv should be gray intensity since a signed -1..1 scale is stored in unsigned bytes.

sebastianbaginski:
(2) fixed in https://github.com/cogneco/ooc-kean/pull/1232

davidpiuva:
Removed (6) since colors don't have to contain other colors.

davidpiuva:
Enough point have been solved for me to consider Color to be good enough and not worth having an issue about.

PR #1232, opened 2016-01-21
corrected some functions in Color classes

sebastianbaginski:
Fixed `set` functions, now they call by reference.
Updated `copy`, removed redundant constructor call.
@marcusnaslund review ?


marcusnaslund:
You may tick the corresponding items in #781

sebastianbaginski:
OK, I will do that after PR is merged.

PR #1236, opened 2016-01-21
fixed possible uses of uninitialized variables

sebastianbaginski:
@marcusnaslund review ?


source/concurrent/SynchronizedQueue.ooc
```diff
@@ -58,7 +58,7 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 		this _populated broadcast()
 	}
 	wait: func (isOk := null as Bool*) -> T {
-		result: T
+		result: T = null
```
marcusnaslund: `result: T = null` and `result := null as T` both work, right? We should decide on one going forward and I mostly like the one you wrote here.

PR #1237, opened 2016-01-21
Fixed some small issues in stb_image

sebastianbaginski:
- possible out of bounds reads
- unchecked error codes from `fseek`
- possible uninitialized array reads

@marcusnaslund

### DIFF
--- a/source/draw/stb_image/stb_image.c
+++ b/source/draw/stb_image/stb_image.c
@@ -446,6 +446,36 @@ static void stbi__start_callbacks(stbi__context *s, stbi_io_callbacks *c, void *
    stbi__refill_buffer(s);
 }
 
+// this is not threadsafe
+static const char *stbi__g_failure_reason;
+
+STBIDEF const char *stbi_failure_reason(void)
+{
+   return stbi__g_failure_reason;
+}
+
+static int stbi__err(const char *str)
+{
+   stbi__g_failure_reason = str;
+   return 0;
+}
+
+// stbi__err - error
+// stbi__errpf - error returning pointer to float
+// stbi__errpuc - error returning pointer to unsigned char
+
+#ifdef STBI_NO_FAILURE_STRINGS
+   #define stbi__err(x,y)  0
+#elif defined(STBI_FAILURE_USERMSG)
+   #define stbi__err(x,y)  stbi__err(y)
+#else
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Looks like the repo this file comes from is undergoing some continous development, should we look at also updating our version with what's new? @emilwestergren @tomasmahlberg

marcusnaslund:
It might be a better idea would be to make this PR to the [original repo](https://github.com/nothings/stb) directly, and then take *that* latest version and merge it here.

sebastianbaginski:
It may or may not be accepted in the original repo, I don't think it is worth waiting, it's a very simple fix. I'd say let's just fix it ourselves here **and** make a pull request to stb repo. Merging this won't be a big problem later, eventually.

marcusnaslund:
If we care about updating, we should at least do that *before* this change. Input @fredrikbryntesson ?

sebastianbaginski:
It is not so easy as simply replacing the files with new versions. Looks like the part of `stbi` library we are interested in became header-only. Trying to replace the headers with new version will cause compiler failure related to multiple definitions of functions from the stbi header. Not mentioning hundreds of warnings when compiling with -Wall, which we probably have to fix ourselves.
My suggestion is to merge this, and create an issue about updating the stbi library, I can do that later.

sebastianbaginski:
https://github.com/magic-lang/ooc-kean/issues/1240

fredrikbryntesson:
Depressing :( but okay. If updating causes too much work it is better to merge our patch with our current version of stb_image. Do the pull request to upstream anyway but maybe it will not be of any use for us because of their new version is so much different from the one we are currently using?

sebastianbaginski:
@fredrikbryntesson : the file we have patched here is gone from stbi :)
Of course the functions are still there, but inlined in the header file.

fredrikbryntesson:
Ok, then I guess it is not so much more to do. But I guess checking if an update is possible or what they have improved should be done in somewhat near future.

marcusnaslund:
I will merge this, and we will discuss what to do with `stb_image` later. Good you made the issue.

sebastianbaginski:
OK, thanks. I think there will be more small fixes soon, I will take care of updating stbi & merging them next week.

PR #1238, opened 2016-01-21
fixed possible race condition in FileUnix.ooc

sebastianbaginski:
Possible security flaw detected by static analysis.
Running this function could result in assigning executable permissions to wrong file.
@marcusnaslund

### DIFF
--- a/source/sdk/io/native/FileUnix.ooc
+++ b/source/sdk/io/native/FileUnix.ooc
@@ -36,12 +36,16 @@ version (unix || apple) {
 
 	TimeT: cover from time_t
 	ModeT: cover from mode_t
+	INodeT: cover from ino_t
 
 	FileStat: cover from struct stat {
 		st_mode: extern ModeT
 		st_size: extern SizeT
+		st_ino: extern INodeT
 		st_atime, st_mtime, st_ctime: extern TimeT
 	}
+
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
Wait a minute I can't build it ...

sebastianbaginski:
OK, ready @marcusnaslund

source/sdk/io/native/FileUnix.ooc
```diff
@@ -181,18 +188,22 @@ version (unix || apple) {
...
+			lstatResult, fstatResult: FileStat
+			res := lstat(path as CString, lstatResult&)
 			if (res != 0) return false // couldn't get file mode
-
-			mode := result st_mode
+			fd := open(path as CString, READWRITE)
+			if (fd == -1) return false
+			res = fstat(fd, fstatResult&)
+			if (res != 0) return false
+			if (fstatResult st_ino != lstatResult st_ino || fstatResult st_mode != lstatResult st_mode)
```
sebastianbaginski: Fix is to use `lstat(path), fd := open(path), fstat(fd)` and compare results of both `*stat` calls. If the file was changed between calls to `lstat` and `fstat`, then it is not safe to set the executable permission.

PR #1239, opened 2016-01-21
String split() now returns VectorList

marcusnaslund:
`CharBuffer split` returns a `VectorList` which doesn't free its contents, and so works just like the old `ArrayList`.


marcusnaslund:
Ready for peer review @sebastianbaginski

PR #1242, opened 2016-01-22
removed some dead code

sebastianbaginski:
Removed some code that could never been reached.
@marcusnaslund


marcusnaslund:
"He's dead, Jim"

PR #1245, opened 2016-01-22
fixed remaining static analysis issues in IO

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/sdk/io/native/FileUnix.ooc
+++ b/source/sdk/io/native/FileUnix.ooc
@@ -65,6 +65,7 @@ version (unix || apple) {
 	close: extern func (Int) -> Int
 	_mkdir: extern (mkdir) func (CString, ModeT) -> Int
 	_mkfifo: extern (mkfifo) func (CString, ModeT) -> Int
+	umask: extern func (ModeT) -> ModeT
 	remove: extern func (path: CString) -> Int
 	_getcwd: extern (getcwd) func (buf: CString, size: SizeT) -> CString
 	ooc_get_cwd: unmangled func -> String {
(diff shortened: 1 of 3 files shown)


source/sdk/io/native/FileUnix.ooc
```diff
@@ -316,5 +320,17 @@ version (unix || apple) {
 		mkfifo: override func ~withMode (mode: Int32) -> Int {
 			_mkfifo(path as CString, mode as ModeT)
 		}
+
+		createTempFile: static func (pattern, mode: String) -> FStream {
+			result: FStream
+			mask := umask(S_IRUSR | S_IWUSR)
+			fd := mkstemp(pattern)
```
sebastianbaginski: We need to specify the access mask explicitly, because:

> In glibc versions 2.06 and earlier, the file is created with
> permissions 0666, that is, read and write for all users.  This old
> behavior may be a security risk (...)

marcusnaslund: Can't find an exact date fpr 2.06, but 2.09 was released in December 1997, do we still care?

sebastianbaginski: It is fixed now anyway. Maybe someday a client will come and say "I have a unix system from 1991, write me a program" :P

sebastianbaginski: I can remove the fix if you prefer and suppress the warning

marcusnaslund: No, the work is done, let's not throw it away. :)

source/sdk/io/native/FileUnix.ooc
```diff
@@ -316,5 +320,17 @@ version (unix || apple) {
...
+			if (fd >= 0)
+				result = fdopen(fd, mode)
+			else
+				result = null
```
marcusnaslund: `result = fd >= 0 ? fdopen(fd, mode) : null`

sebastianbaginski: ok

sebastianbaginski:
updated

PR #1246, opened 2016-01-22
Fixed singularity problem in Quaternion weightedMean

tomasmahlberg:
I have tested this in applied usage and it seems to get rid of the problems that were apparent before.

I could not find an issue for this TODO.

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -237,9 +237,11 @@ Quaternion: cover {
 		// Rotate quaternions if the angle is close to pi to avoid singularity problem
 		centerQuaternion := quaternions[quaternions count / 2]
 		preRotate := centerQuaternion angle(This identity) equals(Float pi, 0.5f)
-		if (preRotate)
+		if (preRotate) {
+			centerQuaternionInverse := centerQuaternion inverse
 			for (index in 0 .. quaternions count)
-				quaternions[index] = centerQuaternion inverse * quaternions[index]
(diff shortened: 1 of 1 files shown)


tomasmahlberg:
@niklasborwell

niklasborwell:
Looks good.

marcusnaslund:
I don't understand how you use `preRotate` as a boolean statement. Isn't it always a quaternion, hence always not null, hence always true?

marcusnaslund:
Wait, it's not a Quaternion. Ignore me. It's friday.

source/geometry/Quaternion.ooc
```diff
@@ -234,7 +234,12 @@ Quaternion: cover {
 		// W - observationVectors
 		// Y - gibbsVector
 		// Z - vectorQuantityZ
-		// TODO: Fix singularity problem when the angle is close to PI, using sequential rotations
+		// Rotate quaternions if the angle is close to pi to avoid singularity problem
+		centerQuaternion := quaternions[quaternions count / 2]
+		preRotate := centerQuaternion angle(This identity) equals(Float pi, 0.5f)
+		if (preRotate)
+			for (index in 0 .. quaternions count)
+				quaternions[index] = centerQuaternion inverse * quaternions[index]
```
marcusnaslund: You calculate the inverse on every iteration now. Wouldn't it be better to do it just once?

tomasmahlberg:
Updated with minor optimization.

Issue #1233, opened 2016-01-21
Get rid of ArrayList<T>


marcusnaslund:
This does not seem too far away anymore. A search reveals only five files (four of them in the SDK) which uses `ArrayList`. Every use should be replaced by `VectorList`. Since it resides in the SDK folder now, that shouldn't be a problem.

Replacing `String` with `Text` would solve some problems, but for now an option is to make sure `String split()` (and possibly other methods) return `VectorList` instead of `ArrayList`.

marcusnaslund:
The only use of `List<T>` is `ArrayList<T>`, so once the latter is gone, the former goes away easily.

marcusnaslund:
#1239 is taking care of a lot by making `String split` return a `VectorList<String>` instead.

marcusnaslund:
#1244 aims at removing `List<T>`.

marcusnaslund:
Update: `List<T>` is now gone. `ArrayList` is only used in `HashMap` (and is imported in `String`, but that can hopefully be removed)

PR #1251, opened 2016-01-25
Removed (some) uses of ArrayList

marcusnaslund:
Part of the work on #1233 

Only `HashMap` left after this. Rock may or may not require the `ArrayList` file to exist, we'll see.

### DIFF
--- a/source/sdk/lang/String.ooc
+++ b/source/sdk/lang/String.ooc
@@ -1,14 +1,7 @@
-/**
- * The String class represents character strings.
- *
- * The String class is immutable by default, this means every writing operation
- * is done on a clone, which is then returned
- */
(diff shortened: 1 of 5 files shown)


source/sdk/lang/String.ooc
```diff
@@ -1,14 +1,7 @@
-/**
- * The String class represents character strings.
- *
- * The String class is immutable by default, this means every writing operation
- * is done on a clone, which is then returned
- */
+import structs/ArrayList // Must be here, otherwise keyEquals in HashMap does not resolve...
```
marcusnaslund: @fredrikbryntesson - Does `rock` use hashmap in some hard-coded way?

marcusnaslund:
Ready for peer review, maybe @thomasfanell ?

PR #1252, opened 2016-01-25
added keyboard events for X11

sebastianbaginski:
It is now possible to listen to keyboard events.
@marcusnaslund

### DIFF
--- a/source/ui/GuiEvent.ooc
+++ b/source/ui/GuiEvent.ooc
@@ -5,16 +5,20 @@ GuiEvent: abstract class {
 	init: func
 }
 
-MouseEvent: class extends GuiEvent {
+InputEvent: abstract class extends GuiEvent {
 	_position: IntPoint2D
 	position ::= this _position
 	init: func (=_position)
 }
 
-KeyboardEvent: class extends GuiEvent {
+MouseEvent: class extends InputEvent {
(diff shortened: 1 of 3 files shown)


source/ui/GuiEvent.ooc
```diff
@@ -5,16 +5,20 @@ GuiEvent: abstract class {
...
-KeyboardEvent: class extends GuiEvent {
+MouseEvent: class extends InputEvent {
+	init: func (position: IntPoint2D) { super(position) }
+}
+
+KeyboardEvent: class extends InputEvent {
 	_key: Char
 	key ::= this _key
-	init: func (=_key)
+	init: func (position: IntPoint2D, =_key){ super(position) }
```
marcusnaslund: What's `position` in terms of a keyboard event?

marcusnaslund:
Magic complains.

source/ui/GuiEvent.ooc
sebastianbaginski: Same as for mouse event, position of the mouse pointer on the screen while the key was pressed / released.

sebastianbaginski:
updated

source/ui/GuiEvent.ooc
marcusnaslund: Right, that makes sense, thanks.

PR #1253, opened 2016-01-26
Made an abstract `Map` class

davidpiuva:
Made an abstract `Map` class to allow having shader pointers in GPU independent code.


davidpiuva:

PR #1254, opened 2016-01-26
removed CharBuffer and FileUnix errors

sebastianbaginski:
Removed new coverity erros from CharBuffer and missed error in FileUnix
@marcusnaslund / @fredrikbryntesson

### DIFF
--- a/source/sdk/CharBuffer.ooc
+++ b/source/sdk/CharBuffer.ooc
@@ -63,7 +63,8 @@ CharBuffer: class extends Iterable<Char> {
 			// if we were coming from a string literal, copy the original data as well (gc_realloc cant work on text segment)
 			if (size > 0 && mallocAddr == null) {
 				rs = 0
-				memcpy(tmp, data, size)
+				if (tmp != null && data != null)
+					memcpy(tmp, data, size)
 			}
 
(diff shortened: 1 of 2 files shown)


fredrikbryntesson:
Looks perfect, I hope it is satisfied now.

sebastianbaginski:
wait a second...

source/sdk/CharBuffer.ooc
```diff
@@ -63,7 +63,8 @@ CharBuffer: class extends Iterable<Char> {
 			// if we were coming from a string literal, copy the original data as well (gc_realloc cant work on text segment)
 			if (size > 0 && mallocAddr == null) {
 				rs = 0
-				memcpy(tmp, data, size)
+				if (tmp != null && data != null)
```
marcusnaslund: You're checking `tmp` both here and on line 61 now.

sebastianbaginski: Yes, this is on purpose. I think coverity engine doesn't understand that this line can be never reached because exception is thrown.

marcusnaslund: Hmm, OK then.

PR #1262, opened 2016-01-27
Removed most return statements

### LINKED ISSUE #1261: Remove return keywords in common files
We should follow the maxim of `a function returns only once, on the last line`.

marcusnaslund:
Fixes #1261 

I've intentionally left some files out, like `ArrayList`, either because we are getting rid of them soon or because they need a much more extensive rewrite in the long run.

### DIFF
--- a/source/io/Pipe.ooc
+++ b/source/io/Pipe.ooc
@@ -13,10 +13,12 @@ Pipe: abstract class {
 	read: func ~string (len: Int) -> String {
 		buf := gc_malloc(len + 1) as CString
 		howmuch := read(buf, len)
-		if (howmuch == -1)
-			return null
-		buf[howmuch] = '\0'
-		buf toString()
+		result: String = null
+		if (howmuch != -1) {
+			buf[howmuch] = '\0'
+			result = buf toString()
+		}
+		result
 	}
 	read: func ~buffer (buf: CharBuffer) -> Int {
 		bytesRead := read(buf data, buf capacity)
@@ -56,14 +58,14 @@ Pipe: abstract class {
 		PipeWriter new(this)
 	}
 	new: static func -> This {
-		version(unix || apple) {
-			return PipeUnix new() as This
-		}
-		version(windows) {
-			return PipeWin32 new() as This
-		}
-		Exception new(This, "Unsupported platform!\n") throw()
-		null
+		result: This = null
+		version(unix || apple)
+			result = PipeUnix new() as This
+		version(windows)
+			result = PipeWin32 new() as This
+		if (result == null)
+			Exception new(This, "Unsupported platform!\n") throw()
+		result
 	}
 }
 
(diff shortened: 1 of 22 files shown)


marcusnaslund:
Ready for peer review @sebastianbaginski

source/io/Pipe.ooc
```diff
@@ -13,10 +13,12 @@ Pipe: abstract class {
...
 		buf := gc_malloc(len + 1) as CString
 		howmuch := read(buf, len)
-		if (howmuch == -1)
-			return null
-		buf[howmuch] = '\0'
-		buf toString()
+		result: String = null
+		if (howmuch != -1) {
+			buf[howmuch] = '\0'
+			result =buf toString()
```
sebastianbaginski: formatting error with ` =buf`

source/io/Pipe.ooc
```diff
@@ -56,14 +58,14 @@ Pipe: abstract class {
...
-			return PipeUnix new() as This
-		}
-		version(windows) {
-			return PipeWin32 new() as This
-		}
-		Exception new(This, "Unsupported platform!\n") throw()
-		null
+		result: This = null
+		version(unix || apple)
+			result = PipeUnix new() as This
```
sebastianbaginski: I think this can generate warnings in coverity about not using old value of `result` before writing new. But we can solve this later.

source/sdk/threading/Thread.ooc
```diff
@@ -13,32 +13,30 @@ Thread: abstract class {
...
-			return ThreadUnix new(_code) as This
-		}
-		version (windows) {
-			return ThreadWin32 new(_code) as This
-		}
-		Exception new(This, "Unsupported platform!\n") throw()
-		null
+		result: This = null
+		version (unix || apple)
+			result = ThreadUnix new(_code) as This
```
sebastianbaginski: Same here with result, and above somewhere, but we can leave this for now

marcusnaslund: It's not possible to do `version() ... else version() ... else`?

sebastianbaginski: I have no idea

marcusnaslund: Let's leave it for the future anyhow.

marcusnaslund:
Oh, tests fail.

sebastianbaginski:
Maybe because of 
> formatting error with =buf

:)

marcusnaslund:
Actually, `magic` didn't catch that. But I forgot the generic part of `ThreadLocal`.

marcusnaslund:
Also an error with how I changed `instanceOf?`. There we go.

marcusnaslund:
All tests pass now. Any other comments?

source/io/native/ProcessWin32.ooc
```diff
@@ -28,14 +28,12 @@ ProcessWin32: class extends Process {
...
 		status := WaitForSingleObject(pi process, duration)
-
-		if (status != WAIT_OBJECT_0) return -1
-
-		exitCode: ULong
-		GetExitCodeProcess(pi process, exitCode&)
-
-		CloseHandle(pi thread)
-		CloseHandle(pi process)
+		exitCode: ULong = -1
```
marcusnaslund: ...

PR #1264, opened 2016-01-27
Projection

davidpiuva:
Applying reference coordinates relative to the target image when reference coordinates are relative to the input image of another resolution caused the need to convert to and from normalized coordinates every time a transform is used. The external interface was improved by allowing both normalized and reference coordinates in the DrawState interface but the initial solution was just another layer of conversion on top of the old code. This PR removes a conversion back to reference coordinates that was canceled out by the projections given to the GPU. DrawState now uses normalized coordinates in the OpenGL implementation to avoid the two conversions and get simpler code.

More testing will be applied once the whole DrawState API is complete since things will change a lot.

### DIFF
--- a/source/draw/gpu/opengl/OpenGLCanvas.ooc
+++ b/source/draw/gpu/opengl/OpenGLCanvas.ooc
@@ -32,16 +32,16 @@ OpenGLCanvas: class extends OpenGLSurface {
 		gpuMap: GpuMap = drawState map as GpuMap ?? this context defaultMap
 		viewport := (drawState viewport hasZeroArea) ? IntBox2D new(this size) : drawState viewport
 		this context backend setViewport(viewport)
-		gpuMap view = _toLocal * drawState getTransformNormalized() normalizedToReference(this size) * _toLocal
+		gpuMap view = _toLocal * drawState getTransformNormalized() * _toLocal
 		if (this _focalLength > 0.0f) {
-			a := 2.0f * this _focalLength / this size x
-			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
+			a := this _focalLength
+			f := -(this _coordinateTransform e as Float) * this _focalLength
 			k := (this _farPlane + this _nearPlane) / (this _farPlane - this _nearPlane)
 			o := 2.0f * this _farPlane * this _nearPlane / (this _farPlane - this _nearPlane)
 			gpuMap projection = FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
 		} else
-			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
-		gpuMap model = this _createModelTransform(IntBox2D new(this size))
+			gpuMap projection = FloatTransform3D createScaling(1.0f, -(this _coordinateTransform e as Float), 1.0f)
+		gpuMap model = FloatTransform3D identity
 		if (drawState opacity < 1.0f)
 			this context backend blend(drawState opacity)
 		else
--- a/source/draw/gpu/opengl/OpenGLCanvas.ooc
+++ b/source/draw/gpu/opengl/OpenGLCanvas.ooc
@@ -32,7 +32,7 @@ OpenGLCanvas: class extends OpenGLSurface {
 		gpuMap: GpuMap = drawState map as GpuMap ?? this context defaultMap
 		viewport := (drawState viewport hasZeroArea) ? IntBox2D new(this size) : drawState viewport
 		this context backend setViewport(viewport)
-		gpuMap model = FloatTransform3D identity // TODO: set using a destination argument in drawState
+		gpuMap model = FloatTransform3D createTranslation(0.0f, 0.0f, this focalLength) // TODO: set using a destination argument in drawState
 		gpuMap view = _toLocal * drawState getTransformNormalized() * _toLocal
 		gpuMap projection = FloatTransform3D createProjection(this _focalLength, this _nearPlane, this _farPlane, this _coordinateTransform)
 		// Set opacity
--- a/source/draw/gpu/opengl/OpenGLCanvas.ooc
+++ b/source/draw/gpu/opengl/OpenGLCanvas.ooc
@@ -32,16 +32,10 @@ OpenGLCanvas: class extends OpenGLSurface {
 		gpuMap: GpuMap = drawState map as GpuMap ?? this context defaultMap
 		viewport := (drawState viewport hasZeroArea) ? IntBox2D new(this size) : drawState viewport
 		this context backend setViewport(viewport)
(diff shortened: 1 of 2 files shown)


davidpiuva:
@bettinaselig Peer review?

source/geometry/FloatTransform3D.ooc
```diff
@@ -253,6 +254,16 @@ FloatTransform3D: cover {
 	createProjection: static func (focalLength: Float) -> This {
 		This new(focalLength, 0, 0, 0, 0, focalLength, 0, 0, 0, 0, focalLength, 1.0f, 0, 0, 0, 0)
 	}
+	createProjection: static func ~flip (focalLength, nearPlane, farPlane: Float, coordinateTransform: IntTransform2D) -> This {
```
davidpiuva: To be reusable enough for `geometry`, I also allow flipping the x axis.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -32,16 +32,10 @@ OpenGLCanvas: class extends OpenGLSurface {
 		gpuMap: GpuMap = drawState map as GpuMap ?? this context defaultMap
 		viewport := (drawState viewport hasZeroArea) ? IntBox2D new(this size) : drawState viewport
 		this context backend setViewport(viewport)
-		gpuMap view = _toLocal * drawState getTransformNormalized() normalizedToReference(this size) * _toLocal
```
davidpiuva: This line used to convert to reference coordinates of the right image size before the matrices below converts back to normalized.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -32,16 +32,10 @@ OpenGLCanvas: class extends OpenGLSurface {
...
-		if (this _focalLength > 0.0f) {
-			a := 2.0f * this _focalLength / this size x
-			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
-			k := (this _farPlane + this _nearPlane) / (this _farPlane - this _nearPlane)
-			o := 2.0f * this _farPlane * this _nearPlane / (this _farPlane - this _nearPlane)
-			gpuMap projection = FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
-		} else
-			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
-		gpuMap model = this _createModelTransform(IntBox2D new(this size))
+		gpuMap model = FloatTransform3D identity // TODO: set using a destination argument in drawState
```
davidpiuva: Now that the given transform is normalized, we can use the identity matrix for the model.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -32,16 +32,10 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		gpuMap view = _toLocal * drawState getTransformNormalized() * _toLocal
```
davidpiuva: The transform is flipped by applying the transform between two flip matrices that are their own inverses.
We multiply in reverse order when using inverse matrices because the coordinate system of each side is flipped but the order does not matter here as both expressions are equivalent.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -32,16 +32,10 @@ OpenGLCanvas: class extends OpenGLSurface {
...
+		gpuMap projection = FloatTransform3D createProjection(this _focalLength, this _nearPlane, this _farPlane, this _coordinateTransform)
```
davidpiuva: The reusable function is called with the same arguments as before but without 2 / size.

source/geometry/FloatTransform3D.ooc
```diff
@@ -253,6 +254,16 @@ FloatTransform3D: cover {
...
+		if (focalLength > 0.0f) {
+			a := (coordinateTransform a as Float) * focalLength
+			f := -(coordinateTransform e as Float) * focalLength
+			k := (farPlane + nearPlane) / (farPlane - nearPlane)
+			o := 2.0f * farPlane * nearPlane / (farPlane - nearPlane)
+			FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
```
davidpiuva: Creating a projection matrix for OpenGL's axis system. Maybe it does not belong here but we already have a projection function above.

source/geometry/FloatTransform3D.ooc
```diff
@@ -253,6 +254,16 @@ FloatTransform3D: cover {
...
+		} else
+			FloatTransform3D createScaling((coordinateTransform a as Float), -(coordinateTransform e as Float), 1.0f)
```
davidpiuva: Just flipping the orthogonal system since there is no perspective.

bettinaselig:
@emilwestergren is doing the peer review instead.

davidpiuva:
ok

source/draw/gpu/opengl/OpenGLCanvas.ooc
emilwestergren: Where do you take `focalLength` into account?

davidpiuva: I probably have to add it here somehow.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -32,16 +32,10 @@ OpenGLCanvas: class extends OpenGLSurface {
...
-		if (this _focalLength > 0.0f) {
-			a := 2.0f * this _focalLength / this size x
-			f := -(this _coordinateTransform e as Float) * 2.0f * this _focalLength / this size y
-			k := (this _farPlane + this _nearPlane) / (this _farPlane - this _nearPlane)
-			o := 2.0f * this _farPlane * this _nearPlane / (this _farPlane - this _nearPlane)
-			gpuMap projection = FloatTransform3D new(a, 0.0f, 0.0f, 0.0f, 0.0f, f, 0.0f, 0.0f, 0.0f, 0.0f, k, -1.0f, 0.0f, 0.0f, o, 0.0f)
-		} else
-			gpuMap projection = FloatTransform3D createScaling(2.0f / this size x, -(this _coordinateTransform e as Float) * 2.0f / this size y, 1.0f)
-		gpuMap model = this _createModelTransform(IntBox2D new(this size))
+		gpuMap model = FloatTransform3D createTranslation(0.0f, 0.0f, this focalLength) // TODO: set using a destination argument in drawState
```
davidpiuva: Does this look better? I modified the formula from createModelTransform to be centered and normalized.

emilwestergren: This is confusing now. Focal length is measured in pixels and is part of the model and projection transform, but you have view transform normalized. Are you sure everything works out as it should?

davidpiuva:
I leave it as is until we have a better idea of how to define the focal depth without depending too much on image resolution or aspect ratio.

Issue #1256, opened 2016-01-26
Abstract shared contexts in OpenGLContext


emilwestergren:
Currently the user needs to manage and create an `OpenGLContext` for each thread where it is used. This could be abstracted by internally creating a new shared backend context for every new thread which uses the `OpenGLContext` (or in practice we want a `register` function since we want all context creation in the `VidhanceContext` constructor).

Each backend context can then hold instances of everything that is needed to maintain thread safety, like default shaders.

Since not every OpenGL resource can be shared between contexts, these need special care. This is the case for Framebuffer Objects (used when rendering to a texture), Vertex Array Objects, Pixel Buffer Objects. This means these need special consideration when recycled and reused in different threads.

@davidpiuva @sebastianbaginski please give me your opinions since it will probably be one of you who implements this

davidpiuva:
Being able to create GPU resources without having to pass a context would be nice but having a pointer passed to function calls is a way to make sure that a context exists before we create things.

sebastianbaginski:
Let's list some cases affected by having multiple threads working in the graph (filter A and filter B are running in different threads TA / TB), maybe it will be easier to think about good solution then:
- A wants to render on image, it calls `image canvas` and a FBO is created in thread TA
- frame propagates to filter B, control switches to thread TB
- B wants to render, but it can't use the same FBO as it was created in TA
- B needs to use own opengl context to create FBO in thread TB, is that right ?

I was thinking about having a base class for thread-aware resources, something like (don't mind the names and lack of obvious mutex locks)
```ooc
ThreadedResource: class {
    HashMap<ThreadID, VectorList<This> >: static perThreadResources
    threadAffinity : ThreadID
    new: static func (some parameters) -> This {
        if (This->perThreadResources[thisThreadId] isEmpty == false && check parameters)
            return this->perThreadResources takeFirst()
        return This new(parameters) // also sets threadAffinity
    }
    free: override func {
        This perThreadResources[thisThreadId] << this
    }
    checkThreadAffinity: func -> Bool {
          this threadAffinity == thisThreadID()
    }
}
```
I think you get the idea.
And then use it as base class for resources that must be recycled in correct threads:
```ooc
FrameBufferObject: abstract class extends ThreadedResource {
}

PixelBufferObject: abstract class extends ThreadedResource {
}
```

So we could have the recycling logic in one place.

davidpiuva:
It seems like Sebastian has more experience with multi threading on the CPU.

[...]

PR #1266, opened 2016-01-27
base class for synchronized resource

sebastianbaginski:
prototype for https://github.com/magic-lang/ooc-kean/issues/1256
not for review yet

### DIFF
--- a/source/concurrent/SynchronizedResource.ooc
+++ b/source/concurrent/SynchronizedResource.ooc
@@ -1,0 +1,81 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use base
+use concurrent
+import threading/Thread
+import threading/Mutex
+import structs/HashMap
+
+SynchronizedResource: abstract class {
+	_threadAffinity := 0L
+	_recycle := true
+	init: func { this _threadAffinity = Thread currentThreadId() }
+	checkThreadAffinity: func -> Bool {
+		Thread equals(this _threadAffinity, Thread currentThreadId())
+	}
+}
+
+SynchronizedResourceRecycler: class {
+	_mutex := Mutex new()
+	_resources := HashMap<Long, VectorList<SynchronizedResource>> new()
+	init: func
+	free: override func {
+		this _clear()
+		this _mutex free()
+		this _resources free()
+	}
+	create: func -> SynchronizedResource {
+		result: SynchronizedResource = null
+		this _mutex lock()
+		objects := this _resources get(Thread currentThreadId())
+		if (objects && objects count > 0)
+			result = objects remove()
+		this _mutex unlock()
+		result
+	}
+	create: func ~withTestFunction (testSettings: Func (SynchronizedResource) -> Bool) -> SynchronizedResource {
+		result: SynchronizedResource = null
+		this _mutex lock()
+		objects := this _resources get(Thread currentThreadId())
+		if (objects)
+			for (i in 0 .. objects count)
+				if (testSettings(objects[i])) {
+					result = objects[i]
+					objects removeAt(i)
+					break
+				}
+		this _mutex unlock()
+		result
+	}
+	recycle: func (object: SynchronizedResource) {
+		this _mutex lock()
+		threadId := Thread currentThreadId()
+		objects := this _resources get(threadId)
+		if (!objects)
+			objects = VectorList<SynchronizedResource> new()
+		objects add(object)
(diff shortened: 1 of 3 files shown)


marcusnaslund:
You crashed `PRInt`. :(

sebastianbaginski:
I just clicked `run tests` manually :D

sebastianbaginski:
Few times in a row, I think :)

marcusnaslund:
> Few times in a row, I think :)

Haha yes, there's a bug when a PR is updated (or tests restarted) twice in a short time. It's back up again now.

sebastianbaginski:
I think it's ready for review. There still will be some code duplication when using this in different classes, but at least it takes care of the resource set management and thread safety.
Basic usage to recycle a class per-thread (complete example with per-object settings is in the test case)
```
Resource: class extends SynchronizedResource {
   recycler := static SynchronizedResourceRecycler new()
   init: func
   free: override func {
		if (this _recycle)
			This recycler recycle(this)
		else
			super()
	}
	new: static func ~recycled -> This {
		result: This
		object := This recycler create()
		if (object)
			result = object as This
		else
			result = This new()
		result
	}
}

// get resource from per-thread set:
object := Resource new~recycled()
// puts the resource back to the set
object free()
```

It can't be implemented simply like 
```
SynchronizedResource: abstract class {
    resourceSet := static set<This>
}

Resource1: class extends SynchronizedResource
Resource2: class extends SynchronizedResource
```
because both classes `Resource1` and `Resource2` would share the same instance of `resourceSet` (it is only one static created for base class, not one for each derived classes) ...

@emilwestergren review ?

sebastianbaginski:
@marcusnaslund review if @emilwestergren is too busy ?

test/concurrent/SynchronizedResourceTest.ooc
```diff
@@ -0,0 +1,105 @@
...
+	new: static func ~recycled (value: Int) -> This {
+		result: This
+		withSettings := func (recycled: SynchronizedResource) -> Bool {
+			(recycled as This) value == value
+		}
+		object := This recycler create(withSettings)
+		if (object)
+			result = object as This
+		else
+			result = This new(value)
```
marcusnaslund: `result = (object != null) ? object as This : This new(value)`

test/concurrent/SynchronizedResourceTest.ooc
```diff
@@ -0,0 +1,105 @@
...
+			expect(recycledObjects != null)
+			expect(recycledObjects count, is equal to(1))
+			object = TestObject new~recycled(3)
+			expect(object checkThreadAffinity())
+			expect(recycledObjects count, is equal to(0))
+			expect(object objectId, is equal to(counter))
+			object free()
+			object = TestObject new~recycled(16)
+			expect(recycledObjects count, is equal to(1))
+			expect(counter != object objectId)
```
marcusnaslund: `is equal to`?

test/concurrent/SynchronizedResourceTest.ooc
```diff
@@ -0,0 +1,105 @@
...
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use base
+use concurrent
+use collections
+use unit
+import threading/Thread
+import structs/HashMap
```
marcusnaslund: Do you use it?

sebastianbaginski: Without it compiler won't let me access `keys count` some lines below.

marcusnaslund: Ah alright.

sebastianbaginski:
updated

source/concurrent/SynchronizedResource.ooc
```diff
@@ -0,0 +1,81 @@
...
+	_recycle := true
+	init: func { this _threadAffinity = Thread currentThreadId() }
+	checkThreadAffinity: func -> Bool {
+		Thread equals(this _threadAffinity, Thread currentThreadId())
+	}
+}
+
+SynchronizedResourceRecycler: class {
+	_mutex := Mutex new()
+	_resources := HashMap<Long, VectorList<SynchronizedResource>> new()
```
marcusnaslund: Because the key is not of type `Int` (or rather doesn't have its size) the hashmap will not use the int hash but rather the more complicated generic `murmurHash`. I don't really know too much about hashing functions to know whether this is good or bad, but it certainly is less efficient.

marcusnaslund: Let's consider it later, though, as there is more to be done on the hashmap. It's nice to see it being used.

sebastianbaginski: Yeah I don't think it really matters compared to, for example, the GPU upload overhead etc.

Issue #1265, opened 2016-01-27
Fixture expect printouts no longer works


sebastianbaginski:
`expect(3, is equal to(4))`
> -> : expected  equal to  '', was ''

@marcusnaslund can you reproduce

marcusnaslund:
Even better: `expected  equal to  '�.i', was '�.i'`

marcusnaslund:
Could be, but hopefully isn't, related to the number system switch.

sebastianbaginski:
OK, so this is about wrong formatting placeholders ?

marcusnaslund:
It seems like we should only use `%`. I will try.

marcusnaslund:
Wait, `"%i %f" printfln(3, 4.2f)` works just fine. So there's something wrong with `Fixture` itself.

thomasfanell:
I got similar results before, and that turned out to be ownership issues with `Text`. In my particular case I solved it by taking ownership `take()`. I haven't scanned through this particular code path, but it could be worth some consideration.

marcusnaslund:
In `Fixture createFailureMessage` where I messed up a test comparing ints:

`(leftValue as Cell<Int>) val toString() println()` works
`(leftValue as Cell) toString() println()` prints nonsense

sebastianbaginski:
It looks like a weird error with type resolution in Cell, it works when I replaced toString in Cell with this
```
toString: func -> String {
		result: String
		if (T inheritsFrom?(Text))
			result = (this val as Text) toString()
		else if (T inheritsFrom?(Bool))
			result = (this val as Bool) toString()
		else if (T inheritsFrom?(Char))
			result = (this val as Char) toString()
		else if (T inheritsFrom?(Int))
			result = (this val as Int) toString()
		else if (T inheritsFrom?(Long))
			result = (this val as Long) toString()
		else if (T inheritsFrom?(LLong))
			result = (this val as LLong) toString()
		else if (T inheritsFrom?(UInt))
			result = (this val as UInt) toString()
		else if (T inheritsFrom?(ULong))
			result = (this val as ULong) toString()
		else if (T inheritsFrom?(ULLong))
			result = (this val as ULLong) toString()
		else if (T inheritsFrom?(Float))
			result = "%.8f" formatFloat(this val as Float)
		else if (T inheritsFrom?(Double))
			result = "%.12f" formatDouble(this val as Double)
		else if (T inheritsFrom?(LDouble))
[...]

PR #1268, opened 2016-01-27
some fixes in Fixture

sebastianbaginski:
Fixes https://github.com/magic-lang/ooc-kean/issues/1265
If someone has better idea how to fix this then please let me know
@marcusnaslund

### DIFF
--- a/source/sdk/lang/types.ooc
+++ b/source/sdk/lang/types.ooc
@@ -128,21 +128,36 @@ Cell: class <T> {
 	get: func -> T { val }
 
 	toString: func -> String {
-		match (this val) {
-			case value: Text => value toString()
-			case value: Bool => value toString()
-			case value: Char => value toString()
-			case value: Int => value toString()
-			case value: Long => value toString()
-			case value: Long => value toString()
-			case value: UInt => value toString()
-			case value: ULong => value toString()
-			case value: ULong => value toString()
-			case value: Float => "%.8f" formatFloat(value)
-			case value: Double => "%.12f" formatDouble(value)
-			case value: LDouble => "%.12f" formatLDouble(value)
-			case => raise("[Cell] toString() is not implemented on the specified type"); null
-		}
+		result: String
+		if (T inheritsFrom?(Text))
+			result = (this val as Text) toString()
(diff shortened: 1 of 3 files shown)


marcusnaslund:
Not too terrible, but it's a shame if we can't trust `match` statements.

marcusnaslund:
Did you try to match on `this val class`, remove `value`, and instead do `this val toString()` on every line?

sebastianbaginski:
Yes, `this val class` crashed at runtime. `T class` caused compilation error.

marcusnaslund:
Alright then. Have you tried this for failing comparisons of `String`?

sebastianbaginski:
I didn't, only for text and some numeric types

marcusnaslund:
It's alright, it wasn't good before and it'll be replaced with `Text` soon.

marcusnaslund:
Wait, there is no `String` clause amongst these `if` statements. That explains a lot. Would you add that too?

sebastianbaginski:
ok

marcusnaslund:
Makes very little sense to have `toString` if it cannot handle `String`. :)

sebastianbaginski:
Updated.

source/sdk/lang/types.ooc
```diff
@@ -128,21 +128,36 @@ Cell: class <T> {
...
-			case value: ULong => value toString()
-			case value: Float => "%.8f" formatFloat(value)
-			case value: Double => "%.12f" formatDouble(value)
-			case value: LDouble => "%.12f" formatLDouble(value)
-			case => raise("[Cell] toString() is not implemented on the specified type"); null
-		}
+		result: String
+		if (T inheritsFrom?(Text))
+			result = (this val as Text) toString()
+		if (T inheritsFrom?(String))
```
marcusnaslund: missing `else`

sebastianbaginski: ...

sebastianbaginski: fixed

marcusnaslund:
Works well now, nice.

PR #1269, opened 2016-01-27
fixed some memory leaks in HashMap

sebastianbaginski:
Fixed memory leaks in `resize`
Also fixes leak related to using iterator in ArrayList :: copy
@marcusnaslund


sebastianbaginski:
Needed for https://github.com/magic-lang/ooc-kean/pull/1266

PR #1270, opened 2016-01-28
`RecycleBin` moved and backend changed to `SynchronizedList`

christopherlagerhult:
`RecycleBin` moved and updated

### DIFF
--- a/source/concurrent/RecycleBin.ooc
+++ b/source/concurrent/RecycleBin.ooc
@@ -11,7 +11,7 @@ import threading/Mutex
 RecycleBin: class <T> {
 	_mutex := Mutex new()
 	_free: Func (T)
-	//TODO: Replace with SynchronizedVectorList
+	//TODO: Replace with SynchronizedList
 	_list: VectorList<T>
 	_size: Int
 	count ::= this _list count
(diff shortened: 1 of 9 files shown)


marcusnaslund:
You should rename this PR to something more appropriate/descriptive.

christopherlagerhult:
@emilwestergren peer review?

source/concurrent/RecycleBin.ooc
```diff
@@ -14,48 +14,39 @@
...
 		if (this isFull)
 			this _free(this _list remove(0))
 		this _list add(object)
-		this _mutex unlock()
 	}
 	search: func (matches: Func (T) -> Bool) -> T {
 		result: T = null
-		this _mutex lock()
 		for (i in 0 .. this _list count) {
 			if (matches(this _list[i])) {
```
emilwestergren: I just realized this will now be much more expensive since the mutex lock in the list will be locked/unlocked twice for every element in the list (`this _list[i]` and `this _list remove(i)`). Previously it was only locked once before and after the loop. Can you use the `apply` function and solve this with a single lock/unlock like before?

source/concurrent/RecycleBin.ooc
```diff
@@ -5,49 +5,34 @@
  * This software may be modified and distributed under the terms
  * of the MIT license.  See the LICENSE file for details.
  */
-
-import threading/Mutex
+use concurrent
```
marcusnaslund: There should be an empty line after the license

test/concurrent/RecycleBinTest.ooc
```diff
@@ -1,6 +1,6 @@
 use unit
```
marcusnaslund: There should be a license here now, rebasing (or adding it manually) will fix the conflicts.

marcusnaslund:
Is there anything left to do here? @emilwestergren - will you finish peer review?

PR #1271, opened 2016-01-28
Fixed toText for all numeric types and replaced Cell toString

### LINKED ISSUE #1267: Replace Cell toString with toText
This is our own function, so should be no problem. I believe it's only used in `Fixture`.

marcusnaslund:
Fixes #1267

### DIFF
--- a/source/sdk/Numbers.ooc
+++ b/source/sdk/Numbers.ooc
@@ -15,6 +15,12 @@ ULLONG_MAX: extern static ULLong
 LLong: cover from signed long long {
 	toString: func -> String { "%lld" formatLLong(this as LLong) }
 	toHexString: func -> String { "%llx" formatLLong(this as LLong) }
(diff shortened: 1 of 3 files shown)


source/sdk/Numbers.ooc
```diff
@@ -95,17 +88,18 @@ Double: cover from double extends LDouble {
 	}
 }
 
+Double: cover from double extends LDouble {
+	isNumber ::= this == this
+	toString: func -> String {
+		"%.2f" formatDouble(this)
```
marcusnaslund: What a confusing diff. Anyway, `"%.2f" formatDouble(this)` is now `toString` of both `LDouble` and `Double`, because of a bug in rock. https://github.com/fasterthanlime/rock/issues/445

marcusnaslund:
Peer review @thomasfanell ?

Issue #741, opened 2015-11-11
Remove VarArgs


marcusnaslund:
There appears to be no way of using these without leaking a lot of memory.

The only use from us is in `SvgWriter2D` - we should be able to easily fix this by using a `VectorList` instead. VarArgs are also used in a few functions in `math.ooc` and in other parts of the SDK, but not much longer.

marcusnaslund:
This will also allow us to simplify calls like `Int maximum~two` to just `Int maximum`.

marcusnaslund:
Or, can we do this? How would we implement `format` for `Text` without it? @sebastianbaginski

sebastianbaginski:
I think we could limit ourselves to finite count of arguments and implement `format` without `VarArgs`.
Something like 
```ooc
format: func ~main <T, U> (fmt: T, arg: U) -> T
format: func ~main2 <T, U, V> (fmt: T, arg1: U, arg2: V) -> T
format: func ~main3 <T, U, V, W, X> (fmt: T, arg1: U, arg2: V, arg3: X) -> T
format: func ~main4 <T, U, V, W, X, Y> (fmt: T, arg1: U, arg2: V, arg3: X, arg4: Y ) -> T

// and so on, until we are happy with count of arguments
```
instead of
```ooc
format: func ~main <T> (fmt: T, args: ... ) -> T {
```

sebastianbaginski:
Of course `~main2` can simply call `~main` twice, `~main3` can call `~main2` and `~main` and so on.

marcusnaslund:
We can get rid of `vwritef` in `BufferWriter` and `writef` in `Writer` without any consequences. Just `Format.ooc` remains then.

PR #1272, opened 2016-01-28
Removed never-used VarArgs functions

marcusnaslund:
In line with #741 

Unlike what the comment says, `writef` does not call `vwritef`. Nothing does. They seem more like old remnants.

Objections @sebastianbaginski ?


sebastianbaginski:
> Objections @sebastianbaginski ?

nope, please continue

PR #1274, opened 2016-01-28
Removed not needed Vector imports

marcusnaslund:


marcusnaslund:
Peer review @thomasfanell

PR #1275, opened 2016-01-28
Removed the Dynlib memory leak present in all ooc/magic programs

marcusnaslund:
Lines 29-37 will definitely set `Dynlib suffix` to something, and the original empty string is thus lost.

Building an empty test program now leaks only 64 bytes instead of 192 bytes.


sebastianbaginski:
OK:)

PR #1276, opened 2016-01-29
Implemented source and destination for DrawState.

davidpiuva:
Implemented the source and destination arguments from the old API.
I have compared the results between the old and new API without seeing any differing pixel so far.

There is a problem with encapsulation because automatically generated getters and setters on covers cannot guarantee that the side effect is not thrown away to a temporary clone of the value.
Callers should be using the `set` functions to change a `DrawState` even if only a single setting is changed so that the internal representation can change freely.

### DIFF
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -12,7 +12,7 @@ import Image
 import Map
 
 // Example
-// DrawState new(myImage) setMap(myShader) draw()
+// DrawState new(targetImage) setMap(shader) draw()
 
 DrawState: cover {
 	target: Image = null
@@ -21,6 +21,8 @@ DrawState: cover {
 	opacity := 1.0f
 	_transformNormalized := FloatTransform3D identity
 	viewport := IntBox2D new(0, 0, 0, 0)
+	destination := IntBox2D new(0, 0, 0, 0) // TODO: Accept FloatBox2D
+	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	init: func@ ~default
 	init: func@ ~target (=target)
 	setTarget: func (target: Image) -> This {
@@ -39,6 +41,21 @@ DrawState: cover {
 		this viewport = viewport
 		this
 	}
+	setDestination: func (destination: IntBox2D) -> This {
+		this destination = destination
+		this
+	}
+	setSource: func ~Int (source: IntBox2D, imageSize: IntVector2D) -> This {
+		this setSource(source toFloatBox2D(), imageSize toFloatVector2D())
+	}
+	setSource: func ~Float (source: FloatBox2D, imageSize: FloatVector2D) -> This {
+		this setSourceNormalized(source / imageSize)
+	}
+	setSourceNormalized: func (source: FloatBox2D) -> This {
+		this _sourceNormalized = source
+		this
+	}
+	getSourceNormalized: func -> FloatBox2D { this _sourceNormalized }
 	setInputImage: func (inputImage: Image) -> This {
 		this inputImage = inputImage
 		this
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -37,39 +37,50 @@ DrawState: cover {
 		this opacity = opacity
 		this
 	}
+	// Local region
 	setViewport: func (viewport: IntBox2D) -> This {
 		this viewport = viewport
 		this
 	}
+	// Local region
 	setDestination: func (destination: IntBox2D) -> This {
 		this destination = destination
 		this
 	}
+	// Local region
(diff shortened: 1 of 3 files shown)


davidpiuva:
@emilwestergren Peer review?

davidpiuva:
@ziquanyu Peer review?

source/draw/DrawState.ooc
```diff
@@ -29,6 +29,8 @@ DrawState: cover {
 	opacity := 1.0f
 	_transformNormalized := FloatTransform3D identity
 	viewport := IntBox2D new(0, 0, 0, 0)
+	destination := IntBox2D new(0, 0, 0, 0) // TODO: Accept FloatBox2D
```
davidpiuva: Should I make this private or just wait for a better type?

davidpiuva:
Rebased on MIT license.

marcusnaslund:
All tests (except license checker) passed. Restarting now because I hadn't merged the new licenses for the test files before this was run, and so this didn't pass. It will be fine now.

davidpiuva:

source/draw/DrawState.ooc
```diff
@@ -39,6 +41,21 @@ DrawState: cover {
 		this viewport = viewport
 		this
 	}
+	setDestination: func (destination: IntBox2D) -> This {
```
davidpiuva: Local coordinates are default for source and destination.

source/draw/DrawState.ooc
sebastianbaginski: Why having a TODO instead of changing this to `FloatBox2D` ?

source/draw/gpu/GpuSurface.ooc
```diff
@@ -82,9 +82,15 @@ GpuSurface: abstract class extends Canvas {
 	}
 	draw: virtual func ~mesh (image: GpuImage, mesh: GpuMesh) { Debug raise("draw~mesh unimplemented!") }
 	readPixels: virtual func -> ByteBuffer { raise("readPixels unimplemented!"); null }
-	_createTextureTransform: static func (imageSize: IntVector2D, box: IntBox2D) -> FloatTransform3D {
-		scaling := FloatTransform3D createScaling(box size x as Float / imageSize x, box size y as Float / imageSize y, 1.0f)
-		translation := FloatTransform3D createTranslation(box leftTop x as Float / imageSize x, box leftTop y as Float / imageSize y, 0.0f)
+	_createTextureTransform: static func ~Int (imageSize: IntVector2D, box: IntBox2D) -> FloatTransform3D {
+		This _createTextureTransform(imageSize toFloatVector2D(), box toFloatBox2D())
+	}
+	_createTextureTransform: static func ~Float (imageSize: FloatVector2D, box: FloatBox2D) -> FloatTransform3D {
```
sebastianbaginski: Maybe this and the function above could also have a `~local` part in the suffix, as `static func ~Normalized` below.

source/draw/DrawState.ooc
davidpiuva: This version is a wrapper on the old API that only takes integers. The next step is much more complex and might break backward compability if I create a bug.

source/draw/DrawState.ooc
sebastianbaginski: OK, I think it could be good to include this information somewhere in the source code.

source/draw/DrawState.ooc
sebastianbaginski: Alright

davidpiuva: I would get one less argument if I give in local coordinates but floating. Someone might however mistake FloatBox2D for normalized coordinates then. If I store in normalized coordinates, I have to send in the size of the image or take it from the DrawState but then I can have (0, 0, 1, 1) as default value instead of checking if the area exists.

davidpiuva:
Updated

marcusnaslund:
`DrawState` has a lot of comments now that, if they're needed, mostly belong somewhere else.

davidpiuva:
Sebastian said that it might be good to explain things there.

sebastianbaginski:
That's true. But now after you have created this documentation in separate PR, maybe it could be a good idea to just a simple comment redirecting there.

marcusnaslund:
> it could be a good idea to have just a simple comment redirecting there.

Agreed.

davidpiuva:
These are specific about each overload which the document sais nothing about.

sebastianbaginski:
OK, so maybe leave these very specific ones about overloads in the source code. I'll let @marcusnaslund decide though.

marcusnaslund:
First order of business: This leaks memory.

davidpiuva:
Now I only mention the coordinate systems for each function. If the readme.md document comes in before https://github.com/magic-lang/ooc-kean/pull/1295 then it can be referenced there.

davidpiuva:
PRInt is done.

PR #1286, opened 2016-01-31
Added WorkerThread class

### LINKED ISSUE #1284: Implement a wrapper class "WorkerThread"
This should use a ThreadPool with one thread as backend.

marcusnaslund:
Fixes #1284

### DIFF
--- a/test/concurrent/WorkerThreadTest.ooc
+++ b/test/concurrent/WorkerThreadTest.ooc
@@ -1,0 +1,65 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use collections
+use concurrent
+use unit
+import threading/Thread
+
+WorkerThreadTest: class extends Fixture {
+	threads := static SynchronizedList<Long> new()
+	sum: static Long = 0
+	taskUp := func {
+		This threads add(Thread currentThreadId())
+		Time sleepMilli(250)
+		This sum += 1
+		//">" print(); fflush(stdout)
+		Time sleepMilli(250)
+		This threads add(Thread currentThreadId())
+	}
+	taskDown := func {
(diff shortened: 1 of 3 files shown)


marcusnaslund:
Peer review/Agree this is a good idea @emilwestergren ?

test/concurrent/WorkerThreadTest.ooc
```diff
@@ -0,0 +1,48 @@
...
+		this add("Create and wait for all", func {
+			thread := WorkerThread new()
+			thread getPromise(this taskUp) wait() . free()
+			expect(this sum, is greater than(0))
+
+			thread add(this taskUp)
+			thread add(this taskUp)
+			thread add(this taskDown)
+			thread add(this taskDown)
+			thread add(this taskUp)
```
sebastianbaginski: Maybe verify that these tasks really run in the same thread, if this is the purpose of this class.

emilwestergren:
Looks good

test/concurrent/WorkerThreadTest.ooc
marcusnaslund: That's a good idea, and it's now done.

marcusnaslund:
You can perhaps take over peer review @sebastianbaginski

test/concurrent/WorkerThreadTest.ooc
sebastianbaginski: ok, nice

sebastianbaginski:
I agree with Emil, it looks ok.
It seems a little weird to have a thread `pool` with just one thread (as `WorkerThread` **is a** `ThreadPool` and can be used whenever ThreadPool is expected) but I understand why you want to do this.

emilwestergren:
I agree that it should probably not *be* a `ThreadPool`, rather *have* a `ThreadPool` as backend. However, when inheriting you get the same interface for free which is nice in this case.

marcusnaslund:
This works locally but PRInt just stops after the license checker for... some reason.

marcusnaslund:
This fails with `gpuOff` for no appearent reason. :/ I'm taking this off final review until I sort it out.

PR #1288, opened 2016-01-31
Removed useless comments related to Backtrace/Exception

marcusnaslund:
Any objections @sebastianbaginski ?


sebastianbaginski:
> Any objections

not really, I don't think we need to copy-paste man pages into source code :)

PR #1291, opened 2016-01-31
Removed ArrayList<T>

### LINKED ISSUE #1233: Get rid of ArrayList<T>

marcusnaslund:
```
I've been waiting for that bright sunshine to show up and shine in my back door someday.
- Luther Allison
```
Say farewell. Fixes #1233 
Peer review @thomasfanell , @fredrikbryntesson , @sebastianbaginski ?


sebastianbaginski:
OK, I don't see any reason for keeping this.

fredrikbryntesson:
Kill it, with fire.

marcusnaslund:
Killed.

Issue #1277, opened 2016-01-29
Fix lingering gcc warnings from shared library


marcusnaslund:
When building this repo as part of a shared library, `gcc` with `-Wall` provides the following output:

```
/home/mnaslund/versioned/ooc/ooc-kean/source/io/native/ProcessUnix.ooc: In function ‘native_ProcessUnix__ProcessUnix_terminate_impl’:
/home/mnaslund/versioned/ooc/ooc-kean/source/io/native/ProcessUnix.ooc:25:9: warning: implicit declaration of function ‘kill’ [-Wimplicit-function-declaration]
    kill(pid, SIGTERM)
         ^
/home/mnaslund/versioned/ooc/ooc-kean/source/sdk/io/native/FileUnix.ooc: In function ‘io_native_FileUnix__FileUnix_createTempFile’:
/home/mnaslund/versioned/ooc/ooc-kean/source/sdk/io/native/FileUnix.ooc:333:5: warning: implicit declaration of function ‘mkstemp’ [-Wimplicit-function-declaration]
    fd := mkstemp(pattern)
     ^
/home/mnaslund/versioned/ooc/ooc-kean/source/sdk/io/native/FileUnix.ooc:334:5: warning: implicit declaration of function ‘fdopen’ [-Wimplicit-function-declaration]
    result: FStream = fd >= 0 ? fdopen(fd, mode) : null
     ^
/home/mnaslund/versioned/ooc/ooc-kean/source/sdk/io/native/FileUnix.ooc:334:98: warning: pointer/integer type mismatch in conditional expression [enabled by default]
    result: FStream = fd >= 0 ? fdopen(fd, mode) : null
                                                                                                  ^
/home/mnaslund/versioned/ooc/ooc-kean/source/sdk/IO.ooc: In function ‘IO__FStream_open_withFlags’:
/home/mnaslund/versioned/ooc/ooc-kean/source/sdk/IO.ooc:97:5: warning: implicit declaration of function ‘fdopen’ [-Wimplicit-function-declaration]
   fd >= 0 ? fdopen(fd, mode) : null
     ^
/home/mnaslund/versioned/ooc/ooc-kean/source/sdk/IO.ooc:97:84: warning: pointer/integer type mismatch in conditional expression [enabled by default]
   fd >= 0 ? fdopen(fd, mode) : null
                                                                                    ^
[...]

PR #1292, opened 2016-01-31
Removed gcc warnings for pthread_kill, kill, usleep

marcusnaslund:
Takes care of three points of #1277 

### DIFF
--- a/source/sdk/Time.ooc
+++ b/source/sdk/Time.ooc
@@ -9,8 +9,9 @@
 import os/win32
 
 version(linux) {
-	include unistd | (__USE_BSD), sys/time | (__USE_BSD), time | (__USE_BSD)
-	//include unistd, sys/time, time
+	include unistd | (__USE_BSD, _BSD_SOURCE, _DEFAULT_SOURCE)
+	include sys/time | (__USE_BSD, _BSD_SOURCE, _DEFAULT_SOURCE)
+	include time | (__USE_BSD, _BSD_SOURCE, _DEFAULT_SOURCE)
 } else {
 	include unistd, sys/time
 }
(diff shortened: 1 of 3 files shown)


source/sdk/Time.ooc
```diff
@@ -9,8 +9,7 @@
 import os/win32
 
 version(linux) {
-	include unistd | (__USE_BSD), sys/time | (__USE_BSD), time | (__USE_BSD)
-	//include unistd, sys/time, time
+	include unistd | (__USE_BSD, _BSD_SOURCE), sys/time | (__USE_BSD, _BSD_SOURCE), time | (__USE_BSD, _BSD_SOURCE)
```
sebastianbaginski: According to http://man7.org/linux/man-pages/man7/feature_test_macros.7.html : 
```
_BSD_SOURCE (deprecated since glibc 2.20)
...
Since glibc 2.20, this macro is deprecated.  It now has the same effect as defining _DEFAULT_SOURCE, but generates a compile-time warning (unless _DEFAULT_SOURCE is also defined).
To allow code that requires _BSD_SOURCE in glibc 2.19 and earlier and _DEFAULT_SOURCE in
glibc 2.20 and later to compile without warnings, define both _BSD_SOURCE and _DEFAULT_SOURCE.
```
can you try with both macros ?

marcusnaslund: Good catch! I'm running 2.19 so I didn't notice the other warning.

sebastianbaginski:
Is it ready now ?

marcusnaslund:
Sorry, yes, I've added the flag as you suggested.

sebastianbaginski:
OK, great

PR #1294, opened 2016-01-31
Cleaned up Char and CString

marcusnaslund:
Looks like a rewrite, but it's only code rearranged. Also added missing `this` and removed all `return`.

Peer review @thomasfanell or @fredrikbryntesson ?

### DIFF
--- a/source/sdk/Character.ooc
+++ b/source/sdk/Character.ooc
@@ -26,75 +26,35 @@ strtod: extern func (Char*, Pointer) -> Double
 version (!cygwin) {
 	strtold: extern func (Char*, Pointer) -> LDouble
 } else {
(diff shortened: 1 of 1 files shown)


source/sdk/Character.ooc
```diff
@@ -127,33 +90,30 @@ operator as (value: CString) -> String { value ? value toString() : null }
...
-		if (other == null) return false
-		l := length()
-		l2 := length()
-		if (l != l2) return false
-
-		for (i in 0 .. l) {
-			if (this[i] != other[i]) return false
+		result := false
+		if (other != null) {
+			(length, otherLength) := (this length(), other length())
```
marcusnaslund: Great example of why this is so useful. The original code doesn't even check `other length`, and would eventually crash if it's shorter.

PR #1295, opened 2016-02-01
Destination

davidpiuva:
Normalized the internal representation of destination so that the same value can be used for multiple resolutions. You can now set a destination for multiple resolutions at the same time. Viewport is however still stored in pixels because it is just defining a subset of the target.
```ooc
// Create a DrawState with destination and transform for both Y and UV
state := DrawState new() setDestination(location, frame size) setTransformReference(rotation, frame size)
// Clone and set targets for Y and UV draw calls
stateY := state setTarget(inputImage y)
stateUV := state setTarget(inputImage uv)
// Use the draw states for drawing to Y and UV
stateY setInputImage(result y) draw()
stateUV setInputImage(result uv) draw()
```

### DIFF
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -14,6 +14,8 @@ import Map
 // Example
 // DrawState new(targetImage) setMap(shader) draw()
 
+// See README.md about input arguments and coordinate systems
+
 DrawState: cover {
 	target: Image = null
 	inputImage: Image = null
@@ -21,7 +23,7 @@ DrawState: cover {
 	opacity := 1.0f
 	_transformNormalized := FloatTransform3D identity
 	viewport := IntBox2D new(0, 0, 0, 0)
-	destination := IntBox2D new(0, 0, 0, 0) // TODO: Accept FloatBox2D
+	_destinationNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	init: func@ ~default
 	init: func@ ~target (=target)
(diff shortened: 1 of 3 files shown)


davidpiuva:

sebastianbaginski:
Sure, I will just wait for the other PR to go upstream, maybe something will have to be changed.

davidpiuva:
ok

davidpiuva:
Added reference to README.md.

davidpiuva:
@sebastianbaginski The other PR is merged and this can be reviewed.

source/draw/DrawState.ooc
```diff
@@ -43,11 +45,25 @@ DrawState: cover {
 		this
 	}
 	// Local region
-	setDestination: func (destination: IntBox2D) -> This {
-		this destination = destination
-		this
+	// Precondition: target exists
```
sebastianbaginski: It could go into `version(safe)` block. @marcusnaslund ?

sebastianbaginski:
I see no problems here, but I don't use the draw API a lot. Maybe @emilwestergren would like to have a look

source/draw/DrawState.ooc
marcusnaslund: Or really not be here at all. Obviously `target` must exist if `this target size` is to make any sense.

davidpiuva: If a crash can occur from something that is not given as argument then a warning is needed.

sebastianbaginski: Yes, but someone might call this function without setting the target.

davidpiuva: One could also mention the precondition for draw but who would draw to nothing.

marcusnaslund: `safe`-block it is then.

source/draw/DrawState.ooc
```diff
@@ -62,17 +78,21 @@ DrawState: cover {
...
 	// Normalized region
 	getSourceNormalized: func -> FloatBox2D { this _sourceNormalized }
-	// Giving a single texture as "texture0"
 	setInputImage: func (inputImage: Image) -> This {
 		this inputImage = inputImage
 		this
 	}
 	// Reference transform
-	setTransformReference: func ~targetSize (transform: FloatTransform3D) -> This {
+	// Precondition: target exists
```
marcusnaslund: Same here.

davidpiuva:
Updated. Even thou a comment would be too obvious, it made sense to have safety for draw too since side effects can be unexpected from a typo.

marcusnaslund:
Yeah, I understand why you wanted to put it there. A version flag seems more like our "style" though.

source/draw/DrawState.ooc
```diff
@@ -82,5 +106,10 @@ DrawState: cover {
 	}
 	// Normalized transform
 	getTransformNormalized: func -> FloatTransform3D { this _transformNormalized }
-	draw: func { this target canvas draw(this) }
+	draw: func {
+		version(safe)
+			if (this target == null)
+				raise("Can't draw without a selected target.")
```
sebastianbaginski: Hmm could we introduce `raise: static func (condition: Bool, message: String) { if (condition) raise(message) }` ?
I think it could look better
```ooc
version(safe)
    raise(this target == null, "Null target")
```
@marcusnaslund what do you think ?

davidpiuva:
```ooc
safeRaise(this target == null, "Null target")
```

marcusnaslund: Hey, I like that. Make an issue for that too, please.

davidpiuva:
PRInt only complains about a file from another PR that failed before.

marcusnaslund:
Sorry, my bad for taking shortcuts. Never again. Upstream is fixed now but we don't have to rerun the tests.

PR #1296, opened 2016-02-01
corrected unmangled function name in promise

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -28,7 +28,7 @@ Promise: abstract class {
 		collector
 	}
 	kean_concurrent_promise_wait: unmangled func { this wait() }
-	kean_concurrent_promise_wait_timeout: unmangled func (timeout: Double) { this wait(timeout) }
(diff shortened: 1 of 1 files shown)


source/concurrent/Promise.ooc
```diff
@@ -28,7 +28,7 @@ Promise: abstract class {
 		collector
 	}
 	kean_concurrent_promise_wait: unmangled func { this wait() }
-	kean_concurrent_promise_wait_timeout: unmangled func (timeout: Double) { this wait(timeout) }
+	kean_concurrent_promise_waitWithTimeout: unmangled func (timeout: Double) { this wait(timeout) }
```
marcusnaslund: Why is using `With` correct, as the original function is `wait~timeout`?

sebastianbaginski: because it is more consistent with other functions in the public header (I can't give specific examples here :) )

marcusnaslund: I'll take your word for it.

Issue #757, opened 2015-11-13
Write tests for, and fix memory leaks in, source/net


marcusnaslund:
The contents of `source/net` are basically just copied from the original sdk and cleaned up to make magic happy. We do not even know if it works since we have never used it and there are no tests. Furthermore, there are most likely several memory leaks and unnecessary dependencies on what's left in the SDK.

Here is what should be done to solve this issue:

- [x] Implement rudimentary tests for what we want to do with the `net` classes
- [x] Verify that the `net` module works as expected and has everything we want for now
- [ ] Reduce memory leaks from the tests (and thus from relevant `net` classes) to zero

More sophisticated tests will follow in a later issue if/when we need it.

marcusnaslund:
Related: https://github.com/fasterthanlime/rock/issues/814 which makes this issue deserve the `bug` label.

sebastianbaginski:
@simonmika can we implement new version of this as extension to the new IO module ? I think sockets fit into the IO component anyway

simonmika:
Yes, thats what we should do.

marcusnaslund:
Updated with a slightly better description of what's needed. @fredrikbryntesson

marcusnaslund:
Should be quite easy to take the examples from https://ooc-lang.org/docs/sdk/net/ and turn into basic tests, connecting the sockets to `localhost`.

marcusnaslund:
@fredrikbryntesson Can we tick off the first box?

marcusnaslund:
Regarding memory leaks (running the net tests):
```
definitely lost: 1,224 bytes in 49 blocks
indirectly lost: 21,672 bytes in 150 blocks
possibly lost: 0 bytes in 0 blocks
```

fredrikbryntesson:
I can check how we can free the sockets better, and yes tick off the first.

PR #1302, opened 2016-02-01
General cleanup in source/net

marcusnaslund:
Updated to conform with new versions of rock and style guides for `magic`, remove gcc warnings, and also includes some bug fixes.

Added a blank test which makes sure the code in `source/net` gets processed when running `./test.sh`. There is already an issue on writing actual tests: #757

### DIFF
--- a/source/net/Address.ooc
+++ b/source/net/Address.ooc
@@ -172,155 +172,136 @@ IP4Address: class extends IPAddress {
 			InvalidAddress new("Could not parse address") throw()
 		}
 	}
-
 	init: func ~wildcard {
 		init("0.0.0.0")
 	}
-
 	init: func ~withAddr (addr: InAddr) {
 		family = AddressFamily IP4
 		memcpy(ai&, addr&, InAddr size)
 	}
-
-	broadcast: func -> Bool { ai s_addr == INADDR_NONE }
-	wildcard: func -> Bool { ai s_addr == INADDR_ANY }
-	globalMulticast: func -> Bool {
+	broadcast: override func -> Bool { ai s_addr == INADDR_NONE }
+	wildcard: override func -> Bool { ai s_addr == INADDR_ANY }
+	globalMulticast: override func -> Bool {
 		addr := ntohl(ai s_addr)
-		return addr >= 0xE0000100 && addr <= 0xEE000000
+		addr >= 0xE0000100 && addr <= 0xEE000000
 	}
-	ip4Compatible: func -> Bool { true }
-	ip4Mapped: func -> Bool { true }
-	linkLocal: func -> Bool { (ntohl(ai s_addr) & 0xFFFF0000) == 0xA9FE0000 }
-	linkLocalMulticast: func -> Bool { (ntohl(ai s_addr) & 0xFF000000) == 0xE0000000 }
-	loopback: func -> Bool { ntohl(ai s_addr) == 0x7F000001 }
-	multicast: func -> Bool { (ntohl(ai s_addr) & 0xF0000000) == 0xE0000000 }
-	nodeLocalMulticast: func -> Bool { false }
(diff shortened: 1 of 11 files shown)


source/net/ServerSocket.ooc
```diff
@@ -118,22 +114,22 @@ ServerSocket: class extends Socket {
...
 	*/
-	accept: func ~withClosure (f: func (TCPServerReaderWriterPair) -> Bool) {
-		if (!listening)
-			listen()
+	accept: func ~withClosure (f: Func (TCPServerReaderWriterPair) -> Bool) {
+		if (!this listening)
+			this listen()
 
-		loop(||
+		while (true) {
```
marcusnaslund: I don't know if `loop` was a keyword that disappeared or what was going on here?

source/net/translation.ooc
```diff
@@ -10,63 +10,66 @@ import berkeley, Socket, Address
...
+		//version (!windows) {
 			// use built-in!
-			return inet_ntop(addressFamily, address, destination, destinationSize)
-		}
+			// TODO: Using the built-in inet_ntop results in a gcc warning because inet_ntop returns a const char*,
+			// and the const qualifier is not supported by rock...
+			//result = inet_ntop(addressFamily, address, destination, destinationSize)
+		//} else {
+			// Roll our own implementation, based on:
+			// https://github.com/pkulchenko/luasocket/blob/5a58786a39bbef7ed4805821cc921f1d40f12068/src/inet.c#L512
```
marcusnaslund: As the comment says, `inet_ntop` returns `const char*`, but the closest thing we have is `CString` which is a cover of `char*`...

source/net/ServerSocket.ooc
thomasfanell: https://github.com/fasterthanlime/rock/blob/master/sdk/lang/Abstractions.ooc

marcusnaslund: Oh yes, I remember that file. Thanks.

marcusnaslund:
Who feels like doing some networking peer review? @jonathanudd maybe?

source/net/Address.ooc
```diff
@@ -172,155 +172,136 @@ IP4Address: class extends IPAddress {
 			InvalidAddress new("Could not parse address") throw()
 		}
```
jonathanudd: Brackets could be removed here and on the if above.

source/net/Address.ooc
```diff
@@ -172,155 +172,136 @@ IP4Address: class extends IPAddress {
...
 	}
-
 	init: func ~wildcard {
```
jonathanudd: Do we want this kind of functions on one line?

thomasfanell: Only if the nearest neighbours also can be compacted to one line.

source/net/Address.ooc
```diff
@@ -350,39 +331,34 @@ SocketAddress: abstract class {
...
 		"[%s]:%d" format(host() toString() toCString(), port())
 	}
 	new: static func (host: IPAddress, nPort: Int) -> This {
+		result: This
 		if (host family == AddressFamily IP4) {
 			ip4Host := host as IP4Address
-			return SocketAddressIP4 new(ip4Host ai, nPort)
+			result = SocketAddressIP4 new(ip4Host ai, nPort)
 		}
 		else if (host family == AddressFamily IP6) {
```
jonathanudd: The `else if` could be moved to the line above. The same applies for the else below.

jonathanudd:
There are some functions that could be converted to one liners but I don't know if that will increase readability.
Apart from the comments this looks good.

marcusnaslund:
Thank you, updated according to comments. Just waiting on PRInt to give a green flag.

PR #1303, opened 2016-02-01
Added moveTo to Buffer

marcusnaslund:
Is this what you had in mind @sebastianbaginski when you mentioned this before?


sebastianbaginski:
I don't remember, but I think you may be right :)

marcusnaslund:
Peer review? :)

PR #1311, opened 2016-02-02
Merged OpelGLSurface into OpenGLCanvas.

davidpiuva:
Merging OpelGLSurface into OpenGLCanvas to make things easier to understand. Emil requested this change before.


davidpiuva:
Binding and unbinding became virtual since the parent class was merged with the most common class to inherit it. This together with following pull requests will keep the bloated inheritance down to the minimum required for polymorphism since it takes way too much time to find the called draw function when there are 5 layers of inheritance to look at.

davidpiuva:

sebastianbaginski:
> it takes way too much time to find the called draw function when there are 5 layers of inheritance to look at

I couldn't agree more.
I'm fine with this PR.

marcusnaslund:
This is causing a segfault when a video has finished playing. :/
It doesn't show up with `-w` which is why it slipped through `PRInt` testing.

PR #1312, opened 2016-02-02
Added empty promise class.

emilwestergren:
Used for creating dummy promises which return immediately.

@marcusnaslund peer

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -30,9 +30,14 @@ Promise: abstract class {
 	kean_concurrent_promise_wait: unmangled func { this wait() }
 	kean_concurrent_promise_waitWithTimeout: unmangled func (timeout: Double) { this wait(timeout) }
 	kean_concurrent_promise_free: unmangled func { this free() }
-	start: static func (action: Func) -> This {
-		_ThreadPromise new(action)
-	}
+	start: static func (action: Func) -> This { _ThreadPromise new(action) }
+	empty: static This { get { _EmptyPromise new() } }
+}
+
+_EmptyPromise: class extends Promise {
+	init: func { super() }
+	wait: override func -> Bool { true }
+	wait: override func ~timeout (seconds: Double) -> Bool { true }
 }
 
 _ThreadPromise: class extends Promise {


source/concurrent/Promise.ooc
```diff
@@ -33,6 +33,13 @@ Promise: abstract class {
 	start: static func (action: Func) -> This {
 		_ThreadPromise new(action)
 	}
+	createEmpty: static func -> This { _EmptyPromise new() }
```
emilwestergren: Don't know if `empty` is a good name, please suggest anything better if possible.

sebastianbaginski: In the other project it is a static `empty` property : `empty := static This new()`

marcusnaslund: From [Wiktionary](https://en.wiktionary.org/wiki/empty_promise):
```
empty promise ‎(plural empty promises)
(idiomatic) A promise that is either not going to be carried out, worthless or meaningless.
```
I guess it works. But it would be nicer with a static `empty` property which returns an `_EmptyPromise` instance, that would be more in line with the style in similar situations.

emilwestergren: I thought we decided not to use only `empty` due to the ambiguity between `empty: Bool` `empty: func` `empty: static func -> This`

sebastianbaginski: I think it is `isEmpty: func bool` . At least in the other project.

emilwestergren: Ok so `empty` is always a static create function?

sebastianbaginski: I don't know, I'm sure it is at least for one class.
@marcusnaslund what does our coding rules have to say about this ?

marcusnaslund: `isEmpty` is a property which returns `Bool`, `empty` returns an empty instance. `Text` is a good example.

emilwestergren: Alright then :smiley:

emilwestergren:
Updated.

marcusnaslund:
Don't mean to add more to this PR, but since one-liners seem to make you happy: `Promise start` would look better as one line now. :)

emilwestergren:
Updated

marcusnaslund:
Perfection.

emilwestergren:
Hold on, should `empty` be a property or function?

marcusnaslund:
Thanks. Property.

emilwestergren:
Updated

PR #1313, opened 2016-02-02
Readpixels fix

emilwestergren:
* Moved the allocation of the target `ByteBuffer` to a higher abstraction level, resulting in more control of where the result is written. This is mainly to prepare for the implementation of `toRaster(target: RasterImage)` for `GpuImage`.

* Removed `readPixels` from `GpuSurface` API since it should not be public.


sebastianbaginski:
Seems OK

marcusnaslund:
This will conflict with #1311 . Which one should I do first?

emilwestergren:
Doesn't matter.

marcusnaslund:
Fortunately no conflicts, nice.

PR #1314, opened 2016-02-02
Renamed file to GpuCanvasYuv420Semiplanar

davidpiuva:
GpuCanvas did not even exist so I renamed GpuSurface into GpuCanvas and moved out GpuCanvasYuv420Semiplanar to its own file.

There might be redundant includes or things copied wrong since I moved a lot of code and Git is not helping at all to see changes.

Git is too primitive to find the path of least change because GpuCanvas.ooc was renamed to GpuCanvasYuv420Semiplanar.ooc and GpuSurface.ooc was renamed to GpuCanvas.ooc as if I made a total change from GpuCanvas.ooc to GpuCanvas.ooc which are two completely different files.


davidpiuva:
@sebastianbaginski Peer review after https://github.com/magic-lang/ooc-kean/pull/1311?

source/draw/gpu/GpuCanvas.ooc
```diff
@@ -8,72 +8,96 @@
```
davidpiuva: This is a false detection between two different files.

sebastianbaginski:
Can you do a PR which simply renames the files and then another one with changes on top of it ?

source/draw/gpu/GpuCanvas.ooc
sebastianbaginski: Yeah, quite hard to review the actual changes now

davidpiuva:
I could change one name at a time but I would still have it detected as new and deleted if anything inside has to change.

davidpiuva:
Git can't even merge because of nonsense conflicts without anything marked as red.

sebastianbaginski:
I don't really know what to do with this one, even if I review it and it's OK, looks like you will still have a hard time rebasing this.
I think you need to do that in a very small steps.

davidpiuva:
Renaming a module apparently causes a segfault with our compiler.

davidpiuva:
Updated.

sebastianbaginski:
OK, great

davidpiuva:
Even after dividing into multiple pull requests, Git won't understand what I did until the previous PR is merged.

davidpiuva:
Replaced by new pull requests.

PR #1315, opened 2016-02-02
Create override fix

emilwestergren:
Still looks like we have missing `override` in the code base. This is not good.

@marcusnaslund peer


marcusnaslund:
@fredrikbryntesson ?

sebastianbaginski:
Wasn't rock supposed to detect it ?

fredrikbryntesson:
Yes it is supposed to do that. I look into it now why these methods behave so different.

marcusnaslund:
Well, this should be here, so I'm merging.

PR #1322, opened 2016-02-03
fixed possible out of bounds access in stbi

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/draw/stb_image/stb_image.h
+++ b/source/draw/stb_image/stb_image.h
@@ -4427,7 +4427,7 @@ static int stbi__parse_png_file(stbi__png *z, int scan, int req_comp)
                p = (stbi_uc *) STBI_REALLOC_SIZED(z->idata, idata_limit_old, idata_limit); if (p == NULL) return stbi__err("outofmem", "Out of memory");
                z->idata = p;
             }
-            if (!stbi__getn(s, z->idata+ioff,c.length)) return stbi__err("outofdata","Corrupt PNG");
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Might be a nice gesture to make a similar PR to upstream.

sebastianbaginski:
I'm not really sure if this is even possible, more like super defensive programming.
But on the other hand, it is not so easy to follow the control flow in this code, this fix might be legit.

fredrikbryntesson:
The upstream version are not working at all for us now?

sebastianbaginski:
It is working. This is to make Coverity happy

fredrikbryntesson:
Ok but it is possible to make upstream Coverity friendly?

fredrikbryntesson:
Or is this a dead end?

sebastianbaginski:
Making Coverity happy is a continuous process :)

fredrikbryntesson:
So try to push it upstream, it would be better in the end.

sebastianbaginski:
OK, first lets merge it here and see if it really fixes the problem.

marcusnaslund:
There is no problem?

sebastianbaginski:
Analysis tool complains that `z->idata + ioff` can exceed the valid memory range.

marcusnaslund:
Right, but it doesn't actually happen to us as far as we know?

sebastianbaginski:
I don't think so.

marcusnaslund:
This only failed because `SynchronizedTest` segfaulted, unfortunately I didn't have time to merge the (hopefully) fix before this. So this goes in.

PR #1323, opened 2016-02-03
Revert "Merged OpelGLSurface into OpenGLCanvas."

marcusnaslund:
Reverts magic-lang/ooc-kean#1311 because of segfaults when closing the window. Will be replaced shortly.

### DIFF
--- a/opengl.use
+++ b/opengl.use
@@ -12,6 +12,7 @@ Imports: OpenGLBgra
 Imports: OpenGLMonochrome
 Imports: OpenGLUv
 Imports: OpenGLCanvas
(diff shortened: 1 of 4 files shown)


marcusnaslund:
@jonathanudd PRInt fails because this PR isn't from a user but from a branch on upstream. Don't know if we can (or should) do anything about that though.

jonathanudd:
I think it fails because print tries to clone a branch based on the user who made the pull request and in this case the branch doesn't exist on the users fork. I don't think there are any easy fix for this at the moment.

PR #1324, opened 2016-02-03
fixed warnings from IO

sebastianbaginski:
Fixes warnings related to `fdopen` when building shared library.
The ones from `FileUnix` seems more ... persistent.
@marcusnaslund

### DIFF
--- a/source/system/IO.ooc
+++ b/source/system/IO.ooc
@@ -6,7 +6,7 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-include stdio, fcntl, unistd
+include stdio | (_POSIX_SOURCE), fcntl, unistd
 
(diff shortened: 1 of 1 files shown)


source/sdk/IO.ooc
```diff
@@ -6,7 +6,7 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-include stdio, fcntl, unistd
+include stdio | (_POSIX_SOURCE), fcntl, unistd
```
marcusnaslund: Does this affect windows builds in any way?

sebastianbaginski: I think the windows build is currently broken even without it :( I have a plan to build and run on windows this week though

marcusnaslund: OK cool.

PR #1325, opened 2016-02-03
Renamed the file containing GpuCanvasYuv420Semiplanar to the name of the class.

davidpiuva:
A new attempt at renaming that is not based on the changes to the window class.


davidpiuva:
@sebastianbaginski Can you review this again?

davidpiuva:
I don't understand why there are still things referring to classes that does not exist but it still compiles and runs.

davidpiuva:
Rebased

PR #1329, opened 2016-02-03
reduced thread count in SynchronizedResource test

sebastianbaginski:
@marcusnaslund


marcusnaslund:
I don't know if this is better since it only happens sometimes, so we'll see. :)

PR #1330, opened 2016-02-03
Color cleanup

thomasfanell:
**Reference: #781**

@davidpiuva peer review on code changes and default values?
@marcusnaslund review style fix?


davidpiuva:
ok

source/draw/Color.ooc
```diff
@@ -14,16 +14,7 @@ ColorMonochrome: cover {
...
-	init: func@ ~uint32 (i: UInt) { this init(i as Byte) }
-	init: func@ ~uint64 (i: ULong) { this init(i as Byte) }
-	init: func@ ~int (i: Int) { this init(i as Byte) }
-	init: func@ ~float (f: Float) { this init(f*255.0f clamp(0.0f, 255.0f) as Byte) }
-	init: func@ ~double (d: Double) { this init(d*255.0f clamp(0.0f, 255.0f) as Byte) }
-	set: func@ (color: This) {
-		this y = color y
-	}
+	set: func@ (color: This) { this y = color y }
 	copy: func -> This { this }
```
davidpiuva: `copy` can also be removed since it is a value type.

source/draw/Color.ooc
```diff
@@ -113,25 +71,16 @@ ColorYuv: cover {
 	toYuva: func -> ColorYuva { ColorYuva new(this, 255) }
 	toBgr: func -> ColorBgr { ColorConvert yuvToBgr(this) }
 	toBgra: func -> ColorBgra { this toBgr() toBgra() }
+	equals: func ~yuv (other: This) -> Bool { this y == other y && this u == other u && this v == other v }
 	blend: func (factor: Float, other: This) -> This {
 		This new((this y * (1 - factor) + other y * factor) as Byte, (this u * (1 - factor) + other u * factor) as Byte, (this v * (1 - factor) + other v * factor) as Byte)
 	}
 	distance: func (other: This) -> Float {
 		((this y - other y) as Float pow(2) + (this u - other u) as Float pow(2) + (this v - other v) as Float pow(2)) / 3.0f sqrt()
```
davidpiuva: This one is still incorrect but I don't know if anything is depending on the behaviour.

source/draw/Color.ooc
```diff
@@ -139,6 +88,7 @@ ColorYuva: cover {
 	alpha: Byte
 	normalized ::= FloatTuple4 new(this yuv y as Float / 255, this yuv u as Float / 255, this yuv v as Float / 255, this alpha as Float / 255)
 	init: func@ (=yuv, =alpha)
+	init: func@ ~default { this init(0, 0, 0, 0) }
```
davidpiuva: Use 0, 127, 127, 255 for black with full opacity.

marcusnaslund:
> @marcusnaslund review style fix?

I always review style, that's my job. :)

source/draw/Color.ooc
```diff
@@ -178,16 +123,13 @@ ColorBgr: cover {
...
 		This new((this blue * (1 - factor) + other blue * factor) as Byte, (this green * (1 - factor) + other green * factor) as Byte, (this red * (1 - factor) + other red * factor) as Byte)
 	}
 	distance: func (other: This) -> Float {
 		((this blue - other blue) as Float pow(2) + (this green - other green) as Float pow(2) + (this red - other red) as Float pow(2)) / 3.0f sqrt()
 	}
-	equals: func ~monochrome (other: ColorMonochrome) -> Bool { false }
-	equals: func ~yuv (other: ColorYuv) -> Bool { false }
-	equals: func ~bgr (other: This) -> Bool { this blue == other blue && this green == other green && this red == other red }
-	equals: func ~bgra (other: ColorBgra) -> Bool { false }
 	svgRGBToString: func -> String {
```
davidpiuva: Scalar vector graphics has nothing to do with color and I tried to move it before in a large PR that was not merged.

emilwestergren:
Now this is a nice cleanup

source/draw/Color.ooc
thomasfanell: At least `RasterYuvPlanar` does

source/draw/Color.ooc
```diff
@@ -227,16 +160,9 @@ ColorBgra: cover {
...
-	}
-	distance: func (other: This) -> Float {
-		(this bgr distance(other bgr) * 3.0f + (this alpha - other alpha) as Float pow(2)) / 4.0f sqrt()
-	}
-	equals: func ~monochrome (other: ColorMonochrome) -> Bool { false }
-	equals: func ~yuv (other: ColorYuv) -> Bool { false }
-	equals: func ~bgr (other: ColorBgr) -> Bool { false }
 	equals: func ~bgra (other: This) -> Bool { this bgr equals(other bgr) && this alpha == other alpha }
+	blend: func (factor: Float, other: This) -> This { This new(this bgr blend(factor, other bgr), (this alpha * (1 - factor) + other alpha * factor) as Byte) }
+	distance: func (other: This) -> Float { (this bgr distance(other bgr) * 3.0f + (this alpha - other alpha) as Float pow(2)) / 4.0f sqrt() }
```
davidpiuva: The bgr distance divided by 3 before taking the square root so it would be `|bgr|^2 * 3` to get the square distance but it is better to just redo both into the real distance in 4 dimensional space unless something depends on the weirdness. `sqrt(b^2 + g^2 + r^2 + a^2)`

marcusnaslund:
I suggest we limit ourselves to mostly what is in this PR for now and maybe @davidpiuva you could make a subsequent PR later which fixes the `distance` methods and other things.

davidpiuva:
ok

source/draw/Color.ooc
thomasfanell: The SVG plotter utilizes these

davidpiuva:
I am done with the first review.
YUVA's default constructor should be consistent with the new YUV default constructor. One way would be to call YUV's default constructor from YUVAs.

source/draw/Color.ooc
thomasfanell: There are some other weird things, such as, for instance:
`ColorYuv toYuv()`, I suppose I could get rid of those too (there are one for each color type), since they are effectively unnecessary?

davidpiuva: Yes, they would only fill a purpose in polymorphism from arbitrary classes.

source/draw/Color.ooc
davidpiuva: I tried running with gpuOff after commenting out the distance functions in BGR and BGRA. All this seems to be dead code that no one uses anymore.

thomasfanell:
@davidpiuva round two?

davidpiuva:
ok

source/draw/Color.ooc
```diff
@@ -14,124 +14,67 @@ ColorMonochrome: cover {
...
-	init: func@ ~uint64 (i: ULong) { this init(i as Byte) }
-	init: func@ ~int (i: Int) { this init(i as Byte) }
-	init: func@ ~float (f: Float) { this init(f*255.0f clamp(0.0f, 255.0f) as Byte) }
-	init: func@ ~double (d: Double) { this init(d*255.0f clamp(0.0f, 255.0f) as Byte) }
-	set: func@ (color: This) {
-		this y = color y
-	}
-	copy: func -> This { this }
-	toMonochrome: func -> This { this copy() }
+	set: func@ (color: This) { this y = color y }
```
davidpiuva: What is the intended use case?
Seems like I might as well use `color y = 255` or `color = ColorMonochrome new(255)`.

thomasfanell: It should really be a property, if anything...

thomasfanell: But yeah, seems a bit odd

thomasfanell:
@marcusnaslund you're up

davidpiuva:
Once merged, we can check off number 1, 3 and 5 in https://github.com/magic-lang/ooc-kean/issues/781. I can take point 4 or we just get rid of the distance function that is not well defined. Point 6 might not be a good idea anymore because of how setters works with value types.

marcusnaslund:
I'm happy too, especially if @davidpiuva fixes the remaining points in another PR soon. :)

davidpiuva:
It seems like raster drawing tests are using the distance functions so that I can't just delete them but it is a lot of bloat in the main code just for testing.

Blend and LinearInterpolation should probably be unified into `mix` as used for both vectors and colors in GLSL.

marcusnaslund:
The draw tests are a very minor part of the `PRInt` runtime.

davidpiuva:
ok

PR #1331, opened 2016-02-03
merged two constructors in AndroidContext

sebastianbaginski:
This will allow to simplify something.
@marcusnaslund review


marcusnaslund:
> simplify something.

I like when descriptions are specific like that.

PR #1336, opened 2016-02-04
Deleted svg specific functions from Color.

davidpiuva:
Inlined the 3 functions to the only place that called them.


davidpiuva:

sebastianbaginski:
Looks ok if we want to get rid of them

PR #1337, opened 2016-02-04
Flattening out all colors for access to elements by reference. 

davidpiuva:
* Flattening out all colors for access to elements by reference. 
* Removed getters and setters.

### DIFF
--- a/source/draw/gpu/opengl/OpenGLMap.ooc
+++ b/source/draw/gpu/opengl/OpenGLMap.ooc
@@ -72,7 +72,7 @@ OpenGLMap: class extends GpuMap {
 					case ColorBgra => color := cell as Cell<ColorBgra> get(); program setUniform(key, color red as Float / 255, color green as Float / 255, color blue as Float / 255, color alpha as Float / 255)
 					case ColorUv => color := cell as Cell<ColorUv> get(); program setUniform(key, color u as Float / 255, color v as Float / 255)
 					case ColorYuv => color := cell as Cell<ColorYuv> get(); program setUniform(key, color y as Float / 255, color u as Float / 255, color v as Float / 255)
-					case ColorYuva => color := cell as Cell<ColorYuva> get(); program setUniform(key, color yuv y as Float / 255, color yuv u as Float / 255, color yuv v as Float / 255, color alpha as Float / 255)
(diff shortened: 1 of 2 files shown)


davidpiuva:
@sebastianbaginski Yet another peer review?

sebastianbaginski:
Maybe @thomasfanell will do that quicker ?
@thomasfanell let me know if you don't have time, I will review then

source/draw/gpu/opengl/OpenGLMap.ooc
```diff
@@ -72,7 +72,7 @@ OpenGLMap: class extends GpuMap {
 					case ColorBgra => color := cell as Cell<ColorBgra> get(); program setUniform(key, color red as Float / 255, color green as Float / 255, color blue as Float / 255, color alpha as Float / 255)
 					case ColorUv => color := cell as Cell<ColorUv> get(); program setUniform(key, color u as Float / 255, color v as Float / 255)
 					case ColorYuv => color := cell as Cell<ColorYuv> get(); program setUniform(key, color y as Float / 255, color u as Float / 255, color v as Float / 255)
-					case ColorYuva => color := cell as Cell<ColorYuva> get(); program setUniform(key, color yuv y as Float / 255, color yuv u as Float / 255, color yuv v as Float / 255, color alpha as Float / 255)
+					case ColorYuva => color := cell as Cell<ColorYuva> get(); program setUniform(key, color y as Float / 255, color u as Float / 255, color v as Float / 255, color alpha as Float / 255)
```
thomasfanell: Are we "allowed" to access fields directly?

davidpiuva: We do it in ColorYuv so we might as well do the same for ColorYuva so that one can change between the types without changing how to access elements. Getters for value types are pointless because you can't pass the result by reference to a function or assignment.

marcusnaslund: Simon told me before public fields in covers are ok, not classes.

davidpiuva: Unless the mangled methods are exposed in the C library, I don't really see a point in using properties for classes either when there is no logic connected since the language allows accessing attributes and properties with the same syntax.

davidpiuva:
@marcusnaslund PRInt tests are completed.

PR #1338, opened 2016-02-04
Added a compact safe-raise operator

### LINKED ISSUE #1309: Compact assertion using safe flag
Problem **1**: Having an extra line for the condition for calling raise takes up space from the actual code.
Problem **2**: Having yet another extra line for only running the test in the safe version takes up more space.
```ooc
version(safe)
	if (condition)
		raise(message)
```

Solution **1**: Placing the condition as an argument to a conditional raise method would get rid of the if statement.
```ooc
version(safe)
	assert(condition, message)
```
Solution **2A**: Implementing a `safe` keyword in the compiler for wrapping the rest of the line in a `-Dsafe` version flag would make it more compact but make the language less reusable.
```ooc
safe assert(condition, message)
```
Solution **2B**:  Just putting the version flag in the called method would allow the compiler to optimize away most of the conditions. Without the ability to enforce functional purity on input argument expressions, this will sometimes leave side effects from accessing things that are not needed in the release version.
```ooc
safeAssert(condition, message)
```
Solution **2C**: Just leave the version block around the assertion after solving problem **1**.
Solution **2D**: Implement a pre-compiler for the language and make macros. This will create even more problems than slerping did with knowing which module to recompile if macros can be used across modules. Macros can dynamically generate the names of other macros using concatenation of partial macro names and then use them in the next iteration until all macro names are gone.

Discussed in https://github.com/magic-lang/ooc-kean/pull/1295

marcusnaslund:
Fixes #1309

### DIFF
--- a/source/collections/Queue.ooc
+++ b/source/collections/Queue.ooc
@@ -79,8 +79,7 @@ VectorQueue: class <T> extends Queue<T> {
 
 	operator [] (index: Int) -> T {
 		version(safe)
-			if (index < -this count || index >= this count)
-				raise("Indexing in get accessor of VectorQueue outside of range.")
+			raise(index < -this count || index >= this count, "Indexing in get accessor of VectorQueue outside of range.")
 		position := (index >= 0 ? this _head + index : this _tail + index) modulo(this _capacity)
 		this _backend[position]
 	}
--- a/source/collections/Vector2D.ooc
+++ b/source/collections/Vector2D.ooc
@@ -77,17 +77,13 @@ Vector2D: class <T> {
 	}
 
 	operator [] (row, column: Int) -> T {
-		version (safe) {
-			if (row >= this rowCount || row < 0 || column >= this columnCount || column < 0)
-				raise("Accessing Vector2D index out of range in get operator")
-		}
+		version (safe)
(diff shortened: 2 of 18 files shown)


source/collections/Queue.ooc
```diff
@@ -79,8 +79,7 @@ VectorQueue: class <T> extends Queue<T> {
 
 	operator [] (index: Int) -> T {
 		version(safe)
-			if (index < -this count || index >= this count)
-				raise("Indexing in get accessor of VectorQueue outside of range.")
+			raise(index < -this count || index >= this count, "Indexing in get accessor of VectorQueue outside of range.")
```
marcusnaslund: @thomasfanell What do you think, should we put these on one line as we do with `debug` version statements in the other project? These lines might get too long though.

thomasfanell: I didn't even know you could make those conditional... nice :+1:
Anyway, I think we can put them on one line just as long they don't exceed 200 or so characters.

marcusnaslund: You couldn't until this PR. :)

marcusnaslund:
Decided against one line as this is more consistent with long and short versions.
Peer review @thomasfanell ?

thomasfanell:
very nice indeed

marcusnaslund:
Passing the compliment along to @sebastianbaginski :+1:

PR #1344, opened 2016-02-04
fixes to build on win32

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/system/Character.ooc
+++ b/source/system/Character.ooc
@@ -6,7 +6,7 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-include stdlib
+include stdlib, stdint
 
 __LINE__: extern Int
 __FILE__: extern CString
--- a/source/system/Exception.ooc
+++ b/source/system/Exception.ooc
@@ -10,7 +10,7 @@ import threading/ThreadLocal
 import structs/Stack
 import Backtrace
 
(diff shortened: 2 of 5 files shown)


marcusnaslund:
The android build fails.

sebastianbaginski:
Android build works now.

source/system/Character.ooc
```diff
@@ -6,7 +6,7 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-include stdlib
+include stdlib, stdint
```
marcusnaslund: This is to make our fixed-width integers work now right? Shouldn't `stdint` be included in the use file under a `windows` version flag? Otherwise, will we need to add this everywhere?

sebastianbaginski: This is to have access to `uint64_t` on windows, i think this header is included by something else on unix, and on windows you need to do that manually.

marcusnaslund: Yes. That wasn't my question. :)

marcusnaslund: Ping. Will we need to add this file everywhere we use 64-bit integers in order to run on windows?

sebastianbaginski: No, here is enough.

marcusnaslund: Alright then.

PR #1345, opened 2016-02-04
A more complete test of System

### LINKED ISSUE #1221: Improve tests of remaining system (SDK) parts
- [x] `sdk/threading`
There are no tests for `RecursiveMutex` or timed join of `Thread`

- [x] `Env`
No tests of `get`, `set` or `unset`

- [x] `Buffer`
No tests for `slice`, `reset`, or the operators

- [x] `Char` (`Character.ooc`)
No tests at all.

marcusnaslund:
Fixes #1221 

### DIFF
--- a/source/system/Buffer.ooc
+++ b/source/system/Buffer.ooc
@@ -49,7 +49,7 @@ Buffer: cover {
 			start = this _size + start
 		if (distance < 0)
 			start -= size
-		start < this size ? This new(this _pointer + start, size minimum(this _size - start)) : This empty
+		start < this size ? This new(this _pointer as Byte* + start, size minimum(this _size - start)) : This empty
 	}
 	copy: func -> This { // call by value -> modifies copy of cover
 		result: This
--- a/source/system/Character.ooc
+++ b/source/system/Character.ooc
@@ -45,13 +45,6 @@ Char: cover from char {
 	whitespace: func -> Bool { this == ' ' || this == '\f' || this == '\n' ||this == '\r' || this == '\t' || this == '\v' }
 	blank: func -> Bool { this == ' ' || this == '\t' }
 	hexDigit: func -> Bool { this digit() || (this >= 'A' && this <= 'F') || (this >= 'a' && this <= 'f') }
-	toInt: func -> Int {
-		result := -1
-		if (this digit())
-			result = (this - '0') as Int
-		result
-	}
-
 	toString: func -> String { String new(this& as CString, 1) }
 	print: func { fputc(this, stdout) }
 	print: func ~withStream (stream: FStream) { fputc(this, stream) }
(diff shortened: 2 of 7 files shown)


source/system/Buffer.ooc
```diff
@@ -49,7 +49,7 @@ Buffer: cover {
 			start = this _size + start
 		if (distance < 0)
 			start -= size
-		start < this size ? This new(this _pointer + start, size minimum(this _size - start)) : This empty
+		start < this size ? This new(this _pointer as Byte* + start, size minimum(this _size - start)) : This empty
```
marcusnaslund: These are the kinds of bugs you discover when writing tests. Pointer arithmetic, appearently, do not work with `Pointer`s...?

test/system/CharacterTest.ooc
```diff
@@ -0,0 +1,27 @@
...
+		super("Env")
+		this add("basic test", func {
+			(a, b, c, d, e) := ('7', 't' 'F', ' ', '9')
+			expect(a digit(), is true)
+			expect(b digit(), is false)
+			expect(b lower(), is true)
+			expect(c lower(), is false)
+			expect(d whitespace(), is true)
+			expect(e alpha(), is false)
+			expect(e alphaNumeric(), is true)
```
marcusnaslund: Yes, there's more stuff in `Char`, but I feel this will do for now.

source/system/Character.ooc
```diff
@@ -78,6 +71,12 @@ Char: cover from char {
 	compareWith: func (compareFunc: Func (Char, Char*, SizeT) -> SSizeT, target: Char*, targetSize: SizeT) -> SSizeT {
 		compareFunc(this, target, targetSize)
 	}
+	toInt: func -> Int {
+		result := -1
+		if (this digit())
+			result = (this - '0') as Int
+		result
+	}
 }
```
fredrikbryntesson: Why was this moved?

fredrikbryntesson:
CharacterTest should probably be extended later and we should look into why not pointer arithmetic fails, else it looks good.

marcusnaslund:
Ah, nothing surprising, `Pointer` is a cover of `Void*`, and arithmetic on those are illegal in `C` too.

source/system/Character.ooc
marcusnaslund: Looks better. :)

source/system/Buffer.ooc
sebastianbaginski: That's right, because the expression `*(p + x)`, where `p` is a pointer `T*` and `x` integer, generates memory access at address `p + x * sizeof(T)`, and for `T = void` `sizeof(void)` is not defined.

sebastianbaginski:
Problem with this is that `sizeof(void)` is not (and cannot be) well defined. Some compilers, like gcc, will allow you to ask for a size of void, but this is not a standard. 
On the other hand, `void*` is fine as long as you are just passing generic data around. When you start accessing it you need to know what type of data is it, thus it is required to cast it to some other type, even when adding an offset.

PR #1346, opened 2016-02-04
Cleaned up some more in HashMap

marcusnaslund:
Gets rid of nearly 20% of the code in `HashMap.ooc` and will make it a lot easier to work on later.

Peer review @jonathanudd ?

### DIFF
--- a/source/system/structs/HashMap.ooc
+++ b/source/system/structs/HashMap.ooc
@@ -6,41 +6,17 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-getStandardEquals: func <T> (k1, k2: T) -> Bool {
-	if (T == String)
-		stringEquals(k1, k2)
-	else if (T == CString)
-		cstringEquals(k1, k2)
-	else if (T size == Pointer size)
-		pointerEquals(k1, k2)
-	else if (T size == UInt size)
-		intEquals(k1, k2)
-	else if (T size == Char size)
-		charEquals(k1, k2)
-	else
-		genericEquals(k1, k2)
-}
-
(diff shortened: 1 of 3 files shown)


source/system/structs/HashMap.ooc
```diff
@@ -6,41 +6,17 @@
...
-intEquals: func <K> (k1, k2: K) -> Bool { k1 as Int == k2 as Int }
-
-charEquals: func <K> (k1, k2: K) -> Bool { k1 as Char == k2 as Char }
-
-genericEquals: func <K> (k1, k2: K) -> Bool { memcmp(k1, k2, K size) == 0 }
-
 HashEntry: cover {
-	key, value: Pointer
+	key: Pointer
+	value: Pointer
```
jonathanudd: Why do we want to put this on two rows instead of one when most of our code guidelines prefers longer rows over more rows?

marcusnaslund: I wonder about that myself. We don't do it inside methods, but all our classes seem to be declared this way, so we're sticking to that.

source/system/structs/HashMap.ooc
```diff
@@ -149,21 +109,17 @@ getStandardHashFunc: func <T> (T: Class) -> Func <T> (T) -> SizeT {
 HashMap: class <K, V> extends BackIterable<V> {
 	_size, capacity: SizeT
```
jonathanudd: Based on the comment earlier this should be placed on two lines.

jonathanudd:
Peer review done.

source/system/structs/HashMap.ooc
marcusnaslund: Indeed.

PR #1349, opened 2016-02-05
corrected some warnings in net code

sebastianbaginski:
Unchecked return code from library call.
Some null ptr dereferences.
@marcusnaslund

### DIFF
--- a/source/net/ServerSocket.ooc
+++ b/source/net/ServerSocket.ooc
@@ -61,7 +61,8 @@ ServerSocket: class extends Socket {
 				case AddressFamily IP6 =>
 					addr = getSocketAddress6(ip, port)
 			}
-			this bind(addr)
+			if (addr)
+				this bind(addr)
 		} else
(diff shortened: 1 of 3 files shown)


marcusnaslund:
Is this from the static analysis?

sebastianbaginski:
Yeah

source/net/ServerSocket.ooc
```diff
@@ -127,7 +130,7 @@ ServerSocket: class extends Socket {
 				shutdown(conn sock descriptor, SHUT_RDWR)
 			}
 			conn close()
-			if ((conn && ret) as Bool)
+			if (ret as Bool)
```
marcusnaslund: Right, after `conn close()`, this memory is forfeited?
You don't need the cast to `Bool` anymore since `f` returns `Bool`.

sebastianbaginski: It was complaining that `conn` is dereferenced even after it is checked for `null`.

sebastianbaginski:
Updated.

PR #1355, opened 2016-02-05
Taming the BlockedQueue beast once and for all

### LINKED ISSUE #646: Rewrite SynchronizedQueue and BlockedQueue
There has for a long time been something wrong with the implementation of `BlockedQueue` (and possibly `SynchronizedQueue`) which seems to manifest itself more often in Windows than Linux, for some reason. This is the reason why the ThreadPool test hangs sometimes on CircleCI (and more often on AppVeyor).

marcusnaslund:
Basically, the infamous deadlock would occur when `cancel` would be called *before* all threads were waiting on the empty queue. Those who were already waiting would receive the broadcast and finish. The other thread(s) would *later* call `wait()`, hold on the `WaitCondition` and never receive the broadcast because it had already been sent. The simple fix is to check the status of `_canceled` and if it's true *not* wait.

This PR...

- Fixes the deadlock that could occur, thereby closes #646 and is cause for a major celebration.
- Reopens the `ThreadPool` test
- Reduces the number of threads created in `BlockedQueueTest`, also does not add `0` to the test queue to make sure it is never returned.

### DIFF
--- a/source/concurrent/SynchronizedQueue.ooc
+++ b/source/concurrent/SynchronizedQueue.ooc
@@ -68,15 +68,12 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 	wait: func (isOk := null as Bool*) -> T {
 		result: T = null
 		this _mutex lock()
-		while (this empty) {
+		while (this empty && !this _canceled)
 			this _populated wait(this _mutex)
-			if (this _canceled) {
-				if (isOk)
-					isOk@ = false
-				break
-			}
-		}
-		if (!this _canceled)
+		if (this _canceled) {
+			if (isOk)
+				isOk@ = false
+		} else
 			result = this _backend dequeue(result)
 		this _mutex unlock()
 		result
--- a/test/concurrent/BlockedQueueTest.ooc
+++ b/test/concurrent/BlockedQueueTest.ooc
@@ -14,16 +14,16 @@ import threading/Thread
 BlockedQueueTest: class extends Fixture {
 	init: func {
 		super("BlockedQueue")
-		version (!windows) { this add("cover", This _testWithCover) }
+		this add("cover", This _testWithCover)
 		this add("class", This _testWithClass)
 	}
 	_testWithCover: static func {
 		queue := BlockedQueue<Int> new()
-		numberOfThreads := 8
+		numberOfThreads := 4
 		countPerThread := 20_000
 		totalCount := numberOfThreads * countPerThread
 		produce := func {
-			for (i in 0 .. totalCount) {
+			for (i in 1 .. totalCount + 1) {
 				queue enqueue(i)
 				Thread yield()
 			}
@@ -32,9 +32,10 @@ BlockedQueueTest: class extends Fixture {
 			for (i in 0 .. countPerThread) {
 				value := queue wait()
 				expect(value < totalCount)
+				expect(value > 0)
 			}
 		}
-		queue enqueue(0)
+		queue enqueue(1)
 		threads := Thread[numberOfThreads] new()
 		for (i in 0 .. numberOfThreads) {
 			threads[i] = Thread new(consume)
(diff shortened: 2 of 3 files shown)


thomasfanell:
Ah, gotta love the printf-debugging... true and pure ;-)

sebastianbaginski:
^_^ https://s-media-cache-ak0.pinimg.com/564x/4d/81/98/4d81987de77e4220af9068912534199c.jpg

marcusnaslund:
Thanks for the peptalk. :) 

I managed to introduce a new race condition but that should be easy to fix on Monday. There are also more mutex locks/unlocks now than before, but those can be simplified away. All-in-all, this looks promising.

emilwestergren:
`wait` locks the mutex before checking the condition, if the condition is false then it unlocks the mutex and goes to sleep until the condition is signaled.

source/concurrent/SynchronizedQueue.ooc
```diff
@@ -12,8 +12,18 @@ import threading/[Thread, Mutex, WaitCondition]
 SynchronizedQueue: class <T> extends Queue<T> {
 	_mutex := Mutex new()
 	_backend: Queue<T>
-	count ::= this _backend count
-	empty ::= this _backend empty
+	count: Int { get {
+		this _mutex lock()
```
emilwestergren: Why locks here? As soon as you unlock the mutex you can't be sure anyway if `count` is still valid.

source/concurrent/SynchronizedQueue.ooc
```diff
@@ -67,16 +85,25 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 	}
 	wait: func (isOk := null as Bool*) -> T {
 		result: T = null
-		this _mutex lock()
-		while (this empty) {
+		(e, c) := (this empty, this canceled)
+		while (e && !c) {
```
emilwestergren: When you think it's empty here, it might actually not be empty since you unlocked the mutex and someone might have added something to the queue immediately after. Then you will sleep and wait to be signaled even though the queue is populated.

emilwestergren:
Can I have a look before you merge this? I'm not really sure about these changes.

marcusnaslund:
@emilwestergren It's always nice to receive feedback, but perhaps wait until at least after the debug code is gone. :) Anyhow, I've now managed to find a more elegant fix and cleaned everything up. I'm no longer able to reproduce the error on my own laptop, but I will do some more stress testing when I'm back at work on Monday before this goes to peer review. You are of course a given candidate for that position.

emilwestergren:
Sure, just wanted to make sure you were on the right track.

marcusnaslund:
Here we go. I've run the `concurrent` tests at least 50 times (with and without `stress`/`PRInt` etc. running in the background) and 0 problems. If I uncomment the `ThreadPoolTest` on `upstream/master` it hangs at least once every five times.

Also seems to run fine on Android. I don't know what else to test so I'm letting this move on to peer review @emilwestergren . If you have time @sebastianbaginski please take a look as well.

test/concurrent/BlockedQueueTest.ooc
```diff
@@ -14,16 +14,16 @@ import threading/Thread
...
 	init: func {
 		super("BlockedQueue")
-		version (!windows) { this add("cover", This _testWithCover) }
+		this add("cover", This _testWithCover)
 		this add("class", This _testWithClass)
 	}
 	_testWithCover: static func {
 		queue := BlockedQueue<Int> new()
-		numberOfThreads := 8
+		numberOfThreads := 4
```
sebastianbaginski: was it failing as the sync resource test because of too many threads ?

marcusnaslund: This never failed in the same way AFAIK, I just changed it to 4 threads to be "consistent" and avoid problems similar to those in the sync resource test. I still don't really know why that test failed but not this one. OK to stick to four for now and then evaluate if we can change all the threading tests later (if we need to)?

sebastianbaginski: Yeah I guess most important is that we use at least two different threads.

marcusnaslund:
How say ye, @emilwestergren ?

emilwestergren:
Looks good, I trust you tested it enough

marcusnaslund:
I will do another round of testing on the phone today. If you're happy, feel free to set the final flag.

marcusnaslund:
Boom.

PR #1356, opened 2016-02-05
merged ThreadLocal classes for unix and windows

sebastianbaginski:
The only difference was the function used to get current thread id.
@marcusnaslund


sebastianbaginski:
I'm not so sure we should even use this class. Thread ID is not guaranted to be unique through the application lifetime, at least in pthreads. When thread is done, its ID can be reused when creating next thread, which makes this class unreliable in my opinion.

marcusnaslund:
If you want to throw it out completely, fine by me.

marcusnaslund:
It's the one thing that stops us moving `system/threading` out to `concurrent`.

sebastianbaginski:
Maybe let's start with this and we can talk about it next week, I gotta go soon.

PR #1359, opened 2016-02-08
fixed datatype in thread local

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/system/threading/ThreadLocal.ooc
+++ b/source/system/threading/ThreadLocal.ooc
@@ -11,7 +11,7 @@ import Thread
 import Mutex
 
 ThreadLocal: class <T> {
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
https://github.com/magic-lang/ooc-kean/issues/1360

marcusnaslund:
This fixes it? Makes no sense.

sebastianbaginski:
Yeah, try it :) it is wrong to use either of these types imho

marcusnaslund:
Right, it works. But, `Long` was the key before as well.. so still makes no sense to me.

sebastianbaginski:
I think it was ULong on linux

PR #1362, opened 2016-02-08
Added Text read() for console input

### LINKED ISSUE #1357: Implement support for reading a Text/String from the console
`ooc` ought to have something better than manually allocating a `Char` array and reading a fixed width using `fgets` and then converting it to a `Text`, all manually.

marcusnaslund:
Fixes #1357

### DIFF
--- a/source/system/Text.ooc
+++ b/source/system/Text.ooc
@@ -172,6 +172,11 @@ Text: cover {
 		this print()
 		t"\n" print()
 	}
+	read: static func (length := 64) -> This {
+		input := CString new(length)
+		fgets(input, length, stdin)
+		This new(input, length + 1)
+	}
 	format: func (args: ...) -> This {
 		string := this take() toString()
 		resultString := string format(args)


marcusnaslund:
Peer review @sebastianbaginski ? Thoughts on default value? I've tried using it and get no leaks.
I guess there's no way of making an automated test.

source/system/Text.ooc
```diff
@@ -172,6 +172,11 @@ Text: cover {
 		this print()
 		t"\n" print()
 	}
+	read: static func (length := 64) -> This {
+		input := CString new(length)
+		fgets(input, length, stdin)
+		This new(input, length + 1)
```
marcusnaslund: `length + 1` because `CString` allocates `length + 1` bytes to make room for `\0`.

sebastianbaginski: you could also pass `stdin` as parameter of type `FStream` and get a general file read function :)

sebastianbaginski:
Try to type in something that has more than 64 characters.

sebastianbaginski:
Also you could make an automated test by redirecting the output of some process like `echo "text" > ./a.out` and writing the result of this function to a file. I doubt it is worth the effort though.

marcusnaslund:
```ooc
stuff := Text read(10)
stuff println()
```
with input `12345678901234567890` gives me `123456789`.

sebastianbaginski:
OK, great.

marcusnaslund:
Let's hold on the automated test, and general file read function until we (if ever) rework all the IO stuff.

PR #1365, opened 2016-02-08
Added ThreadId cover

### LINKED ISSUE #1360: Add ThreadId datatype
should be used instead of `Long` and `ULong` to keep thread id
on unix it should be a cover from `pthread_t`, on windows cover from `DWORD`

marcusnaslund:
Fixes #1360 

Not at all ready, but is this what you had in mind @sebastianbaginski ?

I think we should change the static `equals` away in `Thread` to a `ThreadId` method.

### DIFF
--- a/test/system/ThreadTest.ooc
+++ b/test/system/ThreadTest.ooc
@@ -68,7 +68,7 @@ ThreadTest: class extends Fixture {
 	}
 	_testThreadId: static func {
 		myId := Thread currentThreadId()
-		otherId := Cell<Long> new(0L)
+		otherId := Cell<ThreadId> new(0L)
 		job := func {
 			otherId set(Thread currentThreadId())
 		}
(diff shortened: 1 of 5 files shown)


sebastianbaginski:
Yes, something like this.

marcusnaslund:
Ready for peer review now. @sebastianbaginski

test/system/ThreadTest.ooc
```diff
@@ -76,10 +76,10 @@ ThreadTest: class extends Fixture {
 		expect(thread start())
 		expect(thread wait())
 		thread free()
-		expect(Thread equals(myId, otherId get()) == false)
+		expect(myId equals(otherId get()) == false)
 		otherId free()
 		(job as Closure) free()
-		expect(Thread equals(myId, Thread currentThreadId()))
+		expect(myId equals(Thread currentThreadId()) == true)
```
sebastianbaginski: does not work without `== true` ? :)

marcusnaslund: It does, I just wanted it to look like line 79. :)

source/system/threading/ThreadId.ooc
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+import native/[ThreadUnix, ThreadWin32]
+include pthread | (_POSIX_C_SOURCE=200809L)
+
+version (unix || apple) {
+	ThreadId: cover from PThread {
```
sebastianbaginski: You could also add this directly to Thead.ooc, as it is not very useful without `Thread`

marcusnaslund: Yeah, that's probably a good idea. I will fix.

marcusnaslund:
Updated

sebastianbaginski:
Great.

PR #1367, opened 2016-02-08
Removes the warning from Android build

marcusnaslund:
Removed the warnings that pop up when building for Android.


marcusnaslund:
Well, it *does* remove the warnings. It also destroys everything else.

PR #1369, opened 2016-02-09
Now using TimeSpan in Promise/Future

### LINKED ISSUE #1307: `wait` in Promise should accept TimeSpan instead of double
to improve readability and get rid of ugly constants

marcusnaslund:
Fixes #1307


marcusnaslund:

source/base/TimeSpan.ooc
```diff
@@ -16,6 +16,9 @@ TimeSpan: cover {
 		this _ticks = DateTime timeToTicks(hour, minute, second, millisecond)
 	}
 	negate: func -> This { This new(-1 * this ticks) }
+	elapsedNanoseconds: func -> Long {
+		1_000_000 * this elapsedMilliseconds()
```
sebastianbaginski: Why not `this ticks / DateTime nanosecondsPerTick` ?
Current one will return zero for ticks corresponding to less than 1 ms. (as `elapsedMilliseconds()` will round down to zero).

sebastianbaginski:
> 16:49:17 ThreadPool (SIGSEGV) segmentation fault

:)

As for the PR, no comments other than the above one.

source/base/TimeSpan.ooc
marcusnaslund: Oh, I didn't know we had that, thanks.

marcusnaslund:
Updated as you said, but with a multiplication sign though. And a test for nanoseconds.

sebastianbaginski:
Oh of course, it is division in DateTime :)

PR #1375, opened 2016-02-10
added CircularQueue class

sebastianbaginski:
This class could be useful in some places in the other project.
@marcusnaslund

### DIFF
--- a/source/collections/Queue.ooc
+++ b/source/collections/Queue.ooc
@@ -84,3 +84,44 @@ VectorQueue: class <T> extends Queue<T> {
 		this _backend[position]
 	}
 }
+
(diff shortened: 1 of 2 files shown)


marcusnaslund:
I will do this after this PR, but I plan to split up `Queue.ooc` so that the abstract class goes to `system/structs` and `VectorQueue` + `CircularQueue` end up in separate files in `collections`. Does that make sense?

marcusnaslund:
Looks good otherwise, just waiting on a restarted `PRInt` now.

sebastianbaginski:
> Does that make sense?

Yes, that sounds reasonable.

Issue #908, opened 2015-11-26
toRasterAsync should return Future and be renamed


emilwestergren:
We currently have two versions for downloading images from the GPU to a RasterImage:

```ooc
toRaster: func -> RasterImage
toRasterAsync: func -> (RasterImage, GpuFence)
```

Instead of `toRasterAsync` returning a `GpuFence` we could probably use the new `Future` class. Also the naming of these functions can be discussed.

@davidpiuva @sebastianbaginski @simonmika

sebastianbaginski:
It could be possible to have a `Raster...` subclasses that synchronizes with gpu implicitly as soon as someone tries to access the pixel data. You could have just `toRaster` then.

emilwestergren:
Or a specialized ByteBuffer that can be constructed with a fence/future/promise.

simonmika:
Let's do this now:

- [ ] Make `toRasterAsync` return a `Future<RasterImage>`.
- [ ] Remove `toRaster` and rename `toRasterAsync` into `toRaster`. 

I also like @emilwestergren's suggestion above but let us start with the above.

emilwestergren:
I have implemented this but it doesn't work yet due to unknown ooc error with inheritance and generics.

marcusnaslund:
Can you send a link to the branch to @thomasfanell so he can take a look at it?

emilwestergren:
https://github.com/emilwestergren/ooc-kean/tree/future_fix

In `ToRasterTest.ooc`:

```ooc
			future := gpuImage toRasterAsync()
			future wait()
			raster := future getResult(null)
			println("Raster: %p" format(raster))
			expect(raster != null)
```

`future getResult(null)` always returns `null`, although the value inside the function is not `null`.

thomasfanell:
@marcusnaslund @emilwestergren I'm looking at this. I haven't found a workaround yet though.

emilwestergren:
Did you replicate the problem?

thomasfanell:
@emilwestergren I pulled your branch

emilwestergren:
Yes, but do you understand the bug?

thomasfanell:
Is there something I need to understand other than that you've already outlined here?

emilwestergren:
No, just wanted to make sure you understand the problem

marcusnaslund:
@thomasfanell @emilwestergren 
This works, but it ain't pretty:

```ooc
future := gpuImage toRasterAsync()
future wait()
[...]

PR #1376, opened 2016-02-11
toRasterAsync now returns Future.

emilwestergren:
@marcusnaslund @sebastianbaginski what do you think about this?

### DIFF
--- a/source/draw/gpu/opengl/AndroidContext.ooc
+++ b/source/draw/gpu/opengl/AndroidContext.ooc
@@ -116,25 +116,24 @@ AndroidContext: class extends OpenGLContext {
 			super(source, target)
 		result
 	}
-	toRasterAsync: func ~monochrome (gpuImage: OpenGLMonochrome) -> (RasterImage, GpuFence) {
+	toRasterAsync: func ~monochrome (gpuImage: OpenGLMonochrome) -> Future<RasterImage> {
 		(buffer, fence) := this toBuffer(gpuImage, this _packMonochrome)
-		(RasterMonochrome new(buffer, gpuImage size), fence)
+		_FenceToRasterFuture new(RasterMonochrome new(buffer, gpuImage size), fence)
 	}
-	toRasterAsync: func ~uv (gpuImage: OpenGLUv) -> (RasterImage, GpuFence) {
+	toRasterAsync: func ~uv (gpuImage: OpenGLUv) -> Future<RasterImage> {
 		(buffer, fence) := this toBuffer(gpuImage, this _packUv)
-		(RasterUv new(buffer, gpuImage size), fence)
+		_FenceToRasterFuture new(RasterUv new(buffer, gpuImage size), fence)
 	}
-	toRasterAsync: override func (gpuImage: GpuImage) -> (RasterImage, GpuFence) {
-		rasterResult: RasterImage
-		fenceResult: GpuFence
+	toRasterAsync: override func (gpuImage: GpuImage) -> Future<RasterImage> {
+		result: Future<RasterImage>
 		aligned := this isAligned(gpuImage size x)
 		if (aligned && gpuImage instanceOf(OpenGLMonochrome))
-			(rasterResult, fenceResult) = this toRasterAsync(gpuImage as OpenGLMonochrome)
+			result = this toRasterAsync(gpuImage as OpenGLMonochrome)
 		else if (aligned && gpuImage instanceOf(OpenGLUv))
-			(rasterResult, fenceResult) = this toRasterAsync(gpuImage as OpenGLUv)
+			result = this toRasterAsync(gpuImage as OpenGLUv)
 		else
-			(rasterResult, fenceResult) = super(gpuImage)
-		(rasterResult, fenceResult)
+			result = super(gpuImage)
+		result
 	}
 	unpackBgraToYuv420Semiplanar: func (source: GpuImage, targetSize: IntVector2D, padding := 0) -> GpuYuv420Semiplanar {
 		target := this createYuv420Semiplanar(targetSize) as GpuYuv420Semiplanar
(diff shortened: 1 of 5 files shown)


source/draw/gpu/opengl/OpenGLContext.ooc
```diff
@@ -164,6 +201,7 @@ OpenGLContext: class extends GpuContext {
 		target canvas draw(source, map)
 	}
 	createFence: override func -> GpuFence { OpenGLFence new(this) }
+	toRasterAsync: override func (source: GpuImage) -> Future<RasterImage> { _DummyFuture new(this toRaster(source)) }
```
emilwestergren: The default implementation will actually be synchronized, so we wait on a dummy future.

test/draw/gpu/ToRasterTest.ooc
```diff
@@ -40,15 +40,31 @@ ToRasterTest: class extends Fixture {
 		gpuImage free()
 		raster free()
 	}
+	toRasterAsync: func (sourceImage: RasterImage) {
+		gpuImage := context createImage(sourceImage)
+		future := gpuImage toRasterAsync() as _ToRasterFuture
```
emilwestergren: This cast is necessary, otherwise we always get `null` as result.

marcusnaslund:
Looks good to me. `_DummyFuture` is too generic a name when specialised for a certain type, though.

emilwestergren:
Sure, maybe `_DummyToRasterFuture`

marcusnaslund:
That's good.

emilwestergren:
Do you prefer if `toRasterAsync` returns `ToRasterFuture` instead, and we change the API when the ooc bug is fixed?

marcusnaslund:
It's up to you if you want to be explicit about the type everywhere or not. I don't know how soon we will be able to fix this particular bug/design feature, so perhaps don't base the API on it.

emilwestergren:
Yeah I think it's better to define the API as we want it, and go around the bug until it has been fixed.

sebastianbaginski:
I like it because you got rid of `GpuFence` from the API

emilwestergren:
Yes, after this we will remove `GpuFence` and `OpenGLFence` :smiley:

emilwestergren:
@marcusnaslund peer review

marcusnaslund:
Looks good, but:
```
source/draw/gpu/GpuContext.ooc:59:1: 
error: control reaches end of non-void function [-Werror=return-type] }
```

emilwestergren:
Updated

test/draw/gpu/ToRasterTest.ooc
```diff
@@ -40,15 +40,31 @@ ToRasterTest: class extends Fixture {
...
+		future wait()
+		raster := future getResult(null)
+		expect(raster != null)
+		distance := raster distance(sourceImage)
```
marcusnaslund: You assign `distance` here but never use it on the next line. That's why PRInt fails.

emilwestergren:
I think I've changed my mind. What do you think of `toRaster: func -> ToRasterFuture` instead? This way you don't need to cast it yourself. I don't think the bug in rock will be solved anytime soon anyway.

test/draw/gpu/ToRasterTest.ooc
emilwestergren: Oh I see

emilwestergren:
Updated.

marcusnaslund:
Memory leaks.

source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -116,25 +116,24 @@ AndroidContext: class extends OpenGLContext {
...
+	toRasterAsync: func ~monochrome (gpuImage: OpenGLMonochrome) -> ToRasterFuture {
 		(buffer, fence) := this toBuffer(gpuImage, this _packMonochrome)
-		(RasterMonochrome new(buffer, gpuImage size), fence)
+		_FenceToRasterFuture new(RasterMonochrome new(buffer, gpuImage size), fence)
 	}
-	toRasterAsync: func ~uv (gpuImage: OpenGLUv) -> (RasterImage, GpuFence) {
+	toRasterAsync: func ~uv (gpuImage: OpenGLUv) -> ToRasterFuture {
 		(buffer, fence) := this toBuffer(gpuImage, this _packUv)
-		(RasterUv new(buffer, gpuImage size), fence)
+		_FenceToRasterFuture new(RasterUv new(buffer, gpuImage size), fence)
```
marcusnaslund: Note that the return type is `ToRasterFuture`. When you call `free` on whatever this method returns, and don't explicitly cast (I can't see any explicit cast), you will call `ToRasterFuture::free` which doesn't free the `GpuFence` instance.

emilwestergren: What? Why? Because of the rock bug?

sebastianbaginski: That's so weird.

marcusnaslund: It's the same thing as before, the type is inferred as the specified function return type, not as the object you actually return.

emilwestergren: But `free` is a virtual function, it shouldn't depend on the returned type...

emilwestergren: Even `getResult` actually calls the correct function (i.e. `getResult` in the implemented type), it's just the returned value that's corrupt.

emilwestergren: I think the memory leak is due to my adaption to this change in Vidproc.

marcusnaslund: Oh, alright then. When it comes to rock, I usually assume the worst.

emilwestergren: Sounds like a fair strategy ;)

marcusnaslund:
#908 specifies that `ToRasterAsync` should be renamed, will that be in the next PR?

emilwestergren:
Yes, that will be a separate PR. It's a bit more messy since a lot of tests need to be adapted.

PR #1377, opened 2016-02-11
Separated Queue.ooc into three files

marcusnaslund:
Moved `Queue` to `system/structs`
`VectorQueue` and `CircularQueue` are now two separate files in `collections`.
Also separated the test files to match.


marcusnaslund:
Peer review @thomasfanell ?

PR #1378, opened 2016-02-11
Fixed search function for List.

emilwestergren:
Changed the API to how we want it. It works for lists of covers if you use the cleaner closure syntax: `|instance| instance == 1` instead of `func (instance: Int) -> Bool { instance == 1 }`. This is why it works for similar functions in `VectorList` like `modify`.

@marcusnaslund peer review

### DIFF
--- a/source/concurrent/RecycleBin.ooc
+++ b/source/concurrent/RecycleBin.ooc
@@ -26,9 +26,8 @@ RecycleBin: class <T> {
 			this _free(this _list remove(0))
 		this _list add(object)
 	}
-	search: func (matches: Func (T*) -> Bool) -> T {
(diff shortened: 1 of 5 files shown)


marcusnaslund:
So... it works with `||` but not with `func`? They don't generate the same code? What? @thomasfanell

marcusnaslund:
You must adapt `RecycleBin` as well.

thomasfanell:
@marcusnaslund No, it doesn't generate the same C code.

```ooc
func (instance: Int*) -> Bool { instance@ == 1 }
```
should be equal to the short-form, assuming instance is an `Int`of course.

emilwestergren:
*Adapting RecycleBinTest...*

emilwestergren:
Updated.

PR #1382, opened 2016-02-12
Removed BinarySequence reader and writer

marcusnaslund:
We're not using it, and never have, and it's dependent on the "old" I/O files from the original sdk, which we want to rewrite eventually.

For now, it's safe to just remove this.

Objections @sebastianbaginski ?


sebastianbaginski:
It is great to have such functions ready and tested in the toolbox, I hope they will come back one day.

marcusnaslund:
I could also fix them up a bit and write some basic tests...?

sebastianbaginski:
We do not work so much with data which must not depend on the platform endianess, like portable file formats, network code etc. I'm not sure if we should keep this now.

marcusnaslund:
Then hello, garbage can.

PR #1383, opened 2016-02-12
Removed CharBufferIterator

marcusnaslund:
Well, that was easy. Peer review @sebastianbaginski ?


sebastianbaginski:
And here is another thing that I really like to have in the toolbox - iterators :)
Btw. what's the "global" status on the iterators - should we try to fix / rewrite them at some point or we'll be stuck with `for(i in .. 0 .. vector count)` for the rest of our days ? :)

marcusnaslund:
I agree that `for (elem in list) { ... }` is nice, but at present they leak memory and I don't know how easy it is to fix. I believe we *can* use them by first creating the iterator, using it in the loop, and then manually freeing it. But that's not so much fun anymore.

Either way, this gets rid of a specific iterator we don't want or use, so I'm glad.

sebastianbaginski:
I understand. It is just such a shame that there is no automatic destructor in ooc.

PR #1385, opened 2016-02-12
Removed last extenson of Iterable

marcusnaslund:
`HashMap` no longer inherits from `BackIterable` and `copy` doesn't leak a function.

Peer review @jonathanudd ?


jonathanudd:
Looks Good.

PR #1388, opened 2016-02-15
Created RGB colors.

davidpiuva:
First step of replacing BGR with RGB.

### DIFF
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -17,8 +17,10 @@ ColorMonochrome: cover {
 	toUv: func -> ColorUv { ColorUv new(128, 128) }
 	toYuv: func -> ColorYuv { ColorYuv new(this y, 128, 128) }
 	toYuva: func -> ColorYuva { ColorYuva new(this y, 128, 128, 255) }
-	toBgr: func -> ColorBgr { ColorConvert yuvToBgr(this toYuv()) }
-	toBgra: func -> ColorBgra { this toBgr() toBgra() }
+	toRgb: func -> ColorRgb { this toYuv() toRgb() }
+	toRgba: func -> ColorRgba { this toYuv() toRgba() }
+	toBgr: func -> ColorBgr { this toYuv() toBgr() }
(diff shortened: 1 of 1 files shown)


davidpiuva:

marcusnaslund:
Does not compile.

davidpiuva:
That is because upstream does not compile.

davidpiuva:
Is it safe to fetch again?

source/draw/Color.ooc
```diff
@@ -73,6 +77,8 @@ ColorYuva: cover {
 	toMonochrome: func -> ColorMonochrome { ColorMonochrome new(this y) }
 	toUv: func -> ColorUv { this toYuv() toUv() }
 	toYuv: func -> ColorYuv { ColorYuv new(this y, this u, this v) }
+	toRgb: func -> ColorBgr { this toYuv() toRgb() }
```
marcusnaslund: Upstream runs fine. The problem is here. Return type is `Bgr` but you return `Rgb`.

davidpiuva: Updated. My compiler did not catch that but something about "CharBuffer lastIndexOf()" is instantly crashing every time.

source/draw/Color.ooc
```diff
@@ -73,6 +79,8 @@ ColorYuva: cover {
 	toMonochrome: func -> ColorMonochrome { ColorMonochrome new(this y) }
 	toUv: func -> ColorUv { this toYuv() toUv() }
 	toYuv: func -> ColorYuv { ColorYuv new(this y, this u, this v) }
+	toRgb: func -> ColorRgb { this toYuv() toRgb() }
+	toRgba: func -> ColorRgba { ColorRgba new(this toRgb(), alpha) }
```
sebastianbaginski: `this alpha`
`this` is missing in `toBgra` too.

sebastianbaginski:
Looks good.

source/draw/Color.ooc
davidpiuva: Updated

PR #1389, opened 2016-02-15
fixed leaked Closure in RasterMonochrome

sebastianbaginski:
@marcusnaslund


marcusnaslund:
Are closures always supposed to be freed by the function that uses them?

sebastianbaginski:
I don't know, this seems strange to me that assignment of existing function creates new closure that you need to free explicitly.

marcusnaslund:
I agree, this is better, but we should determine some general rule for passing and freeing closures.

PR #1390, opened 2016-02-15
Added tests for Pipe and Process

### LINKED ISSUE #1220: Write tests for Pipe and Process
These classes are never tested.

- [ ] `Pipe` (including of course `PipeUnix`)
- [ ] `Process` (including of course `ProcessUnix`)

marcusnaslund:
Fixes #1220

### DIFF
--- a/test/io/ProcessTest.ooc
+++ b/test/io/ProcessTest.ooc
@@ -1,0 +1,44 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use unit
+use io
+import io/FileReader
+
+ProcessTest: class extends Fixture {
+	init: func {
+		super("Process")
+		this add("Basic use", func {
+			scriptName: String
+			version (windows)
+				scriptName = "bash test/io/input/pipeprocesstester.sh"
+			else
+				scriptName = "test/io/input/pipeprocesstester.sh"
+
+			scriptArgs := [scriptName, "10000", "abcABC"]
+			process := Process new(scriptArgs)
+			process setStdout(Pipe new())
+			process execute()
+			process args free()
+			process free()
+			scriptArgs free()
+
+			reader := FileReader new(t"test/io/output/sum.txt")
+			expect(reader hasNext())
+			data := CString new(16)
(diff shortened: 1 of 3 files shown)


marcusnaslund:

test/io/ProcessTest.ooc
```diff
@@ -0,0 +1,44 @@
...
+			scriptArgs free()
+
+			reader := FileReader new(t"test/io/output/sum.txt")
+			expect(reader hasNext())
+			data := CString new(16)
+			reader read(data, 0, 15)
+			string := data toString()
+			expect(string, is equal to("50005000 abcABC"))
+			string free()
+			memfree(data)
```
sebastianbaginski: `CString` does not have a `free` function ?

test/io/input/pipeprocesstester.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+sum_to () (
+    set -- $(seq $1)
+    IFS=+
```
sebastianbaginski: When a bash function is changing the internal field separator then I think it does something more than simply `sum_to` :)

test/io/ProcessTest.ooc
marcusnaslund: No, but it's only a `Char*` cover.

sebastianbaginski:
OK, nice test, looks good.

test/io/input/pipeprocesstester.sh
marcusnaslund: I actually have no idea what this is doing, I got it online. Bash is just awful.

sebastianbaginski: What do you want it to do ? :)

sebastianbaginski: I think it sums the numbers by redirecting the string created from numeric values separated by `+` to `bc`. So it calculates `1+2+3+4+5+..` by concatenating and sending this string to separate process to do the calculations :D

marcusnaslund: I just wanted some kind of function that takes some time to run. My only critera was that it does not contain `rm -rf /`. This did the trick. :)

sebastianbaginski: You can simply use sleep - `sleep 0.5` will delay for half second.

marcusnaslund: Oh. Ah well. If this doesn't do anything bad I'm happy this way.

PR #1392, opened 2016-02-15
Added suppression file and updated memcheck script

marcusnaslund:
Copied from some other mysterious project.

I would also like to (for the time being) suppress everything related to the `Raster` classes but not the `RasterTest` itself but I don't yet know how to do that. @sebastianbaginski probably does?

### DIFF
--- a/tools/kean.supp
+++ b/tools/kean.supp
@@ -1,0 +1,36 @@
+{
+   compileshader
+   Memcheck:Leak
+   ...
+   fun:backend_gles3_Gles3ShaderProgram__Gles3ShaderProgram__compileShader
+}
+{
+   fboClear
+   Memcheck:Leak
+   ...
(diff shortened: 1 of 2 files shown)


sebastianbaginski:
You can use regular expressions in suppression file.

marcusnaslund:
> You can use regular expressions in suppression file.

Thank you. And thank you valgrind.

marcusnaslund:
Changed my mind. This is allowed to leak (for now), so let's not suppress anything extra.

fredrikbryntesson:
I thought we suppressed more stuff actually.

fredrikbryntesson:
But looks good.

marcusnaslund:
> I thought we suppressed more stuff actually.

I recently removed two entries. :) I will examine how hard it is to remove the String suppression after my upcoming trip to a certain warm country.

PR #1393, opened 2016-02-16
RGB adapt

davidpiuva:
Pen and plotter uses RGB instead of BGR.

These small changes will break the Kean API for each commit.

PR #1394, opened 2016-02-16
RGB image

davidpiuva:
Adds RGB and RGBA images for CPU and OpenGL.

### DIFF
--- a/source/draw/RasterRgb.ooc
+++ b/source/draw/RasterRgb.ooc
@@ -10,6 +10,7 @@ use geometry
 use base
 import RasterPacked
 import RasterImage
+import io/File
 import StbImage
 import Image
 import Color
@@ -75,6 +76,59 @@ RasterRgb: class extends RasterPacked {
 		result canvas draw(this, IntBox2D new(this size), IntBox2D new(size))
 		result
 	}
+	distance: func (other: Image) -> Float {
+		result := 0.0f
+		if (!other || (this size != other size))
+			result = Float maximumValue
+		else if (!other instanceOf(This)) {
+			converted := This convertFrom(other as RasterImage)
+			result = this distance(converted)
+			converted referenceCount decrease()
+		} else {
+			for (y in 0 .. this size y)
+				for (x in 0 .. this size x) {
+					c := this[x, y]
+					o := (other as This)[x, y]
+					if (c distance(o) > 0) {
+						maximum := o
+						minimum := o
+						for (otherY in 0 maximum(y - this distanceRadius) .. (y + 1 + this distanceRadius) minimum(this size y))
+							for (otherX in 0 maximum(x - this distanceRadius) .. (x + 1 + this distanceRadius) minimum(this size x))
+								if (otherX != x || otherY != y) {
+									pixel := (other as This)[otherX, otherY]
+									if (maximum blue < pixel blue)
+										maximum blue = pixel blue
+									else if (minimum blue > pixel blue)
+										minimum blue = pixel blue
+									if (maximum green < pixel green)
+										maximum green = pixel green
+									else if (minimum green > pixel green)
+										minimum green = pixel green
+									if (maximum red < pixel red)
+										maximum red = pixel red
+									else if (minimum red > pixel red)
+										minimum red = pixel red
+								}
+						distance := 0.0f
+						if (c blue < minimum blue)
+							distance += (minimum blue - c blue) as Float squared
+						else if (c blue > maximum blue)
+							distance += (c blue - maximum blue) as Float squared
+						if (c green < minimum green)
+							distance += (minimum green - c green) as Float squared
+						else if (c green > maximum green)
+							distance += (c green - maximum green) as Float squared
+						if (c red < minimum red)
+							distance += (minimum red - c red) as Float squared
+						else if (c red > maximum red)
+							distance += (c red - maximum red) as Float squared
+						result += (distance) sqrt() / 3
+					}
+				}
+			result /= this size length
+		}
+		result
+	}
 	swapRedBlue: func {
 		this swapChannels(0, 2)
 	}
@@ -90,6 +144,18 @@ RasterRgb: class extends RasterPacked {
 
 	open: static func (filename: String) -> This {
 		This convertFrom(RasterBgr open(filename))
+	}
+	savePacked: func (filename: String) -> Int {
+		file := File new(filename)
+		folder := file parent . mkdirs() . free()
+		file free()
+		StbImage writePng(filename, this size x, this size y, this bytesPerPixel, this buffer pointer, this size x * this bytesPerPixel)
+	}
+	save: override func (filename: String) -> Int {
+		bgr := this redBlueSwapped()
+		result := bgr savePacked(filename)
+		bgr free()
+		result
 	}
 	convertFrom: static func (original: RasterImage) -> This {
 		result: This
--- a/source/draw/RasterRgb.ooc
+++ b/source/draw/RasterRgb.ooc
@@ -10,11 +10,12 @@ use geometry
 use base
 import RasterPacked
 import RasterImage
+import RasterRgba
 import io/File
 import StbImage
 import Image
 import Color
-import Canvas, RasterCanvas, RasterBgr
+import Canvas, RasterCanvas
 
 RasterRgbCanvas: class extends RasterPackedCanvas {
 	target ::= this _target as RasterRgb
@@ -143,7 +144,7 @@ RasterRgb: class extends RasterPacked {
 	operator []= (x, y: Int, value: ColorRgb) { ((this buffer pointer + y * this stride) as ColorRgb* + x)@ = value }
 
 	open: static func (filename: String) -> This {
-		This convertFrom(RasterBgr open(filename))
+		This convertFrom(RasterRgba open(filename))
 	}
 	savePacked: func (filename: String) -> Int {
 		file := File new(filename)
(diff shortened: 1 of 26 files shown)


marcusnaslund:
#1393 will be evaluated right after lunch.

davidpiuva:
Rebased.

davidpiuva:
@emilwestergren Peer review?

marcusnaslund:
I still expect tests for the new `Raster` classes. Should be simple enough to adapt the same tests as those for the `Bgr` versions.

davidpiuva:
If BGR will be removed then I can rewrite them to use RGB. It will be a huge pull request.

emilwestergren:
Looks good, but I agree with @marcusnaslund, can you copy the tests for BGR and then we can simply remove all BGR files in a single PR after this?

marcusnaslund:
In comparison to everything else I see nothing huge about the `RasterBgr` tests.

davidpiuva:
Actually removing BGR should probably go in a separate pull request so that it can be reverted in the future or at least used as a reference if we suddenly want it back.

emilwestergren:
Yes

davidpiuva:
My "Rgb image adapt" pull requests are removing around half of what is using BGR right now.

emilwestergren:
So:
1. Include tests for RGB in this PR
2. Create PR that removes any use of BGR
3. Create PR that removes BGR

davidpiuva:
So I just copy the files and replace BGR with RGB for its own tests while leaving all unrelated tests using BGR?

emilwestergren:
Yes

davidpiuva:
One of the tests can not be implemented for RGB now because saving of PNG images is hard coded for BGR. I have to make some workaround by flipping channels on a copy in order to delete BGR.

davidpiuva:
The thread pool test is still randomly crashing.

davidpiuva:
Now I just need to break the infinite recursion by calling the base save function.
```
save: override func (filename: String) -> Int {
	bgra := this redBlueSwapped()
	result := bgra save(filename)
	bgra free()
	result
}
```

sebastianbaginski:
https://github.com/magic-lang/ooc-kean/issues/1409 it happened to my PRs too few days ago

davidpiuva:
I failed to override save when using super.
The generated virtual function table shows that RasterPacked's save is still called causing the tests to fail. `.save = (void*) RasterPacked__RasterPacked_save_impl`

davidpiuva:
When commenting out the call to super, everything works.
`.save = (void*) RasterRgb__RasterRgb_save_impl`

davidpiuva:
I might have to duplicate the save function from Packed until the compiler is fixed.

davidpiuva:
Updated.

source/draw/RasterRgb.ooc
```diff
@@ -0,0 +1,189 @@
...
+	}
+	_createCanvas: override func -> Canvas { RasterRgbCanvas new(this) }
+
+	operator [] (x, y: Int) -> ColorRgb { this isValidIn(x, y) ? ((this buffer pointer + y * this stride) as ColorRgb* + x)@ : ColorRgb new(0, 0, 0) }
+	operator []= (x, y: Int, value: ColorRgb) { ((this buffer pointer + y * this stride) as ColorRgb* + x)@ = value }
+
+	open: static func (filename: String) -> This {
+		This convertFrom(RasterBgr open(filename))
+	}
+	savePacked: func (filename: String) -> Int {
```
davidpiuva: This will be removed once super works.

source/draw/RasterRgba.ooc
```diff
@@ -0,0 +1,201 @@
...
+			(top * (left * topLeft green + (1 - left) * topRight green) + (1 - top) * (left * bottomLeft green + (1 - left) * bottomRight green)),
+			(top * (left * topLeft blue + (1 - left) * topRight blue) + (1 - top) * (left * bottomLeft blue + (1 - left) * bottomRight blue)),
+			(top * (left * topLeft alpha + (1 - left) * topRight alpha) + (1 - top) * (left * bottomLeft alpha + (1 - left) * bottomRight alpha))
+		)
+	}
+
+	open: static func (filename: String) -> This {
+		This convertFrom(RasterBgra open(filename))
+	}
+	savePacked: func (filename: String) -> Int {
```
davidpiuva: This will be removed once super works.

davidpiuva:
The generated code seems to be completely random for the same input. Probably lazy compilation causing it again.

davidpiuva:
@marcusnaslund All tests passed.

marcusnaslund:
Conflicts.

davidpiuva:
Made interactive rebase with new commits to handle version conflict.

davidpiuva:
@fredrikbryntesson `SynchronizedResourceTest` crashed.

davidpiuva:
@fredrikbryntesson All tests passed.

davidpiuva:
@sebastianbaginski Peer review the last added commit?

source/draw/RasterRgb.ooc
```diff
@@ -0,0 +1,195 @@
...
+	operator []= (x, y: Int, value: ColorRgb) { ((this buffer pointer + y * this stride) as ColorRgb* + x)@ = value }
+
+	open: static func (filename: String) -> This {
+		x, y, imageComponents: Int
+		requiredComponents := 3
+		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
+		buffer := ByteBuffer new(data as Byte*, x * y * requiredComponents, true)
+		result := This new(buffer, IntVector2D new(x, y))
+		result swapRedBlue()
+		result
```
sebastianbaginski: Maybe `This convertFrom(RasterRgba open(filename))`

davidpiuva: A function returning a raw buffer from a file might be able to give the correct `requiredComponents` from the beginning.

sebastianbaginski: It does not matter, this is only for debugging / testing. You were complaining about code bloat in draw module so just use a function that already exists.

davidpiuva: So it is okay if it might crash?

sebastianbaginski: You can of course add this function if you want, but this PR is quite big already.

sebastianbaginski: What might crash ?

davidpiuva: Saying that the image has 4 dimensions when it has only 3. I don't know how the exception is handled in the external library.

sebastianbaginski: If you ask for 4 and image has only 3 then you will get only 3, `imageComponents` returns the value from file.

sebastianbaginski:
OK just leave this as it is for now, I guess we should clean this up anyway.

sebastianbaginski:
Great, PR done.

davidpiuva:
@fredrikbryntesson All checks passed again.

source/draw/RasterRgb.ooc
```diff
@@ -0,0 +1,190 @@
...
+			rgb = RasterRgb convertFrom(image as RasterImage)
+		else
+			Debug error("Unsupported image type in RgbRasterCanvas draw")
+		this _resizePacked(rgb buffer pointer as ColorRgb*, rgb, source, destination)
+		if (rgb != image)
+			rgb referenceCount decrease()
+	}
+}
+
+RasterRgb: class extends RasterPacked {
```
davidpiuva: A copy of BGR with loading and saving changed.

source/draw/RasterRgba.ooc
```diff
@@ -0,0 +1,207 @@
...
+	target ::= this _target as RasterRgba
+	init: func (image: RasterRgba) { super(image) }
+	_drawPoint: override func (x, y: Int) {
+		position := this _map(IntPoint2D new(x, y))
+		if (this target isValidIn(position x, position y))
+			this target[position x, position y] = this target[position x, position y] blend(this pen alphaAsFloat, this pen color)
+	}
+}
+
+RasterRgba: class extends RasterPacked {
```
davidpiuva: A copy of BGRA with loading and saving changed.

source/draw/RasterRgba.ooc
```diff
@@ -0,0 +1,207 @@
...
+	}
+	savePacked: func (filename: String) -> Int {
+		file := File new(filename)
+		folder := file parent . mkdirs() . free()
+		file free()
+		StbImage writePng(filename, this size x, this size y, this bytesPerPixel, this buffer pointer, this size x * this bytesPerPixel)
+	}
+	save: override func (filename: String) -> Int {
+		bgra := this redBlueSwapped()
+		result := bgra savePacked(filename)
```
davidpiuva: Calling super failed to generate correct C code so the code was duplicated.

source/draw/RasterRgb.ooc
```diff
@@ -0,0 +1,190 @@
...
+	}
+	savePacked: func (filename: String) -> Int {
+		file := File new(filename)
+		folder := file parent . mkdirs() . free()
+		file free()
+		StbImage writePng(filename, this size x, this size y, this bytesPerPixel, this buffer pointer, this size x * this bytesPerPixel)
+	}
+	save: override func (filename: String) -> Int {
+		bgr := this redBlueSwapped()
+		result := bgr savePacked(filename)
```
davidpiuva: Calling super failed to generate correct C code so the code was duplicated.

PR #1395, opened 2016-02-16
Changed from YVU to YUV for consistency with semiplanar images.

davidpiuva:
This seems to be an old bug since the planar YUV format has not been used.


davidpiuva:
@emilwestergren Peer review?

PR #1397, opened 2016-02-17
Removed unused YvuPlanar to Bgra shader.

davidpiuva:
It was supposed to be YuvPlanar but that format is not even used anymore.


davidpiuva:
@emilwestergren Peer review?

PR #1398, opened 2016-02-17
fixed memory leak in File getParent

sebastianbaginski:
`File` makes a copy and frees its `path`.
Now nothing leaks from test suite in the other project (except the ordinary stuff).
@marcusnaslund review


marcusnaslund:
`/giphy celebrate`

PR #1399, opened 2016-02-17
Maximum/minimum value for Long/TimeSpan

emilwestergren:
* Added min/max value for `Long`
* Added min/max value for `TimeSpan`

### DIFF
--- a/source/base/TimeSpan.ooc
+++ b/source/base/TimeSpan.ooc
@@ -96,6 +96,8 @@ TimeSpan: cover {
 	hour: static func -> This { This hours(1) }
 	day: static func -> This { This days(1) }
 	week: static func -> This { This weeks(1) }
+	maximumValue ::= static This new(Long maximumValue)
+	minimumValue ::= static This new(Long minimumValue)
 
 	milliseconds: static func (count: Double) -> This { This new(DateTime ticksPerMillisecond * count) }
 	seconds: static func (count: Double) -> This { This new(DateTime ticksPerSecond * count) }
--- a/source/system/math.ooc
+++ b/source/system/math.ooc
@@ -27,6 +27,8 @@ extend Short {
 }
 
 extend Long {
+	maximumValue ::= static INT64_MAX
+	minimumValue ::= static INT64_MIN
 	modulo: func (divisor: This) -> This {
 		result := this - (this / divisor) * divisor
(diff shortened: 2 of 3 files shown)


emilwestergren:
@marcusnaslund peer review

source/system/math.ooc
```diff
@@ -11,6 +11,8 @@ use math-lib
 
 FLT_EPSILON: extern Float
 DBL_EPSILON: extern Double
+INT64_MAX: extern Long
+INT64_MIN: extern Long
```
marcusnaslund: These should be defined in `Numbers.ooc` together with the others.

source/base/TimeSpan.ooc
```diff
@@ -103,6 +103,8 @@ TimeSpan: cover {
 	hours: static func (count: Double) -> This { This new(DateTime ticksPerHour * count) }
 	days: static func (count: Double) -> This { This new(DateTime ticksPerDay * count) }
 	weeks: static func (count: Double) -> This { This new(DateTime ticksPerWeek * count) }
+	maximumValue ::= static This new(Long maximumValue)
+	minimumValue ::= static This new(Long minimumValue)
```
marcusnaslund: These make more sense above line 93 instead

source/system/math.ooc
```diff
@@ -33,6 +35,8 @@ extend Long {
 	}
 	maximum: func (other: This) -> This { this > other ? this : other }
 	minimum: func (other: This) -> This { this < other ? this : other }
+	maximumValue ::= static INT64_MAX
+	minimumValue ::= static INT64_MIN
```
marcusnaslund: These should be at the top (like in `Int` and others)

marcusnaslund:
Besides changing everything, this looks good.

source/system/math.ooc
emilwestergren: I thought static members should always be at the bottom?

marcusnaslund: Hmm generally yeah, except we (meaning I) seem to have always put them at the top in this file. Let's be consistent.

emilwestergren: Sort of breaks the point of the general policy though, but alright I don't mind.

emilwestergren:
Updated

marcusnaslund:
Excellent.

PR #1406, opened 2016-02-17
Removed GpuMap.

emilwestergren:
No need for it when we have a base class `Map`.

@davidpiuva peer review


davidpiuva:
Sounds like a good idea. I will have a look tomorrow.

source/draw/Map.ooc
```diff
@@ -8,9 +8,73 @@
...
+	_ownsObject: Bool
+	init: func (=key, =value, =_ownsObject)
+	free: override func {
+		if (this _ownsObject)
+			this value free()
+		super()
+	}
+}
+
+_Map: class {
```
davidpiuva: Maybe call it `PropertyBag` or `Dictionary` to avoid confusion with shaders.

source/draw/Map.ooc
```diff
@@ -8,9 +8,73 @@
...
+	}
+	use: virtual func
+	add: func <T> (key: String, value: T) {
+		if (T inheritsFrom(Object))
+			this _bindings add(key, value as Object, false)
+		else
+			this _bindings add(key, Cell<T> new(value), true)
+	}
+	get: func (key: String) -> Object { this _bindings get(key) value }
+	apply: func (action: Func (String, Object)) { this _bindings apply(action) }
```
davidpiuva: Most of these methods are just wrappers for _Map. Can we merge the classes or is purity important when passing _Map as an argument?

source/draw/Map.ooc
```diff
@@ -8,9 +8,73 @@
 
 use geometry
 
+_MapEntry: class {
```
davidpiuva: A bit overkill to have a new class for each entry when we already have cells around each value. Can this inherit from Cell?

davidpiuva:
Naming could be improved. Allocation might be possible to simplify but this can be another PR if a simpler solution is found.

emilwestergren:
Yes, all valid points but should be in a separate PR since this one is only about removing/moving code.

marcusnaslund:
So either 1) Get on that right away, or 2) Make an issue about it

emilwestergren:
@davidpiuva create issue if you think it's necessary, I don't mind the current solution

PR #1408, opened 2016-02-18
Removed a third of the repo

marcusnaslund:
Title was chosen for maximal effect.

Basically, this file takes up over a third of the repo size (15 MB), but **isn't used anywhere**.

Any objections @emilwestergren @niklasborwell @davidpiuva ? We talked yesterday about having large test-files in a separate folder so this seems like a candidate for that, if we ever need it.


davidpiuva:
It is okay.

niklasborwell:
DO EIT!

Issue #1403, opened 2016-02-17
Correct enum values in GpuImage.ooc


sebastianbaginski:
```ooc
GpuImageType: enum {
	monochrome
	rgba
	rgb
	bgr
	bgra
	uv
	yuvSemiplanar
	yuvPlanar
	yuv422
}
```
should they start with upper case ? @marcusnaslund @emilwestergren

marcusnaslund:
Yes. `GpuImageType` together with `Order` are the only exceptions.
Also, `TextureType` have almost the same elements and they start with upper case.

PR #1413, opened 2016-02-19
Capitalized entries in the Order enum

marcusnaslund:
All others are capitalized except this one (and `GpuImageType`, see #1403 )

Peer review @thomasfanell ?


thomasfanell:
This is some tricky code you're sending my way...

source/base/Order.ooc
```diff
@@ -7,7 +7,7 @@
  */
 
 Order: enum {
-	less = -1
-	equal = 0
-	greater = 1
+	Less = -1
+	Equal = 0
```
thomasfanell: By the way, `Equal` and `Greater` need not be explicit.

PR #1416, opened 2016-02-19
Properly defined maximum and minimum values

### LINKED ISSUE #1402: Define maximumValue and minimumValue properly
As of #1399,  `maximumValue` and `minimumValue` are defined on `Long`, and they are defined on the `int64_t` sizes, as it should be.

We need to change the values of `Int minimumValue` etc. to conform to the exact platform-independent sizes.

We also need to add these properties to the unsigned versions.

marcusnaslund:
Fixes #1402

### DIFF
--- a/source/system/math.ooc
+++ b/source/system/math.ooc
@@ -21,14 +21,7 @@ asin: extern (asinf) func (Float) -> Float
 atan: extern (atanf) func (Float) -> Float
 sqrt: extern (sqrtf) func (Float) -> Float
 
-extend Short {
-	minimumValue ::= static SHRT_MIN
-	maximumValue ::= static SHRT_MAX
-}
-
 extend Long {
-	maximumValue ::= static INT64_MAX
-	minimumValue ::= static INT64_MIN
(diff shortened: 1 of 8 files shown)


marcusnaslund:
Peer review @thomasfanell ? Slightly more exciting code than last time. :)

source/system/math.ooc
```diff
@@ -42,11 +35,6 @@ extend ULong {
 }
 
 extend Int {
-	negativeInfinity ::= static INT_MIN
-	positiveInfinity ::= static INT_MAX
-	epsilon ::= static 1
```
marcusnaslund: `Int epsilon` just kind of makes no sense.

marcusnaslund:
@thomasfanell Magic fails at

`slice: func (start: Int, distance := Int maximumValue) -> This`

with the message

`[107, 58].....redundant type information, reduce to (start, distance, maximumValue: Int)`

thomasfanell:
@marcusnaslund a fix is on the way

PR #1420, opened 2016-02-22
string_literal module to workaround memory leak caused by 'makeStringLiteral'

sebastianbaginski:
In order to fix this memory leak I needed to create a separate module
in plain C, just to prevent circular references to static variables
declared in `String` class.
What I mean is that it is (I think) impossible to fix this at ooc level,
as declaring the `string literal` arrays directly in `String` class causes
rock to generate references to `String__String__class()` when accessing them:
```ooc
String: class {
    _stringLiterals := static This* = null
    _stringLiteralsCount := static 0
...
}

makeStringLiteral: func (...) {
   // this line will cause `String__String__class()` access,
   // which will again call `makeStringLiteral` at some point
    if (String _stringLiteralsCount >= ...) {
        String _stringLiterals = realloc( ... ) // last realloc result always leaked
        ...
    }
}

```
This caused the `_stringLiterals` array to leak for some reason,
even if it was explicitly released at program exit.
I struggled to fix this at ooc level for some time, other idea was to use global variables
instead of class statics, but this is even worse as global variables are not initialized,
even when explicitly assigned:
```ooc
stringLiterals := static String* = null

String: class {...
```
generates following C code:
```C
// at file scope 'String.c'
static String__String* String__stringLiterals; // not initialized anywhere, contains garbage
...
```

In fact my first idea was just to use a static `VectorList` in `String` class, forget it, VectorList class init functions calls `makeStringLiteral` too.

So in the end I decided to write logically equavalent code but in plain C and
manage string literal memory in separate module referenced via extern functions. Of course it worked without problems.

tl;dr : I needed to create a C "plugin" for `String` in order to fix the leaked string literals

@fredrikbryntesson review
@marcusnaslund @fredrikbryntesson @thomasfanell  I know it would be the best to fix such things at ooc level, but believe me, I spent far too many hours trying to fix this problem in ooc. We already have hand-written C code for `Arrays`, so I think this one is not that bad. What do you think about this solution ?

### DIFF
--- a/source/system/String.ooc
+++ b/source/system/String.ooc
@@ -6,29 +6,31 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
+include ./string/string_literal
+
+string_literal_new: extern func (String)
+string_literal_free: extern func (String)
+string_literal_free_all: extern func
+
 String: class {
 	// Avoid direct buffer access, as it breaks immutability.
 	_buffer: CharBuffer
-
 	size ::= _buffer size
 
 	init: func ~withBuffer (=_buffer)
-
-	init: func ~withCStr (s: CString) {
-		init(s, s length())
-	}
-
+	init: func ~withCStr (s: CString) { init(s, s length()) }
 	init: func ~withCStrAndLength (s: CString, length: Int) {
 		_buffer = CharBuffer new(s, length)
 	}
 	free: override func {
+		if (this _buffer mallocAddr == 0)
+			string_literal_free(this)
 		this _buffer free()
 		super()
 	}
 	length: func -> Int {
 		_buffer size
 	}
-
 	equals: final func (other: This) -> Bool {
 		result := false
 		if (this == null)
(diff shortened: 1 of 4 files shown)


source/system/String.ooc
```diff
@@ -22,6 +28,8 @@ String: class {
 		_buffer = CharBuffer new(s, length)
 	}
 	free: override func {
+		if (this _buffer mallocAddr == 0)
```
sebastianbaginski: `buffer mallocAddr` is `0` only for string literals

thomasfanell:
This is similar to something i've considered doing myself, so I think it's a good solution. Great job!

sebastianbaginski:
> This is similar to something i've considered doing myself

Thanks, good to know that.  Have you also tried to workaround this in ooc ?

thomasfanell:
When I had a look, I came to the conclusion that the string stuff was so deeply nested it wouldn't really be worth it. Judging what you wrote here, I think you came to the same conclusion.

source/system/string/string_literal.h
```diff
@@ -0,0 +1,8 @@
+#ifndef __STRINGLITERAL_FIX_H__
+#define __STRINGLITERAL_FIX_H__
+
```
fredrikbryntesson: _FIX_ makes it feel like a temporary hack that should be changed very soon.

fredrikbryntesson:
String.ooc and string_literal.c has a lot of empty lines.

marcusnaslund:
Why not rather try to get rid of String all together?

sebastianbaginski:
> Why not rather try to get rid of String all together?

So why do we have an issue addressing this memory leak specifically ?
Also, isn't `makeStringLiteral` hard-coded in the compiler ?

source/system/string/string_literal.h
sebastianbaginski: That was the idea, I'm happy that you thought so. This is a hack and should be removed as soon as we get rid of String class.

fredrikbryntesson:
Yes, or the c-function is actually.

sebastianbaginski:
@fredrikbryntesson newlines in .c removed
do you want me to remove them all from String.ooc in this PR ?

sebastianbaginski:
OK it is done.

PR #1422, opened 2016-02-23
Remove bgr

davidpiuva:
Removing BGR completely. This final step breaks backward compatibility to prevent anyone from creating new dependencies to the BGR weirdness.


davidpiuva:
Tests fails now because there is no branch with the same name in the other repository.

davidpiuva:
@thomasfanell Peer review?

thomasfanell:
Looks good to me

marcusnaslund:
Good job.

PR #1425, opened 2016-02-24
more tests for FloatTransform2D

molhamkhalil:
more test sets added to FloatTransform2DTest.
two test sets were commented, since problems were faced during setRotation at those commented ones.


molhamkhalil:
using branch math_test instead

Issue #1427, opened 2016-02-24
OpenGL safety


davidpiuva:
* Prevent any input texture from also being the output texture. This may crash on phones.
* Prevent any shader from reading a texture that has not been assigned recently.
* Prevent uniforms from leaking to another draw call.

davidpiuva:
@emilwestergren @sebastianbaginski @fredrikbryntesson any suggestions?

emilwestergren:
I don't think this should be the priority right now, but we can keep the issue and discuss it later.

PR #1429, opened 2016-02-26
Safety

davidpiuva:
Since the full safety feature takes too long time, I made a lightweight safety feature to begin with.

* Giving more relevant information when drawing with DrawState to an unsupported target.
* Implemented a safety feature from https://github.com/magic-lang/ooc-kean/issues/1427. This check is not compatible with the legacy calls becuase they are in `OpenGLSurface` but using it as a canvas without access to `_target`.

### DIFF
--- a/source/draw/gpu/opengl/OpenGLCanvas.ooc
+++ b/source/draw/gpu/opengl/OpenGLCanvas.ooc
@@ -40,7 +40,7 @@ OpenGLCanvas: class extends OpenGLSurface {
 			this context backend enableBlend(false)
 		if (drawState inputImage)
 			gpuMap add("texture0", drawState inputImage)
-		gpuMap use()
+		gpuMap use(this _target)
 		this _bind()
 		this context drawQuad()
 		this _unbind()
--- a/source/draw/gpu/opengl/OpenGLContext.ooc
+++ b/source/draw/gpu/opengl/OpenGLContext.ooc
@@ -77,7 +77,7 @@ OpenGLContext: class extends GpuContext {
 		positions := pointList pointer as Float*
 		this _linesShader projection = projection
 		this _linesShader add("color", pen color normalized)
-		this _linesShader use()
+		this _linesShader use(null)
 		this _renderer drawLines(positions, pointList count, 2, pen width)
 	}
 	drawPoints: func (pointList: VectorList<FloatPoint2D>, projection: FloatTransform3D, pen: Pen) {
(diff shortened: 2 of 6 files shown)


source/draw/gpu/opengl/OpenGLSurface.ooc
```diff
@@ -37,7 +37,7 @@ OpenGLSurface: abstract class extends GpuCanvas {
 		map model = this _createModelTransform(destination)
 		map view = this _view
 		map projection = this _projection
-		map use()
+		map use(null)
```
davidpiuva: `null` should actually be `this _target` that does not exist in `OpenGLSurface`.

source/draw/gpu/opengl/OpenGLSurface.ooc
```diff
@@ -60,7 +60,7 @@ OpenGLSurface: abstract class extends GpuCanvas {
 		f := func {
 			this context meshShader add("texture0", image)
 			this context meshShader projection = this _projection
-			this context meshShader use()
+			this context meshShader use(null)
```
davidpiuva: `null` should actually be `this _target` that does not exist in `OpenGLSurface`.

source/draw/gpu/opengl/OpenGLContext.ooc
```diff
@@ -77,7 +77,7 @@ OpenGLContext: class extends GpuContext {
 		positions := pointList pointer as Float*
 		this _linesShader projection = projection
 		this _linesShader add("color", pen color normalized)
-		this _linesShader use()
+		this _linesShader use(null)
```
davidpiuva: No need to compare input textures when none are given.

source/draw/gpu/opengl/OpenGLContext.ooc
```diff
@@ -85,7 +85,7 @@ OpenGLContext: class extends GpuContext {
 		this _pointsShader add("color", pen color normalized)
 		this _pointsShader add("pointSize", pen width)
 		this _pointsShader projection = projection
-		this _pointsShader use()
+		this _pointsShader use(null)
```
davidpiuva: No need to compare input textures when none are given.

source/draw/gpu/opengl/OpenGLCanvas.ooc
```diff
@@ -40,7 +40,7 @@ OpenGLCanvas: class extends OpenGLSurface {
 			this context backend enableBlend(false)
 		if (drawState inputImage)
 			gpuMap add("texture0", drawState inputImage)
-		gpuMap use()
+		gpuMap use(this _target)
```
davidpiuva: Tell the uploading of arguments that target is reserved for output.

davidpiuva:
@thomasfanell Peer review?

source/draw/gpu/opengl/OpenGLSurface.ooc
davidpiuva: It might be possible to call a function that is overridden in `OpenGLCanvas` but adding complexity to support a safety feature in legacy code is not worth it.

PR #1430, opened 2016-02-26
Draw state mesh

davidpiuva:
Using a mesh with DrawState should now be identical to the old version.

### DIFF
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -10,6 +10,7 @@ use geometry
 import Pen
 import Image
 import Map
+import Mesh
 
 // Example
 // DrawState new(targetImage) setMap(shader) draw()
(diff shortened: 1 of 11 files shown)


davidpiuva:
@emilwestergren Peer review?

source/draw/Mesh.ooc
```diff
@@ -6,9 +6,7 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-version(!gpuOff) {
```
davidpiuva: The version flag was preventing `GpuMesh` from being used in `DrawState`. A mesh could in theory exist for CPU rendering too even if we don't have it now.

emilwestergren: Yes, we want to remove as much as possible of the `Gpu` abstraction layer anyway.

source/draw/DrawState.ooc
```diff
@@ -35,6 +37,10 @@ DrawState: cover {
 		this map = map
 		this
 	}
+	setMesh: func (mesh: Mesh) -> This {
+		this mesh = mesh
+		this
+	}
```
davidpiuva: This is the new interface for mesh.

PR #1431, opened 2016-02-29
Create output folder in PipeTest

fredrikbryntesson:
Create output folder to remove requirement that ProcessTest or CsvWriterTest must be executed before PipeTest.


marcusnaslund:
This looks ready but is not in review so, to quote a famous rapper, I can't touch this.

fredrikbryntesson:
@jonathanudd

PR #1432, opened 2016-02-29
Moved WaitCondition and Thread to source/concurrent

### LINKED ISSUE #659: Move system/threading to source/concurrent
There is only one obstacle in the way: `Exception`. It uses `ThreadLocal` to handle exceptions from different threads. (It's unclear whether this actually works.)

Unfortunately, moving the folder out and putting `use ooc-threading` in `Exception.ooc` causes a circular dependancy.

marcusnaslund:
Totally breaks the threading/concurrent API but fixes #659


marcusnaslund:
@fredrikbryntesson There are three classes in `source/base` that use `Mutex` and because things in `concurrent` use things in `base`, we cannot move `Mutex` to `concurrent` without getting circular dependencies. I suggest we keep `Mutex` inside `source/system` for now. Agree?

fredrikbryntesson:
Perfect

marcusnaslund:

PR #1435, opened 2016-03-01
Changed red,green,blue to r,g,b in RGB(a) Colors

marcusnaslund:
First part of solving #1412, rename `red`, `green`, `blue` to `r`, `g`, `b`. The former names will be used as static members in a later PR.

PR #1436, opened 2016-03-01
Added default static colors

### LINKED ISSUE #1412: Static default colors in color covers
`black ::= static ColorRgba new(0, 0, 0, 1)`
`white ::= static ColorRgba new(1, 1, 1, 1)`

would be nice. One problem is that `red/green/blue` are already members in `ColorRgba` so we can't define those default colors unless we change the naming.

@davidpiuva @marcusnaslund what do you think?

marcusnaslund:
Depends on #1435 

Fixes #1412

### DIFF
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -12,6 +12,9 @@ use geometry
 ColorMonochrome: cover {
 	y: Byte
 	normalized ::= this y as Float / 255
+	black ::= static This new(0)
+	white ::= static This new(255)
+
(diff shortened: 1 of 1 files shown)


marcusnaslund:
I've added `black`, `white`, `red`, `green`, and `blue`. The yuv components I calculated using some online rgb->yuv converter. Peer review @davidpiuva or @emilwestergren , whoever comes first.

davidpiuva:
ok

emilwestergren:
yay

PR #1437, opened 2016-03-01
Reorganized the Timer classes

### LINKED ISSUE #1415: Rename ClockTimer to CpuClockTimer
What it really measures is the CPU usage for the process, it can cause confusion when used in multi-threaded environment.

marcusnaslund:
Fixes #1415 

Instead of `Timer` and `ClockTimer`, there is now an abstract `Timer` class from which `WallTimer` and `CpuTimer` are derived. What do you think about this @sebastianbaginski @fredrikbryntesson @thomasfanell ?

### DIFF
--- a/source/base/Profiler.ooc
+++ b/source/base/Profiler.ooc
@@ -14,7 +14,7 @@ import Timer
 
 Profiler: class {
 	_name: String
-	_timer := Timer new()
+	_timer := WallTimer new()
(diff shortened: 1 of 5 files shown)


fredrikbryntesson:
Makes sense.

marcusnaslund:
Then peer review @thomasfanell ?

PR #1440, opened 2016-03-01
CharacterTest now actually says it tests Character

marcusnaslund:
The person in charge of reviewing pull requests is obviously not doing his job, whoever it is. Peer review @jonathanudd ?


jonathanudd:
I wonder who it could be. Looks better now at least :)

PR #1443, opened 2016-03-01
General cleanup in source/net

marcusnaslund:
Removed the specialized exceptions (which were in implementation all identical and did nothing special), removed some comments, and compactified a few methods.

Peer reivew @jonathanudd ?


jonathanudd:
Looks good. I didn't read all the comments that are still there but I guess they are of some value to understand the code.

marcusnaslund:
>  I didn't read all the comments that are still there but I guess they are of some value to understand the code.

Well, yes and no. Mostly they're there to make it easier to clean everything up once we know what parts we use and not.

PR #1444, opened 2016-03-01
Fixed depth for projective transforms.

davidpiuva:
Fixed depth for projective transforms. Z is now defined in the same scale as X to allow using the same normalised projective transform between resolutions of the same aspect ratio. Changing a rotation axis just because the image has a different resolution made no sense since we only scale the rotation when aspect ratio changes.

This fixes a bug seen with DrawState when giving a reference transform that is normalized from full resolution and then brought back to reference coordinates in half resolution for the UV channels. This operation whould take half translation while preserving rotation but the rotation became distorted from the old conversion.

Old test cases based on incorrect results had to be removed. Having a test where you can't see what property is fulfilled only makes sure that it does not change and transforms are not matured enough to lock the implementation.


davidpiuva:
@emilwestergren and @niklasborwell Peer review?

test/draw/DrawStateTest.ooc
```diff
@@ -27,6 +27,22 @@ DrawStateTest: class extends Fixture {
 	}
 	init: func {
 		super("DrawState")
+		this add("uniform scaling does not affect rotation", func {
+			small := FloatVector2D new(150, 100)
+			large := FloatVector2D new(300, 200)
+			degreeToRadian := 0.017453293f
+			rotateX45 := FloatTransform3D createRotationX(45.0f * degreeToRadian)
```
marcusnaslund: `45.0f toRadians()` and likewise below.

source/geometry/FloatTransform3D.ooc
```diff
@@ -244,16 +244,16 @@ FloatTransform3D: cover {
 		This new(focalLength, 0, 0, 0, 0, focalLength, 0, 0, 0, 0, focalLength, 1.0f, 0, 0, 0, 0)
 	}
 	referenceToNormalized: func ~float (imageSize: FloatVector2D) -> This {
-		toReference := This createScaling(imageSize x / 2.0f, imageSize y / 2.0f, 1.0f)
-		toNormalized := This createScaling(2.0f / imageSize x, 2.0f / imageSize y, 1.0f)
+		toReference := This createScaling(imageSize x / 2.0f, imageSize y / 2.0f, imageSize x / 2.0f)
```
marcusnaslund: I don't really know too much about what's going on, but I'm curious, why scale in `z` direction with `x`?

davidpiuva: Because when you apply uniform scaling on the resolution and the transform affects depth, you want to preserve the rotation along X and Y so that 45 degrees to the right remains 45 degrees to the right no matter the resolution of the image on the plane. I chose X because I will soon define normalized focal length relative to the width of the image.

marcusnaslund: Alright, I'm confident you know what you're doing. :)

davidpiuva:
Updated.

marcusnaslund:
Sorry, it should also be `This` before `compareTransform`. Then I'm happy.

davidpiuva:
Updated

PR #1445, opened 2016-03-01
Perspective fix

davidpiuva:
DrawState now handles focalLength and stores it in normalized coordinates so that it will automatically adapt when drawing to YUV images. This will obsolete `focalLength` in canvas so that you do not have side effects.


davidpiuva:
@emilwestergren Peer review?

Issue #1043, opened 2015-12-07
Implement class holding WaitCondition and Mutex


emilwestergren:
Since a `Mutex` is always needed when using a `WaitCondition`, we might as well create a class for this.

Name suggestions so far:
* `SynchronizationLock`
* `WaitLock`
* `BlockLock`
* `MutexCondition`

The interface is the same as for `WaitCondition`, i.e.
`signal`
`broadcast`
`wait`

@sebastianbaginski @simonmika @marcusnaslund 
Let's reach a decision here and then someone can implement it.

sebastianbaginski:
I think `lock` and `unlock` will have to be in public interface too as sometimes you want to wait only after testing a condition protected by the mutex.

sebastianbaginski:
As for the name definitely `BlockLock`, as this one is a very unique name :)

emilwestergren:
Yes, you're right, unless we can hide this instead.
```ooc
wait: func (condition: Func -> Bool)
```

simonmika:
I suggest that what today is called `Mutex` should be called a `Lock` and that the combination of a mutex and wait condition is called a `WaitLock`. They should all have the same API though. We will then end up width:

- [ ] `Lock` - Abstract base class.
- [ ] `UnsafeLock` - Dummy not performing any locking.
- [ ] `SafeLock` - Uses regular mutex.
- [ ] `GlobalLock` - Uses one global lock for all locking.
- [ ] `RecursiveLock` - Can be locked multiple times by the same thread.
- [ ] `WaitLock` - Implements a more efficient wait method.

simonmika:
Moving it all to `kean/concurrent` would be great while we are at it.

simonmika:
And start using `free(Owner Receiver)` for closures.

emilwestergren:
But `WaitLock` will be the only one with `signal`, `broadcast` and `wait` ?

simonmika:
No, but it might offer the most efficient implementation. And I'm not sure about including `signal` and `broadcast` in the api at all.

emilwestergren:
So how do you suggest the signaling of the lock is done?

simonmika:
Can you provide a use case?

emilwestergren:
Thread 1:
```ooc
while (!completed)
  waitLock wait()
```

Thread 2:
```ooc
complete()
completed = true
//Wake up thread 1 in some way...
```

[...]

PR #1447, opened 2016-03-02
Added new class WaitLock

erikhagglund:
part of #1043
@emilwestergren @marcusnaslund 

Suggestion on merging  WaitCondition and WaitLock to the same class since WaitCondition cannot exist without a mutex and WaitLock is a combination of a Waitcondition and a Mutex.

Doing this does will however limit the useage of the functionality in WaitCondition. For example in the case where we hwve multiple threads waiting for the same signal but have separate mutexes since they do not share resources with eachother. By merging the two classes we will lose the N to N mapping possibility between WaitCondition and Mutex and will ot be able to interact with WaitCondition on a lower level.

According to man for pthread_cond_wait you should not use the same mutex for multiple conditions

"The effect of using more than one mutex for concurrent pthread_cond_wait() or pthread_cond_timedwait() operations on the same condition variable is undefined; that is, a condition variable becomes bound to a unique mutex when a thread waits on the condition variable, and this (dynamic) binding ends when the wait returns."

So in reality on Unix you will have a N to 1 mapping between WaitCondition and Mutex which is what the New WaitLock class fullfills with the overloaded constructor where you can supply a Mutex

The decision was to create a new class

### DIFF
--- a/source/concurrent/SynchronizedQueue.ooc
+++ b/source/concurrent/SynchronizedQueue.ooc
@@ -48,11 +48,11 @@ SynchronizedQueue: class <T> extends Queue<T> {
 }
 
 BlockedQueue: class <T> extends SynchronizedQueue<T> {
-	_waitLock : WaitLock
+	_waitLock: WaitLock
 	_canceled := false
 	init: func {
 		super()
-		this _waitLock = WaitLock new(_mutex)
+		this _waitLock = WaitLock new(this _mutex)
 	}
 	free: override func {
 		this _waitLock free()
@@ -62,13 +62,10 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 		super(item)
 		this _waitLock wake()
 	}
-	cancel: func {
-		this _waitLock with(|| this _canceled = true)
-		this _waitLock wakeAll()
-	}
+	cancel: func { this _waitLock with(|| this _canceled = true) . wakeAll() }
 	wait: func (isOk := null as Bool*) -> T {
 		result: T = null
-		this _waitLock lockWhen(|| !this empty || this _canceled)
+		this _waitLock lockWhen(func -> Bool { !this empty || this _canceled })
 		if (this _canceled) {
 			if (isOk)
 				isOk@ = false
--- a/source/concurrent/SynchronizedQueue.ooc
+++ b/source/concurrent/SynchronizedQueue.ooc
@@ -48,34 +48,33 @@ SynchronizedQueue: class <T> extends Queue<T> {
 }
 
 BlockedQueue: class <T> extends SynchronizedQueue<T> {
-	_populated := WaitCondition new()
+	_waitLock : WaitLock
 	_canceled := false
-	init: func { super() }
+	init: func {
+		super()
+		this _waitLock = WaitLock new(_mutex)
+	}
 	free: override func {
-		this _populated free()
+		this _waitLock free()
 		super()
 	}
 	enqueue: override func (item: T) {
 		super(item)
-		this _populated signal()
+		this _waitLock wake()
 	}
 	cancel: func {
-		this _mutex lock()
-		this _canceled = true
-		this _mutex unlock()
-		this _populated broadcast()
+		this _waitLock with(|| this _canceled = true)
+		this _waitLock wakeAll()
 	}
 	wait: func (isOk := null as Bool*) -> T {
 		result: T = null
-		this _mutex lock()
-		while (this empty && !this _canceled)
-			this _populated wait(this _mutex)
+		this _waitLock lockWhen(|| !this empty || this _canceled)
 		if (this _canceled) {
 			if (isOk)
 				isOk@ = false
 		} else
 			result = this _backend dequeue(result)
-		this _mutex unlock()
+		this _waitLock unlock()
 		result
 	}
 }
--- a/source/concurrent/WaitLock.ooc
+++ b/source/concurrent/WaitLock.ooc
@@ -1,0 +1,53 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use concurrent
+
+WaitLock: class {
+	_waitCondition := WaitCondition new()
+	_mutex : Mutex
+	_ownsMutex : Bool
+	init: func {
+		this _mutex = Mutex new()
+		this _ownsMutex = true
+	}
+	init: func ~WithMutex (mutex: Mutex)
+	{
+		this _mutex = mutex
+		this _ownsMutex = false
+	}
+	free: override func {
+		if(this _ownsMutex) {
+			this _mutex free()
+			this _mutex = null
+		}
+		this _waitCondition free()
+		this _waitCondition = null
+		super()
+	}
+	lock: func { this _mutex lock() }
+	unlock: func { this _mutex unlock() }
+	lockWhen: func (condition: Func -> Bool)
+	{
+		this lock()
+		while (!condition())
+			this _waitCondition wait(this _mutex)
+		(condition as Closure) free(Owner Receiver)
+	}
+	with: func (f: Func) {
+		this _mutex with(f)
+	}
+	wake: func -> Bool
+	{
+		this _waitCondition signal()
+	}
+	wakeAll: func -> Bool
+	{
+		this _waitCondition broadcast()
+	}
+}
--- a/source/concurrent/WaitLock.ooc
+++ b/source/concurrent/WaitLock.ooc
@@ -10,19 +10,14 @@ use concurrent
 
 WaitLock: class {
 	_waitCondition := WaitCondition new()
-	_mutex : Mutex
-	_ownsMutex : Bool
-	init: func {
-		this _mutex = Mutex new()
-		this _ownsMutex = true
-	}
-	init: func ~WithMutex (mutex: Mutex)
-	{
-		this _mutex = mutex
-		this _ownsMutex = false
+	_mutex: Mutex
+	_ownsMutex: Bool
+	init: func (mutex: Mutex = null) {
+		this _ownsMutex = mutex == null
+		this _mutex = this _ownsMutex ? Mutex new() : mutex
 	}
 	free: override func {
-		if(this _ownsMutex) {
+		if (this _ownsMutex) {
 			this _mutex free()
 			this _mutex = null
 		}
@@ -32,22 +27,13 @@ WaitLock: class {
 	}
 	lock: func { this _mutex lock() }
 	unlock: func { this _mutex unlock() }
-	lockWhen: func (condition: Func -> Bool)
-	{
+	lockWhen: func (condition: Func -> Bool) {
 		this lock()
 		while (!condition())
 			this _waitCondition wait(this _mutex)
 		(condition as Closure) free(Owner Receiver)
 	}
-	with: func (f: Func) {
-		this _mutex with(f)
-	}
-	wake: func -> Bool
-	{
-		this _waitCondition signal()
-	}
-	wakeAll: func -> Bool
-	{
-		this _waitCondition broadcast()
-	}
+	with: func (f: Func) { this _mutex with(f) }
+	wake: func -> Bool { this _waitCondition signal() }
+	wakeAll: func -> Bool { this _waitCondition broadcast() }
 }
--- a/test/concurrent/WaitLockTest.ooc
+++ b/test/concurrent/WaitLockTest.ooc
@@ -1,0 +1,113 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use concurrent
+use unit
+
+WaitLockTest: class extends Fixture {
+	init: func {
+		super("WithLock")
+		this add("_testWithMutexOwnership", This _testWithMutexOwnership)
+		this add("_testWithoutMutexOwnership", This _testWithoutMutexOwnership)
+		this add("_testWakeWithFailingCondition", This _testWakeWithFailingCondition)
+		this add("_testWakeWithPassingCondition", This _testWakeWithPassingCondition)
+	}
+	_testWithMutexOwnership: static func {
+		waitLock := WaitLock new()
+		waitLock free()
+	}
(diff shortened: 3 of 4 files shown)


source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,53 @@
...
+WaitLock: class {
+	_waitCondition := WaitCondition new()
+	_mutex : Mutex
+	_ownsMutex : Bool
+	init: func {
+		this _mutex = Mutex new()
+		this _ownsMutex = true
+	}
+	init: func ~WithMutex (mutex: Mutex)
+	{
```
marcusnaslund: This bracket belongs on the line above

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,53 @@
...
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use concurrent
+
+WaitLock: class {
+	_waitCondition := WaitCondition new()
+	_mutex : Mutex
+	_ownsMutex : Bool
```
marcusnaslund: No spaces before `:`

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,53 @@
...
+		this _mutex = Mutex new()
+		this _ownsMutex = true
+	}
+	init: func ~WithMutex (mutex: Mutex)
+	{
+		this _mutex = mutex
+		this _ownsMutex = false
+	}
+	free: override func {
+		if(this _ownsMutex) {
```
marcusnaslund: space between `if` and `(`

test/concurrent/WaitLockTest.ooc
```diff
@@ -0,0 +1,113 @@
...
+		waitLock free()
+		mutex free()
+	}
+	_testWakeWithFailingCondition: static func {
+		timesTriggered := 0
+		timesTriggeredRef := timesTriggered&
+		waitLock := WaitLock new()
+		waitingThread := Thread new(||
+			waitLock lockWhen(||
+				timesTriggeredRef@ = timesTriggeredRef@ + 1
```
marcusnaslund: Why is this done by reference? Am I missing something?

source/concurrent/SynchronizedQueue.ooc
```diff
@@ -48,34 +48,33 @@ SynchronizedQueue: class <T> extends Queue<T> {
...
-		this _populated broadcast()
+		this _waitLock with(|| this _canceled = true)
+		this _waitLock wakeAll()
 	}
 	wait: func (isOk := null as Bool*) -> T {
 		result: T = null
-		this _mutex lock()
-		while (this empty && !this _canceled)
-			this _populated wait(this _mutex)
+		this _waitLock lockWhen(|| !this empty || this _canceled)
```
marcusnaslund: The first `||` operator thus means something other than the second, that doesn't look good.

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,53 @@
...
+	_ownsMutex : Bool
+	init: func {
+		this _mutex = Mutex new()
+		this _ownsMutex = true
+	}
+	init: func ~WithMutex (mutex: Mutex)
+	{
+		this _mutex = mutex
+		this _ownsMutex = false
+	}
```
marcusnaslund: You could have `_ownsMutex` to default to `false` and do something like
```ooc
init: func (mutex: Mutex = null) {
    if (mutex)
        this _mutex = mutex
    else {
        this _mutex = Mutex new()
        this _ownsMutex = true
    }
}
```
to avoid the need for two different constructors

marcusnaslund:
I gave some quick comments before this went into review. Normally I wouldn't until it reaches final review, but since you're new, I thought I'd give you some heads up.

source/concurrent/WaitLock.ooc
marcusnaslund: Same on line 36, 46, 50

test/concurrent/WaitLockTest.ooc
shamanas: Probably some confusion regarding closure captures?  
rock should be able to mark and capture `timesTriggered` by reference if the line is changed to `timesTriggered = timesTriggered + 1`  

@erikhagglund If rock was not able to do that, it should be reported as a bug, otherwise just using the original variable is probably clearer.

source/concurrent/SynchronizedQueue.ooc
```diff
@@ -48,34 +48,33 @@ SynchronizedQueue: class <T> extends Queue<T> {
 }
 
 BlockedQueue: class <T> extends SynchronizedQueue<T> {
-	_populated := WaitCondition new()
+	_waitLock : WaitLock
 	_canceled := false
-	init: func { super() }
+	init: func {
+		super()
+		this _waitLock = WaitLock new(_mutex)
```
emilwestergren: `this _mutex`

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,53 @@
...
+	unlock: func { this _mutex unlock() }
+	lockWhen: func (condition: Func -> Bool)
+	{
+		this lock()
+		while (!condition())
+			this _waitCondition wait(this _mutex)
+		(condition as Closure) free(Owner Receiver)
+	}
+	with: func (f: Func) {
+		this _mutex with(f)
```
emilwestergren: Is this closure freed somewhere?

test/concurrent/WaitLockTest.ooc
erikhagglund: It has to do with the closure variables not being passed by reference (could be due to it being a local variable?), therefore if the variable timesTriggered is updated the closure will not know about it.

source/concurrent/SynchronizedQueue.ooc
erikhagglund: yes...

test/concurrent/WaitLockTest.ooc
shamanas: It should be marked by reference automatically, I will look into it.

source/concurrent/WaitLock.ooc
erikhagglund: Not in Mutex as it seems..

Mutex: abstract class {
	lock: abstract func
	unlock: abstract func

	with: func (f: Func) {
		this lock()
		f()
		this unlock()
	}

erikhagglund: Seems like this is the only placed it is used so adding handling for it in Mutex

marcusnaslund: That crashes the Mutex tests, though, since they're not expecting it.

erikhagglund: yes, I noticed :), will update Mutex in a separate pull request first

marcusnaslund: > will update Mutex in a separate pull request first

Excellent.

marcusnaslund: That PR is now merged, so time to rebase this PR.

test/concurrent/WaitLockTest.ooc
marcusnaslund: Nice to see you around here @shamanas :)

source/concurrent/SynchronizedQueue.ooc
marcusnaslund: So, I would suggest something like
```ooc
populatedOrCanceled := func -> Bool { !this empty || this _canceled }
this _waitLock lockWhen(populatedOrCanceled)
```
to avoid the double `||`.

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,50 @@
...
+	}
+	wake: func -> Bool {
+		this _waitCondition signal()
+	}
+	wakeAll: func -> Bool {
+		this _waitCondition broadcast()
+	}
```
marcusnaslund: The three methods above could all be written on one line each.

source/concurrent/SynchronizedQueue.ooc
```diff
@@ -48,34 +48,32 @@ SynchronizedQueue: class <T> extends Queue<T> {
...
 		super(item)
-		this _populated signal()
+		this _waitLock wake()
 	}
 	cancel: func {
-		this _mutex lock()
-		this _canceled = true
-		this _mutex unlock()
-		this _populated broadcast()
+		this _waitLock with(|| this _canceled = true). wakeAll()
```
emilwestergren: Space before `.`

marcusnaslund:
You're not the first person to do it, but try to avoid commit messages like "Changes according to review comments". It means nothing if one has to review the git log later. "Style fixes" is OK if it's just style fixes, otherwise try to be a bit more descriptive. Of course, don't overdo it either. :)

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+use concurrent
+
+WaitLock: class {
+	_waitCondition := WaitCondition new()
+	_mutex: Mutex
+	_ownsMutex := false
+	init: func (mutex: Mutex = null) {
+		if (mutex == null) {
```
emilwestergren: This can be written as:
```ooc
this _ownsMutex = mutex == null
this _mutex = this _ownsMutex ? Mutex new() : mutex
```

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,50 @@
...
+		this lock()
+		while (!condition())
+			this _waitCondition wait(this _mutex)
+		(condition as Closure) free(Owner Receiver)
+	}
+	with: func (f: Func) {
+		this _mutex with(f)
+	}
+	wake: func -> Bool {
+		this _waitCondition signal()
```
emilwestergren: Put this on a single line

source/concurrent/WaitLock.ooc
```diff
@@ -0,0 +1,50 @@
...
+	}
+	wakeAll: func -> Bool {
+		this _waitCondition broadcast()
```
emilwestergren: Put this on a single line

source/concurrent/WaitLock.ooc
marcusnaslund: Right. I think I suggested the current way. Good catch.

source/concurrent/SynchronizedQueue.ooc
emilwestergren: I'm fine with the first way, you have the same problem with things like `x& & y` anyway

emilwestergren: But @marcusnaslund has the final say :)

marcusnaslund: Have we ever written something like `x& & y`?

emilwestergren: Yes, every day ;)

marcusnaslund: :)

marcusnaslund: Anyway, as @emilwestergren said, I have unlimited and awesome power in style issues, and I say we go with what I suggested. :)
Unless 
`this _waitLock lockWhen(func -> Bool { !this empty || this _canceled })` 
works, but I don't think it does? Maybe 
`this _waitLock lockWhen((f := func -> Bool { !this empty || this _canceled }))`
?

erikhagglund: both those suggestions work

marcusnaslund: The shorter one is better, then.

marcusnaslund:
@emilwestergren added final review and removed final review 4 hours ago. Was it an accidental add or accidental removal?

emilwestergren:
Accidental add.

marcusnaslund:
Alright. All comments appear to be fixed, so please comment whatever is holding this back.

test/concurrent/WaitLockTest.ooc
```diff
@@ -0,0 +1,111 @@
...
+		waitLock free()
+	}
+	_testWithoutMutexOwnership: static func {
+		mutex := Mutex new()
+		waitLock := WaitLock new(mutex)
+		waitLock free()
+		mutex free()
+	}
+	_testWakeWithFailingCondition: static func {
+		timesTriggered := 0
```
marcusnaslund: I tried adding `timesTriggered: static Int = 0` to the top of the class and then changing all `timesTriggeredRef@` to `This timesTriggered` and it worked. Please do that.

test/concurrent/WaitLockTest.ooc
```diff
@@ -0,0 +1,111 @@
...
+		)
+		testThread := Thread new(||
+			while (true) {
+				waitLock lock()
+				if (timesTriggeredRef@ >= 1)
+					break
+				waitLock unlock()
+				Thread yield()
+			}
+			expect(timesTriggeredRef@ == 1)
```
marcusnaslund: And then these should be `expect(This timesTriggered, is equal to(1))` etc.

test/concurrent/WaitLockTest.ooc
```diff
@@ -0,0 +1,111 @@
...
+			expect(timesTriggeredRef@ == 2)
+			waitLock unlock()
+			while (!waitingThread wait(0.05))
+				Thread yield()
+		)
+		testThread start()
+		waitingThread start()
+		expect(testThread wait(0.1))
+		testThread free()
+		expect(!waitingThread alive())
```
marcusnaslund: `expect(waitingThread alive(), is false)` for better output in case something fails

test/concurrent/WaitLockTest.ooc
```diff
@@ -0,0 +1,111 @@
...
+		this add("_testWakeWithPassingCondition", This _testWakeWithPassingCondition)
+	}
+	_testWithMutexOwnership: static func {
+		waitLock := WaitLock new()
+		waitLock free()
+	}
+	_testWithoutMutexOwnership: static func {
+		mutex := Mutex new()
+		waitLock := WaitLock new(mutex)
+		waitLock free()
```
marcusnaslund: You could throw in an `expect(mutex, is notNull)` here.

marcusnaslund:
Looks pretty good, the tests just need some small polishing.

test/concurrent/WaitLockTest.ooc
```diff
@@ -0,0 +1,111 @@
...
+		expect(mutex, is notNull)
+		mutex free()
+	}
+	timesTriggered: static Int = 0
```
marcusnaslund: This belongs at the top of the class. Look for example at `test/base/SynchronizedTest`

marcusnaslund:
Nothing left to say except... Welcome to the "My PR has been merged to upstream" club!

PR #1452, opened 2016-03-02
Major overhaul of the unit test framework

### LINKED ISSUE #539: Fixture has problems with is not equal, is not not true, etc.
This seems to be an old problem, but why it happens remains a mystery. The following expect calls do not work:
```ooc
this add("true is not equal to(false)", func { expect(true, is not equal to(false)) })
this add("true is not not true", func { expect(true, is not not true) })
```
But the following does work:
```ooc
notEqualToFalse := is not equal to(false)
this add("true is not not true", func { expect(true, notEqualToFalse) })
notNotTrue := is not not false
this add("false is not not false", func { expect(false, notNotTrue) })
```
A big difference is that, in the latter case, `notEqualToFalse` and `notNotTrue` are resolved before the current test class is even initialized. Otherwise, they are resolved in the `add` call. The error message is as follows:
```error call expect(true, Fixture __getis__() __getnot__() __getnot__() __gettrue__()) waiting on static Fixture expect(value: Object, constraint: Constraint) to resolve.```

### PR DESCRIPTION
**Pull request by marcusnaslund**
First step towards #539 - address headaches associated with our unit tests. In summary:

- `Contraints.ooc` was split up into two files (the new one is `Modifiers.ooc`)
- *Serious* cleanup of code (why have we lived with this mess for so long?)
- Removed the `EmptyModifier` (it only worked with `String` and was never used anyway)
- Removed the `NotModifier` and instead added `notEqual` and `notNull` keywords
- Renamed `Modifier` to `ExpectModifier` (because the name clashed badly with `Modifier` in another project which rock couldn't handle when they were moved to another file.)

A subsequent PR will add `lessThanOrEqual` and `greaterThanOrEqual` keywords, as well as utilize the `equals`/`lessThan`/... methods defined on our numeric floating-point types, which will hopefully seal my last problems with `Fixture`.

Basically, this PR will allow us to do

```ooc
expect(0, is equal to(0))
expect(1, is notEqual to(0))
expect(0.f, is equal to(0.f) within(tolerance))
expect(1.f, is notEqual to(0.f) within(tolerance))
```

without weird bugs. The only drawback is we won't be able to do stuff like `is not not true`, but I don't really see a universe where you would want to. Also, it didn't really work as expected either, so...


source/unit/Constraints.ooc
```diff
@@ -7,261 +7,66 @@
...
-		CompareConstraint new(this, Cell<ULong> new(right), f, ComparisonType GreaterThan)
-	}
-	than: func ~llong (right: LLong) -> CompareConstraint {
-		f := func (value, c: Object) -> Bool { (value as Cell<LLong> get()) > (c as Cell<LLong> get()) }
-		CompareConstraint new(this, Cell<LLong> new(right), f, ComparisonType LessThan)
-	}
-	than: func ~ullong (right: ULLong) -> CompareConstraint {
-		f := func (value, c: Object) -> Bool { (value as Cell<ULLong> get()) > (c as Cell<ULLong> get()) }
-		CompareConstraint new(this, Cell<ULLong> new(right), f, ComparisonType LessThan)
-	}
```
marcusnaslund: All of this was moved to `Modifiers.ooc`.

source/unit/Constraints.ooc
```diff
@@ -272,46 +77,36 @@ CompareConstraint: class extends Constraint {
...
-	true ::= TrueConstraint new(this)
-	false ::= FalseConstraint new(this)
-	Null ::= NullConstraint new(this)
-	empty ::= EmptyConstraint new(this)
-	not ::= NotModifier new(this)
-	equal ::= EqualModifier new(this)
-	less ::= LessModifier new(this)
-	greater ::= GreaterModifier new(this)
-	init: func
-}
```
marcusnaslund: Moved to the top of the file

marcusnaslund:
Peer review @thomasfanell ? :)

thomasfanell:
I'll review on an hour or so

marcusnaslund:
Or @sebastianbaginski , whoever comes first. :)

thomasfanell:
Very nice indeed, good job

PR #1456, opened 2016-03-03
Fixed memory leak in "Mutex With"

erikhagglund:
Memory leak found in review for "Added new class WaitLock #1447"
@marcusnaslund @emilwestergren


marcusnaslund:
Very good.

PR #1460, opened 2016-03-03
Fixed OpenGLContext constructors.

emilwestergren:
Null check for shared context is now done in OpenGLContext. @sebastianbaginski peer review


sebastianbaginski:
Nice.

PR #1462, opened 2016-03-03
More cleanup and bug fix in Fixture

### LINKED ISSUE #539: Fixture has problems with is not equal, is not not true, etc.

marcusnaslund:
Fixes #539 

- [x] Adds nice error messages for when `is null`, `is notNull`, `is true`, `is false` are used
- [x] Even more cleanup of code
- [x] Now uses `equals()` instead of `==` by default when comparing floating point types
- [x] Implemented `greaterOrEqual` and `lessOrEqual` keywords

### DIFF
--- a/source/unit/Constraints.ooc
+++ b/source/unit/Constraints.ooc
@@ -14,9 +14,9 @@ ComparisonType: enum {
 	Equal
 	NotEqual
 	LessThan
-	LessThanOrEqual // TODO: Implement
+	LessOrEqual
 	GreaterThan
-	GreaterThanOrEqual // TODO: Implement
+	GreaterOrEqual
 	Within
 	NotWithin
 }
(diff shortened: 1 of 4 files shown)


marcusnaslund:
Peer review @thomasfanell ?

PR #1463, opened 2016-03-04
fixes for ThreadID on windows

sebastianbaginski:
Fixes problems with Thread and some tests in concurrent on windows.
@marcusnaslund

### DIFF
--- a/source/concurrent/Thread.ooc
+++ b/source/concurrent/Thread.ooc
@@ -40,7 +40,7 @@ Thread: abstract class {
 		version (unix || apple)
 			result = pthread_self() as ThreadId
 		version (windows)
-			result = GetCurrentThread() as ThreadId
+			result = GetCurrentThreadId() as ThreadId
 		result
 	}
 	yield: static func -> Bool {
--- a/test/concurrent/WorkerThreadTest.ooc
+++ b/test/concurrent/WorkerThreadTest.ooc
@@ -48,8 +48,8 @@ WorkerThreadTest: class extends Fixture {
 		this add("Correct threading", func {
 			mainThread := Thread currentThreadId()
 			for (i in 0 .. This threads count - 1) {
-				expect(mainThread != This threads[i])
-				expect(This threads[i], is equal to(This threads[i + 1]))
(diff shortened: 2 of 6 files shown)


source/concurrent/Thread.ooc
```diff
@@ -64,9 +64,9 @@ version (unix || apple) {
 }
 
 version (windows) {
-	include windows
+	include inttypes
 
-	ThreadId: cover from DWORD {
+	ThreadId: cover from UInt {
```
sebastianbaginski: According to windows docs, DWORD is 32 bit type. For some reasons it was typedef'ed to UInt64, so I needed to change it to UInt.

marcusnaslund: You should probably change `DWORD: extern cover from Long` in `system/os/win32.ooc` instead.

sebastianbaginski: Ah, so there it is:)

sebastianbaginski:
Updated. Removed redundant DWORD typedef in Exception.

marcusnaslund:
So you can define the same thing (differently) twice in global scope and rock goes "OK"...?

marcusnaslund:
Ah, you need to either explicitly cast to `Int` (should work) or change to use `==` in the tests.

test/concurrent/WorkerThreadTest.ooc
```diff
@@ -48,8 +48,8 @@ WorkerThreadTest: class extends Fixture {
 		this add("Correct threading", func {
 			mainThread := Thread currentThreadId()
 			for (i in 0 .. This threads count - 1) {
-				expect(mainThread != This threads[i])
-				expect(This threads[i], is equal to(This threads[i + 1]))
+				expect(mainThread equals(This threads[i]) == false)
+				expect(This threads[i] equals(This threads[i + 1]))
```
sebastianbaginski: @marcusnaslund you meant this one ? I could not see the PRInt details correctly on windows

marcusnaslund: I can't see the details either since it restarted, but yes, it seems to run fine now. :)

PR #1470, opened 2016-03-04
Context bugfix

emilwestergren:
Even though it **should** be legit for several contexts to call `eglTerminate(display)`, this causes a seg fault on some drivers. I solved this by counting the number of created contexts and only terminating when the last context is freed. Not a beautiful solution but it does the job.

@marcusnaslund peer review

### DIFF
--- a/source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
+++ b/source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
@@ -18,6 +18,7 @@ Gles3Context: class extends GLContext {
 	_eglContext: Pointer
 	_eglSurface: Pointer
 	_contextCount := static 0
+	_mutex := static Mutex new()
 	init: func
 	free: override func {
 		status := eglMakeCurrent(this _eglDisplay, null, null, null)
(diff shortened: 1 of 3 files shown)


marcusnaslund:
Do you really have to do the intermediate store to `result` four times?

emilwestergren:
What?

marcusnaslund:
Sorry, I mean `status`.

marcusnaslund:
Oh yes you do. Nevermind. It's friday again.

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -17,13 +17,24 @@ import Gles3Debug, Gles3Fence, Gles3FramebufferObject, Gles3Quad, Gles3Renderer,
...
+			Debug print("eglDestroyContext failed with error code %d" format(status))
+		status = eglDestroySurface(this _eglDisplay, this _eglSurface)
+		if (status != EGL_TRUE)
+			Debug print("eglDestroySurface failed with error code %d" format(status))
+		if (This _contextCount == 1) {
+			status = eglTerminate(this _eglDisplay)
+			if (status != EGL_TRUE)
+				Debug print("eglTerminate failed with error code %d" format(status))
+		}
+		This _contextCount -= 1
```
emilwestergren: This actually needs a mutex lock.

PR #1471, opened 2016-03-04
Updated Raster classes to throw an exception when 'StbImage load' fai…

### LINKED ISSUE #1468: When 'StbImage load' fails to load an image it returns Null which is left unhandled
The suggestion is to raise an exception if the Image failed to load.

erikhagglund:
…ls to open the image

Fixes #1468
@emilwestergren

### DIFF
--- a/source/draw/RasterImage.ooc
+++ b/source/draw/RasterImage.ooc
@@ -57,19 +57,15 @@ RasterImage: abstract class extends Image {
 		pathString free()
 	}
 	open: static func ~unknownType (filename: String) -> This {
-		x, y, imageComponents: Int
-		data := StbImage load(filename, x&, y&, imageComponents&, 0)
-		if (data == null)
-			Exception new("Failed to load image: " + filename) throw()
 		result: This
-		buffer := ByteBuffer new(data as Byte*, x * y * imageComponents, true)
+		(buffer, size, imageComponents) := StbImage load(filename)
 		match (imageComponents) {
 			case 1 =>
-				result = RasterMonochrome new(buffer, IntVector2D new(x, y))
+				result = RasterMonochrome new(buffer, size)
 			case 3 =>
-				result = RasterRgb new(buffer, IntVector2D new(x, y))
+				result = RasterRgb new(buffer, size)
 			case 4 =>
-				result = RasterRgba new(buffer, IntVector2D new(x, y))
+				result = RasterRgba new(buffer, size)
 			case =>
 				buffer free()
 				raise("Unsupported number of channels in image")
--- a/source/draw/RasterImage.ooc
+++ b/source/draw/RasterImage.ooc
@@ -59,6 +59,8 @@ RasterImage: abstract class extends Image {
 	open: static func ~unknownType (filename: String) -> This {
 		x, y, imageComponents: Int
 		data := StbImage load(filename, x&, y&, imageComponents&, 0)
+		if (data == null)
+			Exception new("Failed to load image: " + filename) throw()
 		result: This
 		buffer := ByteBuffer new(data as Byte*, x * y * imageComponents, true)
 		match (imageComponents) {
--- a/source/draw/RasterImage.ooc
+++ b/source/draw/RasterImage.ooc
@@ -66,6 +66,7 @@ RasterImage: abstract class extends Image {
 				result = RasterRgb new(buffer, size)
 			case 4 =>
 				result = RasterRgba new(buffer, size)
+				(result as RasterRgba) swapRedBlue()
 			case =>
 				buffer free()
 				raise("Unsupported number of channels in image")
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -195,13 +195,9 @@ RasterMonochrome: class extends RasterPacked {
 	}
 
 	open: static func (filename: String) -> This {
-		x, y, imageComponents: Int
 		requiredComponents := 1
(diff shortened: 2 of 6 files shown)


source/draw/RasterImage.ooc
```diff
@@ -59,6 +59,9 @@ RasterImage: abstract class extends Image {
 	open: static func ~unknownType (filename: String) -> This {
 		x, y, imageComponents: Int
 		data := StbImage load(filename, x&, y&, imageComponents&, 0)
+		if (data == null) {
```
emilwestergren: You don't need brackets here.

source/draw/RasterMonochrome.ooc
```diff
@@ -198,6 +198,9 @@ RasterMonochrome: class extends RasterPacked {
 		x, y, imageComponents: Int
 		requiredComponents := 1
 		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
+		if (data == null) {
```
emilwestergren: Same here

source/draw/RasterRgba.ooc
```diff
@@ -153,6 +153,9 @@ RasterRgba: class extends RasterPacked {
 		x, y, imageComponents: Int
 		requiredComponents := 4
 		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
+		if (data == null) {
```
emilwestergren: Same here

emilwestergren:
Looks good.

test/draw/RasterRgbaTest.ooc
```diff
@@ -16,6 +16,15 @@ RasterRgbaTest: class extends Fixture {
 	sourceFlower := "test/draw/input/Flower.png"
 	init: func {
 		super("RasterRgbaTest")
+		this add("Test file not found", func {
+			try {
+				image := RasterRgba open("nonExistingFile")
+				expect(false)
```
marcusnaslund: Why is this line here?

source/draw/RasterRgba.ooc
```diff
@@ -153,6 +153,8 @@ RasterRgba: class extends RasterPacked {
 		x, y, imageComponents: Int
 		requiredComponents := 4
 		data := StbImage load(filename, x&, y&, imageComponents&, requiredComponents)
+		if (data == null)
+			Exception new("Failed to load image: " + filename) throw()
```
marcusnaslund: `raise("Failed to load image: " + filename)` is cleaner

sebastianbaginski: We could also throw directly from `StbImage load` and avoid some duplicated code.

test/draw/RasterRgbaTest.ooc
```diff
@@ -16,6 +16,15 @@ RasterRgbaTest: class extends Fixture {
...
+				expect(image, is equal to (null as RasterRgba))
+			} catch (e: Exception) {
+				expect(e message contains("Failed to load image: "))
```
sebastianbaginski: I'd rather avoid testing for the specific error message, I think it would be enough just to expect an exception.

source/draw/RasterRgba.ooc
marcusnaslund: Very good, let's do that.

marcusnaslund:
Put on hold until comments are resolved. You may add the `final review` flag again once that's done.

test/draw/RasterRgbaTest.ooc
erikhagglund: I agree that testing for a specific string isn't optimal, but to test that an exception is triggered will trigger on any type of exception and not necessary due to the failure we want to test. Ideally I guess you would want a separate exception type and trigger on that

test/draw/RasterRgbaTest.ooc
erikhagglund: I guess a comment would be good to add here, that line is there since that code should be unreachable since we are expecting an exception to trigger on the line above. So fi that row is executed the test should fail. (The row below is due to that the compiler says the variable Image is "unused" and I'm not sure to what level the compiler optimizes. I don't want the critical part of the test to be optimized away)

test/draw/RasterRgbaTest.ooc
sebastianbaginski: OK, we already have some Exception subtypes, maybe we can update it later.

test/draw/RasterRgbaTest.ooc
sebastianbaginski: >  I don't want the critical part of the test to be optimized away

I don't think gcc will completely remove a call to external function with possible side effects even with -O3, but I think you are right, we should remove this warning.

marcusnaslund: Couldn't you just declare the `image` before the `try-block` and define it as `null`?

marcusnaslund: Now I understand, thanks. The above comment applies to the line below.

test/draw/RasterRgbaTest.ooc
```diff
@@ -16,6 +16,15 @@ RasterRgbaTest: class extends Fixture {
 	sourceFlower := "test/draw/input/Flower.png"
 	init: func {
 		super("RasterRgbaTest")
+		this add("Test file not found", func {
+			try {
+				image := RasterRgba open("nonExistingFile")
+				expect(false) // Unreachable code since we are expecting an exception
+				expect(image, is equal to (null as RasterRgba))
```
marcusnaslund: `expect(image, is Null)` works now.

PR #1472, opened 2016-03-06
Added GlobalCleanup class

marcusnaslund:
An upcoming PR will make sure this class is actually used.


sebastianbaginski:
Looks OK, that was the idea.

PR #1475, opened 2016-03-06
Added tests for LDouble, LLong and ULLong

marcusnaslund:
We should have had these ever since we redefined the numeric types, but better late than never. This is also useful to make sure the unit test framework works with all types.

### DIFF
--- a/test/unit/LLongTest.ooc
+++ b/test/unit/LLongTest.ooc
@@ -1,0 +1,25 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use unit
+
+LLongTest: class extends Fixture {
+	init: func {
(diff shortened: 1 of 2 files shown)


test/unit/LLongTest.ooc
```diff
@@ -0,0 +1,25 @@
...
+
+use unit
+
+LLongTest: class extends Fixture {
+	init: func {
+		super("LLong, ULLong")
+
+		this add("1_000_000_000 is equal to 1_000_000_000", func { expect(1_000_000_000 as LLong, is equal to(1_000_000_000 as LLong)) })
+		this add("-1_000_000_001 is less than 1_000_000_000", func { expect(-1_000_000_001 as LLong, is less than(-1_000_000_000 as LLong)) })
+		this add("100_000_000_000 is greater than 99_999_999_999", func { expect(100_000_000_000 as LLong, is greater than(99_999_999_999 as LLong)) })
```
sebastianbaginski: Does it work with the `LL` / `ULL` suffixes ?

marcusnaslund: Yes, fixed.

PR #1476, opened 2016-03-06
Made one Vector::copy abstract and removed StackVector

marcusnaslund:
This method is never used, and also make little sense (calling `copy()` on a `StackVector` would give you a `HeapVector`). If anything, we could make it abstract and have the actual implementations in each derived class. Comments @sebastianbaginski ?


sebastianbaginski:
>  we could make it abstract and have the actual implementations in each derived class

This :+1:

marcusnaslund:
I don't really understand this `StackVector` - Why do we have it? We never use it. It seems to me like it's just like `HeapVector` except the constructor takes a backend that's somehow stack-allocated. How can I be sure to allocate on the stack for the copy?

sebastianbaginski:
> How can I be sure to allocate on the stack for the copy?

Stack memory can be allocated with `alloca`, but the memory is freed when the calling function returns, so I think it is not suitable for data structures.

marcusnaslund:
@fredrikbryntesson You appear to have written the `StackVector` class some 15 months ago. :) Can you shed some light on the intended use case?

fredrikbryntesson:
I think it was something that @niklasborwell needed, I do not remember what it was. We never used it.

marcusnaslund:
I feel like we can remove it, then?

fredrikbryntesson:
Kill it, with fire.

niklasborwell:
I don't remember any of this. If it's not used, bin it!

marcusnaslund:
Thank you all. 

marcusnaslund:
Killed. With considerable amounts of fire.

PR #1477, opened 2016-03-06
Added some missing unit tests

marcusnaslund:


marcusnaslund:
Peer review @thomasfanell ?

test/system/SystemTest.ooc
```diff
@@ -0,0 +1,25 @@
...
+use unit
+
+SystemTest: class extends Fixture {
+	init: func {
+		super("System")
+		this add("numProcessors, hostname", func {
+			processors := System numProcessors()
+			hostname := System hostname()
+
+			expect(processors, is greaterOrEqual than(1))
```
marcusnaslund: I guess we can't do better than this

marcusnaslund:
Taken off review stage for a while, I want to add some more.

source/system/math.ooc
```diff
@@ -35,8 +35,6 @@ extend ULong {
 }
 
 extend Int {
-	pi ::= static 3
-	e ::= static 2
```
marcusnaslund: I'm pretty sure I removed these before, but here we go.

marcusnaslund:
There we go.

PR #1481, opened 2016-03-07
Removed PointerVector and PointerVectorList

### LINKED ISSUE #1474: PointerVector and PointerVectorList
These classes lack tests, but are also not used anywhere. Are we going to use them? Have we ever used them? Should they be removed?

@sebastianbaginski @emilwestergren @fredrikbryntesson ?

marcusnaslund:
These were never used. Fixes #1474 


marcusnaslund:
Moving this to final review as @fredrikbryntesson gave his OK in #1474

PR #1487, opened 2016-03-07
Using 'is Null' and 'is notNull' in tests

marcusnaslund:
The advantage is you get output like `expected null, was 0x153ac56f` or `expected not null, was null`.

Peer review @jonathanudd ?


jonathanudd:
Looks good

PR #1488, opened 2016-03-07
Added functionality to GlobalCleanup

### LINKED ISSUE #1421: Global free~all function
We should have a single funciton that collects all cleanup code from kean

marcusnaslund:
Fixes #1421


source/system/String.ooc
```diff
@@ -311,8 +311,9 @@ strArrayFromCString: func ~hack (argc: Int, argv: String*) -> String[] {
...
 cStringPtrToStringPtr: func (cstr: CString*, len: Int) -> String* {
 	toRet: String* = calloc(len, Pointer size) // otherwise the pointers are stack-allocated
-	for (i in 0 .. len) {
+	for (i in 0 .. len)
 		toRet[i] = makeStringLiteral(cstr[i], cstr[i] length())
-	}
 	toRet
 }
+
+//GlobalCleanup register(|| String free~all())
```
marcusnaslund: @sebastianbaginski `String free~all()` seems to always crash with memory errors no matter where/how I call it, including like this.

sebastianbaginski: But it is actually called now on upstream, at the end of `main()`. I can download your branches and check it out today if you want.

marcusnaslund: Yep, you're rightm but I can't get it to work in this repo alone.

sebastianbaginski: It crashes because something in Fixture is trying to use string literals after the cleanup function is called.
Comment out the content of `printFailures` and it will not crash on this branch.
For me it crashed in this line 
```
This _print(t"Total time: %.2f s\n" format(This totalTime))
```

marcusnaslund: Oh yeah, we're doing some printouts registered on `atexit`... I'll have to circumvent that somehow.

test/base/ProfilerTest.ooc
```diff
@@ -34,7 +34,6 @@ ProfilerTest: class extends Fixture {
 			profiler start() . stop()
 			profilerToFree := Profiler new("to free")
 			profilerToFree free()
-			Profiler free~all()
```
marcusnaslund: Will be handled by `GlobalCleanup`

marcusnaslund:
This seems to work fine with the other project. I will incorporate this with the test framework in the next PR (there are some problems with the freeing of `String` and/or `ByteBuffer` because we have other things registered on the exit handler). Shouldn't be too difficult to fix.

sebastianbaginski:
Looks good

PR #1492, opened 2016-03-08
Updated VS Code settings

thomasfanell:
Always terminate lines with LF


thomasfanell:
@marcusnaslund dare to review?

Issue #1409, opened 2016-02-18
ThreadPool test crashes randomly


sebastianbaginski:
As reported here https://github.com/magic-lang/ooc-kean/pull/1394
This needs to be verified as a bug and solved or confirmed as unrelated to our code (related to PRInt workload ?). Someone needs to work on this anyway.

davidpiuva:
The synchronized resource test also crashes randomly.

marcusnaslund:
The `SynchronizedResourceTest` fails a lot less now that we're not creating 8 threads at once and not running 4 jobs in parallell on print (also, now we're running on a workstation with better performance and I haven't yet seen the problem since I got back with either test). So it seems to be more about PRInt than our code, but I'm not sure how to verify it.

davidpiuva:
It sounds like a race condition to me if more performance makes it go away.

marcusnaslund:
I just stress tested the concurrent tests a lot (400 times, some with `-x`, running four at once, including `stress` on cpu+memory+io+hdd).

- **0** problems with `ThreadPool` whatsoever :/
- **3** problems with `SynchronizedResource` and `SynchronizedQueue`, they always happened together and the backtrace looked like this: 
```
11:31:08 SynchronizedResource .f failed
  -> multiple threads : expected not null was null
11:31:08 SynchronizedQueue (SIGSEGV) segmentation fault
.[fancy backtrace]
0     BacktraceHandler backtrace()     in Backtrace  (at /home/mnaslund/versioned/ooc/kean2/source/system/Backtrace.ooc:26) 
1     Exception getCurrentBacktrace()  in Exception  (at /home/mnaslund/versioned/ooc/kean2/source/system/Exception.ooc:173)
2     _signalHandler()                 in Exception  (at /home/mnaslund/versioned/ooc/kean2/source/system/Exception.ooc:237)
3     killpg()                         in            (at (null):0)                                                          
.Total time: 12.00 s
Failed tests: 1 [SynchronizedResource]
```
- **2** problems with the new `WaitLock`, but that's easily fixed.

sebastianbaginski:
That's great news, keep it up.
I'm a bit surprised that `ThreadPool` didn't fail, I remember it failing a lot more than `SynchronizedQueue`.

marcusnaslund:
I'm adding a script to help us do this more often and easily. #1524

marcusnaslund:
Removing all `pool free()` in the test fixes this issue...

[...]

PR #1495, opened 2016-03-08
Changed wait timeouts in WaitLock test

marcusnaslund:
As reported in #1409 the `WaitLock` tests failed sometimes during stress testing, as `0.1` seconds was a bit too conservative under heavy load. This should fix that.

Peer review @erikhagglund ?


erikhagglund:
Looks good.

marcusnaslund:
*Looks* good, but `PRInt` seems to hang on this PR for unknown reason which causes it to lock up its place in the queue forever... @jonathanudd I am unable to reproduce this on my computer

marcusnaslund:
Turns out it was an issue with PRInt. It's restarted now.

PR #1496, opened 2016-03-08
Added maximum difference function to FloatVectorList.

niklasborwell:
@marcusnaslund peer review?

### DIFF
--- a/test/math/FloatVectorListTest.ooc
+++ b/test/math/FloatVectorListTest.ooc
@@ -446,6 +446,27 @@ FloatVectorListTest: class extends Fixture {
 			zeros free()
 			ones free()
 		})
+		this add("maximum difference", func {
+			list1 := FloatVectorList new()
+			list2 := FloatVectorList new()
+			list1 add(-1.1f)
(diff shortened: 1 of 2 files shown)


test/math/FloatVectorListTest.ooc
```diff
@@ -446,6 +446,27 @@ FloatVectorListTest: class extends Fixture {
...
+			list1 add(8.4f)
+			list2 add(8.4f)
+			list1 add(-21.0f)
+			list2 add(-21.0f)
+			expect(list1 maximumDifference(list2), is equal to(0.f) within(tolerance))
+			list1 add(2.1f)
+			list2 add(2.2f)
+			expect(list1 maximumDifference(list2), is equal to(0.1f) within(tolerance))
+			list1 add(-3.1f)
+			list2 add(2.2f)
```
marcusnaslund: Would you insert these somewhere in the list to make sure we're not just taking the difference of the last elements?  `insert(index, value)`

niklasborwell:
Done.

PR #1497, opened 2016-03-08
Now using the improved 'is true' and 'is false' methods

marcusnaslund:
Provides better output if a test fails.
Peer review @jonathanudd ?


jonathanudd:
Looks good except that one test seems to fail on print. I guess it didn't fail when run locally?

marcusnaslund:
I didn't try locally. But Print is right. I'm fixing it now.

marcusnaslund:
There we go. I have to explicitly cast the return from `contains()` to `Bool` in the `uri` tests for... some reason. Very strange.

PR #1498, opened 2016-03-09
Fixed override signature mismatch warnings

thomasfanell:
Reference: [override-signatures](https://github.com/magic-lang/bounty/tree/master/override-signatures)
@marcusnaslund review?

### DIFF
--- a/source/ui/win32/Win32DisplayWindow.ooc
+++ b/source/ui/win32/Win32DisplayWindow.ooc
@@ -15,7 +15,7 @@ version(windows) {
 Win32DisplayWindow: class extends DisplayWindow {
 	_backend: Win32Window
 	init: func (size: IntVector2D, title: String) {
-		super(size, title)
+		super()
 		this _backend = Win32Window new(size, title)
 	}
 	free: override func {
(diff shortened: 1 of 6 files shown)


source/ui/win32/Win32DisplayWindow.ooc
```diff
@@ -15,7 +15,7 @@ version(windows) {
 Win32DisplayWindow: class extends DisplayWindow {
 	_backend: Win32Window
 	init: func (size: IntVector2D, title: String) {
-		super()
+		super(size, title)
```
marcusnaslund: Interesting. `DisplayWindow` has `init: func (size: IntVector2D, title: String)` which does... nothing. I guess we should rather change its constructor to just be `init: func`. @sebastianbaginski ?

thomasfanell: I say we should

sebastianbaginski: It used at least the size last time I checked. Like two months ago.

sebastianbaginski: Or was it UnixWindow... nevermind, we can remove it now

thomasfanell: It does in the platform specific implementations

thomasfanell: I'll remove the parameters where applicable, it's useless to have them there

sebastianbaginski: OK

PR #1499, opened 2016-03-09
Made a fill function taking color by argument

davidpiuva:
Made a way to fill without pen so that the older fill and clear methods can later be removed. The old calls are now deprecated but should work exactly as before.

### DIFF
--- a/source/draw/Canvas.ooc
+++ b/source/draw/Canvas.ooc
@@ -9,6 +9,7 @@
 use base
 use geometry
 use collections
+use draw
 import Image
 import Pen
 import DrawState
@@ -61,7 +62,9 @@ Canvas: abstract class {
 		this drawLines(positions)
 		positions free()
 	}
-	fill: abstract func
+	fill: abstract func ~color (color: ColorRgba)
+	fill: func ~pen { this fill(this pen color) } // Deprecated
+	clear: func { this fill() } // Deprecated
 	draw: virtual func ~DrawState (drawState: DrawState) { Debug error("draw~DrawState unimplemented for class " + this class name + "!") }
 	draw: abstract func ~ImageSourceDestination (image: Image, source, destination: IntBox2D)
 	draw: func ~ImageDestination (image: Image, destination: IntBox2D) { this draw(image, IntBox2D new(image size), destination) }
--- a/source/draw/RasterCanvas.ooc
+++ b/source/draw/RasterCanvas.ooc
@@ -9,6 +9,7 @@
 use base
 use geometry
 use collections
+use draw
 import Image
 import Canvas
 import Pen
@@ -18,7 +19,7 @@ import RasterImage
 RasterCanvas: abstract class extends Canvas {
 	_target: RasterImage
 	init: func (=_target) { super(this _target size) }
-	fill: override func { raise("RasterCanvas fill unimplemented!") }
+	fill: override func ~color (color: ColorRgba) { raise("RasterCanvas fill unimplemented!") }
 	draw: override func ~ImageSourceDestination (image: Image, source, destination: IntBox2D) { Debug error("RasterCanvas draw~ImageSourceDestination unimplemented!") }
 	drawPoint: override func (point: FloatPoint2D) { this _drawPoint(point x as Int, point y as Int) }
 	_drawPoint: abstract func (x, y: Int)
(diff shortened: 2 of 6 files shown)


davidpiuva:
@emilwestergren Peer review?

davidpiuva:
Do we need a `fillBlack` method to replace `clear` or will we have a named color constant for black?

emilwestergren:
No, if you want to clear to black then you simply do `canvas clear(ColorBgra black)`

davidpiuva:
ok

source/draw/Canvas.ooc
```diff
@@ -61,7 +62,9 @@ Canvas: abstract class {
 		this drawLines(positions)
 		positions free()
 	}
-	fill: abstract func
+	fill: abstract func ~color (color: ColorRgba)
+	fill: func ~pen { this fill(this pen color) } // Deprecated
+	clear: func { this fill() } // Deprecated
```
marcusnaslund: Will you be removing them shortly?

davidpiuva: As soon as all callers are gone.

marcusnaslund: Cool.

source/draw/RasterCanvas.ooc
```diff
@@ -18,7 +19,7 @@ import RasterImage
 RasterCanvas: abstract class extends Canvas {
 	_target: RasterImage
 	init: func (=_target) { super(this _target size) }
-	fill: override func { raise("RasterCanvas fill unimplemented!") }
+	fill: override func ~color (color: ColorRgba) { raise("RasterCanvas fill unimplemented!") }
```
marcusnaslund: Will you ever `fill` with something else? Why the `~color` suffix?

davidpiuva: Because it would not compile without separating `~color` from `~pen`. I can remove it once the old version is gone.

marcusnaslund: Ah, right.

source/draw/gpu/GpuCanvasYuv420Semiplanar.ooc
```diff
@@ -76,9 +76,10 @@ GpuCanvasYuv420Semiplanar: class extends GpuCanvas {
 		uvLines free()
 	}
 	drawPoints: override func (pointList: VectorList<FloatPoint2D>) { this _target y canvas drawPoints(pointList) }
-	fill: override func {
-		this _target y canvas fill()
-		this _target uv canvas fill()
+	fill: override func ~color (color: ColorRgba) {
+		yuv := color toYuv()
+		this _target y canvas fill(ColorRgba new(yuv y, 0, 0, 255))
+		this _target uv canvas fill(ColorRgba new(yuv u, yuv v, 0, 255))
```
marcusnaslund: I'm no color expert, so please explain why you take an `rgba` argument, convert it to `yuv`, then fill `y` and `uv` channels with `rgba` colors using the `yuv` components in red and green channels.

davidpiuva: It is copied from the Pen setter to maintain full backward compatibility. Y and UV are stored in RGBA just like on the GPU instead of being converted from RGBA on the planar level because a monochrome could just as well be used to represent a planar RGB image or an intensity map. It is an ugly solution but I don't see a better way since each packed image knows nothing about the planar image.
```ooc
 pen: Pen {
 	get { this _pen }
	set(value) {
		this _pen = value
 		yuv := value color toYuv()
 		this _target y canvas pen = Pen new(ColorRgba new(yuv y, 0, 0, 255), value width)
 		this _target uv canvas pen = Pen new(ColorRgba new(yuv u, yuv v, 0, 255), value width)
	}
}
```

marcusnaslund: Alright. :)

emilwestergren: Shouldn't this conversion be applied in the canvas for Y and UV?
```ooc
fill: func (color: ColorRgba) {
    yuv := color toYuv()
    this _backend setClearColor(yuv y)
}
```
something like that.

davidpiuva: The hard thing is that monochrome and UV does not always belong to a YUV image. UV has become a default format for any packed two channel image and can represent hue and saturation from the HSV color system.

marcusnaslund:
Tested and ready but awaiting results from the discussion above.

davidpiuva:
If we do change existing color conversion then it should be a new PR since it breaks backward compatibility. One ugly but working solution would be to not use virtual fill methods and take the native color given with the correct type declaration.

marcusnaslund:
@emilwestergren Objections?

emilwestergren:
No

Issue #1500, opened 2016-03-09
Add support for comparing relevant classes in unit test framework. 


niklasborwell:
We need to be able to use the `is equal to` framework with all relevant classes.

Higher priority, i.e. used a lot:
- [x] FloatVector2D
- [x] FloatVector3D
- [x] FloatPoint2D
- [x] FloatPoint3D
- [x] Quaternion
- [x] FloatEuclidTransform
- [x] FloatRotation3D
- [x] FloatTransform2D
- [x] FloatTransform3D

Lower priority:
- [ ] FloatBox2D
- [ ] FloatConvexHull2D
- [ ] FloatShell2D
- [ ] IntBox2D
- [x] IntPoint2D
- [x] IntPoint3D
- [ ] IntShell2D
- [ ] IntTransform2D
- [x] IntVector2D
- [x] IntVector3D
- [ ] FloatComplex
- [ ] FloatMatrix

niklasborwell:
@marcusnaslund

marcusnaslund:
Would `FloatComplex` and `FloatMatrix` from `math` be suitable candidates, also?

niklasborwell:
I guess so, but in the low priority category.

marcusnaslund:
@niklasborwell I should have the high-priority items working before lunch tomorrow.

niklasborwell:
Sounds great.

marcusnaslund:
There we go. #1518 will take care of the remaining items.

PR #1505, opened 2016-03-10
Added support for extending equal to() operator with geometry/math covers

marcusnaslund:
Required some re-structuring of `CompareWithinConstraint` since we have to match on the compared type and not on the tolerance type. Otherwise we'd have to do something like

```ooc
expect(vector1, is equal to(vector2) within(FloatVector2D new(0.00001f, 0.00001f))
``` 
instead of
```ooc
expect(vector1, is equal to(vector2) within(0.00001f))
```

The new system also makes it very easy to add new types.

A future PR will take care of remaining types in #1500 ... for now this is a good starting step.

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -8,6 +8,8 @@
 
 use base
 use collections
+use geometry
+use math
 import Constraints
 import Modifiers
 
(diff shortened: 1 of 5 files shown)


source/unit/Fixture.ooc
```diff
@@ -84,32 +86,26 @@ Fixture: abstract class {
...
+			case ComparisonType NotWithin => result append(t"not equal to")
 		}
-		result append(expectedValue toText())
-		result append(t"was")
-		result append(testedValue toText())
+		expectedText, testedText: Text
+		match (expectedValue T) {
+			case FloatVector2D => (expectedText, testedText) = (expectedValue toText~floatvector2d(), testedValue toText~floatvector2d())
+			case => (expectedText, testedText) = (expectedValue toText(), testedValue toText())
+		}
```
marcusnaslund: I wish `rock` could figure out which `toText` to use, but sadly, I have to be explicit.

marcusnaslund:

sebastianbaginski:
Great.

PR #1510, opened 2016-03-10
Added support for high-priority geometry covers in Fixture

marcusnaslund:
Depends on #1505 

After this we can check all the high-priority items in #1500

### DIFF
--- a/source/geometry/FloatEuclidTransform.ooc
+++ b/source/geometry/FloatEuclidTransform.ooc
@@ -65,3 +65,7 @@ FloatEuclidTransform: cover {
 		result
 	}
 }
+
+extend Cell<FloatEuclidTransform> {
+	toText: func ~floateuclidtransform -> Text { (this val as FloatEuclidTransform) toText() }
+}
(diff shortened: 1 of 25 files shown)


marcusnaslund:
@niklasborwell I'm not sure how to apply `within()` for `FloatEuclidTransform` or `FloatTransform2D`/`3D`. Do we need it or is it enough to compare with `equal to()` which uses the `==` operators for these classes?

niklasborwell:
It will still use a proper float comparison, right? Like less than some default threshold

marcusnaslund:
It will use whatever the `==` operators use, which at this moment mostly is *not* proper float comparison. We should probably change that, right?

niklasborwell:
Ok, I see. Yes, definitely. But I thought the == operator would be changed? When would you ever not want to use the proper version?

marcusnaslund:
It should be, but it hasn't. I will make an issue and fix it.

marcusnaslund:
#1516

marcusnaslund:
Peer review @sebastianbaginski ? :)

sebastianbaginski:
ok

sebastianbaginski:
Generic programming does not exist in ooc and this pull request proves it.

sebastianbaginski:
Of course not your fault @marcusnaslund , you did the best you could here :)

PR #1512, opened 2016-03-10
Removed memory leaks in Debug class

### LINKED ISSUE #1509: Debug has no free
There should be a `free~all()` function like, for example, with `Profiler`.

This should be added to `GlobalCleanup`.

marcusnaslund:
Fixes #1509

### DIFF
--- a/source/base/Debug.ooc
+++ b/source/base/Debug.ooc
@@ -23,6 +23,7 @@ Debug: class {
 	_level: static DebugLevel = DebugLevel Everything
 	_printFunction: static Func (String) = func (s: String) { println(s) }
 	initialize: static func (f: Func (String)) {
+		(This _printFunction as Closure) free()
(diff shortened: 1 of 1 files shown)


marcusnaslund:

source/base/Debug.ooc
```diff
@@ -23,6 +23,7 @@ Debug: class {
 	_level: static DebugLevel = DebugLevel Everything
 	_printFunction: static Func (String) = func (s: String) { println(s) }
 	initialize: static func (f: Func (String)) {
+		(This _printFunction as Closure) free()
```
sebastianbaginski: Maybe `free(Owner Receiver)` ?

marcusnaslund: Yeah, *maybe* is the real word here.

PR #1517, opened 2016-03-11
Modified the == operator in geometry and math

### LINKED ISSUE #1516: The == operator in geometry and math covers should use proper float comparison
That means `equals()` instead of `==`. This applies to all `Float`-based covers.

marcusnaslund:
Fixes #1516 

Changed the use of `==` to use `equals()` when it's floating-point numbers. 

Also added `==` and `!=` operators for `FloatMatrix`.

Peer review @niklasborwell ?

### DIFF
--- a/source/geometry/FloatEuclidTransform.ooc
+++ b/source/geometry/FloatEuclidTransform.ooc
@@ -41,7 +41,7 @@ FloatEuclidTransform: cover {
 	operator == (other: This) -> Bool {
 		this translation == other translation &&
 		this rotation == other rotation &&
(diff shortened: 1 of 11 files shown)


marcusnaslund:
What the... the new `FloatMatrix` test works locally every time, but fails on PRInt every time... @jonathanudd ? Can you try it on your machine?

jonathanudd:
I cloned and checked out the branch and then ran all the tests which results in the same failed test as PRInt.

marcusnaslund:
There we go, let's see if PRInt says OK now.

PR #1521, opened 2016-03-11
Now using GlobalCleanup in test framework

marcusnaslund:


source/base/ByteBuffer.ooc
```diff
@@ -52,10 +52,10 @@ ByteBuffer: class {
 	new: static func ~recover (pointer: Byte*, size: Int, recover: Func (This) -> Bool) -> This {
 		_RecoverableByteBuffer new(pointer, size, recover)
 	}
-	free: static func ~all { _RecyclableByteBuffer _free~all() }
+	free: static func ~all { "\n\tByteBuffer" println(); _RecyclableByteBuffer _free~all() }
 }
 
-GlobalCleanup register(|| ByteBuffer free~all())
+//GlobalCleanup register(|| ByteBuffer free~all(), true)
```
marcusnaslund: Everything else seems to work, but there's no way of calling `ByteBuffer free~all()` in tests without segfaulting. @sebastianbaginski

sebastianbaginski: I have no idea, what does gdb say ?

PR #1527, opened 2016-03-13
Removed incrementers and decrementers

### LINKED ISSUE #1526: Remove all ++ and -- operators
`++variable` is semantically +(+(variable)) and generates the C code `++variable` because of no special-casing. On the other hand, `variable++` simply doesn't parse.

The fact that `++variable` actually works is a mistake/bug in rock and will be fixed soon - see https://github.com/ooc-lang/rock/issues/982

Therefore, we should avoid it all-together. It is confusing in itself that `++variable` works but `variable++` does not.

marcusnaslund:
Along with #1525 this fixes #1526 and the motivation for this PR is also there.

### DIFF
--- a/source/base/TextBuilder.ooc
+++ b/source/base/TextBuilder.ooc
@@ -94,7 +94,7 @@ TextBuilder: class {
 		result := this count == other count
 		i := 0
 		while (i < this count && (result &= this[i] == other[i]))
-			++i
+			i += 1
 		result
 	}
 	operator [] (index: Int) -> Char {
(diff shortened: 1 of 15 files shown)


shamanas:
Sorry about this one, it's completely our fault.  
It never occurred to me to test for `++` and `--` since I knew it was a design decision to not have those when unary `+` and `-` were added.

marcusnaslund:
> Sorry about this one, it's completely our fault.

Not at all, this is a good example of how we benefit from each other. :)

marcusnaslund:

PR #1531, opened 2016-03-14
Bug fixes in GlobalCleanup

marcusnaslund:
- Changed the data structure to a `VectorList` to allow for registering callbacks at the bottom, not just at the top.
- `String free~all()` and `ByteBuffer free~all()` is now put at the bottom
- `MutexGlobal free~all()` is now actually called, it was not before. Thank you rock.

### DIFF
--- a/source/system/Memory.ooc
+++ b/source/system/Memory.ooc
@@ -25,22 +25,24 @@ alloca: extern func (SizeT) -> Pointer
 
 // Used for executing any/all cleanup (free~all) functions before program exit
 GlobalCleanup: class {
-	_functionPointers: static Stack<Closure> = null
+	_functionPointers: static VectorList<Closure> = null
 	init: func
-	register: static func (pointer: Func) {
+	register: static func (pointer: Func, last := false) {
 		if (This _functionPointers == null)
-			This _functionPointers = Stack<Closure> new()
-		This _functionPointers push(pointer as Closure)
+			This _functionPointers = VectorList<Closure> new()
+		if (last)
+			This _functionPointers insert(0, pointer as Closure)
+		else
+			This _functionPointers add(pointer as Closure)
 	}
 	run: static func {
 		if (This _functionPointers != null) {
-			while (!This _functionPointers isEmpty) {
-				next := This _functionPointers pop()
+			while (!This _functionPointers empty) {
+				next := This _functionPointers remove(This _functionPointers count - 1)
 				(next as Func)()
 				(next) free()
 			}
-			This _functionPointers free()
-			This _functionPointers = null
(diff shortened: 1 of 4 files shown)


source/system/Memory.ooc
```diff
@@ -50,3 +52,6 @@ GlobalCleanup: class {
 		}
 	}
 }
+
+// Must be here because Mutex.ooc is loaded before Memory.ooc
```
sebastianbaginski: Oh yeah, never assume anything about the order of initialization of static data. Maybe it was not that good idea to have it registered in every module, maybe we should collect them all in `GlobalCleanup` ?

sebastianbaginski:
Looks good.

source/system/Memory.ooc
marcusnaslund: I don't know, the only problems we could have are in `system` where there's only `Mutex` and `String`, and only the first posed problems.

sebastianbaginski: OK, I guess we can leave it like this for now.

marcusnaslund:
@sebastianbaginski This PR seems to fix the problem I just told you about.

sebastianbaginski:
Great.

source/system/Mutex.ooc
```diff
@@ -56,8 +56,6 @@ MutexGlobal: class extends Mutex {
 	lock: override func { This _globalMutex lock() }
 	unlock: override func { This _globalMutex unlock() }
 	free: static func ~all { This _globalMutex free() }
-
-	GlobalCleanup register(|| This free~all())
```
shamanas: @marcusnaslund  
Code in a class body gets added to its default function, if I am not mistaken.  
That means it is executed with each constructor call.  

I don't know if this is related to the bug you were talking about since this seems like a class that would only be initialized once but it could be, so I just wanted to point it out so there is no confusion :)

marcusnaslund: It's actually a different bug we're talking about. :)

But you are correct, this was run once for each class instance, which wasn't what we wanted either. This line was temporarily here because putting it in global scope in `Mutex.ooc` caused the line not to be executed at all. C code looks fine and there's no error, it just didn't do anything. I'm not sure if this is an ooc bug or how to make a good example.

shamanas: Good, just wanted to make sure.  
If you run into that bug again, please ping me and I will take a look into it (two instances will probably be enough to figure out what is happening).

marcusnaslund: Thanks! Putting the line in `Memory.ooc` worked, so I think it might be caused by the order that static memory is initialized.

PR #1532, opened 2016-03-14
Not using exit handler in Fixture

marcusnaslund:
Required for using `GlobalCleanup` in tests. Also, not having to use `atexit()` feels cleaner.


sebastianbaginski:
Yes this is better.

PR #1538, opened 2016-03-15
Removed Pipe and Process tests from android

### LINKED ISSUE #1535: Use android flag in test sources
It could be nice to mark test code not suitable for running on android with `version(!android)` flag.
For now I see two candidates:
- [ ] PipeTest (because of bash script)
- [ ] ProcessTest (same as above)

marcusnaslund:
Fixes #1535 unless something more comes up. @sebastianbaginski ?

### DIFF
--- a/test/io/PipeTest.ooc
+++ b/test/io/PipeTest.ooc
@@ -10,6 +10,7 @@ use unit
 use io
 import io/File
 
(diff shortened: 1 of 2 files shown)


sebastianbaginski:
Splendid.

PR #1539, opened 2016-03-15
do not test thread cancelation on android

sebastianbaginski:
It always fail as `cancel` is not implemented for android
@marcusnaslund


sebastianbaginski:
Which makes me wonder if we should roll out our own implementation

marcusnaslund:
This is OK for now.

marcusnaslund:
Kean tests passed on PRInt, that's all that matters.

PR #1543, opened 2016-03-16
Frequent includes are now imports

### LINKED ISSUE #1484: Change most/all includes to imports
For example, we should never do `include stdlib` but instead have an `stdlib.ooc` defined (and it already is!) and do `import stdlib`. This has the advantage of not having to add a certain include flag every time, and let's us define external functions just once. We already do this for some files, just need to make it consistent.

We don't need to do this for specific files only used once, but rather for headers included in many places.

- [x] pthread
- [x] windows
- [x] time
- [x] signal
- [x] unistd
- [x] ~~stdint~~
- [x] stdlib
- [x] errno

marcusnaslund:
First (and easiest) step of #1484


source/system/external/windows.ooc
```diff
@@ -0,0 +1,105 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+version(windows) {
```
marcusnaslund: Code in this file is identical to before, I just rearranged some stuff to get variables and covers/classes at the top, and removed indentation.

marcusnaslund:

system.use
```diff
@@ -17,16 +17,16 @@ Imports: Numbers
...
 Imports: Owner
 Imports: OwnedBuffer
-Imports: stdlib
 Imports: TextBuffer
 Imports: Text
 Imports: Time
 Imports: VarArgs
+Imports: external/stdlib
+Imports: external/unistd
+Imports: external/windows
```
sebastianbaginski: Nice, much better now

PR #1552, opened 2016-03-17
free leaked error message from stb wrapper

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/draw/StbImage.ooc
+++ b/source/draw/StbImage.ooc
@@ -18,7 +18,8 @@ StbImage: class {
 	load: static func (filename: String, requiredComponents: Int = 0) -> (ByteBuffer, IntVector2D, Int) {
 		x, y, imageComponents: Int
 		data := This _load (filename toCString(), x&, y&, imageComponents&, requiredComponents)
-		raise(data == null, "Failed to load image: " + filename, This)
+		if (data == null)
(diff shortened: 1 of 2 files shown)


marcusnaslund:
What? We have to do this?

marcusnaslund:
What if we hide the `raise` call in an if statement?

sebastianbaginski:
Yes, parameters are always evaluated. If statement would fix this too.

marcusnaslund:
I vote for an if-statement then.

sebastianbaginski:
I mean, it would leak only in case of exception.

sebastianbaginski:
... which would create a leak in this test
```ooc
this add("Test file not found", func {
			try {
				image := RasterRgba open("nonExistingFile")
				expect(false) // Unreachable code since we are expecting an exception
				expect(image, is Null)
			} catch (e: Exception) {
				expect(e message contains("Failed to load image: "))
			}
		})
```

marcusnaslund:
Can't we do something like `e message free()` or whatever in the test?

marcusnaslund:
I don't like adding extra memory management code to free things for events that shouldn't happen.

sebastianbaginski:
ok, updated

marcusnaslund:
Thanks.

PR #1554, opened 2016-03-17
Cleaned up pthread includes

marcusnaslund:
Resolves issues with all includes of `pthread`. We only had three different include statements. But no duplicated definitions, at least. :)

### DIFF
--- a/source/concurrent/Thread.ooc
+++ b/source/concurrent/Thread.ooc
@@ -54,8 +54,8 @@ Thread: abstract class {
 }
 
 version (unix || apple) {
-	include pthread | (_POSIX_C_SOURCE=200809L)
-
+	// This should be in external/pthread.ooc but fails for unknown reason
+	include pthread | (_XOPEN_SOURCE=500, _POSIX_C_SOURCE=200809L)
 	ThreadId: cover from PThread {
 		equals: func (other: This) -> Bool {
 			pthread_equal(this as PThread, other as PThread) != 0
--- a/source/concurrent/native/ThreadUnix.ooc
+++ b/source/concurrent/native/ThreadUnix.ooc
@@ -75,26 +75,18 @@ ThreadUnix: class extends Thread {
 }
 
 // C interface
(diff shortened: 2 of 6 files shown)


source/concurrent/native/ThreadUnix.ooc
```diff
@@ -75,26 +75,18 @@ ThreadUnix: class extends Thread {
...
 	tv_sec: extern TimeT
 	tv_nsec: extern Long
 }
+
+sched_yield: extern func -> Int
+
 version (!apple && !android) {
 	// Using proto here as defining '_GNU_SOURCE' seems to cause more trouble than anything else...
 	pthread_timedjoin_np: extern proto func (thread: PThread, retval: Pointer, abstime: TimeSpec*) -> Int
 }
```
marcusnaslund: Remaining externs here will be cleaned up in a later PR.

source/concurrent/Thread.ooc
```diff
@@ -54,8 +54,8 @@ Thread: abstract class {
 }
 
 version (unix || apple) {
-	include pthread | (_POSIX_C_SOURCE=200809L)
-
+	// This should be in external/pthread.ooc but fails for unknown reason
+	include pthread | (_XOPEN_SOURCE=500, _POSIX_C_SOURCE=200809L)
```
marcusnaslund: I get errors like "Cannot match `ThreadId`, nearest match is `ThreadId`" and "missing function `set(ThreadId)` in `Cell<ThreadId>`, nearest match is `set(ThreadId)`"

Go home rock, you're drunk.

marcusnaslund:

source/concurrent/Thread.ooc
shamanas: Wat.
I'll try to reproduce this later.

marcusnaslund: Thanks. :) Just move this block to `source/system/external/pthread.ooc` and you should get it too.

PR #1555, opened 2016-03-18
Moved all time-related includes to a separate file

marcusnaslund:
`TimeT` was only defined three times, but at least as the same type every time.

Also removed two `include stdio` in places where it wasn't used.

Possibly the last step of #1484 - I will evaluate after this PR


marcusnaslund:

sebastianbaginski:
> TimeT was only defined three times, but at least as the same type every time.

Nice :)

PR #1558, opened 2016-03-18
Added external/errno.ooc

marcusnaslund:
And thus removed some double-definition and weirdness related to error handling.

Fixes the last part of #1484 - hooray. That was a lot easier than expected.


marcusnaslund:
Well, we filled the hard drive of the PRInt computer. Nice.

PR #1562, opened 2016-03-21
Added missing this in math tests

davidpiuva:
Added lots and lots of missing `this`. Starting with the math folders.


davidpiuva:

test/math/FloatComplexVectorListTest.ooc
```diff
@@ -19,39 +19,39 @@ FloatComplexVectorListTest: class extends Fixture {
...
 
 	complexNumberArray := FloatComplexVectorList new(4)
-	complexNumberArray add(complexNumber0)
-	complexNumberArray add(complexNumber1)
-	complexNumberArray add(complexNumber2)
-	complexNumberArray add(complexNumber3)
+	this complexNumberArray add(this complexNumber0)
+	this complexNumberArray add(this complexNumber1)
+	this complexNumberArray add(this complexNumber2)
+	this complexNumberArray add(this complexNumber3)
```
davidpiuva: Should void expressions with order dependent side effects be allowed in the declaration section?

shamanas: For what it's worth, all rock does is place them in the `__defaults__` function in order, so the order of evaluation is guaranteed.  

Style wise, I wouldn't recommend it but that's a personal / per project thing :)

davidpiuva: I suspected that and this code should probably to to a constructor or the same line as the declaration.

davidpiuva:

davidpiuva:
Static calls were missed because the compiler resolved it as a namespace. More might come.

davidpiuva:
Updated with missing `This` for the `expect` method.

davidpiuva:
Updated tolerance.

PR #1568, opened 2016-03-23
Added missing this in source/draw

davidpiuva:


davidpiuva:

PR #1574, opened 2016-03-24
Fixed bug in toRaster~target.

emilwestergren:


marcusnaslund:
```
AndroidContext.ooc:106:16: error: assignment from incompatible pointer type [-Werror]
result = super(source, target)
```

marcusnaslund:
Makes sense now.

PR #1578, opened 2016-03-29
Fixed confusing String capitalize()

marcusnaslund:
Maybe just me, but I got too confused by the original to not change it.


sebastianbaginski:
I agree, `this[1, -1]` was not very self-explanatory.

davidpiuva:
I recognized -1 for infinity from RedFaction's level editor.

PR #1579, opened 2016-03-29
Fixed "tolerance: nan" in some tests

### LINKED ISSUE #1575: Double check that correct tolerance values are passed to Fixture...
...for all types. The `Quaternion` test fails with tolerance nan on windows, for example.

( @fredrikbryntesson @sebastianbaginski There is something to be said about testing on windows, it helps us detect bugs like these.)

marcusnaslund:
Turns out it was not a problem with the testing, but rather that `Fixture` tried to read `tolerance` even though it did not exist because `within()` was never called.

Also added tests for all types supported by `Fixture`.

Closes #1575

### DIFF
--- a/source/unit/Constraints.ooc
+++ b/source/unit/Constraints.ooc
@@ -83,7 +83,7 @@ CompareConstraint: class extends Constraint {
 
 CompareWithinConstraint: class extends CompareConstraint {
 	actualType: Class
-	precision: LDouble
+	precision: LDouble = -1.0L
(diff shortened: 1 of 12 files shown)


marcusnaslund:

source/unit/Constraints.ooc
```diff
@@ -83,7 +83,7 @@ CompareConstraint: class extends Constraint {
 
 CompareWithinConstraint: class extends CompareConstraint {
 	actualType: Class
-	precision: LDouble
+	precision: LDouble = -1.0L
```
sebastianbaginski: great :+1:

PR #1581, opened 2016-03-29
Added 'is within(...)' method to unit tests

marcusnaslund:
Instead of
```ooc
expect(myValue >= 4.f && myValue < 12.f)
```
it is now possible to do stuff like 
```ooc
expect(myValue, is within(4 .. 12))
```
which is cleaner and gives correct error messages if the test fails, for example:
```
  -> uniform : expected within 12 and 16 was 4.00
```

Also:
- added `toText` for `Range` although it wasn't needed in the final result, but it doesn't hurt to have.
- added the `ìn` method for float-point types as for integer types.

### DIFF
--- a/source/system/Numbers.ooc
+++ b/source/system/Numbers.ooc
@@ -114,6 +114,9 @@ LDouble: cover from long double {
 		string free()
 		result
 	}
+	in: func (range: Range) -> Bool {
+		this >= range min && this < range max
+	}
 }
 
 Double: cover from double extends LDouble {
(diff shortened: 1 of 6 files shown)


sebastianbaginski:
Very nice.

source/system/Numbers.ooc
```diff
@@ -114,6 +114,9 @@ LDouble: cover from long double {
 		string free()
 		result
 	}
+	in: func (range: Range) -> Bool {
+		this >= range min && this < range max
```
sebastianbaginski: Should this use floating point comparison functions instead of operators ?

marcusnaslund: I don't think so since `range min` and `range max` are integers.

sebastianbaginski: Oh.

PR #1582, opened 2016-03-29
 Made expect() a global function

marcusnaslund:
Your idea @davidpiuva , so I leave the peer review to you. This should drastically reduce the number of implicit `This`, right?

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -161,7 +161,7 @@ Fixture: abstract class {
 		text print()
 		fflush(stdout)
 	}
-	expect: static func (value: Object, constraint: Constraint) {
+	verify: static func (value: Object, constraint: Constraint) {
 		This _expectCount += 1
 		if (!constraint verify(value))
 			TestFailedException new(value, constraint, This _expectCount) throw()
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Hang on, not finished.

davidpiuva:
I will have a look.

source/unit/Fixture.ooc
```diff
@@ -210,3 +175,39 @@ Test: class {
...
+expect: func ~intvector3d (value: IntVector3D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~intpoint2d (value: IntPoint2D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~intpoint3d (value: IntPoint3D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~floatcomplex (value: FloatComplex, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~floatmatrix (value: FloatMatrix, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~inttransform2d (value: IntTransform2D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~floatbox2d (value: FloatBox2D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~intbox2d (value: IntBox2D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~floatshell2d (value: FloatShell2D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
+expect: func ~intshell2d (value: IntShell2D, constraint: Constraint) { Fixture expect(Cell new(value), constraint) }
```
davidpiuva: Can we replace all these overloads with a generic function for any type?

marcusnaslund: No. :/

marcusnaslund: At least I couldn't find a way before.

davidpiuva: If we can't then it might be a bug fix or feature request for rock. I should have a try at it.

davidpiuva: `expect: static func ~any <T> (value: T, constraint: Constraint) { This expect(Cell new(value), constraint) }`
Looks like a bug in Rock since the C compiler bailed out.
`error: lvalue required as unary ‘&’ operand`
@fredrikbryntesson Is this a priority?

davidpiuva:
99 complaints about missing this remains for expect. I have no idea why.

marcusnaslund:
How many were there originally?

davidpiuva:
Around 5300

davidpiuva:
Just adding `This` everywhere would remove all complaints.

marcusnaslund:
Alright, I think I know what's wrong, hang on.

marcusnaslund:
How about now?

davidpiuva:
0 detections. :)

marcusnaslund:
`/giphy celebrate`

davidpiuva:
Peer review is done.

Issue #1523, opened 2016-03-11
Double free or corruption in BlockedQueue test


marcusnaslund:
Hadn't seen it before in the office, but on my home laptop I keep getting

```
BlockedQueue *** Error in `./Tests': double free or corruption (!prev): 0x0000000001313eb0 ***
```

pretty often, when running the concurrent tests.
This is a test not running properly on Windows, so I guess now we know why.

marcusnaslund:
@niklasborwell Is this the problem you were talking about?

niklasborwell:
It's hard to say since these kind of errors are so hard to debug, but it's the same kind of error. I encounter it randomly when running Vidhance.

marcusnaslund:
Changing `BlockedQueue enqueue` to
```ooc
enqueue: override func (item: __onheap__ T) {
		this _mutex lock()
		this _backend enqueue(item)
		this _waitLock wake()
		this _mutex unlock()
	}
```
removes the problem on my laptop, as per http://stackoverflow.com/questions/1640389/pthreads-pthread-cond-signal-from-within-critical-section

Must be tested some more though.

marcusnaslund:
Fixed by #1603

PR #1585, opened 2016-03-30
Cleaned up unit test framework

### LINKED ISSUE #1584: Memory issues with making Fixture is a global member
Only one major obstacle stands in the way of removing all implicit `this`/`This`. In every `expect` call we use the static property `is` defined in `Fixture` as:

```ooc
is ::= static IsConstraints new()
```

When moving this line to global scope, `is := IsConstraints new()`, tests always crash. This is because, when a test gets freed, the `free` call propagates up to freeing `is` which then gets doubly freed if a test has more than one `expect` call. This can be "fixed" by adding

```ooc
free: override func
```

to `IsConstraints` in `source/unit/Constraints.ooc` which prevents it from getting freed (the `super` call never happens). Exactly why this works when `is` is a static property of `Fixture` I do not know, but someone well-versed in ooc/rock might have a better answer.

However, this approach still does not work with the concurrent tests (but seems to work fine with all others). Running `./test.sh concurrent` with a global `is`, even when `free` has been changed, results in all kinds of `SIGSEGV`s and `SIGABRT`s. 

It might be that our use of `is` has some memory and/or thread-safety issues, which might be the reason for the uncommon (but still existing) problems on upstream with the `BlockedQueue` and `ThreadPool` tests (#1523), as well as random memory problems with `clang` and under Windows (#1466).

Long story short, this needs a closer examination. Ideas and inputs are very welcome @thomasfanell @fredrikbryntesson @sebastianbaginski @davidpiuva

marcusnaslund:
Done here:
- Removed the useless `Constraint` class.
- Removed the `Range` constructor from `RangeConstraint`.
- Removed all `~parent` suffixes.
- Made sure noone has `Is` as parent, so that it isn't freed after every expect call.
- Properly using `ComparisonType` when creating `lessOrEqual` and `greaterOrEqual` constraints.
- Renamed `IsConstraints` to `is` and made it abstract
- Removed the `is` property from `Fixture`
- Through the above, removed the need for implicit `This` when using `is`
- Moderate speedup of unit tests since we're not creating and freeing an `IsConstraints` and an extra `ExpectModifier` instance on every `expect` call.
- Verify that no new memory leaks have been introduced
- Fixes #1584 

Remaining work:
- Check if all this resolves any/all problems in #1409 or #1523

### DIFF
--- a/source/unit/Constraints.ooc
+++ b/source/unit/Constraints.ooc
@@ -23,61 +23,50 @@ ComparisonType: enum {
 	NotWithin
 }
 
-IsConstraints: class extends ExpectModifier {
-	true ::= TrueConstraint new(this)
-	false ::= FalseConstraint new(this)
-	Null ::= NullConstraint new(this)
-	notNull ::= NotNullConstraint new(this)
-	equal ::= EqualModifier new(this)
-	notEqual ::= EqualModifier new(this, ComparisonType NotEqual)
-	less ::= LessModifier new(this)
-	lessOrEqual ::= LessModifier new(this, true)
-	greater ::= GreaterModifier new(this)
-	greaterOrEqual ::= GreaterModifier new(this, true)
+is: abstract class {
+	true ::= static TrueConstraint new()
+	false ::= static FalseConstraint new()
+	Null ::= static NullConstraint new()
+	notNull ::= static NotNullConstraint new()
+	equal ::= static EqualModifier new()
+	notEqual ::= static EqualModifier new(ComparisonType NotEqual)
+	less ::= static LessModifier new()
+	lessOrEqual ::= static LessModifier new(ComparisonType LessOrEqual)
+	greater ::= static GreaterModifier new()
+	greaterOrEqual ::= static GreaterModifier new(ComparisonType GreaterOrEqual)
+	within: static func ~range (range: Range) -> RangeConstraint { RangeConstraint new(range min, range max) }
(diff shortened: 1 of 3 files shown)


marcusnaslund:
Peer review @fredrikbryntesson or @sebastianbaginski ?

@davidpiuva - can you give me some nice numbers on the number of reduced implicit `This`?

davidpiuva:
I am on it.

sebastianbaginski:
Looks good, but maybe better have @thomasfanell to check it as well.

davidpiuva:
Both upstream and this branch has 150 warnings.

marcusnaslund:
> Both upstream and this branch has 150 warnings. The content looks identical.

How can that be? `is` was a property of `Fixture` earlier.

davidpiuva:
I have sent you a file with warnings.

marcusnaslund:
> I have sent you a file with warnings.

I believe you, I just don't understand. It should catch implicit `This` as well as implicit `this`, right?

davidpiuva:
Missing `This` is found for functions but detecting static variables is not yet implemented.

marcusnaslund:
Ah, OK, thanks.

fredrikbryntesson:
Looks okay.

marcusnaslund:
By sheer awesomeness I move this to final review then. This only affects unit tests, so fixing anything after the fact is easy. All tests pass just fine on PRInt and locally.

PR #1589, opened 2016-03-31
Raster* distance() only checks 50% of pixels

### LINKED ISSUE #1551: Improve distance function between two raster images
At the moment this function is very slow and can consume several seconds per image comparison. It should be possible to optimize this function, for example by parallelizing it.

marcusnaslund:
Reduces the runtime of the `draw` tests by about 15-20%.
Fixes #1551 (?)

### DIFF
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -87,8 +87,9 @@ RasterMonochrome: class extends RasterPacked {
 			result = this distance(converted)
 			converted referenceCount decrease()
 		} else {
-			for (y in 0 .. this size y)
-				for (x in 0 .. this size x) {
+			for (y in 0 .. this size y) {
+				x := y % 2
+				while (x < this size x) {
 					c := this[x, y]
 					o := (other as This)[x, y]
 					if (c distance(o) > 0) {
@@ -110,8 +111,10 @@ RasterMonochrome: class extends RasterPacked {
 							distance += (c y - maximum y) as Float squared
 						result += distance sqrt()
 					}
+					x += 2
 				}
-			result /= (this size x squared + this size y squared as Float sqrt())
+			}
+			result /= ((this size x squared + this size y squared) as Float sqrt())
 		}
 		result
 	}
--- a/source/draw/RasterRgb.ooc
+++ b/source/draw/RasterRgb.ooc
@@ -82,8 +82,9 @@ RasterRgb: class extends RasterPacked {
 			result = this distance(converted)
 			converted referenceCount decrease()
 		} else {
-			for (y in 0 .. this size y)
-				for (x in 0 .. this size x) {
+			for (y in 0 .. this size y) {
+				x := y % 2
+				while (x < this size x) {
 					c := this[x, y]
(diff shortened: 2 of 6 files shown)


source/draw/RasterMonochrome.ooc
```diff
@@ -111,7 +112,7 @@ RasterMonochrome: class extends RasterPacked {
 						result += distance sqrt()
 					}
 				}
-			result /= (this size x squared + this size y squared as Float sqrt())
+			result /= ((this size x squared + this size y squared) as Float sqrt())
```
marcusnaslund: Also fixed this bug, as a bonus.

source/draw/RasterMonochrome.ooc
```diff
@@ -88,7 +88,8 @@ RasterMonochrome: class extends RasterPacked {
 			converted referenceCount decrease()
 		} else {
 			for (y in 0 .. this size y)
-				for (x in 0 .. this size x) {
+				for (index in y % 2 .. this size x / 2) {
+					x := index * 2
```
davidpiuva: This code will not give a checker pattern since the initial offset is multiplied by two. Add two to the counter instead and let it end below size x as usual.

marcusnaslund: Ahhh thanks.

marcusnaslund:
Peer review @tomasmahlberg ?

source/draw/RasterMonochrome.ooc
```diff
@@ -87,8 +87,9 @@ RasterMonochrome: class extends RasterPacked {
 			result = this distance(converted)
 			converted referenceCount decrease()
 		} else {
-			for (y in 0 .. this size y)
-				for (x in 0 .. this size x) {
+			for (y in 0 .. this size y) {
+				x := y % 2
+				for (_ in 0 .. this size x / 2) {
```
davidpiuva: So now you only compare the left side of the image?

marcusnaslund: No, line 114 increases `x` by 2 each iteration.

davidpiuva: Why not just use x in a while loop instead of having two variables?

marcusnaslund: Good thinking.

marcusnaslund:
Updated.

tomasmahlberg:
Seems like @davidpiuva already led this PR in the right direction so I'll just approve.

niklasborwell:
This seems like a strange fix to me. 
First of all, we shouldn't lower the accuracy of the basic function in case we need the full accuracy. In that way it would be better to create another function that returns an approximate distance. A general function like this would be expected to check every pixel. 
Another way would be to let time consuming tests check scale down images before comparing. In that way we could use the same function and the tests would still be a lot faster.
Secondly, I think this will effectively lower the threshold for current tests that already have a threshold value set based on this function, so the thresholds in those tests would have to be recalibrated. That's because the value returned scales with resolution.

marcusnaslund:
Let's hold on this then.

PR #1594, opened 2016-03-31
fixed wake and wakeAll in WaitLock

sebastianbaginski:
According to pthread documentation, 
```
lock()
broadcast() / signal()
unlock()
```
is a better way to signal waiting threads: 
>  if predictable scheduling behavior is required, then that mutex shall be locked by the thread calling pthread_cond_broadcast() or pthread_cond_signal()
( http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_cond_broadcast.html ) 

Qt implementation of Wait Condition for reference:
http://code.qt.io/cgit/qt/qt.git/tree/src/corelib/thread/qwaitcondition_unix.cpp#n129

@marcusnaslund peer review

### DIFF
--- a/source/concurrent/ThreadPool.ooc
+++ b/source/concurrent/ThreadPool.ooc
@@ -55,8 +55,8 @@ _Task: abstract class {
 			this _mutex unlock()
 			this free()
 		} else {
-			this _mutex unlock()
 			this _waitCondition broadcast()
+			this _mutex unlock()
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Very good.

sebastianbaginski:
```
[FAIL] gcc: error: rock_tmp/<something>.c: Filen eller katalogen finns inte gcc: fatal error: no input files
```
is there something wrong with PRInt ?

marcusnaslund:
PRInt is never wrong, it's just misunderstood.

marcusnaslund:
But yes, we are patching some memory leaks. I will restart in a short while.

PR #1595, opened 2016-03-31
removed static variable from ReferenceCounter

sebastianbaginski:
`count: static Int` in ReferenceCounter was never used and accessed simultaneously without any lock.
helgrind reported possible data races related to string literals.
@marcusnaslund


marcusnaslund:
So stringliterals are dependant on pthread? What will happen on windows?

sebastianbaginski:
I guess it won't compile :dolphin:

sebastianbaginski:
So let's go with only the static variable part for now.

PR #1598, opened 2016-03-31
Using PtrDiff in Raster* convertFrom()

### LINKED ISSUE #1597: Raster* convertFrom() will yield warnings on 32-bit compilers
The problem comes from 
```ooc
row := result buffer pointer as Long
```
which must be changed to 
```ooc
row := result buffer pointer as Int
```
on, for example, our Windows build. This can be fixed by reimplementing `PtrDiff`.

marcusnaslund:
Fixes #1597 

You know what it's about @sebastianbaginski

### DIFF
--- a/source/system/Numbers.ooc
+++ b/source/system/Numbers.ooc
@@ -137,6 +137,7 @@ SizeT: cover from size_t extends ULLong
 SSizeT: cover from ssize_t extends LLong {
 	toString: func -> String { "%u" formatSSizeT(this) }
 }
+PtrDiff: cover from ptrdiff_t extends SSizeT
 
 Range: cover {
 	min, max: Int
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -206,16 +206,16 @@ RasterMonochrome: class extends RasterPacked {
 			result = (original as This) copy()
 		else {
 			result = This new(original)
-			row := result buffer pointer as Long
+			row := result buffer pointer as PtrDiff
 			rowLength := result stride
 			rowEnd := row + rowLength
-			destination := row
+			destination := row as ColorMonochrome*
 			f := func (color: ColorMonochrome) {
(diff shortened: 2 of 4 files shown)


source/system/Numbers.ooc
```diff
@@ -137,6 +137,7 @@ SizeT: cover from size_t extends ULLong
 SSizeT: cover from ssize_t extends LLong {
 	toString: func -> String { "%u" formatSSizeT(this) }
 }
+PtrDiff: cover from ptrdiff_t extends SSizeT
```
sebastianbaginski: `cover from ptrdiff_t extends SSizeT`
C : cover from Small extends Big
but `sizeof(C) == sizeof(Small)` ? Confusing

marcusnaslund: Do not use `sizeof(...)`, you will get the size of the pointer to the class. Use `C size` and `Small size`.

sebastianbaginski: I mean in pseudocode ;)

marcusnaslund: Ah - easily confusing since `sizeof(C) == sizeof(Small)` is not only valid ooc, but also true. :)
Anyway, I took the definition of `PtrDiff` as it appears in the original sdk. I don't understand why you assume ooc to be non-confusing.

sebastianbaginski: What I mean is that `extend` indicates a `is-a` relation (or maybe I got that part wrong), so is should be possible to use the result type in any place where `Big` is expected ? and they have different sizes... or does it work differently for covers ?

sebastianbaginski: OK, I guess we can just move on :)

marcusnaslund: Since covers don't support inheritance, I don't think it shouldn't be thought of in that way.

sebastianbaginski: `covers don't support inheritance`
I learn something new every day :/

PR #1599, opened 2016-03-31
CPU shaders

davidpiuva:
Made an untested prototype of CPU shaders to discuss the concept before making a fully working version.

Concept:
Instead of trying to fully emulate shaders all the time, one can define everything from rasterization to blending in the same `render` function to get more performance from the CPU. Shaders that almost fully emulate the CPU will only be used as a reference implementation for asserting correctness of GPU shaders.

### DIFF
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -224,4 +224,10 @@ RasterMonochrome: class extends RasterPacked {
 		}
 		result
 	}
+	writePixelGpu: override func (x, y: Int, color: ColorRgba) {
+		this[x, y] = ColorMonochrome new(color r)
+	}
+	readPixelGpu: override func (x, y: Int) -> ColorRgba {
+		ColorRgba new(this[x, y] y, 0, 0, 255)
+	}
 }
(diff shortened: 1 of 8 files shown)


davidpiuva:
@emilwestergren What do you think about this interface for DrawState on the CPU?

source/draw/RasterMonochrome.ooc
```diff
@@ -224,4 +224,10 @@ RasterMonochrome: class extends RasterPacked {
 		}
 		result
 	}
+	writePixelGpu: override func (x, y: Int, color: ColorRgba) {
```
emilwestergren: Why include `Gpu` in the naming?

davidpiuva: Its purpose is to emulate how the GPU works for portability when moving code. Otherwise, someone will instantly change how it behaves and break portability.

emilwestergren: I don't see anything in these functions that's unique to a GPU. Isn't it better to leave the `Raster` classes clean and implement the emulation in `CpuMap` or in another class?

davidpiuva: I would prefer not to have switch cases that fails at runtime since it is not modular but if they can be named in a better way that states behaviour explicitly then I am open for suggestions.

davidpiuva:
I should make a better suggestion at some other time.

PR #1603, opened 2016-03-31
Removed simultaneous expect calls

### LINKED ISSUE #1592: Update concurrency tests to not use `expect` simultaneously
There are many tests in the `concurrent` module calling `expect` from different threads simultaneously. This can cause data races as `verify` in Fixture has global data and is not protected by a lock.
After consulting with @marcusnaslund  it seems reasonable to collect the results and move the `expect` calls out of the thread functions. Introducing a global lock in Fixture might change the expected behaviour in concurrent tests.

marcusnaslund:
Fixes #1592 - calling `expect` from several threads at the same time may cause race conditions in `Fixture`.

As a bonus, this also removes the constant crash of the BlockedQueueTest on windows.


marcusnaslund:
Feel free to suggest a better approach.

marcusnaslund:
Oh, he's not here. Peer review @fredrikbryntesson ?

sebastianbaginski:
I think this is better now, use only thread local variables to verify the expected result, and lock global mutex and call `expect` only on failure. Good job :)

Issue #1466, opened 2016-03-04
Some test cases fails on windows


sebastianbaginski:
- [ ] Pipe
- [ ] Process
- [ ] Env

marcusnaslund:
All three classes taken directly from the original SDK, "cleaned up", and never tested on Windows before.

marcusnaslund:
I took the liberty of updating this.

marcusnaslund:
The problem with `BlockedQueueTest` can be reproduced on Linux as well unfortunately... #1523

marcusnaslund:
Added `Net`. The `Quaternion` test fails for me as well, but seperate issue opened on probable cause #1575

marcusnaslund:
I've got the windows build up and running. Updated the description.

sebastianbaginski:
I'm not even sure `cancel` should be in the Thread's interface now. I think we should implement cancelling logic in the client code and use a flag to just exit the thread function, without relying on the OS facilities (which are not always there or are unsafe). Writing non-blocking client code should be preferred over forcing thread termination.

What I mean is, there is no safe way to terminate code such as
```ooc
thread := Thread new(func { while(1) doSomething() })
```
The client code should provide some cancellation points. Ideally every thread task could be a subclass of some abstract Task interface like
```ooc
AbstractThreadTask: abstract class {
    run: abstract func
    cancel: virtual func { lock(), this _runFlag = false, unlock() }
    isCancelled: virtual func -> Bool { lock(), result := this _runFlag, unlock(), return result }
}
```
and clients implementing the interface should check if the task was cancelled inside the `run` implementation, so the previous "unsafe" example becomes
```ooc
MyThreadTask: class extends AbstractThreadTask {
    run: override func {
        while(1){ 
           if( this isCancelled() ) {
               doCleanup()
               return
          }
          doSomething()
      }
    }
}
thread := Thread new(MyThreadTask new())
```
and now the `Thread` class can have a safer `cancel` implementation, which does not rely on the OS:

```ooc
Thread: abstract class {
    _task: AbstractThreadTask
    _code: func { this _task run() }
    cancel: func {
       this _task cancel() // simply sets the flag, should be safe to simply join or detach the thread now
    }
   ...
}
```
[ btw. Ideally we'd like to define the task subclasses in place, like in Java :) ]

I'm aware that this idea makes the Thread class less low-level and maybe slightly less convenient to use. I'm also not saying that we should do this now, it's just maybe something to consider.
[...]

PR #1604, opened 2016-04-01
 Fixed roundoff errors in SvgPlotTest

marcusnaslund:
Fixes the failing `SvgPlot` test on windows and clang. (Part of #1466 )

This seems like a neat and simple enough solution.

### DIFF
--- a/test/plot/SvgPlotTest.ooc
+++ b/test/plot/SvgPlotTest.ooc
@@ -113,22 +113,19 @@ SvgPlotTest: class extends Fixture {
 			writer free()
 		})
 		this add("Output check", func {
-			generated := t"test/plot/output/example.svg"
-			comparison := t"test/plot/input/exampleComparison.svg"
-			generatedFile := FileReader new(generated)
-			comparisonFile := FileReader new(comparison)
-			generatedBuffer := CharBuffer new()
-			comparisonBuffer := CharBuffer new()
-
+			generatedFile := FileReader new(t"test/plot/output/example.svg")
+			comparisonFile := FileReader new(t"test/plot/input/exampleComparison.svg")
 			while (generatedFile hasNext() && comparisonFile hasNext()) {
-				generatedFile read(generatedBuffer)
-				comparisonFile read(comparisonBuffer)
-				expect(generatedBuffer == comparisonBuffer)
+				(generatedString, comparisonString) := (generatedFile readUntil(' '), comparisonFile readUntil(' '))
+				// If they differ, it can only be due to (small) roundoff errors
+				if (generatedString != comparisonString) {
+					(generatedFloat, comparisonFloat) := (generatedString toFloat(), comparisonString toFloat())
+					expect(generatedFloat, is equal to(comparisonFloat) within(0.05))
+				}
+				(generatedString, comparisonString) free()
 			}
-
 			expect(generatedFile hasNext(), is false)
 			expect(comparisonFile hasNext(), is false)
-
 			(generatedFile, comparisonFile) free()
 		})
 	}


marcusnaslund:
Peer review @davidpiuva ?

davidpiuva:
Ok

marcusnaslund:
Any comments?

davidpiuva:
I haven't found anything yet.

marcusnaslund:
For how long do you plan to peer review then?

davidpiuva:
I might be done within 15 minutes. Tokenization and float parsing can be tricky.

davidpiuva:
* I do not know if ooc uses fixed internationalization or if it suddenly can expect french decimals.
* Parsing svg should probably separate at the ' sign used for quoting all numbers.
* There should be a check for if the values are numbers `(-|0..9|.)*` before parsing in case that two invalid strings both return 0.0 for not being floats at all.

marcusnaslund:
Regarding the first two points, it's a test, so should that be a problem we can cross that bridge if/when we come to it.
The third point is valid, though. I will add some sort of number check.

davidpiuva:
So `stroke-opacity='255.00'` is one token to be compared as a float value?
Is there something I missed?

marcusnaslund:
`stroke-opacity='255.0'` will be compared as a string. If it differs it will *(soon)* be checked if it's a number, and since it isn't, fail.

`249.00` and `249.01` will be compared as strings, since they differ they will be checked to see if they're numbers, and since they are, converted to numbers and floating point comparison is made.

The only failing numbers on windows were from those big blobs of many numbers in a row, that's what we want to catch (and this solves that). Other than that this test will be no worse than before.

davidpiuva:
Then all will be string comparisons since all floats are in ' quotes and this PR does not change the behaviour.

marcusnaslund:
>  all floats are in ' quotes

That's not true.

davidpiuva:
At which line in example.svg did you find an unquoted float separated by spaces?

davidpiuva:
Line 1048 has spaces within the quote but I can't find any other exception.

marcusnaslund:
Line 726, 727, 728, 838, 926, etc.
I only need to catch these.

davidpiuva:
Okay.

marcusnaslund:
I will fix the `isNumber` check after the weekend.

test/plot/SvgPlotTest.ooc
```diff
@@ -113,22 +113,19 @@ SvgPlotTest: class extends Fixture {
...
+			generatedFile := FileReader new(t"test/plot/output/example.svg")
+			comparisonFile := FileReader new(t"test/plot/input/exampleComparison.svg")
 			while (generatedFile hasNext() && comparisonFile hasNext()) {
-				generatedFile read(generatedBuffer)
-				comparisonFile read(comparisonBuffer)
-				expect(generatedBuffer == comparisonBuffer)
+				(generatedString, comparisonString) := (generatedFile readUntil(' '), comparisonFile readUntil(' '))
+				// If they differ, it can only be due to (small) roundoff errors
+				if (generatedString != comparisonString) {
+					(generatedFloat, comparisonFloat) := (generatedString toFloat(), comparisonString toFloat())
```
davidpiuva: I can approve after making the float detection.

marcusnaslund:
It might work just checking that no float generated is equal to `0.0f` if there are none in the comparison file.

davidpiuva:
It would be nice to return NaN from `toFloat` if a non math-character is detected. We can assume that there is no scientific notation.

davidpiuva:
Seems like there is no need to check since `strtof` already has the feature.
If any of the values can not be evaluated, NaN will give false for the comparison.
http://www.cplusplus.com/reference/cstdlib/strtof/

PR #1606, opened 2016-04-01
Fixed ProcessTest on Windows

marcusnaslund:
The script was too complex for the bash clone on windows, and to be honest it didn't need any complexity at all. `ProcessTest` now passes.

The `PipeTest` no longer crashes, which is nice, but only gets the first data from the pipe for some reason.


marcusnaslund:
Peer review @jonathanudd ?

marcusnaslund:
Or @sebastianbaginski

PR #1608, opened 2016-04-03
Removed most memory leaks in net test

marcusnaslund:
Reduced from
```
definitely lost: 1,224 bytes in 49 blocks
indirectly lost: 21,672 bytes in 150 blocks
```
to
```
definitely lost: 576 bytes in 22 blocks
indirectly lost: 816 bytes in 29 blocks
```


marcusnaslund:

PR #1609, opened 2016-04-04
Removes error message from ProcessTest

marcusnaslund:
If the test hasn't been run before, this would give an error message since `sum.txt` didn't exist.


sebastianbaginski:
Are you sure I should PR this ? You seem undecided with this one :)

marcusnaslund:
Haha sorry, PRInt didn't catch this and I had to reopen it to make it appear in the list. :)

PR #1612, opened 2016-04-04
I already know it's all OK

### LINKED ISSUE #1409: ThreadPool test crashes randomly

marcusnaslund:
I might be responsible for writing this during my first time here, but I honestly don't know why. We might as well check that `job != null`.

And removing the `isOk` parameter actually fixes #1409, so, hurray.

### DIFF
--- a/source/concurrent/SynchronizedQueue.ooc
+++ b/source/concurrent/SynchronizedQueue.ooc
@@ -62,13 +62,10 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 		this _waitLock wake()
 	}
 	cancel: func { this _waitLock with(|| this _canceled = true) . wakeAll() }
-	wait: func (isOk := null as Bool*) -> T {
+	wait: func -> T {
 		result: T = null
 		this _waitLock lockWhen(func -> Bool { !this empty || this _canceled })
-		if (this _canceled) {
-			if (isOk)
-				isOk@ = false
-		} else
+		if (!this _canceled)
 			result = this _backend dequeue(result)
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Peer review @emilwestergren ?

source/concurrent/SynchronizedQueue.ooc
```diff
@@ -62,13 +62,10 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
...
 	cancel: func { this _waitLock with(|| this _canceled = true) . wakeAll() }
-	wait: func (isOk := null as Bool*) -> T {
+	wait: func -> T {
 		result: T = null
 		this _waitLock lockWhen(func -> Bool { !this empty || this _canceled })
-		if (this _canceled) {
-			if (isOk)
-				isOk@ = false
-		} else
+		if (!this _canceled)
```
emilwestergren: You could replace this with
```ooc
result := !this _canceled ? this _backend dequeue(null) : null
```

marcusnaslund: I wish. `error Using different types in a ternary expression is forbidden. Find another way :)`
and if I change it to `null as T` I get a segfault.

emilwestergren: Nevermind then :)

PR #1616, opened 2016-04-05
Removed almost all remaining memory leaks in Net

marcusnaslund:
Only 190 bytes in 8 blocks left, but they're a bit more complicated (related to the use of `LinkedList` or `TCPReaderWriterPair`.


marcusnaslund:

PR #1618, opened 2016-04-06
No more leaks in Net module

### LINKED ISSUE #757: Write tests for, and fix memory leaks in, source/net

marcusnaslund:
That's the last of 'em.

Fixes #757

### DIFF
--- a/source/net/ServerSocket.ooc
+++ b/source/net/ServerSocket.ooc
@@ -103,14 +103,14 @@ ServerSocket: class extends Socket {
 		This method will normally block if no connection is
 		available immediately.
 	*/
-	accept: func -> TCPServerReaderWriterPair {
+	accept: func -> TCPReaderWriterPair {
 		addr: SockAddr
 		addrSize: UInt = SockAddr size
 		conn := accept(descriptor, addr&, addrSize&)
 		if (conn == -1)
 			raise("Failed to accept an incoming connection.")
 		sock := TCPSocket new(SocketAddress newFromSock(addr&, addrSize), conn)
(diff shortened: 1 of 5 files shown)


source/net/ServerSocket.ooc
```diff
@@ -138,8 +138,3 @@ ServerSocket: class extends Socket {
 		}
 	}
 }
-
-// Workaround to let TCPReaderWriterPair be in this file
```
marcusnaslund: Looks like a workaround for an older rock.

test/net/NetTest.ooc
```diff
@@ -32,7 +32,8 @@ NetTest: class extends Fixture {
 					}
 				}
 				ipaddress free()
-				clientSocket close() . free()
+				clientSocket close()
+				clientSocket readerWriter free()
```
marcusnaslund: Yes, this isn't as pretty as freeing the socket itself, but it works.

marcusnaslund:

PR #1619, opened 2016-04-06
Implement safe cancel of threads

### LINKED ISSUE #1588: Safer way of cancelling threads
I'm not sure `cancel` should be in the Thread's interface now. I think we should implement cancelling logic in the client code and use a flag to just exit the thread function, without relying on the OS facilities (which are not always there or are unsafe). Writing non-blocking client code should be preferred over forcing thread termination.

What I mean is, there is no safe way to terminate code such as
```ooc
thread := Thread new(func { while(1) doSomething() })
```
The client code should provide some cancellation points. Ideally every thread task could be a subclass of some abstract Task interface like
```ooc
AbstractThreadTask: abstract class {
    run: abstract func
    cancel: virtual func { lock(), this _runFlag = false, unlock() }
    isCancelled: virtual func -> Bool { lock(), result := this _runFlag, unlock(), return result }
}
```
and clients implementing the interface should check if the task was cancelled inside the `run` implementation, so the previous "unsafe" example becomes
```ooc
MyThreadTask: class extends AbstractThreadTask {
    run: override func {
        while(1){ 
           if( this isCancelled() ) {
               doCleanup()
               return
          }
          doSomething()
      }
    }
}
thread := Thread new(MyThreadTask new())
```
and now the `Thread` class can have a safer `cancel` implementation, which does not rely on the OS:

```ooc
Thread: abstract class {
    _task: AbstractThreadTask
    _code: func { this _task run() }
    cancel: func {
       this _task cancel() // simply sets the flag, should be safe to join or detach the thread now
    }
   ...
}
```
[ btw. Ideally we'd like to define the task subclasses in place, like in Java :) ]

I'm aware that this idea makes the Thread class less low-level and maybe slightly less convenient to use (but simple tasks could be automatically wrapped around the ThreadTask object, the use of cancellation points should be encouraged, not enforced).

marcusnaslund:
Fixes #1588

### DIFF
--- a/source/concurrent/Thread.ooc
+++ b/source/concurrent/Thread.ooc
@@ -11,20 +11,33 @@ include stdint
 import native/[ThreadUnix, ThreadWin32]
 
 Thread: abstract class {
+	counter: static Int
+	_id: Int
 	_code: Func
+	_canceled := false
+	_cancelMutex: Mutex = null
+	_ownsCancelMutex: Bool
+
+	free: override func {
+		if (this _ownsCancelMutex && this _cancelMutex != null)
+			this _cancelMutex free()
+		super()
+	}
 
 	start: abstract func -> Bool
+	cancel: abstract func -> Bool
+	isCanceled: abstract func -> Bool
 	wait: abstract func -> Bool
 	wait: abstract func ~timed (seconds: Double) -> Bool
 	detach: abstract func -> Bool
-	cancel: abstract func -> Bool
 	alive: abstract func -> Bool
-	new: static func (._code) -> This {
+
+	new: static func (._code, cancelable := false, cancelMutex := null as Mutex) -> This {
 		result: This = null
 		version (unix || apple)
-			result = ThreadUnix new(_code) as This
+			result = ThreadUnix new(_code, cancelable, cancelMutex) as This
 		version (windows)
-			result = ThreadWin32 new(_code) as This
+			result = ThreadWin32 new(_code, cancelable, cancelMutex) as This
 		if (result == null)
 			Exception new(This, "Unsupported platform!\n") throw()
 		result
--- a/test/concurrent/ThreadTest.ooc
+++ b/test/concurrent/ThreadTest.ooc
@@ -33,7 +33,7 @@ ThreadTest: class extends Fixture {
 		threadStarted free()
 	}
 	_testCancelation: static func {
-		expectedValue := 1_000
+		expectedValue := 10
 		value := Cell<Int> new(0)
 		mutex := Mutex new()
 		startedCondition := WaitCondition new()
@@ -45,15 +45,17 @@ ThreadTest: class extends Fixture {
 				value set(value get() + 1)
 				Thread yield()
 				Time sleepMilli(1)
+				if (Thread currentThread() isCanceled())
+					break
 			}
 		}
-		thread := Thread new(job)
+		/*thread := Thread new(job)
 		expect(thread start())
 		thread wait()
 		expect(value get(), is equal to(expectedValue))
 		thread free()
-		value set(0)
-		thread = Thread new(job)
+		value set(0)*/
+		thread := Thread new(job, true)
 		mutex lock()
 		expect(thread start())
 		startedCondition wait(mutex)
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -27,7 +27,9 @@ Promise: abstract class {
 		collector add(other)
 		collector
 	}
-	start: static func (action: Func) -> This { _ThreadPromise new(action) }
+	start: static func (action: Func, cancelable := false, cancelMutex := null as Mutex) -> This {
+		_ThreadPromise new(action, cancelable, cancelMutex)
+	}
 	empty: static This { get { _EmptyPromise new() } }
 }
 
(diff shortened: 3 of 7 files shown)


source/concurrent/Thread.ooc
```diff
@@ -12,19 +12,39 @@ import native/[ThreadUnix, ThreadWin32]
 
 Thread: abstract class {
 	_code: Func
+	_canceled := false
+	_cancelable: Bool
+	_cancelMutex: Mutex = null
+	_ownsCancelMutex: Bool
```
sebastianbaginski: I'm still not sure if we shouldn't move all of this to a separate `Task` class, this seems like quite a lot of data members to add in `Thread`. This way Thread could accept a Task object, which wraps the cancelling / mutex ownership logic internally. Maybe another PR later ?

marcusnaslund: Three bools and a mutex seems reasonable to me. Let me get it working and we can evaluate later.

marcusnaslund: We don't need `_cancelable: Bool`, we can just check if `_cancelMutex != null`

sebastianbaginski: OK, great. I see this does not even have the PR flag, I won't bother you again:)

test/concurrent/ThreadTest.ooc
```diff
@@ -45,15 +45,17 @@ ThreadTest: class extends Fixture {
 				value set(value get() + 1)
 				Thread yield()
 				Time sleepMilli(1)
+				if (Thread currentThread() isCanceled())
```
marcusnaslund: We've got a problem here. `currentThread()` is
```ooc
thread := This new(func)
thread pthread = pthread_self()
thread
```
meaning it, first of all, creates a new `ThreadUnix` instance every time (which must be freed). That's OK.

But more importantly here, this new instance will not have the *same* `cancelMutex` or `canceled` members, and so checking if it's canceled will always return false... :( I can solve it by adding some reference variable which, if not null, propagates the `isCanceled` call to that thread instance. That's not too pretty though. Any other idea? @sebastianbaginski

sebastianbaginski: To be honest I don't like implementing it in the base Thead class and solving this by introducing reference variables or global Thread tables seems kind of hackish. Implement it in the `ThreadTask` abstract class instead and move the responsibility onto the user, this way you will have a clear abstraction layer where we can add more functionallity later. By default the task can be wrapped around a stub object without the cancel functionality so you can even keep the current interface (and behaviour too in most cases, as cancelling does not work on android). It will be quite easy to write a Task subclass with shared mutex for the `TheadPool` too.

sebastianbaginski: By `global thread tables` I mean that `Thread currentThread()` could return the same `instance` each time, assuming that instances are stored in a `thread Id -> thread instance` global hash table. Do you find my idea as ugly as I do @marcusnaslund ? :)

marcusnaslund: > Do you find my idea as ugly as I do

Yes. :) 

I have an idea about creating a `ThreadData` class which could contain the mutex and the flags and all the associated logic, and then `currentThread()` would just pass that instance along to the new thread instance. No fuss and the thread class itself is kept clean. 

I'm not saying your idea isn't good, but I'm not completely sold as it's more complex to implement and the java approach just isn't as great without interfaces (`Runnable`).

emilwestergren: What about something like this:
```ooc
Thread: class
     init: func (Func (currentThread: Thread))
     run: func { this _code(this) }

t := Thread new( |currentThread|
        while(!currentThead cancelled)
        //do stuff

```

marcusnaslund: @emilwestergren Hmm, that works too. Creating a `Thread` with a `Func` would mean an uncancelable thread, and a `Thread` with a `Func (Thread)` would create a cancelable thread.

The limitations are that only that func can check if the thread is canceled. If it calls any other functions, it has to pass the thread instance along in order to allow them to check if the thread is canceled. I haven't checked how pretty this integrates with the threadpool either.

sebastianbaginski: @marcusnaslund I don't think it is complex, as it requires to introduce one class, if we make the base ThreadTask not abstract and not cancellable. C++ has no interfaces too and such approach would work just fine. I don't care if it is an interface or something else, I just want it to be modular and flexible.

marcusnaslund: @sebastianbaginski It's not too complex to implement, but it requires more effort/complexity to actually get something cancelable running on a thread. I prefered my approach because it would allow you to just supply a func the same way as before, but I see now we will need a global `ThreadTable` to handle it, and the beauty goes away.

I guess we can keep the `Func` constructor for `Thread` that automatically creates a (non-cancelable) task and disposes it when the thread is freed, that way we don't need to change as much in the code. Cancelable tasks would need to be manually created and supplied to the thread. We already have a `Task` system for the `ThreadPool` that may or may be partially replaced by this new `ThreadTask`.

I guess I will close this for now and we can pick up the discussion in #1588 or IRL.

sebastianbaginski: I understand. I just wanted to avoid hard-coding it or changing the base `Thread` class too much, because I feel this can be moved to a separate reusable abstraction layer (merging with already existing Tasks from ThreadPool would be ideal).

PR #1620, opened 2016-04-06
Fixed error code mess in test.sh

marcusnaslund:
Checking `if [[ $? -eq 0 && -f ./Tests ]]; then` would generate a new exit code that was always 0, it needs to be stored and updated inside the if-statement if neccessary.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -41,10 +41,12 @@ done
 echo "Main: ./test/Tests.ooc" >> "$TESTS_USE_FILE"
 rm -f .libs/tests-linux64.*
 rock -q +-Wall $ARGS $FLAGS $FEATURES $TESTS_USE_FILE
-if [[ $? -eq 0 && -f ./Tests ]]; then
+exitcode=$?
+if [[ $exitcode -eq 0 && -f ./Tests ]]; then
 	./Tests
+	exitcode=$?
 fi
-if [[ !( $? == 0 ) ]]
+if [[ !( $exitcode == 0 ) ]]
 then
 	exit 1
 else


marcusnaslund:
Check the PRInt status for the kean tests of #1619 for when the previous way failed.

test.sh
```diff
@@ -41,10 +41,12 @@ done
 echo "Main: ./test/Tests.ooc" >> "$TESTS_USE_FILE"
 rm -f .libs/tests-linux64.*
 rock -q +-Wall $ARGS $FLAGS $FEATURES $TESTS_USE_FILE
-if [[ $? -eq 0 && -f ./Tests ]]; then
+exitcode=$?
+if [[ $lol -eq 0 && -f ./Tests ]]; then
```
sebastianbaginski: `$lol` ?

marcusnaslund: Hahahah *oops*

marcusnaslund: Updated.

test.sh
```diff
@@ -41,10 +41,12 @@ done
...
 	./Tests
+	exitcode=$?
 fi
-if [[ !( $? == 0 ) ]]
+if [[ !( $exitcode == 0 ) ]]
```
marcusnaslund: Why do we do this? Why not just `exit $exitcode`? @thomasfanell , you wrote this block eight months ago, do you know?

thomasfanell: I can't really remember to be honest, but if I had to guess, we had reason to either return `1` or `0` to keep it simple.

PR #1622, opened 2016-04-07
Writing text to images

### LINKED ISSUE #1438: Render text
It would be useful to be able to render text onto images.

@davidpiuva @sebastianbaginski

davidpiuva:
Writing text to images using multiple draw calls. CPU and OpenGL implementations can override the virtual method in canvas for better performance. Trying to write text on the CPU will currently fail since DrawState is not yet implemented for the CPU.

DrawState contains a new setting called origin because setting a destination of zero area has too much risk of missunderstandings. The write method that does all the work is placed in canvas taking only the arguments that are actually used to reduce the input space to test and it is virtual to allow specific CPU and OpenGL implementations to be made while comparing with the reference implementation in Canvas. The write method in Canvas can get more input arguments since it will only be called by the wrapper function in DrawState.

Fixes #1438


source/draw/DrawState.ooc
```diff
@@ -132,9 +134,29 @@ DrawState: cover {
...
+	}
+	// Example: DrawState new(targetImage) setMap(shader) draw()
 	draw: func {
 		version(safe)
-			raise(this target == null, "Can't draw without a selected target.")
+			raise(this target == null, "Can't draw without a target.")
 		this target canvas draw(this)
 	}
+	// Example: DrawState new(targetImage) setOrigin(point) setInputImage(context getDefaultFont()) write(t"Hello world!")
+	write: func ~String (message: Text) {
```
davidpiuva: Wrapper to allow extending with more settings without changing each caller.

source/draw/gpu/GpuContext.ooc
```diff
@@ -35,7 +35,20 @@ ToRasterFuture: class extends Future<RasterImage> {
 
 GpuContext: abstract class extends DrawContext {
 	defaultMap ::= null as Map
+	_defaultFontGpu: GpuImage = null
+	defaultFontGpu: GpuImage {
```
davidpiuva: Create the GPU font from the CPU font if needed and return it.

source/draw/DrawContext.ooc
```diff
@@ -17,7 +18,20 @@ AlignWidth: enum {
 }
 
 DrawContext: abstract class {
+	_defaultFontRaster: RasterMonochrome = null
+	defaultFontRaster: RasterMonochrome {
```
davidpiuva: Create the CPU font from the ascii image if needed and return it.

source/draw/Canvas.ooc
```diff
@@ -67,6 +67,30 @@ Canvas: abstract class {
 	}
 	fill: abstract func (color: ColorRgba)
 	draw: virtual func ~DrawState (drawState: DrawState) { Debug error("draw~DrawState unimplemented for class " + this class name + "!") }
+	write: virtual func (target, fontAtlas: Image, localOrigin: IntPoint2D, message: Text) {
```
davidpiuva: The actual implementation that makes testing easier by only taking used settings from DrawState. This function will have settings added when needed instead of overloading. DrawState will handle all alternative data-types and coordinate-systems.

source/draw/DrawState.ooc
```diff
@@ -22,11 +20,15 @@ BlendMode: enum {
 }
 
 DrawState: cover {
+	// Used by all
 	target: Image = null
 	inputImage: Image = null
+	// Only used by the write function
+	_originReference := FloatPoint2D new(0.0f, 0.0f) // Start of text
```
davidpiuva: The text will have the origin as the upper left corner of the first character in reference coordinates.

davidpiuva:

sebastianbaginski:
This is very nice, but I feel like @emilwestergren is a better candidate for this PR.

davidpiuva:
ok

davidpiuva:
@emilwestergren Time for a peer review?

source/draw/images/DefaultFont.ooc
```diff
@@ -0,0 +1,206 @@
...
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+// Original typeface for printing debug messages that was crudely drawn by David Piuva in a few minutes because it was quicker than understanding the license terms of existing fonts.
+
+GlobalCleanup register(|| DefaultFontAsciiImage free())
+DefaultFontAsciiImage := "
```
davidpiuva: The style is a bit inconsistent because only `I` and `J` uses serifs. `I` has to differ from `l`, `J` has to be consistent with `I` and characters have fixed width rendering so that it can later be GPU optimized.

davidpiuva:
Seems like Emil is not here today.

davidpiuva:

source/draw/Canvas.ooc
```diff
@@ -67,6 +67,30 @@ Canvas: abstract class {
...
+		// Generic reference implementation with all inputs that are used from DrawState to simplify input space partitioning
+		skippedRows := 2 // Skip invisible commands
+		visibleRows := 6 // Use 6 lines of printable characters
+		columns := 16 // Use 16 characters per row
```
marcusnaslund: Call it `charactersPerRow` and you don't need the comment

source/draw/DrawContext.ooc
```diff
@@ -17,7 +18,20 @@ AlignWidth: enum {
...
+		get {
+			if (this _defaultFontRaster == null)
+				this _defaultFontRaster = RasterMonochrome fromAscii(DefaultFontAsciiImage)
+			this _defaultFontRaster
+		}
+	}
```
marcusnaslund: The correct way to write this is
```ooc
property: Type { get {
    ...
}}
```

source/draw/gpu/GpuContext.ooc
```diff
@@ -35,7 +35,20 @@ ToRasterFuture: class extends Future<RasterImage> {
...
+		get {
+			if (this _defaultFontGpu == null)
+				this _defaultFontGpu = this createImage(this defaultFontRaster)
+			this _defaultFontGpu
+		}
+	}
```
marcusnaslund: The correct way to write this is
```ooc
property: Type { get {
    ...
}}
```

source/draw/images/DefaultFont.ooc
marcusnaslund: `defaultFontAsciiImage` since it's a variable?

source/draw/DrawState.ooc
```diff
@@ -22,8 +20,12 @@ BlendMode: enum {
...
+	// Only used by the draw function
```
marcusnaslund: These variable-use-comments shouldn't be here

davidpiuva: Then where?

marcusnaslund: Is there a need for them at all?

davidpiuva: Trying to draw a quad with different origin points and see no effect at all would be very confusing since covers don't have inheritance.

source/draw/Canvas.ooc
davidpiuva: That could indicate that the output uses word wrapping at 16 characters if there is no reference to a 2D atlas.

davidpiuva:
Partially updated.

source/draw/Canvas.ooc
```diff
@@ -67,6 +67,30 @@ Canvas: abstract class {
 	}
 	fill: abstract func (color: ColorRgba)
 	draw: virtual func ~DrawState (drawState: DrawState) { Debug error("draw~DrawState unimplemented for class " + this class name + "!") }
+	write: virtual func (target, fontAtlas: Image, localOrigin: IntPoint2D, message: Text) {
+		// Generic reference implementation with all inputs that are used from DrawState to simplify input space partitioning
+		skippedRows := 2
+		visibleRows := 6
+		columns := 16
```
davidpiuva: I just removed the comments since we can be without them.

source/draw/DrawState.ooc
davidpiuva: How about a table in markdown showing availability of settings for different drawing commands?

marcusnaslund: We have some documentation for the draw stuff already, right? That would be more appropriate there, then.

source/draw/README.md
```diff
@@ -44,6 +44,22 @@ Transforms go from zero in the center and increase with either pixels or pixels
 Using `DrawState` you can draw things without having to call the canvas directly.
 This is convenient if you have a lot of things to draw that have settings in common.
 
+## Availability of settings
```
davidpiuva: I now have a table.

marcusnaslund: Excellent.

source/draw/Canvas.ooc
```diff
@@ -67,6 +67,30 @@ Canvas: abstract class {
...
+			charCode := message[i] as Int
+			sourceX := charCode % columns
+			sourceY := (charCode / columns) - skippedRows
+			source := FloatBox2D new((sourceX as Float) / columns, (sourceY as Float) / visibleRows, 1.0f / columns, 1.0f / visibleRows)
+			if (charCode > 32 && charCode < 127)
+				characterDrawState setViewport(viewport + (targetOffset * charSize)) setSourceNormalized(source) draw()
+			targetOffset x += 1
+			if (charCode == '\n') {
+				targetOffset x = 0 // Carriage return
+				targetOffset y += 1 // Line feed
```
fredrikbryntesson: Carriage return, love it.

source/draw/DrawState.ooc
```diff
@@ -132,9 +132,29 @@ DrawState: cover {
...
+	// Reference point
+	setOrigin: func (origin: FloatPoint2D) -> This {
+		this _originReference = origin
+		this
+	}
+	// Local point
+	getOriginLocal: func -> FloatPoint2D {
+		version(safe)
+			raise(this target == null, "Can't get local origin relative to a target that does not exist.")
+		this _originReference + ((this target size toFloatVector2D()) / 2.0f)
```
fredrikbryntesson: The three commens above, do we need them?

davidpiuva: We need them where coordinate system is implicit and the rest is for consistent style.

fredrikbryntesson: The method is   `getTransformNormalized: func -> FloatTransform3D { this _transformNormalized }`, I do not see how `// Normalized transform` give me any new information. Same thing with   `getOriginLocal: func -> FloatPoint2D` and  `setOrigin: func (origin: FloatPoint2D) -> This `

davidpiuva: I don't know where to move the information if I can't even document implicit method names.

davidpiuva: Do I have to remove all comments? That getTransformNormalized is visible here is just random.

marcusnaslund:
Does this fix #1438 ?

davidpiuva:
Yes, mostly. It can be made faster but it is already fast enough for our intended use cases.

marcusnaslund:
Awesome. It's a good idea to add "Fixes #1438" to the PR description, then that issue is automatically closed when this PR is merged.

PR #1623, opened 2016-04-07
Added point+vector arithmetic in 3D

marcusnaslund:
This was possible in 2D, but not in 3D. Will be useful in the other project. Peer review @bettinaselig ?

### DIFF
--- a/source/geometry/FloatPoint3D.ooc
+++ b/source/geometry/FloatPoint3D.ooc
@@ -9,6 +9,7 @@
 use math
 import FloatPoint2D
 import IntPoint3D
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Vector + Point = Vector, which doesn't make a lot of sense to me, but it's the same way in 2D.

bettinaselig:
Do you need that for our project?

bettinaselig:
It's a bit strange as you say, but if you need that I don't mind :-)

marcusnaslund:
> Do you need that for our project?

Yes, exactly. :)

PR #1626, opened 2016-04-08
removed possible data races related to string literals

sebastianbaginski:
Attempt to make string literals initialization thread safe. Fixes some warnings in helgrind.
@marcusnaslund what do you think about hard-coding these simple mutex operations here ? maybe you have a better alternative ?

### DIFF
--- a/source/system/string/string_literal.c
+++ b/source/system/string/string_literal.c
@@ -1,13 +1,37 @@
 #include "string_literal.h"
 #include <assert.h>
 #include <ooc/system/String-fwd.h>
+#include <ooc/system/Mutex.h>
 
+#if defined(__WIN32__) || defined(__WIN64__)
+	typedef Handle mutex_t;
+	#define MUTEX_INIT NULL
+	static void mutex_lock(mutex_t* m) {
+		if (*m == NULL) {
+			HANDLE mutex = CreateMutex(NULL, FALSE, NULL);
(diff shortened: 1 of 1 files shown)


marcusnaslund:
I see the point of it, but I'm not all to keen about having String literals being dependent on our mutex implementation.

sebastianbaginski:
You mean the `#include <ooc/system/Mutex.h>` ? I think I should change this to pthread.h and windows.h in the ifdef blocks.

marcusnaslund:
Yeah. When are stringliterals created? Are there any possible performance impacts from this?

sebastianbaginski:
They are created when the corresponding class (in which literals are used) is first loaded, this is done once per run.

source/system/string/string_literal.c
```diff
@@ -1,31 +1,58 @@
 #include "string_literal.h"
 #include <assert.h>
 #include <ooc/system/String-fwd.h>
+#include <ooc/system/Mutex.h>
```
marcusnaslund: This is probably better actually, then we get the correct `define`s for `pthread.h` and `windows.h` (as they are written in `system/external`).

marcusnaslund:
That's what I thought. Seems reasonable then, and it's not like possible data races are pleasant either.
I will let @emilwestergren peer review this, though.

PR #1628, opened 2016-04-08
Using RecycleBin in AndroidContext and OpenGLContext

### LINKED ISSUE #1374: Take care of _RecycleBin
The class in `source/draw/gpu/opengl/_RecycleBin.ooc` has a terrible name, but the underscore is there so that it does not conflict with the newer `source/concurrent/RecycleBin.ooc`. We should fix this, obviously.

- [ ] `OpenGLContext`
- [ ] `AndroidContext`

marcusnaslund:
Removes the old `_RecycleBin` and implements `concurrent/RecycleBin` in `AndroidContext` and `OpenGLContext`.

Fixes #1374 

The only functional difference is that each of the four bins in `OpenGLContext` has a separate mutex. Previously they shared one. If need be, we can solve this by letting the `RecycleBin` take a mutex as argument.


source/system/structs/VectorList.ooc
```diff
@@ -68,7 +68,7 @@ VectorList: class <T> extends List<T>{
 	search: override func (matches: Func (T) -> Bool) -> Int {
 		result := -1
 		for (index in 0 .. this count)
-			if (matches(this[index]&)) {
+			if (matches(this[index])) {
```
marcusnaslund: The fact that the old way worked is.. bizarre.

source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -19,7 +19,7 @@ AndroidContext: class extends OpenGLContext {
 	_unpackRgbaToMonochrome := OpenGLMap new(slurp("shaders/unpack.vert"), slurp("shaders/unpackRgbaToMonochrome.frag"), this)
 	_unpackRgbaToUv := OpenGLMap new(slurp("shaders/unpack.vert"), slurp("shaders/unpackRgbaToUv.frag"), this)
 	_unpackRgbaToUvPadded := OpenGLMap new(slurp("shaders/unpack.vert"), slurp("shaders/unpackRgbaToUvPadded.frag"), this)
-	_packers := VectorList<EGLRgba> new()
+	_packers := RecycleBin<EGLRgba> new(32, func (image: EGLRgba) { image free() })
```
marcusnaslund: Or `referenceCount decrease()`?

source/concurrent/SynchronizedList.ooc
```diff
@@ -66,7 +66,7 @@ SynchronizedList: class <T> extends List<T> {
 		this _mutex unlock()
 		result
 	}
-	search: override func (matches: Func (T*) -> Bool) -> Int {
+	search: override func (matches: Func (T) -> Bool) -> Int {
```
marcusnaslund: The next version of rock will thankfully catch these override mismatches.

marcusnaslund:
Peer review @emilwestergren ?

source/concurrent/RecycleBin.ooc
```diff
@@ -19,6 +19,7 @@ RecycleBin: class <T> {
 	free: override func {
 		this clear()
 		this _list free()
+		(this _free as Closure) free()
```
emilwestergren: `(this _free as Closure) free(Owner Receiver)`

source/draw/gpu/opengl/OpenGLContext.ooc
```diff
@@ -38,13 +38,32 @@ OpenGLContext: class extends GpuContext {
 	_pointsShader: OpenGLMap
 	_meshShader: OpenGLMapMesh
 	_renderer: GLRenderer
-	_recycleBin := _RecycleBin new()
+	_recycleBinMonochrome: RecycleBin<OpenGLMonochrome>
```
emilwestergren: Do we really need one for each type? Seems like a premature optimization to me.

emilwestergren: You don't have do fix that here, just for the future.

marcusnaslund: Not sure, but it's not like ooc generics have a reputation for being reliable. I can experiment later for a future PR if you want.

source/concurrent/RecycleBin.ooc
marcusnaslund: Updated.

emilwestergren:
Just make sure it is well tested on a device.

source/draw/gpu/opengl/OpenGLContext.ooc
emilwestergren: `_bin: RecycleBin<OpenGLPacked>`

`this _bin search(|image| image instanceOf(OpenGLMonochrome) && image size == size)`

Something like that.

marcusnaslund: The downside of that is that it's harder to limit to exactly 15 of each type. We would recycle 60 of any type.

marcusnaslund:
Runs fine on Android as far as I can tell (been recording many and long clips) and no leaks problems on PC either.

source/draw/gpu/opengl/OpenGLContext.ooc
emilwestergren: That's not an issue at the moment, that's why I feel it's a premature optimization. I think we can go even further and begin with a single bin and optimize it when we need to. It would make the code a lot cleaner.

marcusnaslund: I will look at it

PR #1629, opened 2016-04-08
Added scalar product of two FloatVectorLists

tomasmahlberg:


tomasmahlberg:
@marcusnaslund

test/math/FloatVectorListTest.ooc
```diff
@@ -144,6 +144,23 @@ FloatVectorListTest: class extends Fixture {
 			list1 free()
 			list2 free()
 		})
+		this add("scalar product", func {
+			list1 := FloatVectorList new()
+			list2 := FloatVectorList new()
+			list1 add(1.0f)
+			list1 add(2.0f)
```
marcusnaslund: `list1 add(1.f) . add(2.f)` and similarly on two other places here

test/math/FloatVectorListTest.ooc
```diff
@@ -144,6 +144,23 @@ FloatVectorListTest: class extends Fixture {
...
+			list2 add(11.0f)
+			result := list1 scalarProduct(list2)
+			expect(result, is equal to(32.f) within(tolerance))
+			list1 add(3.0f)
+			list1 add(4.0f)
+			list2 add(12.0f)
+			result = list1 scalarProduct(list2)
+			expect(result, is equal to(68.f) within(tolerance))
+			list1 free()
+			list2 free()
```
marcusnaslund: `(list1, list2) free()?`

test/math/FloatVectorListTest.ooc
```diff
@@ -144,6 +144,23 @@ FloatVectorListTest: class extends Fixture {
...
 			list2 free()
 		})
+		this add("scalar product", func {
+			list1 := FloatVectorList new()
+			list2 := FloatVectorList new()
+			list1 add(1.0f)
+			list1 add(2.0f)
+			list2 add(10.0f)
+			list2 add(11.0f)
+			result := list1 scalarProduct(list2)
```
marcusnaslund: Do you have to save it to a temp variable?

source/math/FloatVectorList.ooc
```diff
@@ -306,6 +306,12 @@ FloatVectorList: class extends VectorList<Float> {
 			result = this copy()
 		result
 	}
+	scalarProduct: func (other: This) -> Float {
+		result := 0.f
+		for (i in 0 .. this count minimum(other count))
+			result += this[i] * other[i]
+		result
```
marcusnaslund: We *could* also do this by
```ooc
product := this * other
product sum
```

tomasmahlberg: Yes, but then you have to instantiate a new `VectorList` that you have to free. This seems like a useful function to have to me.

marcusnaslund: Sure :+1:

test/math/FloatVectorListTest.ooc
tomasmahlberg: I can change it. A general question though; even though this follows the style guide, the multiple places where this is done in the rest of the file does not do it like this. Would it always be advised to follow the style guide in this case or keep the style consistent with the rest of the file?

test/math/FloatVectorListTest.ooc
tomasmahlberg: Probably not!

test/math/FloatVectorListTest.ooc
tomasmahlberg: See above.

test/math/FloatVectorListTest.ooc
marcusnaslund: The reason it's not like that elsewhere in the file is because this is a very new concept and I've prioritized fixing files in source over tests. The tests will be fixed eventually though, so I'd rather see people getting used to this approach.

tomasmahlberg:
Updated.

test/math/FloatVectorListTest.ooc
```diff
@@ -144,6 +144,17 @@ FloatVectorListTest: class extends Fixture {
...
+			list1 := FloatVectorList new()
+			list2 := FloatVectorList new()
+			list1 add(1.0f) . add(2.0f)
+			list2 add(10.0f) . add(11.0f)
+			expect(list1 scalarProduct(list2), is equal to(32.f) within(tolerance))
+			list1 add(3.0f) . add(4.0f)
+			list2 add(12.0f)
+			expect(list1 scalarProduct(list2), is equal to(68.f) within(tolerance))
+			(list1, list2) free()
+		})
```
marcusnaslund: Much better, thanks. :)

PR #1630, opened 2016-04-08
Cleaned up test.sh

marcusnaslund:
I like how we had three styles for if-statements. Now just one.

Peer review @fredrikbryntesson

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -2,9 +2,9 @@
 export OOC_LIBS=$(dirname `pwd`)
 ARGS=""
 TESTS_USE_FILE="tests.use"
-if [ -e rock_arguments.txt ]
-	then
-		ARGS=$(cat rock_arguments.txt)
+if [[ -e rock_arguments.txt ]]
+then
+	ARGS=$(cat rock_arguments.txt)
 fi
 FLAGS=${@:2}
 TESTS=$@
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
Why this style ? I think keeping `if ; then` on a single line is more readable. Kind of like
```C
if () {
}
```
vs
```C
if ()
{
}
```

marcusnaslund:
Sure, but there were a lot more of this style than the other. I can change them to one-line if that's "better".

sebastianbaginski:
Looks like it's time for a `bash` chapter in the style guide ;)

marcusnaslund:
Actually, in that case, we should do `for` and `while` loops on the same line as well. Should we? Can we?

sebastianbaginski:
I think we can, but it sounds like something to discuss with @thomasfanell and @fredrikbryntesson ?

marcusnaslund:
Agreed. Postponed until monday. Have a nice weekend. :)

marcusnaslund:
It's decided. Peer review @sebastianbaginski ?

test.sh
```diff
@@ -40,15 +42,12 @@ do
...
 	./Tests
 	exitcode=$?
 fi
-if [[ !( $exitcode == 0 ) ]]
-then
-	exit 1
-else
-	exit 0
-fi
+exit $exitcode
```
thomasfanell: You might want to change `test/Tests.ooc` to reflect this change then, since it wants either a 0 or a 1.

marcusnaslund: How do you mean? `Tests.ooc` does not read output from `test.sh`, it's the other way around.

thomasfanell: Well that was embarrassing. Pardon me while I go inject 10 cups of coffee directly into my brain.

marcusnaslund: Everything is forgiven on Mondays. :)

PR #1632, opened 2016-04-09
wrapper for basic openssl functionality

sebastianbaginski:
- base file for openssl types and functions
- test cases for AES encryption (both hard-coded input and random data)
- test case for SHA-256 (hard-coded input)
- updated `Cell toText` to support `Byte`

@marcusnaslund What do you think, could we have this in the sdk ?

### DIFF
--- a/source/ssl/include/ssl.ooc
+++ b/source/ssl/include/ssl.ooc
@@ -1,0 +1,30 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+include openssl/rand, openssl/aes, openssl/crypto, openssl/evp
+
+AesKey: cover from AES_KEY
+
+AES_BLOCK_SIZE,
+AES_ENCRYPT,
+AES_DECRYPT: extern const Int
+
+OPENSSL_VERSION_NUMBER: extern const Long
+
+SSLEAY_VERSION,
+SSLEAY_CFLAGS,
+SSLEAY_BUILT_ON,
+SSLEAY_PLATFORM,
+SSLEAY_DIR: extern const Int
+
+AES_set_encrypt_key: extern func (keyData: Byte*, keyLength: Int, aesKey: AesKey*)
+AES_set_decrypt_key: extern func (keyData: Byte*, keyLength: Int, aesKey: AesKey*)
+AES_cbc_encrypt: extern func (input: const Byte*, output: Byte*, inputLength: SizeT, key: AesKey*, initVector: Byte*, mode: Int)
+RAND_bytes: extern func (output: Byte*, length: Int) -> Int
+SSLeay: extern func -> Long
+SSLeay_version: extern func (type: Int) -> CString
--- a/source/ssl/include/ssl.ooc
+++ b/source/ssl/include/ssl.ooc
@@ -6,9 +6,12 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-include openssl/rand, openssl/aes, openssl/crypto, openssl/evp
+include openssl/aes, openssl/crypto, openssl/evp, openssl/rand
 
 AesKey: cover from AES_KEY
+
+EvpMessageDigestContext: cover from EVP_MD_CTX
+EvpMessageDigest: cover from EVP_MD
 
 AES_BLOCK_SIZE,
 AES_ENCRYPT,
@@ -25,6 +28,18 @@ SSLEAY_DIR: extern const Int
 AES_set_encrypt_key: extern func (keyData: Byte*, keyLength: Int, aesKey: AesKey*)
 AES_set_decrypt_key: extern func (keyData: Byte*, keyLength: Int, aesKey: AesKey*)
 AES_cbc_encrypt: extern func (input: const Byte*, output: Byte*, inputLength: SizeT, key: AesKey*, initVector: Byte*, mode: Int)
+
+EVP_MD_CTX_create: extern func -> EvpMessageDigestContext*
+EVP_MD_CTX_destroy: extern func (context: EvpMessageDigestContext*)
+EVP_DigestInit: extern func (context: EvpMessageDigestContext*, type: EvpMessageDigest*) -> Int
+EVP_DigestUpdate: extern func (context: EvpMessageDigestContext*, data: Void*, size: SizeT) -> Int
+EVP_DigestFinal: extern func (context: EvpMessageDigestContext*, output: Byte*, bytesWritten: UInt*) -> Int
+
+EVP_md5: extern func -> EvpMessageDigest*
+EVP_sha1: extern func -> EvpMessageDigest*
+EVP_sha256: extern func -> EvpMessageDigest*
+EVP_sha512: extern func -> EvpMessageDigest*
+
 RAND_bytes: extern func (output: Byte*, length: Int) -> Int
 SSLeay: extern func -> Long
 SSLeay_version: extern func (type: Int) -> CString
--- a/test/ssl/OpenSslTest.ooc
+++ b/test/ssl/OpenSslTest.ooc
@@ -1,0 +1,100 @@
+use base
+use ssl
+use unit
+
+import include/ssl
+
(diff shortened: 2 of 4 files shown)


test/ssl/OpenSslTest.ooc
```diff
@@ -0,0 +1,121 @@
...
+		this add("random", func {
+			length := 16
+			buffer1: Byte[length]
+			buffer2: Byte[length]
+			for (i in 0 .. 32) {
+				expect(RAND_bytes(buffer1[0]&, length), is equal to(1))
+				expect(RAND_bytes(buffer2[0]&, length), is equal to(1))
+				outputDifferent := false
+				for (b in 0 .. length)
+					if (buffer1[b] != buffer2[b]) {
```
marcusnaslund: Hmm, there should be `==` operator for arrays, can we implement that?

test/ssl/OpenSslTest.ooc
```diff
@@ -0,0 +1,121 @@
...
+	init: func {
+		super("OpenSsl")
+		this add("version", func {
+			expect(OPENSSL_VERSION_NUMBER != 0)
+			expect(OPENSSL_VERSION_NUMBER, is equal to(SSLeay()))
+		})
+		this add("random", func {
+			length := 16
+			buffer1: Byte[length]
+			buffer2: Byte[length]
```
marcusnaslund: Both on one line, please

test/ssl/OpenSslTest.ooc
```diff
@@ -0,0 +1,121 @@
...
+						isEncrypted = true
+						break
+					}
+				expect(isEncrypted)
+
+				AES_set_decrypt_key(password[0]&, keyLengthBits, decryptionKey&)
+				AES_cbc_encrypt(cipherText[0]&, decipheredText[0]&, encryptedLength, decryptionKey&, initVectorDecrypt[0]&, AES_DECRYPT)
+
+				for (i in 0 .. inputLength)
+					if (plainText[i] != decipheredText[i])
```
marcusnaslund: `expect(plainText[i] == decipheredText[i]` saves you one line. :)

test/ssl/OpenSslTest.ooc
```diff
@@ -0,0 +1,121 @@
...
+				0x79, 0x0d, 0x35, 0x1c,
+				0x96, 0xc4, 0x7c, 0x02,
+				0x19, 0x0e, 0xcf, 0x1e,
+				0x43, 0x06, 0x35, 0xab ]
+			context := EVP_MD_CTX_create()
+			expect(EVP_DigestInit(context, EVP_sha256()), is equal to(1))
+			expect(EVP_DigestUpdate(context, c"abcdefgh", 8), is equal to(1))
+			result: Byte[32]
+			outputSize: UInt
+			expect(EVP_DigestFinal(context, result[0]&, outputSize&), is equal to(1))
```
marcusnaslund: Isn't it possible to do `result` instead of `result[0]&`?

source/ssl/include/ssl.ooc
```diff
@@ -0,0 +1,45 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
```
marcusnaslund: So `source/ssl` contains... just a folder? I first of all think the folder should rather be named `external` (like in `system`), but more importantly I'd be OK with just having this file reside in `source/ssl` directly.

source/ssl/include/ssl.ooc
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+include openssl/aes, openssl/crypto, openssl/evp, openssl/rand
+
+AesKey: cover from AES_KEY
```
marcusnaslund: I would expect `AES` and `SSL`, since they are acronyms, to be treated differently from names. Thus `AESKey`.

source/ssl/include/ssl.ooc
```diff
@@ -0,0 +1,45 @@
...
+
+OPENSSL_VERSION_NUMBER: extern const Long
+
+SSLEAY_VERSION,
+SSLEAY_CFLAGS,
+SSLEAY_BUILT_ON,
+SSLEAY_PLATFORM,
+SSLEAY_DIR: extern const Int
+
+AES_set_encrypt_key: extern func (keyData: Byte*, keyLength: Int, aesKey: AesKey*)
```
marcusnaslund: I'm OK with keeping function names from C libraries to make it easier to look stuff up, on the other hand they generally follow a quite different naming convention from us. What's your opinion? We already rename classes/covers.

marcusnaslund:
Looks good besides the comments (and whatever PRInt is complaining about that I can't see at the moment).

Why do we want/need this? What's your opinion @fredrikbryntesson ?

sebastianbaginski:
> Why do we want/need this?

Not sure, I just think I saw comments about the need for encryption in some other issue.
I was doing some openssl stuff this weekend anyway, I needed a simple program to generate test data and I thought "why not in ooc" :)
Thanks for the comments, I will update the code tomorrow, if we decide to take it.

source/ssl/include/ssl.ooc
sebastianbaginski: Good quesion, here I just copy-pasted the symbols from openssl api reference.

marcusnaslund:
> I thought "why not in ooc"

Yeah, right. :)

test/ssl/OpenSslTest.ooc
sebastianbaginski: Unfortunately no, compiler complains about passing `Byte[]` instead of `Byte*`.

source/ssl/include/ssl.ooc
sebastianbaginski: Yeah, I agree this looks funny now. I though that this file could be a simple wrapper around the raw openssl api, and we could have more convenient higher level classes here in the future. Something like for opengl stuff.

sebastianbaginski:
So the other project compiled fine on PRInt machine, but kean compilation failed ? Interesting...

marcusnaslund:
This fails because

```
fatal error: openssl/aes.h: Filen eller katalogen finns inte #include <openssl/aes.h>
```

in the ssl test. Since it's not used in the other project, it compiles just fine.

sebastianbaginski:
I always thought `/usr/include/` is searched by default ? Is there no openssl on the PRInt machine ?

marcusnaslund:
I tested this on my machine and it runs fine. Hm. @jonathanudd ?

sebastianbaginski:
@thomasfanell we are wondering if this code is needed now. I think it was you who wrote something about the need for encrypted string class. What do you think, should we have this in the sdk ?

thomasfanell:
@sebastianbaginski Yeah I want it. I am thinking of using this for `encryptSlurp()` or somethings similar in rock. Good job! :+1:

source/ssl/include/ssl.ooc
sebastianbaginski: I have no idea what should we do with this. It is nice to have consistent naming style, but I think it's also good to keep as close as possible to original api in such low-level interfaces. I think in such cases we will build a second higher level interface layer anyway. @fredrikbryntesson @thomasfanell ?

marcusnaslund:
Well then, the only question mark is whether this introduces more dependencies. And how does it work on windows?

thomasfanell:
@marcusnaslund probably not hassle-free :)

thomasfanell:
I only need this in rock though, not here in magic-sdk (rock is using the original sdk). So if I am the only one that is currently in need of this, I see little value in adding it to magic-idk at this point.

marcusnaslund:
So maybe redirect the PR to the rock repo, and we can copy it here if we need it later?

marcusnaslund:
Be sure to update the "building rock on windows" markdown file though

sebastianbaginski:
OK, sounds good. @fredrikbryntesson @thomasfanell please suggest folder structure in the rock repo. Where should this be located ? `sdk/ssl/ssl.ooc` ?

fredrikbryntesson:
`sdk/ssl/ssl.ooc` should be fine.

marcusnaslund:
Works with `apt-get install libssl-dev`

PR #1635, opened 2016-04-10
Generalized some FloatMatrix methods

### LINKED ISSUE #745: Generalize FloatMatrix cofactors(), adjugate(), and determinant()
These three methods **only** work for `3x3` matrices right now.

- [ ] `cofactors`
- [ ] `adjugate`
- [ ] `determinant`

marcusnaslund:
I needed to take determinants for some larger matrices than 3x3 so figured I might as well generalize both `cofactors()` and `determinants()`.

For performance reasons, the explicit calculations for 3x3 matrices are kept intact.

Fixes #745

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -293,47 +293,92 @@ FloatMatrix: cover {
 		result
 	}
 	cofactors: func -> This {
-		// TODO: At some point, when needed, implement for the general case.
 		t := this take()
-		version(safe) {
+		version(safe)
 			raise(!t isSquare, "Matrix must be square in FloatMatrix cofactors")
-			raise(t width != 3, "Cofactors implemented only for 3x3 matrices in FloatMatrix")
-		}
 		result := t create()
-		result[0, 0] = t[1, 1] * t[2, 2] - t[2, 1] * t[1, 2]
-		result[1, 0] = - (t[0, 1] * t[2, 2] - t[2, 1] * t[0, 2])
-		result[2, 0] = t[0, 1] * t[1, 2] - t[1, 1] * t[0, 2]
-		result[0, 1] = - (t[1, 0] * t[2, 2] - t[2, 0] * t[1, 2])
-		result[1, 1] = t[0, 0] * t[2, 2] - t[2, 0] * t[0, 2]
-		result[2, 1] = - (t[0, 0] * t[1, 2] - t[1, 0] * t[0, 2])
-		result[0, 2] = t[1, 0] * t[2, 1] - t[2, 0] * t[1, 1]
-		result[1, 2] = - (t[0, 0] * t[2, 1] - t[2, 0] * t[0, 1])
-		result[2, 2] = t[0, 0] * t[1, 1] - t[1, 0] * t[0, 1]
+		matrixOrder := t order
+		if (matrixOrder == 1)
+			result[0, 0] = t[0, 0]
+		else if (matrixOrder == 2) {
+			result[0, 0] = t[1, 1]
+			result[0, 1] = -t[1, 0]
+			result[1, 0] = -t[0, 1]
+			result[1, 1] = t[0, 0]
+		}
+		else if (matrixOrder == 3) {
+			result[0, 0] = t[1, 1] * t[2, 2] - t[2, 1] * t[1, 2]
+			result[1, 0] = - (t[0, 1] * t[2, 2] - t[2, 1] * t[0, 2])
+			result[2, 0] = t[0, 1] * t[1, 2] - t[1, 1] * t[0, 2]
+			result[0, 1] = - (t[1, 0] * t[2, 2] - t[2, 0] * t[1, 2])
+			result[1, 1] = t[0, 0] * t[2, 2] - t[2, 0] * t[0, 2]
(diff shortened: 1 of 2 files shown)


test/math/FloatMatrixTest.ooc
```diff
@@ -130,7 +130,19 @@ FloatMatrixTest: class extends Fixture {
...
+			m := this createMatrix(4, 4, [1.f, 5.f, 3.f, 2.f, 7.f, 6.f, 8.f, 9.f, 2.f, 0.f, -3.f, 4.f, 4.f, 1.f, 2.f, 3.f])
+			this checkAllElements(m adjugate(), [-43.f, 67.f, 11.f, -187.f, -105.f, 45.f, -15.f, -45.f, 50.f, -50.f, 50.f, 50.f, 59.f, -71.f, -43.f, 131.f])
+		})
+		this add("cofactors (4x4)", func {
+			m := this createMatrix(4, 4, [1.f, 5.f, 3.f, 2.f, 7.f, 6.f, 8.f, 9.f, 2.f, 0.f, -3.f, 4.f, 4.f, 1.f, 2.f, 3.f])
+			this checkAllElements(m cofactors(), [-43.f, -105.f, 50.f, 59.f, 67.f, 45.f, -50.f, -71.f, 11.f, -15.f, 50.f, -43.f, -187.f, -45.f, 50.f, 131.f])
+		})
+		this add("determinant (4x4)", func {
+			m := this createMatrix(4, 4, [1.f, 5.f, 3.f, 2.f, 7.f, 6.f, 8.f, 9.f, 2.f, 0.f, -3.f, 4.f, 4.f, 1.f, 2.f, 3.f])
+			expect(m determinant(), is equal to(-300.0f) within(tolerance))
```
marcusnaslund: I just picked numbers at random, and got this nice even determinant. I'm a genius. All answers verified with Wolfram Alpha.

marcusnaslund:

source/math/FloatMatrix.ooc
```diff
@@ -293,48 +293,90 @@ FloatMatrix: cover {
...
+				submatrix := t _assembleWithoutRowCol(0, column) take()
+				result += t[0, column] * sign * submatrix determinant()
+				sign = -sign
+				submatrix free(Owner Sender)
+			}
+		}
 		this free(Owner Receiver)
 		result
 	}
+	_assembleWithoutRowCol: func (row, column: Int) -> This {
```
sebastianbaginski: Are we allowed to use abbreviations like `row, col` ?
Maybe not that important now, but simple optimization could be to add `submatrix` as third parameter to this function and let it fill in the values instead of allocating, this way you could avoid `N - 1` allocations when calculating determinant for the general case.
I mean this
```ooc
submatrix := FloatMatrix new(matrixOrder - 1, matrixOrder - 1)
for (column in 0 .. matrixOrder) {
    t _assembleWithoutRowCol(0, column, submatrix&)
    result += t[0, column] * sign * submatrix determinant()
....
```

source/math/FloatMatrix.ooc
```diff
@@ -293,48 +293,90 @@ FloatMatrix: cover {
...
+			result = t[0, 0] * t[1, 1] - t[0, 1] * t[1, 0]
+		else if (matrixOrder == 3)
+			result = t[0, 0] * (t[1, 1] * t[2, 2] - t[2, 1] * t[1, 2]) -
 			t[1, 0] * (t[0, 1] * t[2, 2] - t[2, 1] * t[0, 2]) +
 			t[2, 0] * (t[0, 1] * t[1, 2] - t[1, 1] * t[0, 2])
+		else {
+			sign := 1.f
+			for (column in 0 .. matrixOrder) {
+				submatrix := t _assembleWithoutRowCol(0, column) take()
+				result += t[0, column] * sign * submatrix determinant()
```
sebastianbaginski: Maybe 
```ooc
value := t[0, column]
if (!value equals(0.0f))
    result += value * sign * submatrix determinant()
```
?

marcusnaslund: Good thinking.

source/math/FloatMatrix.ooc
marcusnaslund: Names should be self-explanatory, and in the presence of `row`, certainly `col` is self-explanatory.
I see what you're saying, but we will do it if it becomes necessary.

source/math/FloatMatrix.ooc
```diff
@@ -293,48 +293,92 @@ FloatMatrix: cover {
...
-		result[1, 1] = t[0, 0] * t[2, 2] - t[2, 0] * t[0, 2]
-		result[2, 1] = - (t[0, 0] * t[1, 2] - t[1, 0] * t[0, 2])
-		result[0, 2] = t[1, 0] * t[2, 1] - t[2, 0] * t[1, 1]
-		result[1, 2] = - (t[0, 0] * t[2, 1] - t[2, 0] * t[0, 1])
-		result[2, 2] = t[0, 0] * t[1, 1] - t[1, 0] * t[0, 1]
+		matrixOrder := t order
+		if (matrixOrder == 1)
+			result[0, 0] = t[0, 0]
+		else if (matrixOrder == 2)
+			result[0, 0] = t[0, 0] * t[1, 1] - t[0, 1] * t[1, 0]
```
marcusnaslund: Hang on, this makes no sense

PR #1638, opened 2016-04-11
Fixed the swapped color bug

davidpiuva:
Red and blue were swapped when running without GPU because this line was a remain from when we stored the converted image from YUV in BGR format.


davidpiuva:

marcusnaslund:
I find it very amusing that we've been looking for a problem with swapped red and blue channels, and the bug was a call to a function explicitly named `swapRedBlue()`.

marcusnaslund:
I wonder if the `ThreadPool` has an overlooked `randomCrash()` function somewhere.

sebastianbaginski:
I think it was a leftover after the recent Bgr -> Rgb conversion.

PR #1641, opened 2016-04-12
Added scalar division and inverse to FloatMatrix

tomasmahlberg:
We might as well, since all the dirty work is already done.


tomasmahlberg:
@marcusnaslund

PR #1643, opened 2016-04-12
Significant line count reduction in tests

marcusnaslund:
Using method on tuples.
Also removed all semicolon use.


marcusnaslund:
Peer review @thomasfanell ?

PR #1650, opened 2016-04-14
Moved default font atlas to OpenGLContext

davidpiuva:
Any textures in GpuContext cannot be freed because it is called from super in OpenGLContext's destructor after the recycle bin and GLContext used to free textures has been freed.

### DIFF
--- a/source/draw/gpu/GpuContext.ooc
+++ b/source/draw/gpu/GpuContext.ooc
@@ -35,18 +35,7 @@ ToRasterFuture: class extends Future<RasterImage> {
 
 GpuContext: abstract class extends DrawContext {
 	defaultMap ::= null as Map
-	_defaultFontGpu: GpuImage = null
-	defaultFontGpu: GpuImage { get {
-		if (this _defaultFontGpu == null)
-			this _defaultFontGpu = this createImage(this defaultFontRaster)
-		this _defaultFontGpu
-	}}
 	init: func
-	free: override func {
-		if (this _defaultFontGpu != null)
-			this _defaultFontGpu free()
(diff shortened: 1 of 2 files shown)


davidpiuva:
@emilwestergren Peer review?

source/draw/gpu/opengl/OpenGLContext.ooc
```diff
@@ -66,6 +71,8 @@ OpenGLContext: class extends GpuContext {
 	}
 	init: func ~window (display: Pointer, nativeBackend: Long) { this init(GLContext createContext(display, nativeBackend)) }
 	free: override func {
+		if (this _defaultFontGpu != null)
+			this _defaultFontGpu free()
```
davidpiuva: Freeing textures can't be done after freeing the recycle bin but super has to be at the end of the destructor.

source/draw/gpu/GpuContext.ooc
```diff
@@ -35,18 +35,7 @@ ToRasterFuture: class extends Future<RasterImage> {
 
 GpuContext: abstract class extends DrawContext {
 	defaultMap ::= null as Map
-	_defaultFontGpu: GpuImage = null
-	defaultFontGpu: GpuImage { get {
-		if (this _defaultFontGpu == null)
-			this _defaultFontGpu = this createImage(this defaultFontRaster)
-		this _defaultFontGpu
-	}}
```
davidpiuva: Having the atlas here would be a better design but destructors don't allow it and having two layers of destructors would be confusing.

davidpiuva:

PR #1652, opened 2016-04-14
FloatMatrix cofactors fix

tomasmahlberg:
Should be 1 for a 1x1 "matrix". Also minor update to the inverse method.

### DIFF
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -381,9 +381,12 @@ FloatMatrix: cover {
 		submatrix
 	}
 	inverse: func -> This {
-		version(safe)
-			raise(!this take() isSquare, "Matrix must be square in FloatMatrix inverse")
-		result := this take() cofactors() transpose() / this determinant()
+		t := this take()
+		version(safe) {
+			raise(!t isSquare, "Matrix must be square in FloatMatrix inverse")
+			raise(t determinant() equals(0.f), "Matrix is singular in FloatMatrix inverse")
+		}
+		result := t cofactors() transpose() / t determinant()
 		this free(Owner Receiver)
 		result
 	}
--- a/source/math/FloatMatrix.ooc
+++ b/source/math/FloatMatrix.ooc
@@ -381,11 +381,12 @@ FloatMatrix: cover {
 	}
 	inverse: func -> This {
 		t := this take()
+		determinant := t determinant()
 		version(safe) {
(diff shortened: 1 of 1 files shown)


tomasmahlberg:
@marcusnaslund

source/math/FloatMatrix.ooc
```diff
@@ -381,9 +380,12 @@ FloatMatrix: cover {
 		submatrix
 	}
 	inverse: func -> This {
-		version(safe)
-			raise(!this take() isSquare, "Matrix must be square in FloatMatrix inverse")
-		result := this take() cofactors() transpose() / this determinant()
+		t := this take()
+		version(safe) {
+			raise(!t isSquare, "Matrix must be square in FloatMatrix inverse")
+			raise(t determinant() equals(0.f), "Matrix is singular in FloatMatrix inverse")
```
marcusnaslund: With `-Dsafe` we will now calculate the determinant twice. Put `determinant := t determinant()` below line 383 and use it here and on line 388.

source/math/FloatMatrix.ooc
```diff
@@ -299,14 +299,13 @@ FloatMatrix: cover {
 		result := t create()
 		matrixOrder := t order
 		if (matrixOrder == 1)
-			result[0, 0] = t[0, 0]
+			result[0, 0] = 1.f
```
marcusnaslund: Thanks.

source/math/FloatMatrix.ooc
tomasmahlberg: Won't this cause extra penalty (albeit minimal) on the non-safe version? Do we really care about optimizing the safe version that much?

marcusnaslund: What's the penalty?

tomasmahlberg: I'm not sure, it was kind of a question. I was thinking about the introduction of and assignment to a variable.

marcusnaslund: Ah alright. It's common to think that way, but it's not a penalty. Since we have to do `t determinant()` later anyway, its value will need to be put *somewhere* on the stack, regardless of whether we do it explicitly (with variable assignment) or not.

Also, no we don't really need to care about optimizing the safe version, but on the other hand, determinant calculation can be costly and this seems simple enough to fix.

tomasmahlberg: Gotcha.

tomasmahlberg:
Updated.

PR #1654, opened 2016-04-14
Primary feature1.1

emilwestergren:
Necessary changes for release of primary feature 1.1

@marcusnaslund peer review


marcusnaslund:
PRInt fails, but I think that's a bug in PRInt, and I think @fredrikbryntesson is on it.

marcusnaslund:
Otherwise, approved.

fredrikbryntesson:
Temporary close

marcusnaslund:
Can we confirm all commits are present here?

emilwestergren:
Yes, confirmed.

PR #1656, opened 2016-04-15
Cleaned up memory management

marcusnaslund:
Memory management, since it *has* to be explicit everywhere, should take up as little space as possible.


marcusnaslund:
Peer review @jonathanudd ? :)

jonathanudd:
Reducing number of code lines step by step :) Looks good.

PR #1657, opened 2016-04-18
Increased far plane distance.

emilwestergren:
Sometimes too low. Is this exposed in the API?
@davidpiuva peer review


davidpiuva:
If you set near plane to 1 then you can multiply the far clip plane by 100 without losing quality.

marcusnaslund:
12500 is like 25% better than 10000, so merge.

PR #1659, opened 2016-04-21
static File remove method

sebastianbaginski:
@marcusnaslund

### DIFF
--- a/source/system/io/File.ooc
+++ b/source/system/io/File.ooc
@@ -302,6 +302,12 @@ File: abstract class {
 	getCwd: static func -> String {
 		ooc_get_cwd()
 	}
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Haha. You're too fast for me.

marcusnaslund:
But while you're at it - are there any other static methods like this we need?

marcusnaslund:
Also - I expect a test for the function.

sebastianbaginski:
I think not now, but maybe `copy`, `exists` and `rename` ?

marcusnaslund:
Sounds good to me. Something like `exists` already exists (heh), I believe.

sebastianbaginski:
Updated with test.

PR #1661, opened 2016-04-21
Added missing static File methods

### LINKED ISSUE #1660: Add relevant static functions to File
Right now, the only way to remove a file is to instantiate a `File` object, call `rm`, and then free it. This is horrible.

- [x] `remove`
- [ ] `copy`
- [ ] `exists`
- [ ] `rename`

marcusnaslund:
Builds on #1659 and closes #1660 

Might also include some more cleanup of the File class.

### DIFF
--- a/source/system/io/File.ooc
+++ b/source/system/io/File.ooc
@@ -114,8 +114,7 @@ File: abstract class {
 	}
 	parentName: func -> String {
 		idx := this path lastIndexOf(This separator)
-		if (idx == -1) return null
-		return this path substring(0, idx)
+		idx == -1 ? null as String : this path substring(0, idx)
 	}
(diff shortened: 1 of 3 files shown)


source/system/io/File.ooc
```diff
@@ -299,9 +299,31 @@ File: abstract class {
...
+	}
+	exists: static func ~file (path: String) -> Bool {
+		file := This new(path)
+		result := file exists()
+		file free()
+		result
+	}
+	rename: static func (original, new: String) {
+		This copy(original, new)
+		This rm(original)
```
sebastianbaginski: There is a `rename` function in `stdio` ( http://linux.die.net/man/3/rename ), maybe we could use it ?

marcusnaslund:

Issue #1662, opened 2016-04-22
Clean up File, FileUnix and FileWin32 classes


marcusnaslund:
For example, rename `rm` and `rmrf` methods, they're ugly.
There is too much indentation in the latter classes, and a lot of methods we don't use, and a lot that can be moved to properties (or shouldn't be both method and property).
There's also a big link to a commit on upstream rock.

PR #1663, opened 2016-04-22
Cleanup in native File classes

marcusnaslund:
As a part of #1662 - changing as much as possible without affecting the API. Can't believe I haven't touched these files before.

Remember to append `?w=1` to the URL when viewing the diff.

### DIFF
--- a/source/system/native/FileUnix.ooc
+++ b/source/system/native/FileUnix.ooc
@@ -9,10 +9,13 @@
 import io/File
 include dirent
 
-/*
- * Directory covers
- */
+version (linux) {
+	include sys/stat | (__USE_BSD), sys/types | (__USE_BSD)
+} else {
+	include sys/stat, sys/types
+}
 
+version (unix || apple) {
 DIR: extern cover
 
 DirEnt: cover from struct dirent {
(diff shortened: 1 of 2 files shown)


marcusnaslund:

source/system/native/FileUnix.ooc
```diff
@@ -27,313 +30,219 @@ readdir_r: extern func (DIR*, DirEnt*, DirEnt**) -> Int
...
+		if (exec)
+			mode |= (S_IXUSR | S_IXGRP | S_IXOTH)
+		else
+			mode &= ~(S_IXUSR | S_IXGRP | S_IXOTH)
+		result := fchmod(fd, mode) == 0
+		if (close(fd) != 0)
+			result = false
+		result
+	}
+	lastAccessed: override func -> Long {
```
sebastianbaginski: Maybe we should change these functions to return `DateTime` instead ? Not this PR of course.

marcusnaslund: Good idea.

marcusnaslund: Will require us to move `DateTime` away from `base`, though.

sebastianbaginski: You decide if it's worth it, I don't think anyone uses these functions now.

marcusnaslund: I think not now. We will revisit this later if we need it.

PR #1666, opened 2016-04-25
Added fromText to ColorRgba

marcusnaslund:
Useful in the other project.
Peer review @davidpiuva ?

### DIFF
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -153,6 +153,34 @@ ColorRgba: cover {
 	distance: func (other: This) -> Float { (this toRgb() distance(other toRgb()) * 3.0f + (this a - other a) as Float pow(2)) / 4.0f sqrt() }
 	operator == (other: This) -> Bool { this equals(other) }
 	operator != (other: This) -> Bool { !this equals(other) }
+
+	fromText: static func (text: Text) -> This {
+		channels := text take() split(',')
+		result: This
+		match (channels count) {
+			case 1 => // Y
+				luma := channels[0] take() toInt()
+				result = This new(ColorConvert monochromeToRgb(ColorMonochrome new(luma)), 255)
+			case 2 => // YA
+				luma := channels[0] take() toInt()
+				alpha := channels[1] take() toInt()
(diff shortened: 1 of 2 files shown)


source/draw/Color.ooc
```diff
@@ -153,6 +153,33 @@ ColorRgba: cover {
 	distance: func (other: This) -> Float { (this toRgb() distance(other toRgb()) * 3.0f + (this a - other a) as Float pow(2)) / 4.0f sqrt() }
 	operator == (other: This) -> Bool { this equals(other) }
 	operator != (other: This) -> Bool { !this equals(other) }
+
+	fromText: static func (text: Text) -> This {
```
davidpiuva: I guess that you just moved this function but if you use Text, you have to free it when given ownership.

marcusnaslund: Yes, I just moved it. The caller will free `text`.

test/draw/ColorConvertTest.ooc
```diff
@@ -44,6 +44,10 @@ ColorConvertTest: class extends Fixture {
 			expect(color toRgb() equals(ColorRgb new(78, 99, 118)), is true)
 			expect(color toRgba() equals(ColorRgba new(78, 99, 118, 255)), is true)
 		})
+		this add("RGBA from text", func {
+			color := ColorRgba fromText(t"12,34,56,78")
+			expect(color equals(ColorRgba new(12, 34, 56, 78)))
```
davidpiuva: Monochrome, Monochrome+Alpha and RGB needs to be tested too.

source/draw/Color.ooc
davidpiuva: That was the convention for String but with Text we can't assume that.

marcusnaslund: True, there should be `text free(Owner Receiver)` there.

test/draw/ColorConvertTest.ooc
marcusnaslund: Fair enough.

marcusnaslund:
Updated.

davidpiuva:
Looks good

PR #1670, opened 2016-04-25
Resolved some memory leaks in unit tests

marcusnaslund:
Before:
```
definitely lost: 12,256 bytes in 486 blocks
indirectly lost: 54,795 bytes in 503 blocks
```

After:
```
definitely lost: 7,912 bytes in 352 blocks
indirectly lost: 27,535 bytes in 447 blocks
```

Evil villains which still leak a lot or in many places seem to be `LinkedListTest`, `HashDictionaryTest` and some geometry covers.


marcusnaslund:

test/base/DebugTest.ooc
```diff
@@ -46,6 +46,9 @@ DebugTest: class extends Fixture {
 			}
 		})
 	}
+	free: override func {
+		this outputString free()
```
sebastianbaginski: `super()`

sebastianbaginski:
Nice.

test/base/DebugTest.ooc
marcusnaslund: That's just super of you

PR #1674, opened 2016-04-26
Cleaned up c- and h-files according to style guide

marcusnaslund:


marcusnaslund:
Any objections @sebastianbaginski ?

source/system/Array.h
```diff
@@ -1,5 +1,5 @@
-#ifndef OOC_LANG_ARRAY_H
-#define OOC_LANG_ARRAY_H
+#ifndef ARRAY_H_
```
sebastianbaginski: I'd leave the longer macro guard here, `ARRAY_H_` seems not unique enough to me.

marcusnaslund: Good point.

marcusnaslund: Changed it to `OOC_ARRAY_H`, since the lang namespace is gone.

sebastianbaginski: Great, thanks.

PR #1678, opened 2016-04-28
Added optional information message for expect calls

### LINKED ISSUE #1677: Add (optional) message to expect() in unit tests
It would be useful to in some cases be able to call, for example

```ooc
expect(myValue, is equal to(someValue), "Make sure myValue is someValue")
```

so that, if this expect call fails, we get `"Testing myValue is someValue"` output to console, to make it easier to understand what went wrong.

marcusnaslund:
Useful for really long test cases where it may be unclear, if something fails, which expect call actually failed. For example if a variable is compared against another variable, where both may change unexpectedly (or even non-deterministically).

- [x] Add the functionality
- [x] Verify old code works as before
- [x] Verify it works as expected with different types
- [x] Make sure no memory leaks are introduced

Fixes #1677 

---
**EXAMPLE**
Now, a test like:
```ooc
this add("test", func {
	expect(1.f, is equal to(2.f) within(0.1f), "important message")
})
```
Will yield:
```
-> test (important message) : expected equal to 2.00 was 1.00  [tolerance: 0.10000000]
```

### DIFF
--- a/source/unit/Fixture.ooc
+++ b/source/unit/Fixture.ooc
@@ -39,7 +39,7 @@ Fixture: abstract class {
 				test run()
 				test free()
 			} catch (e: TestFailedException) {
-				e message = test name
+				e message = e description != null ? test name + " (" + e description + ")" : test name
 				result = r = false
 				failures add(e)
 			}
@@ -143,7 +143,7 @@ Fixture: abstract class {
 	_testsFailed: static Bool
 	totalTime: static Double
 	failureNames: static VectorList<String>
-	testsFailed: static Bool { get { This _testsFailed } }
+	testsFailed ::= static This _testsFailed
 
 	printFailures: static func {
 		This _print(t"Total time: %.2f s\n" format(This totalTime))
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Peer review @thomasfanell ?

marcusnaslund:
(This should make you happy @patrikmalm - it'll hopefully be in `develop` by Monday)

source/unit/Fixture.ooc
```diff
@@ -170,9 +170,10 @@ Fixture: abstract class {
 }
 
 TestFailedException: class extends Exception {
+	description: String
```
marcusnaslund: This variable will leak *if* the string passed is not a string literal, but if the user really needs a customized description, they will have to memory management it themselves.

sebastianbaginski: Wasn't it safe to run `free` multiple times on string literals ? Maybe it was a dream.

marcusnaslund: It worked some of the time, but not always.
An idea is to have a `_literal` member in `String` which defaults to `false`, and only `makeStringLiteral(...)` sets it to `true`, and then a corresponding to property to use. Don't know how useful it is (or if there are cases where this isn't good).

sebastianbaginski: I think now there is some kind of a buffer pointer, which is null only for string literals, so it works kind of like a boolean flag. I may be wrong, though.

marcusnaslund: The `CharBuffer` constructor takes a `stringLiteral` parameter, just saving it would be even easier.

patrikmalm:
Great addition @marcusnaslund  ;-)

PR #1689, opened 2016-05-03
Implicit canvas

davidpiuva:
Filling, line drawing and point drawing can now be accessed directly from images. There shall be no occurrences of the deprecated identifier "canvas" outside of Kean.

After a while when no one uses the canvas in new pull requests, the methods can be moved from canvas to image. Draw will still exist in canvas for the display window until we can find a way to completely remove canvas.

### DIFF
--- a/source/draw/RasterCanvas.ooc
+++ b/source/draw/RasterCanvas.ooc
@@ -56,11 +56,11 @@ RasterCanvas: abstract class extends Canvas {
 			}
 		}
 	}
-	drawLines: override func ~explicit (lines: VectorList<FloatPoint2D>, pen: Pen) {
-		if (lines count > 1)
-			for (i in 0 .. lines count - 1) {
-				start := IntPoint2D new(lines[i] x, lines[i] y)
-				end := IntPoint2D new(lines[i + 1] x, lines[i + 1] y)
+	drawLines: override func ~explicit (pointList: VectorList<FloatPoint2D>, pen: Pen) {
+		if (pointList count > 1)
+			for (i in 0 .. pointList count - 1) {
+				start := IntPoint2D new(pointList[i] x, pointList[i] y)
+				end := IntPoint2D new(pointList[i + 1] x, pointList[i + 1] y)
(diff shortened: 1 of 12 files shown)


davidpiuva:
@emilwestergren Peer review?

source/draw/RasterCanvas.ooc
```diff
@@ -56,11 +56,11 @@ RasterCanvas: abstract class extends Canvas {
 			}
 		}
 	}
-	drawLines: override func ~explicit (lines: VectorList<FloatPoint2D>, pen: Pen) {
-		if (lines count > 1)
-			for (i in 0 .. lines count - 1) {
-				start := IntPoint2D new(lines[i] x, lines[i] y)
-				end := IntPoint2D new(lines[i + 1] x, lines[i + 1] y)
+	drawLines: override func ~explicit (pointList: VectorList<FloatPoint2D>, pen: Pen) {
+		if (pointList count > 1)
```
marcusnaslund: Not introduced here, but since it's all about cleaning up the `draw` namespace - this if statement is not needed. If `pointList count` is 1 or less, the for loop will not iterate.

source/draw/RasterCanvas.ooc
```diff
@@ -56,13 +56,12 @@ RasterCanvas: abstract class extends Canvas {
...
-				start := IntPoint2D new(lines[i] x, lines[i] y)
-				end := IntPoint2D new(lines[i + 1] x, lines[i + 1] y)
-				this _drawLine(start, end, pen)
-			}
+	drawLines: override func ~explicit (pointList: VectorList<FloatPoint2D>, pen: Pen) {
+		for (i in 0 .. pointList count - 1) {
+			start := IntPoint2D new(pointList[i] x, pointList[i] y)
+			end := IntPoint2D new(pointList[i + 1] x, pointList[i + 1] y)
+			this _drawLine(start, end, pen)
+		}
```
davidpiuva: Updated

marcusnaslund:
Thank you.

PR #1692, opened 2016-05-03
Deleted image formats that we have not used for over a year

davidpiuva:
Deleted the odd image formats that would not even work because there is no GPU uploading. It is cheaper to implement GPU uploading directly from ByteBuffer when needed than maintaining the odd formats for internal use every time something changes.


davidpiuva:
@emilwestergren Peer review?

marcusnaslund:
Looks great.

PR #1694, opened 2016-05-04
Fixing coordinate systems

davidpiuva:
All tests fail as expected since the GPU has never handled correction of input coordinates in flipped systems.

Solutions:
* To solve the problem efficiently, each fragment shader needs to be compiled for up to 4 different input coordinate systems for the affected image. Up to 16 combinations if two images are used. Each shader taking the default system will not have any transform. The shaders applying a transform can apply 1-x without additional machine instructions using the ALU. We cannot assume that the input will always be in the same coordinate system because the same shader can be used for many different things and the same module may get different coordinates at runtime.

* If the Y image is flipped but UV is not then even more combinations have to be considered. To handle combinations of multiple input images, texture transforms have to be applied dynamically using matrix multiplications in each pixel even for the majority that does not even need a transform. The performance cost of this is totally unacceptable.

* A cheap and dirty hack without multiple fragment shaders would be to offset all texture samplings by (1, 1) and use the padding behaviour (clamp/repeat) for S and T to mirror the image but then we cannot get clamped coordinates natively from the sampler state and actually using repeat will be impossible.

A lot of other options have been suggested on forums but they are not applicable to image analysis:
* Applying a transform on texture coordinates in vertex shaders only works for point operations while excluding global and even local operations.
* Transforming images on the CPU is unacceptable since we have a live video feed.

Since no single solution is acceptable, I might have to combine method one and two so that I have one shader with no transforms and one shader taking arbitrary transforms. Each uploaded image will have a coordinate transform uploaded with a mangled name for the sampling macro to use. A MAD instruction should be faster than a matrix multiplication in this special case where we only flip the images.


davidpiuva:
The current plan has been abandoned for not being safe at all.

PR #1695, opened 2016-05-06
Removed optiGraphicbufferupload flag

emilwestergren:
This is not the place for it.
@marcusnaslund peer review

### DIFF
--- a/source/draw/gpu/opengl/AndroidContext.ooc
+++ b/source/draw/gpu/opengl/AndroidContext.ooc
@@ -27,16 +27,14 @@ AndroidContext: class extends OpenGLContext {
 		super()
 	}
 	createImage: override func (rasterImage: RasterImage) -> GpuImage {
-		result: GpuImage
-		if (rasterImage instanceOf(GraphicBufferYuv420Semiplanar)) {
-			graphicBufferImage := rasterImage as GraphicBufferYuv420Semiplanar
-			rgba := graphicBufferImage toRgba(this)
-			result = this _unpackRgbaToYuv420Semiplanar(rgba, rasterImage size, graphicBufferImage uvPadding % graphicBufferImage stride)
-			rgba referenceCount decrease()
+		match(rasterImage) {
+			case (graphicBufferImage: GraphicBufferYuv420Semiplanar) =>
+				rgba := graphicBufferImage toRgba(this)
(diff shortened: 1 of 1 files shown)


source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -27,19 +27,14 @@ AndroidContext: class extends OpenGLContext {
...
 				rgba referenceCount decrease()
-			}
-			else
-				result = super(rasterImage)
-		} else
-			result = super(rasterImage)
-		result
+				result
+			case => super(rasterImage)
+		}
```
marcusnaslund: This method isn't "simple" enough to warrant returning in two places.
`result := match (rasterImage) { ...`
and then
`result`
will suffice.

emilwestergren: It doesn't return in two places, `match` always returns a single value.

marcusnaslund: Yes, but depending on what you are matching, you may either return the `result` variable declared on line 33 or the result from the super function on line 36.

marcusnaslund: The `FloatTuple` covers for example return using a `match` statement like this. The difference is that each case statement there is a single line, so it's easy to follow.

emilwestergren: Why not simply always allow this? I don't think it improves readability by adding a `result` variable to be fair.

marcusnaslund: What about if other `case`s are introduced? Or if other code is added besides the `match`?
Enforcing this rule does not add anything in *this* particular case, I agree, but it does future-proof the code. And helps consistency.

marcusnaslund: Actually, I think I'll withdraw my position on this. The futureproof argunent doesn't really hold.

emilwestergren: Your call, I've voiced my opinion at least :)

marcusnaslund: And it's been processed. Thank's for the discussion, we will go with this.

Issue #1699, opened 2016-05-09
Remove all memory leaks in unit tests


marcusnaslund:
This helps a lot when it comes to bug fixing.

- [x] `base`
- [x] `cli`
- [x] `collections` (only `HashDictionary` remains - see #694 )
- [x] `concurrent`
- [x] `draw`
- [x] `geometry`
- [x] `io`
- [x] `math`
- [x] `net`
- [x] `plot`
- [x] `system`
- [x] `unit`
- [x] `uri`

marcusnaslund:
Updated with status after #1703

marcusnaslund:
Leaks in concurrent fixed as of #1714 and so the only remaining leak is in `HashDictionary` which has a separate issue. Yay!

PR #1702, opened 2016-05-09
Fixed memory leaks in system

marcusnaslund:
Part of #1699 

Fixed a leak in `String`, another in `Exception`, and two in `File`.

### DIFF
--- a/test/system/VectorListTest.ooc
+++ b/test/system/VectorListTest.ooc
@@ -10,6 +10,8 @@ use unit
 use collections
 
 VectorListTest: class extends Fixture {
+	staticlist: static VectorList<Int>
+	count: static Int = 0
 	init: func {
 		super("VectorList")
 		tolerance := 0.00001f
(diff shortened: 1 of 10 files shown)


test/system/VectorListTest.ooc
```diff
@@ -227,21 +213,10 @@ VectorListTest: class extends Fixture {
...
-		this add("Iterator leak", func {
-			list := VectorList<Int> new()
-			list add(1)
-			list add(2)
-			list add(4)
-			// Convenient, but leaks the iterator instance.
-			for ((index, item) in list)
-				expect(item, is equal to(list[index]))
-			list free()
-		})
```
marcusnaslund: Having a test that is made to leak is bad. I moved a comment about it to the next test.

test/system/VectorListTest.ooc
```diff
@@ -137,34 +136,24 @@ VectorListTest: class extends Fixture {
...
+			This staticlist free()
 		})
 		this add("VectorList modify", func {
 			list := VectorList<Int> new()
-			list add(0)
-			list add(1)
-			list add(2)
+			list add(0) . add(1) . add(2)
 
 			list modify(|value| value += 1)
```
marcusnaslund: The closure passed to `modify` somehow doesn't leak, in contrast to the argument to `apply`...

marcusnaslund:

PR #1703, opened 2016-05-09
Fixed memory leaks in geometry

marcusnaslund:
Note: `"#{this x toString()}, #{this y toString()}, #{this z toString()}"` is converted to `this x toString() + ", " + this y toString() + ", " + this z toString()`, which of course (and unfortunately) leaks.

### DIFF
--- a/source/geometry/FloatBox2D.ooc
+++ b/source/geometry/FloatBox2D.ooc
@@ -122,7 +122,7 @@ FloatBox2D: cover {
 		newSize := FloatVector2D linearInterpolation(this size, other size, weight)
 		This createAround(newCenter, newSize)
 	}
-	toString: func -> String { "#{this leftTop toString()}, #{this size toString()}" }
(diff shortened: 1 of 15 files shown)


marcusnaslund:

source/geometry/FloatBox2D.ooc
```diff
@@ -122,7 +122,7 @@ FloatBox2D: cover {
 		newSize := FloatVector2D linearInterpolation(this size, other size, weight)
 		This createAround(newCenter, newSize)
 	}
-	toString: func -> String { "#{this leftTop toString()}, #{this size toString()}" }
+	toString: func -> String { this leftTop toString() & ", " clone() & this size toString() }
```
sebastianbaginski: Maybe `(this leftTop toString() >> ", ") & this size toString()` instead of `clone` ?

marcusnaslund: Excellent

marcusnaslund:
Updated with suggestion.

PR #1705, opened 2016-05-09
Fixed memory leaks in concurrent tests

marcusnaslund:
Part of the work on #1699 

Eliminating these wasteful leaks have shown some leaks in `Promise`, `PromiseCollector`, and `ThreadPool` that would occur if they're used in ways not currently used today. They will be fixed in a later PR.

### DIFF
--- a/test/concurrent/RecycleBinTest.ooc
+++ b/test/concurrent/RecycleBinTest.ooc
@@ -13,13 +13,15 @@ use concurrent
 MyClass: class {
 	content: Int
 	bin: static RecycleBin<This>
-	freed: static VectorList<This>
 	allocated: static Int = 0
+	freed: static Int = 0
 	init: func (=content) { This allocated += 1 }
 	free: override func {
 		// < 0 means force free
-		if (this content < 0)
-			freed add(this)
+		if (this content < 0) {
+			This freed += 1
+			super()
+		}
 		else
 			this bin add(this)
 	}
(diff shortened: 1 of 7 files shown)


test/concurrent/RecycleBinTest.ooc
```diff
@@ -13,13 +13,15 @@ use concurrent
 MyClass: class {
 	content: Int
 	bin: static RecycleBin<This>
-	freed: static VectorList<This>
 	allocated: static Int = 0
+	freed: static Int = 0
```
marcusnaslund: This way we can actually remove the created classes, otherwise they could never be freed.

marcusnaslund:
Peer review @erikhagglund ?

marcusnaslund:
Taking this off review again, I will add some more stuff.

marcusnaslund:
There we go. Peer review @erikhagglund or @sebastianbaginski , whoever has time

test/concurrent/ThreadTest.ooc
```diff
@@ -84,6 +84,8 @@ ThreadTest: class extends Fixture {
 		otherId free()
 		(job as Closure) free()
 		expect(myId equals(Thread currentThreadId()), is true)
+		thisThreadInstance free()
+		otherThreadInstance free()
```
sebastianbaginski: why not one line ?

marcusnaslund: thanks

PR #1707, opened 2016-05-10
Added static File methods for directory creation

### LINKED ISSUE #1701: Add static createDirectory + createDirectories methods to File
And use them in unit tests.

marcusnaslund:
Fixes #1701

### DIFF
--- a/source/system/io/File.ooc
+++ b/source/system/io/File.ooc
@@ -115,7 +115,7 @@ File: abstract class {
 	createDirectory: func -> Int { this createDirectory(0c755) }
 	createDirectory: abstract func ~withMode (mode: Int) -> Int
 	// Create a directory at the path specified by this file and all the parent directories if needed
-	createDirectories: func { this createDirectories(0c755) }
+	createDirectories: func -> Int { this createDirectories(0c755) }
 	createDirectories: func ~withMode (mode: Int) -> Int {
 		p := this parent
 		if (p)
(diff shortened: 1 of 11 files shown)


source/system/io/File.ooc
```diff
@@ -278,15 +278,27 @@ File: abstract class {
 	rename: static func (originalPath, newPath: String) {
 		rename(originalPath toCString(), newPath toCString())
 	}
-	getCwd: static func -> String {
-		ooc_get_cwd()
+	createDirectory: static func ~file (path: String) -> Int {
```
sebastianbaginski: @marcusnaslund  do we even need this one ?

marcusnaslund: Not really, but it is the helper function that `createDirectories` uses, and it's different for the unix and win32 implementations. We can probably make it better though, but not now.

marcusnaslund: Oh, sorry, now I see where you commented. You're right, we can remove this one.

marcusnaslund:
Since you're already here, peer review @sebastianbaginski ?

sebastianbaginski:
Looks good.
In the future I guess it could be convenient to have another `Dir` class which could handle all the stuff related to directory access (instead of having this inside `File`). I mean, I think it's better to have a `Dir` object with specialized interface (like listing entries, iterating etc.) than a `File` object with `file isDirectory == true`.

sebastianbaginski:
Sorry I can't see what's wrong with PRInt from here.

source/system/io/File.ooc
```diff
@@ -278,15 +278,28 @@ File: abstract class {
...
+	createDirectories: static func ~file (path: String) -> Int {
+		file := This new(path)
+		result := file createDirectories()
+		file free()
+		result
+	}
+	createParentDirectories: static func ~file (path: String) -> Int {
+		file := This new(path)
+		fileParent := file parent
+		result := fileParent createParentDirectories()
```
marcusnaslund: Oops, the error is here, `createDirectories` should be called.

marcusnaslund:
> it could be convenient to have another `Dir` class

Absolutely agree.

marcusnaslund:
As an added bonus, this fixes some memory leaks in the draw tests.

PR #1710, opened 2016-05-11
Added option in Fixture to hide debug output

marcusnaslund:
This allows us to add `this hideDebugOutput()` to a test to remove debug output in a test, which is useful to get clean test output in other projects (like `No sensor set, defaulting to...` and `File not found...`)


marcusnaslund:
From the recent discussion on slack. Peer review @sebastianbaginski ?

PR #1712, opened 2016-05-12
Removed PromiseCollector

### LINKED ISSUE #1711: Retire PromiseCollector
It's a nice idea and all (or is it? I'm not even sure anymore) but it has never been used in anything, we don't need it, and it leaks memory. It's confusing that you can chain `Promise`s like that, but not `Future`s. 

And without GC it's complicated to properly define `+` and `+=` between two `PromiseCollector`s without leaking memory (and, as stated above, they currently do if used "incorrectly"). You have to free these two (or more!) instances, but their contents can of course only be freed by one of them.

Thoughts, opinions, objections @sebastianbaginski @fredrikbryntesson @thomasfanell ?

marcusnaslund:
Fixes #1711


marcusnaslund:

PR #1714, opened 2016-05-12
Using onheap keyword in concurrent classes

marcusnaslund:
This is better than creating a new `Cell` instance every time (which also leaks in certain circumstances, as memchecking the tests show).

This PR fixes *all* (yay!) memory leaks in concurrent tests as part of #1699 together with #1712 

Will have to be tested thoroughly.

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -103,7 +103,7 @@ Future: abstract class <T> {
 }
 
 _ThreadFuture: class <T> extends Future<T> {
-	_result: Object
+	_result: __onheap__ T
 	_action: Func
 	_thread: Thread
 	_threadAlive := true
@@ -113,10 +113,7 @@ _ThreadFuture: class <T> extends Future<T> {
 		this _state = _PromiseState Unfinished
 		this _action = func {
 			temporary := task()
-			if (T inheritsFrom(Object))
-				this _result = temporary
-			else
-				this _result = Cell<T> new(temporary)
+			this _result = temporary
 			if (this _state != _PromiseState Cancelled)
 				this _state = _PromiseState Finished
 			if (this _freeOnCompletion) {
@@ -132,6 +129,7 @@ _ThreadFuture: class <T> extends Future<T> {
 			this _freeOnCompletion = true
 			this _thread detach()
 		} else {
+			memfree(this _result)
 			this _thread free()
 			(this _action as Closure) free()
 			super()
@@ -152,10 +150,7 @@ _ThreadFuture: class <T> extends Future<T> {
 	getResult: override func (defaultValue: T) -> T {
 		result := defaultValue
 		if (this _state == _PromiseState Finished && this _result != null)
-			if (T inheritsFrom(Object))
-				result = this _result
-			else
-				result = (this _result as Cell<T>) get()
+			result = this _result
 		result
 	}
 	cancel: override func -> Bool {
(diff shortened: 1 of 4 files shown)


source/concurrent/Promise.ooc
```diff
@@ -109,7 +109,7 @@ Future: abstract class <T> {
 }
 
 _ThreadFuture: class <T> extends Future<T> {
-	_result: Object
+	_result: __onheap__ T
```
sebastianbaginski: And just when I thought ooc code can't get more confusing than it already is... Is inline assembly the next step ? :P

marcusnaslund: This is actually exactly the keyword that `Cell` uses to make sure its contents doesn't disappear in a long lost stack frame.
We should really change the keyword to something prettier, though.

test/concurrent/ThreadPoolTest.ooc
```diff
@@ -65,6 +65,8 @@ ThreadPoolTest: class extends Fixture {
 			comparison := t"fail"
 			result := future wait(comparison)
 			expect(result == comparison)
+			Time sleepMilli(1000)
```
marcusnaslund: We can currently not `wait` on a `Future` we have previously called `cancel` on. This is ugly but works for now. We will get something better with #1588 or similar.

test/concurrent/SynchronizedListTest.ooc
```diff
@@ -10,6 +10,12 @@ use base
 use concurrent
 use unit
 
+applyIndex: static Int = 0
+applyTestFunc := func (value: Int*) {
+	expect(value@, is equal to(applyIndex + 1))
+	applyIndex = applyIndex + 1
+}
```
marcusnaslund: Does this `func` leak? No. Why? Who knows.

source/concurrent/Promise.ooc
sebastianbaginski: >  keyword that Cell uses

Yeah, and it's slowly making it's way to higher level classes :)

test/concurrent/SynchronizedListTest.ooc
sebastianbaginski: Maybe it's compiled to a static C function, without creating a Closure object for it ?

shamanas: `applyIndex` is static so as @sebastianbaginski said, I suspect no context is generated, so no leaks :)

marcusnaslund: Wonderful, thanks @shamanas

source/concurrent/Promise.ooc
shamanas: I've pointed out on [rock#48](http://github.com/magic-lang/rock/issues/48) that this could be automatically done by the compiler, would be happy to oblige if you think it's worth it now that it isn't used in many places.  

(So I can remove the keyword from the language too at the same time now that it's only used in a couple of places)

marcusnaslund: Nice, I hadn't seen that. I would certainly prefer to not have to manually add this keyword, but I will need to look into it some more. @thomasfanell and @fredrikbryntesson , thoughts?

sebastianbaginski: Thanks, I think we should be less explicit about memory management if we can.
I'd love to see this done automatically, but I think @thomasfanell or @fredrikbryntesson should be the one to decide.

thomasfanell:
Obviously, if we could have rock generate the equivalent of ``__onheap__`` and it is reliable, we should definitely do it.

marcusnaslund:

marcusnaslund:
:+1: Will be doing some stress tests before I merge this.

marcusnaslund:
Works like a charm.

PR #1715, opened 2016-05-13
Fixed tolerance in relational operators

tomasmahlberg:
These functions had a `tolerance` argument but it was not used. This would make more sense, right?


tomasmahlberg:
@marcusnaslund

marcusnaslund:
Good catch.

PR #1716, opened 2016-05-15
simple benchmark() function for Profiler

sebastianbaginski:
Just a simple idea for basic performance measurements. Use like:
```ooc
use base
include sched

"average : %f ms" printfln(Profiler benchmark(func {
	Time sleepMilli(300)
}))
```

> average : 300.116500 ms

@marcusnaslund what do you think about this ?

### DIFF
--- a/source/base/Profiler.ooc
+++ b/source/base/Profiler.ooc
@@ -50,6 +50,19 @@ Profiler: class {
 		This _profilers free()
 		This _profilers = null
 	}
+	benchmark: static func (task: Func, timeoutMilliseconds := 1000.0) -> Double {
+		numberOfIterations := 0
+		timer := WallTimer new()
+		totalTime := 0.0
+		while (totalTime < timeoutMilliseconds * 1000.0) {
+			numberOfIterations += 1
+			timer start()
+			task()
+			totalTime += timer stop()
+		}
+		timer free()
+		totalTime / (numberOfIterations * 1000)
+	}
 	_logResults: static func (logMethod: Func (String)) {
 		for (i in 0 .. This _profilers count) {
 			profiler := This _profilers[i]
--- a/test/base/ProfilerTest.ooc
+++ b/test/base/ProfilerTest.ooc
@@ -27,12 +27,19 @@ ProfilerTest: class extends Fixture {
 			expect(content empty(), is false)
 			(profiler, content, file) free()
 		})
+		this add("benchmark", This _testBenchmark)
 		this add("cleanup", func {
 			profiler := Profiler new("for cleanup")
 			profiler start() . stop()
 			profilerToFree := Profiler new("to free")
 			profilerToFree free()
 		})
+	}
+	_testBenchmark: static func {
+		slowTask := func { Time sleepMilli(1) }
+		fastTask := func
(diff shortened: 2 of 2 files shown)


marcusnaslund:
Looks good.

But why should it reside in `Fixture`? Isn't this sort of what `base/Profiler` is for? Maybe we should extend that class instead.

sebastianbaginski:
You're absolutely right, updated.

marcusnaslund:
Alright, that's better. Is this ready for review?

sebastianbaginski:
I guess so. Would you be so kind ?

source/base/Profiler.ooc
```diff
@@ -50,6 +50,19 @@ Profiler: class {
 		This _profilers free()
 		This _profilers = null
 	}
+	benchmark: static func (task: Func, timeoutMilliseconds := 1000.0) -> Double {
+		numberOfIterations := 0
+		timer := WallTimer new()
+		totalTime := 0.0
+		while (totalTime < timeoutMilliseconds * 1000.0) {
+			++numberOfIterations
```
marcusnaslund: Use ´+= 1` instead of `++`

marcusnaslund:
Woah, the code updates faster than I can comment. :D

source/base/Profiler.ooc
sebastianbaginski: Right, I forgot about this. Maybe `magic` could detect this ?

source/base/Profiler.ooc
```diff
@@ -50,6 +50,19 @@ Profiler: class {
 		This _profilers free()
 		This _profilers = null
 	}
+	benchmark: static func (task: Func, timeoutMilliseconds := 1000.0) -> Double {
```
marcusnaslund: I'm not sure what's appropriate, but maybe we should pass and return `TimeSpan`s instead?

marcusnaslund:
Also - be sure to add a simple test that uses the function.

sebastianbaginski:
Updated with test. I was thinking about the `TimeSpan` too, but in the end I decided to keep this as simple as possible, as I think it will be used only for debugging anyway. I can change to TimeSpans if you prefer them though.

marcusnaslund:
Let's stick to a `Double` for now, we can discuss this in a general sense later.

test/base/ProfilerTest.ooc
```diff
@@ -27,13 +27,20 @@ ProfilerTest: class extends Fixture {
...
 			profilerToFree := Profiler new("to free")
 			profilerToFree free()
 		})
 	}
+	_testBenchmark: static func {
+		slowTask := func { Time sleepMilli(1) }
+		fastTask := func
+		expect(Profiler benchmark(slowTask, 5), is greater than(Profiler benchmark(fastTask, 5)))
+		(slowTask as Closure, fastTask as Closure) free()
+	}
```
marcusnaslund: Does it not work if the code is part of the `add` call, as for the other tests?

sebastianbaginski: It didn't work.

marcusnaslund: ...

PR #1717, opened 2016-05-15
Wrote a new HashMap

### LINKED ISSUE #694: Implement efficient HashMap/HashDictionary with no memory leaks

marcusnaslund:
Fixes #694 

Replaces the old and unstable `HashMap`. Uses the same underlying hashing functions (`acX31` for `String` (and now also `Text`) and `murmur` for "other").

### DIFF
--- a/source/system/structs/HashMap.ooc
+++ b/source/system/structs/HashMap.ooc
@@ -6,85 +6,50 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-HashEntry: cover {
-	key: Pointer
-	value: Pointer
-	next: This* = null
-	init: func@ { this init(null, null) }
-	init: func@ ~keyVal (=key, =value)
-	free: func {
-		if (this key != null)
-			memfree(this key)
-		if (this value != null)
-			memfree(this value)
-		if (this next != null) {
-			temp := this next@
-			temp free()
-			memfree(this next)
+_HashEntry: class <K, V> {
+	_key: __onheap__ K
+	_value: __onheap__ V
+	_next: This<K, V> = null
+	init: func (=_key, =_value)
+	free: override func {
+		memfree(this _key)
+		memfree(this _value)
+		if (this _next != null) {
+			this _next free()
+			this _next = null
 		}
+		super()
 	}
 }
 
-intHash: func <K> (key: K) -> SizeT { (key as Int) as SizeT }
-pointerHash: func <K> (key: K) -> SizeT { (key as Pointer) as SizeT }
-
-/**
-	Port of Austin Appleby's Murmur Hash implementation
-	http://code.google.com/p/smhasher/
-*/
-murmurHash: func <K> (keyTagazok: K) -> SizeT {
-	seed: SizeT = 1
-
-	len := K size
-	m = 0x5bd1e995: SizeT
-	r = 24: SSizeT
-	l := len
-
-	h : SizeT = seed ^ len
-	data := (keyTagazok&) as Byte*
-
-	while (true) {
+// Austin Appleby's Murmur Hash implementation: http://code.google.com/p/smhasher/
+_murmurHash: func <K> (key: K) -> Int {
+	(len, m, r, l, h, t) := (K size, 0x5bd1e995 as SizeT, 24 as SizeT, len, (1 as SizeT) ^ len as SizeT, 0)
+	data := (key&) as Byte*
+	while (len >= 4) {
 		k := (data as SizeT*)@
-
 		k *= m
 		k ^= k >> r
 		k *= m
-
 		h *= m
 		h ^= k
-
 		data += 4
 		if (len < 4) break
 		len -= 4
 	}
-
-	t := 0
-
 	if (len == 3) h ^= data[2] << 16
 	if (len == 2) h ^= data[1] << 8
 	if (len == 1) h ^= data[0]
-
 	t *= m; t ^= t >> r; t *= m; h *= m; h ^= t
 	l *= m; l ^= l >> r; l *= m; h *= m; h ^= l
-
-	h ^= h >> 13
-	h *= m
-	h ^= h >> 15
-
-	h
+	h ^= h >> 13; h *= m; h ^= h >> 15
+	h as Int
 }
 
-/**
- * khash's ac_X31_hash_string
- * http://attractivechaos.awardspace.com/khash.h.html
- * @access private
- * @param s The string to hash
- * @return UInt
- */
-acX31Hash: func <K> (key: K) -> SizeT {
-	raise(key == null, "key in acX31Hash must be non-null!")
-	s : Char* = (K == String) ? (key as String) toCString() as Char* : key as Char*
-	h = s@ : SizeT
+// khash's ac_X31_hash_string: http://attractivechaos.awardspace.com/khash.h.html
+_acX31Hash: func <K> (key: K) -> Int {
+	s: Char* = (K == String) ? (key as String) toCString() as Char* : key as Char*
+	h = s@ : Int
 	if (h) {
 		s += 1
 		while (s@) {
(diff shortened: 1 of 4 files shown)


source/system/structs/HashMap.ooc
```diff
@@ -6,85 +6,50 @@
...
 	t *= m; t ^= t >> r; t *= m; h *= m; h ^= t
 	l *= m; l ^= l >> r; l *= m; h *= m; h ^= l
-
-	h ^= h >> 13
-	h *= m
-	h ^= h >> 15
-
-	h
+	h ^= h >> 13; h *= m; h ^= h >> 15
+	h as Int
```
marcusnaslund: All of this has been compacted, otherwise it's the same as before (but it returns `Int`).

sebastianbaginski: For a moment I thought you wrote some perl code.

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,149 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
+		existingValue: V = this remove(key)
+		this _count += 1
+		hashkey := This hash(key) % this capacity
+		entry := _HashEntry<K, V> new(key, value)
+		entry _next = this _buckets[hashkey]
+		this _buckets[hashkey] = entry
+		existingValue
+	}
+	get: func ~nulldefault (key: K) -> V {
+		this get(key, null)
```
marcusnaslund: Just one method with `get: func (key: K, defaultValue: V = null) -> V {` would have been nice but rock wouldn't parse it.

source/system/structs/HashMap.ooc
```diff
@@ -6,85 +6,50 @@
...
- * @access private
- * @param s The string to hash
- * @return UInt
- */
-acX31Hash: func <K> (key: K) -> SizeT {
-	raise(key == null, "key in acX31Hash must be non-null!")
-	s : Char* = (K == String) ? (key as String) toCString() as Char* : key as Char*
-	h = s@ : SizeT
+// khash's ac_X31_hash_string: http://attractivechaos.awardspace.com/khash.h.html
+_acX31Hash: func <K> (key: K) -> Int {
```
marcusnaslund: I don't know enough about hashing functions to decide, but maybe this function is enough for general purposes instead of murmur...

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,149 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
 			case Int => first as Int == second as Int
-			case => memcmp(first, second, T size) == 0
+			case String => first as String equals(second as String)
+			case Text => first as Text == second as Text
+			case => memcmp(first, second, K size) == 0
 		}
 	}
-}
 
-HashMapValueIterator: class <K, T> extends BackIterator<T> {
```
marcusnaslund: I got rid of all the iterator nonsense and suddenly the file got a lot smaller.

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,149 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
-	else
-		murmurHash
-}
-
 HashMap: class <K, V> {
-	_count: SizeT
-	capacity: SizeT
-	hashKey: Func <K> (K) -> SizeT
-	buckets: HashEntry[]
-	keys: VectorList<K>
```
marcusnaslund: I don't know if it's a good idea or not to maintain a list of keys as well. It's still possible to iterate over the elements using `each(...)`. @sebastianbaginski ?

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,149 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
-		this each(f)
-		(f as Closure) free()
-		copy
-	}
-	merge: func (other: This<K, V>) -> This<K, V> {
-		c := this copy()
-		c merge~inplace(other)
-		c
-	}
-	merge: func ~inplace (other: This<K, V>) -> This<K, V> {
```
marcusnaslund: My implementation also lacks `copy` and `merge` functions, we can add those later (including the hassle of getting them to work properly) later if we need them.

marcusnaslund:
I believe this is ready for peer review. Will you do me the honors, @sebastianbaginski ?

source/system/structs/HashMap.ooc
sebastianbaginski: I don't think so. You can search our codebase and see what's the most common use case. If people call `hash keys()` a lot, then it's probably smart to optimize it somehow. Otherwise I think your implementation is ok.

marcusnaslund: People don't actually need it at all, so that's why it was easy to remove. :)

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,149 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
 			entry := oldBuckets[i]
-			if (entry key != null) {
-				this put(entry key as K, entry value as V)
-				old := entry
-				while (entry next) {
-					entry = entry next@
-					this put(entry key as K, entry value as V)
-				}
-				old free()
+			while (entry != null) {
```
sebastianbaginski: Can you `memcpy(newBuckets, oldBuckets, oldCapacity)` instead of looping ?

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,149 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
+			entry := this _buckets[index]
+			while (entry != null) {
+				action(entry _key&, entry _value&)
+				entry = entry _next
 			}
 		}
-		result
 	}
-	resize: func (_capacity: SizeT) {
+	resize: func (newCapacity: Int) {
```
sebastianbaginski: What happens if I use this function to shrink the hash map ?

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,149 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
-		match (T) {
-			case String => first as String equals(second as String)
-			case Pointer => first as Pointer == second as Pointer
+	_keyEquals: func (first, second: K) -> Bool {
+		match (K) {
 			case Int => first as Int == second as Int
-			case => memcmp(first, second, T size) == 0
+			case String => first as String equals(second as String)
+			case Text => first as Text == second as Text
+			case => memcmp(first, second, K size) == 0
```
sebastianbaginski: Will it work for covers / other int types ? Just curious.

sebastianbaginski:
This looks a lot better now. Nicely done.

source/system/structs/HashMap.ooc
marcusnaslund: No, since `hash % newCapacity` will not follow any pattern similar to `hash % oldCapacity`, so the items will be organized completely differently in the "new" buckets.

source/system/structs/HashMap.ooc
marcusnaslund: Already part of the test, and it works just fine.

marcusnaslund: Note that it's the number of buckets, not the number of elements the map can store (which is "unlimited").

source/system/structs/HashMap.ooc
marcusnaslund: It does. I should add a test with another type.

source/system/structs/HashMap.ooc
sebastianbaginski: True.

source/system/structs/HashMap.ooc
marcusnaslund: Sounds like an insult.

source/system/structs/HashMap.ooc
marcusnaslund: Do you have any clue @thomasfanell ?

thomasfanell: I don't feel I know about hashing to comment, sorry...

marcusnaslund: @davidpiuva ?

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,156 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
+	_buckets: _HashEntry<K, V>[]
+
 	count ::= this _count
-	isEmpty ::= this keys empty
+	isEmpty ::= this _count == 0
+	capacity ::= this _capacity
 
-	init: func { init(3) }
-	init: func ~withCapacity (=capacity) {
+	init: func ~default { this init(32) }
```
davidpiuva: Should be 31

source/system/structs/HashMap.ooc
```diff
@@ -6,85 +6,50 @@
...
-
-	h : SizeT = seed ^ len
-	data := (keyTagazok&) as Byte*
-
-	while (true) {
+// Austin Appleby's Murmur Hash implementation: http://code.google.com/p/smhasher/
+_murmurHash: func <K> (key: K) -> Int {
+	(len, m, r, l, h, t) := (K size, 0x5bd1e995 as SizeT, 24 as SizeT, len, (1 as SizeT) ^ len as SizeT, 0)
+	data := (key&) as Byte*
+	while (len >= 4) {
```
marcusnaslund: @davidpiuva I changed this so that it works for types that are < 4 bytes (like `Char` in one of the tests).

source/system/structs/HashMap.ooc
```diff
@@ -95,262 +60,155 @@ acX31Hash: func <K> (key: K) -> SizeT {
...
+	_buckets: _HashEntry<K, V>[]
+
 	count ::= this _count
-	isEmpty ::= this keys empty
+	isEmpty ::= this _count == 0
+	capacity ::= this _capacity
 
-	init: func { init(3) }
-	init: func ~withCapacity (=capacity) {
+	init: func ~default { this init(31) }
```
marcusnaslund: Thanks @davidpiuva , updated.

PR #1718, opened 2016-05-15
Removed HashDictionary

marcusnaslund:
We don't use it (because the current `HashMap` is broken) and hopefully #1717 will make the map good enough to not require any wrapper of this kind.


sebastianbaginski:
Sometimes _the only way to fix it is to flush it all away_ ...

PR #1721, opened 2016-05-16
Added file methods for HashMap

### LINKED ISSUE #1720: Add readFromFile and writeToFile feature to HashMap
With a new and functioning HashMap as of #1717 we will also need to implement files for reading from and writing to file. An example use case:

```ooc
map := HashMap readFromFile("data.txt")

box := FloatBox2D parse(map get("myboxdata"))
values := map get("a list") split(",")
doSomeCalculation(values[0] toFloat(), values[5] toFloat())
```

marcusnaslund:
Fixes #1720 

It is now possible to write the contents of a `<String, String>`-hashmap to file. They will be saved in the very simple format
```
key1:value1
key2:value2
key3:value3
...
```
These files can also be read in to create a new `HashMap<String, String>`. Lines with zero (or more than one) colon will be ignored.

---

This is pretty neat:
```ooc
numbers := HashMap readFromFile("mynumbers.txt")
evenNumberString := numbers get("even") split(',')
evenNumbers := IntVectorList new()
for (i in 0 .. evenNumberString count)
    evenNumbers add(evenNumbers[i] toInt())
```

---

Yes, `readFromFile` takes a `Text` argument and `writeToFile` takes a `String` argument... since this is what's expected from `FileReader` and `FileWriter`, respectively. This is a mess we need to figure out.

### DIFF
--- a/source/system/structs/HashMap.ooc
+++ b/source/system/structs/HashMap.ooc
@@ -6,10 +6,13 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
+ import io/[File, FileWriter, FileReader]
+
 _HashEntry: class <K, V> {
 	_key: __onheap__ K
 	_value: __onheap__ V
 	_next: This<K, V> = null
+
 	init: func (=_key, =_value)
 	free: override func {
 		memfree(this _key)
@@ -19,6 +22,12 @@ _HashEntry: class <K, V> {
 			this _next = null
 		}
 		super()
+	}
+	freeKeyValuePair: func {
+		if (K inheritsFrom(Object))
+			(this _key as Object) free()
+		if (V inheritsFrom(Object))
+			(this _value as Object) free()
 	}
 }
 
@@ -78,6 +87,15 @@ HashMap: class <K, V> {
 		this clear()
 		this _buckets free()
 		super()
+	}
+	freeContent: func {
+		for (i in 0 .. this capacity) {
+			entry := this _buckets[i]
+			while (entry != null) {
+				entry freeKeyValuePair()
+				entry = entry _next
+			}
+		}
 	}
 	put: func (key: K, value: V) -> V {
 		if ((this _count as Float / this _capacity as Float) > 0.75f)
@@ -193,6 +211,19 @@ HashMap: class <K, V> {
 		}
 		oldBuckets free()
 	}
+	writeToFile: func (filename: String) {
+		raise(K != String || V != String, "Key and value must be of type String!")
+		File createParentDirectories(filename)
+		fileWriter := FileWriter new(filename)
+		writerFunc := func (keyPtr, valuePtr: String*) {
+			key := (keyPtr as String*)@
+			value := (valuePtr as String*)@
+			fileWriter write(key) . write(":") . write(value) . write("\n")
+		}
+		this each(writerFunc)
+		(writerFunc as Closure) free()
+		fileWriter close() . free()
+	}
 	_keyEquals: func (first, second: K) -> Bool {
 		match (K) {
 			case Int => first as Int == second as Int
@@ -200,6 +231,19 @@ HashMap: class <K, V> {
 			case Text => first as Text == second as Text
 			case => memcmp(first, second, K size) == 0
 		}
+	}
+	readFromFile: static func (filename: Text) -> This<String, String> {
+		reader := FileReader new(filename, t"r")
+		result := This<String, String> new()
+		while (reader hasNext()) {
+			string := reader readLine()
+			pair := string split(':')
+			if (pair count == 2)
+				result put(pair[0] clone(), pair[1] clone())
+			(string, pair) free()
+		}
+		reader close() . free()
+		result
 	}
 	hash: static func <K> (key: K) -> Int {
 		result := match (K) {
--- a/test/system/HashMapTest.ooc
+++ b/test/system/HashMapTest.ooc
@@ -177,6 +177,30 @@ HashMapTest: class extends Fixture {
 
 			hashmap free()
 		})
+		this add("to and from file", func {
+			hashmap := HashMap<String, String> new()
+			hashmap put("pi", "3.14159")
+			hashmap put("e", "2.71828")
+			hashmap put("sqrt(2)", "1.41459")
+			hashmap put("primes", "2,3,5,7,11,13,17")
+			hashmap writeToFile("test/system/output/mathmap.txt")
+			hashmap free()
+
+			tolerance := 0.00001f
+			input := HashMap readFromFile(t"test/system/output/mathmap.txt")
+			expect(input, is notNull)
+			expect(input count, is equal to(4))
+			(piString, eString, sqrtTwoString, primeString) := (input get("pi"), input get("e"), input get("sqrt(2)"), input get("primes"))
+			(pi, e, sqrtTwo) := (piString, eString, sqrtTwoString) toFloat()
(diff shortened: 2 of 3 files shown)


marcusnaslund:
Not ready yet, but is this the functionality you need @patrikmalm or is something missing?

patrikmalm:
Could you post an example of how the hash map will be initialized and populated? The file interaction functionality looks good.

marcusnaslund:
Example:
```ooc
importantBox := FloatBox2D new(23, 45, 12, 3)
boxes := HashMap<String, String> new()
boxes put("VIP box", box toString())
boxes writeToFile("boxes.txt")

...

data := boxes readFromFile("boxes.txt")
THEBOX := FloatBox2D parse(data get("VIP box"))
```

patrikmalm:
I am much pleased with this :)

marcusnaslund:
Peer review maybe @thomasfanell ?

marcusnaslund:
Or maybe @patrikmalm - if you have more time.

test/system/HashMapTest.ooc
```diff
@@ -177,6 +182,33 @@ HashMapTest: class extends Fixture {
...
+			(pi, e, sqrtTwo) := (piString, eString, sqrtTwoString) toFloat()
+			expect(pi, is equal to(3.14159f) within(tolerance))
+			expect(e, is equal to(2.71828f) within(tolerance))
+			expect(sqrtTwo, is equal to(1.41459f) within(tolerance))
+			primes := primeString split(',')
+			expect(primes count, is equal to(7))
+			primes free()
+
+			// Since keys/values are not string literals we need to manually free
+			input each(freeFunc)
```
patrikmalm: So this will be needed everytime we want to use the read/write functionality? If that is the case it really adds to the complexity of using it and will produce hard to find leaks. Considering OOC memory management is already incredibly opaque is there some other way to manage the destruction of the map that doesn't put as high requirement on the understanding of OOC voodo on the user?

patrikmalm:
A general comment. Would be nice to have some negative testing as well. Especially at it might highlight any limitations of the functionality to anyone trying to understand the class by reading the test suite (which is how we do it since documentation is somewhat scarce).

patrikmalm:
No further comments, your honour.

test/system/HashMapTest.ooc
marcusnaslund: I see your point, but...

> So this will be needed everytime

Yes, but it's not very different from other pieces of ooc: If you throw classes in a generic something, you are also responsible for dealing with their destruction. Having a function which iterates over all entries and frees the key+value pair I thought was actually pretty neat. :)

> is there some other way

No. :( I mean, we could have a method in the map like `freeAllKeyValuePairs()` which would contain a huge switch statement and try to free "all" types of keys and values... but that's not too practical. We *could* have a specialized one for just the <String, String> type, but that may be a bit too much specialization. What do you say?

marcusnaslund:
> Would be nice to have some negative testing as well.

You mean like trying to write a `HashMap<Int, Float>` and watching it fail?

test/system/HashMapTest.ooc
marcusnaslund: Any better ideas @sebastianbaginski @thomasfanell ?

sebastianbaginski: Not really :/ Why do we need a `huge switch statement` instead of something like:
```ooc

_HashEntry: class <K, V> {
    freeContent() {
         if (K inheritsFrom(Object))
            (_key as Object*)@ free()
         if (V inheritsFrom(Object))
            (_value as Object*)@ free()
         this free()
    }
}

HashMap: class ... {
    freeAll: func {
        for (this _buckets)
           this _bucket[i] freeContent()
    }
}
```
?

marcusnaslund: Does `(_key as Object*)@ free()` really work? Won't it free it as an `Object` rather than the actual class it is? Didn't we talk about this earlier @erikhagglund ?

sebastianbaginski: I'm not sure, but
```ooc
use base

MyClass: class {
	init: func
	free: override func {
		"free" println()
		super()
	}
}

x := MyClass new()
(x& as Object*)@ free()
```
prints
> free

marcusnaslund: This makes me very happy.

patrikmalm: I defer to your experience in this case.

patrikmalm:
Yeah, although that might be unnecessary. But perhaps you could try different kinds of strings that should and should not work

test/system/HashMapTest.ooc
marcusnaslund: I will experiment a bit.

erikhagglund: Like Sebastian mentions, it should work. What we talked about earlier was regarding "vector clear()" not freeing its elements.

marcusnaslund:
> But perhaps you could try different kinds of strings that should and should not work

source/system/structs/HashMap.ooc
```diff
@@ -20,6 +23,12 @@ _HashEntry: class <K, V> {
 		}
 		super()
 	}
+	freeKeyValuePair: func {
+		if (K inheritsFrom(Object))
+			(this _key as Object) free()
+		if (V inheritsFrom(Object))
+			(this _value as Object) free()
+	}
```
marcusnaslund: Thanks @sebastianbaginski - this works great (didn't even need the `*@` pointer bonanza).

test/system/HashMapTest.ooc
```diff
@@ -177,6 +177,30 @@ HashMapTest: class extends Fixture {
...
+			(piString, eString, sqrtTwoString, primeString) := (input get("pi"), input get("e"), input get("sqrt(2)"), input get("primes"))
+			(pi, e, sqrtTwo) := (piString, eString, sqrtTwoString) toFloat()
+			expect(pi, is equal to(3.14159f) within(tolerance))
+			expect(e, is equal to(2.71828f) within(tolerance))
+			expect(sqrtTwo, is equal to(1.41459f) within(tolerance))
+			primes := primeString split(',')
+			expect(primes count, is equal to(7))
+			primes free()
+
+			input freeContent() . free()
```
marcusnaslund: What do you think about this @patrikmalm ? :)

test/system/HashMapTest.ooc
shamanas: `(_key as Object*)@ free()` should work since all functions are virtual by default.

marcusnaslund: Thanks @shamanas - `(_key as Object) free()` worked too.

source/system/structs/HashMap.ooc
```diff
@@ -193,6 +211,19 @@ HashMap: class <K, V> {
 		}
 		oldBuckets free()
 	}
+	writeToFile: func (filename: String) {
```
thomasfanell: `filename` is a `String` while `filename` in `readFromFile` is a `Text`? I would expect both methods to take the same type.

thomasfanell: Ok, I saw that you made a note up there about this

marcusnaslund: Yeah, `String` vs `Text` needs to be sorted out when we have time.

test/system/HashMapTest.ooc
patrikmalm: Much better

PR #1723, opened 2016-05-18
Only drawing to packed images

davidpiuva:
Drawing to YUV images has been removed so that DrawState can take multiple input images with explicit indices to reduce confusion in the most common cases.
`getFirstPacked` and `getSecondPacked` were created to handle this transition because some calls were CPU/GPU agnostic and there is no type for YUV semiplanar on both CPU and GPU. The Y and UV channels cannot be mentioned in `Image`.

Later I will remove the drawing from YUV image and start to take all input textures by indices in DrawState so that the interface is consistent with one instead of three ways of doing the same thing. Then I can easily add more safety features against uninitialized memory and crashes that depend on drivers and hardware.

### DIFF
--- a/source/draw/Image.ooc
+++ b/source/draw/Image.ooc
@@ -43,6 +43,8 @@ Image: abstract class {
 		this _canvas = null
 		super()
 	}
+	getFirstPacked: abstract func -> This
+	getSecondPacked: abstract func -> This
 	drawPoint: func ~white (position: FloatPoint2D) { this canvas drawPoint~white(position) }
 	drawPoint: func ~explicit (position: FloatPoint2D, pen: Pen) { this canvas drawPoint~explicit(position, pen) }
 	drawLine: func ~white (start, end: FloatPoint2D) { this canvas drawLine~white(start, end) }
--- a/source/draw/RasterPacked.ooc
+++ b/source/draw/RasterPacked.ooc
@@ -127,6 +127,8 @@ RasterPacked: abstract class extends RasterImage {
 		this _buffer = null
 		super()
 	}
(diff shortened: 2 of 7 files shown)


davidpiuva:
@emilwestergren Peer review?

source/draw/Image.ooc
```diff
@@ -43,6 +43,8 @@ Image: abstract class {
 		this _canvas = null
 		super()
 	}
+	getFirstPacked: virtual func -> This { this }
```
emilwestergren: Why this?

davidpiuva: The first packed image of a packed image is itself. Works to get monochrome for both YUV and grayscale images.

emilwestergren: I don't think this belongs here... Packed is a subclass of `Image`.

davidpiuva: All images contains one or more packed images and Image is the only common parent of both YUV images.

davidpiuva: I could make the function abstract but most image types are packed.

source/draw/Image.ooc
```diff
@@ -43,6 +43,8 @@ Image: abstract class {
 		this _canvas = null
 		super()
 	}
+	getFirstPacked: abstract func -> This
+	getSecondPacked: abstract func -> This
```
davidpiuva: Updated

emilwestergren:
I'm not so sure about this solution, seems a bit hackish to me.

davidpiuva:
I have to close this because of major version conflicts with other's more important pull requests in the other repo.

PR #1725, opened 2016-05-18
Adjusted the sometimes failing Promise test

marcusnaslund:
Going straight to final review as this is just a test and I have so much power... and I want to see if this works during the nightly memcheck tonight.


marcusnaslund:
But this won't help.

PR #1729, opened 2016-05-20
Removed unused AlphaByteString generator

davidpiuva:
This feature was replaced by ascii images.


davidpiuva:
@emilwestergren Peer review?

emilwestergren:
This is for creating a string image with the alpha channel from an RGBA image (like the Vidhance logo), are you sure this won't be needed?

davidpiuva:
We can load a grayscale image and get it as an ascii string. The new way does not require compiling the data as code to work.

PR #1727, opened 2016-05-19
No canvas

davidpiuva:
There is no more canvas.

After this PR, I will look for redundancy in the code that can be removed.
This PR is just moving things with the minimum possible modification.

Builds on https://github.com/magic-lang/ooc-kean/pull/1726.

PR #1731, opened 2016-05-20
YUV refactor

davidpiuva:
All raster YUV images that remained from the odd formats are merged into one.


source/draw/RasterPlanar.ooc
```diff
@@ -1,18 +0,0 @@
...
- *
- * Copyright (C) 2016 magic-lang
- *
- * This software may be modified and distributed under the terms
- * of the MIT license.  See the LICENSE file for details.
- */
-
-use geometry
-use base
-import ByteBuffer into ByteBuffer
```
marcusnaslund: Didn't know you could do this. Or... what it really means in ooc.

marcusnaslund:
#1727 is merged.

davidpiuva:
@emilwestergren Peer review?

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -35,12 +61,19 @@ RasterYuv420Semiplanar: class extends RasterYuvSemiplanar {
...
 	init: func ~fromByteBuffer (buffer: ByteBuffer, size: IntVector2D, stride: UInt, uvOffset: UInt) {
 		(yImage, uvImage) := This _createSubimages(buffer, size, stride, uvOffset)
 		this init(yImage, uvImage)
 	}
+	free: override func {
+		(this y, this uv) referenceCount decrease()
+		super()
+	}
+	distance: override func (other: Image) -> Float {
+		(this y distance((other as This) y) + this uv distance((other as This) uv)) / 2.0f
```
davidpiuva: This was simplified and should be a lot faster by not reading each UV pixel 4 times in Y resolution.

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -203,7 +236,7 @@ RasterYuv420Semiplanar: class extends RasterYuvSemiplanar {
 		(RasterMonochrome new(buffer slice(0, yLength), size, stride), RasterUv new(buffer slice(uvOffset, uvLength), This _uvSize(size), stride))
 	}
 	_uvSize: static func (size: IntVector2D) -> IntVector2D {
-		IntVector2D new(size x / 2 + (size x isOdd ? 1 : 0), size y / 2 + (size y isOdd ? 1 : 0))
+		IntVector2D new((size x + 1) / 2, (size y + 1) / 2)
```
davidpiuva: Integer division can be used to get the same result without if statements.

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -24,9 +23,36 @@ import io/FileReader
...
+		this _uv = uvImage
+		this _uv referenceCount increase()
+	}
+	init: func ~fromYuvSemiplanar (original: This, y: RasterMonochrome, uv: RasterUv) {
+		super(original)
+		this _y = y
+		this _y referenceCount increase()
+		this _uv = uv
+		this _uv referenceCount increase()
+	}
```
davidpiuva: These 16 constructors in different layers of inheritance are very redundant but I can't find a way to get rid of them.

source/draw/Image.ooc
```diff
@@ -17,15 +17,13 @@ Image: abstract class {
 	width ::= this size x
 	height ::= this size y
 	crop: IntShell2D { get set }
```
davidpiuva: Crop does not belong here no matter what it is used for.

source/draw/Image.ooc
```diff
@@ -17,15 +17,13 @@ Image: abstract class {
...
-	wrap: Bool { get set }
```
davidpiuva: Wrap was not even used.

source/draw/Image.ooc
```diff
@@ -16,16 +16,12 @@ Image: abstract class {
 	size ::= this _size
 	width ::= this size x
 	height ::= this size y
-	crop: IntShell2D { get set }
-	wrap: Bool { get set }
```
davidpiuva: Both were dead code.

davidpiuva:
@ziquanyu Peer review?

source/draw/RasterYuv420Semiplanar.ooc
ziquanyu: Will this work on every phone platform?

source/draw/Image.ooc
ziquanyu: If you have tested they are truly dead. I don't have any problem with this.

source/draw/RasterYuv420Semiplanar.ooc
davidpiuva: I don't know but it should be mathematically equivalent to the previous formula.

ziquanyu: Thanks @sebastianbaginski for the info.
according to the C99 standard, the expression (a/b)*b + a%b shall equal a, so if (5/2) == 3, then 3*2 + 1 != 5. It must be 2 according to the C99 : [http://port70.net/~nsz/c/c99/n1256.html#6.5.5p6](url)

@davidpiuva, i'm fine with this one now.

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -24,9 +23,26 @@ import io/FileReader
...
 
-RasterYuv420Semiplanar: class extends RasterYuvSemiplanar {
+RasterYuv420Semiplanar: class extends RasterImage {
+	_y: RasterMonochrome
+	_uv: RasterUv
+	y ::= this _y
+	uv ::= this _uv
 	stride ::= this _y stride
-	init: func ~fromRasterImages (yImage: RasterMonochrome, uvImage: RasterUv) { super(yImage, uvImage) }
+	init: func ~fromRasterImages (yImage: RasterMonochrome, uvImage: RasterUv) {
```
marcusnaslund: Is it not possible to do `(=_y, =_uv)` in the constructor? Or does the `super` call have to occur first?

marcusnaslund: Also maybe `(this_y, this _uv) referenceCount increase()` would save some space.

davidpiuva: I will think about it for next PR since this one is getting quite big.

marcusnaslund: Sure.

marcusnaslund:
Looks good.

PR #1734, opened 2016-05-23
Fixed bug when unpacking texture using AndroidContext.

emilwestergren:
Temporary fix to make it compatible with reference platform.

@davidpiuva peer review


davidpiuva:
Since the previous way did not work, this should at least be an improvement.

marcusnaslund:
From `test.sh with -Dsafe`:
```
16:35:55 BufferAngularVelocitySensor .Warning: GyroSensor returned 0 samples! . done
```
Is this new?

PR #1739, opened 2016-05-25
test for binary search with covers

sebastianbaginski:
I had some weird problems yesterday when trying to use it on a vector list with covers, in the other project. Seems to work fine here though.
@marcusnaslund review


marcusnaslund:
Have you run `memcheck.sh` to make sure there are no invalid read/writes etc. going on?

sebastianbaginski:
valgrind does not report any problems.

sebastianbaginski:
@marcusnaslund updated `upperBound`

marcusnaslund:
Great, thanks.

PR #1741, opened 2016-05-25
Cleaning up image constructors

davidpiuva:
Removed some of the image constructors that took an image just to get the same size.
I am trying to override the copy function instead of using the constructors for everything but since there are 13 constructors to choose from, any remaining call to the removed constructor will always match another constructor with a completely different meaning.

### DIFF
--- a/source/draw/Image.ooc
+++ b/source/draw/Image.ooc
@@ -20,9 +20,6 @@ Image: abstract class {
 	init: func (=_size) {
 		this _referenceCount = ReferenceCounter new(this)
 	}
-	init: func ~fromImage (original: This) {
-		this init(original size)
-	}
 	free: override func {
 		if (this referenceCount != null)
 			this referenceCount free()
--- a/source/draw/RasterYuv420Semiplanar.ooc
+++ b/source/draw/RasterYuv420Semiplanar.ooc
@@ -29,19 +29,9 @@ RasterYuv420Semiplanar: class extends RasterImage {
 	y ::= this _y
 	uv ::= this _uv
 	stride ::= this _y stride
-	init: func ~fromRasterImages (yImage: RasterMonochrome, uvImage: RasterUv) {
-		super(yImage size)
-		this _y = yImage
-		this _y referenceCount increase()
-		this _uv = uvImage
-		this _uv referenceCount increase()
-	}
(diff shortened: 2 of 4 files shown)


source/draw/Image.ooc
```diff
@@ -20,9 +20,6 @@ Image: abstract class {
 	init: func (=_size) {
 		this _referenceCount = ReferenceCounter new(this)
 	}
-	init: func ~fromImage (original: This) {
-		this init(original size)
-	}
```
davidpiuva: This is all that remained after removing crop and wrap from Image so now it has no purpose.

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -29,19 +29,9 @@ RasterYuv420Semiplanar: class extends RasterImage {
...
-	}
-	init: func ~fromYuvSemiplanar (original: This, y: RasterMonochrome, uv: RasterUv) {
-		super(original)
-		this _y = y
-		this _y referenceCount increase()
-		this _uv = uv
-		this _uv referenceCount increase()
+	init: func ~fromRasterImages (=_y, =_uv) {
+		super(this _y size)
+		(this _y, this _uv) referenceCount increase()
```
davidpiuva: Two constructors were merged into one because they both just gave a size to the super constructor.

source/draw/RasterYuv420Semiplanar.ooc
```diff
@@ -29,19 +29,9 @@ RasterYuv420Semiplanar: class extends RasterImage {
...
 	}
```
davidpiuva: This merge assumes that no one uses the stride from the YUV image as a whole since the getter for stride in RasterImage is meant to be overridden in RasterPacked. Removing stride from Image would however break the API.

davidpiuva:

marcusnaslund:
Brilliant.

sebastianbaginski:
Great that someone started cleaning up this mess :+1:

davidpiuva:
The worst still remains since naming the copy constructors will just crash on every attempt.

marcusnaslund:
*The best is yet to come...* :notes:

PR #1742, opened 2016-05-26
Removing more image constructors

davidpiuva:
Removing two image constructors that were almost impossible to remove. I had to remove the constructor taking RasterImage before I could remove the constructor taking RasterPacked. When done in the wrong order, calls with RasterPacked were automatically accepted as the unpadded non-copying method accepting RasterPacked as RasterImage leading to segmentation fault and missing data.

### DIFF
--- a/source/draw/RasterPacked.ooc
+++ b/source/draw/RasterPacked.ooc
@@ -32,11 +32,6 @@ RasterPacked: abstract class extends RasterImage {
 		stride := this bytesPerPixel * size x
 		this init(ByteBuffer new(stride * size y), size, stride)
 	}
-	init: func ~fromOriginal (original: This) {
-		super(original size)
-		this _buffer = original buffer copy()
-		this _stride = original stride
-	}
 	free: override func {
 		if (this _buffer != null)
 			this _buffer referenceCount decrease()
--- a/source/draw/RasterPacked.ooc
+++ b/source/draw/RasterPacked.ooc
@@ -37,11 +37,6 @@ RasterPacked: abstract class extends RasterImage {
 		this _buffer = original buffer copy()
 		this _stride = original stride
 	}
-	init: func ~fromRasterImage (original: RasterImage) {
-		super(original size)
-		this _stride = this bytesPerPixel * original width
-		this _buffer = ByteBuffer new(this stride * original height)
-	}
 	free: override func {
 		if (this _buffer != null)
 			this _buffer referenceCount decrease()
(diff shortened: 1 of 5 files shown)


source/draw/RasterRgb.ooc
```diff
@@ -162,7 +160,7 @@ RasterRgb: class extends RasterPacked {
 		if (original instanceOf(This))
 			result = (original as This) copy()
 		else {
-			result = This new(original)
+			result = This new(original size)
```
davidpiuva: When the copied image was not packed, padding would be removed to handle sub-images.
When the copied image was packed, padding was preserved.

To remove both constructors, both cases will have to remove the padding to be on the safe side with YUV images.

source/draw/RasterRgb.ooc
```diff
@@ -51,7 +49,7 @@ RasterRgb: class extends RasterPacked {
 			for (x in 0 .. this size x)
 				this[x, y] = color toRgb()
 	}
-	copy: override func -> This { This new(this) }
+	copy: override func -> This { This new(this buffer copy(), this size, this stride) }
```
davidpiuva: Stride is preserved when copying because that is how it used to work.

source/draw/RasterPacked.ooc
```diff
@@ -32,16 +32,6 @@ RasterPacked: abstract class extends RasterImage {
...
-	init: func ~fromOriginal (original: This) {
-		super(original size)
-		this _buffer = original buffer copy()
-		this _stride = original stride
-	}
-	init: func ~fromRasterImage (original: RasterImage) {
-		super(original size)
-		this _stride = this bytesPerPixel * original width
-		this _buffer = ByteBuffer new(this stride * original height)
-	}
```
davidpiuva: These two constructors had to be removed because they accept the same input types and the choice between them is decided at runtime depending on if original is packed or planar. `~fromOriginal` made a copy of the image while preserving padding while `~fromRasterImage` made a new image from the size without preserving padding.

davidpiuva:

source/draw/RasterPacked.ooc
sebastianbaginski: I agree, copy should be explicit.

sebastianbaginski:
This looks good and I assume you have tested the other project manually as well.

marcusnaslund:
Let's have @emilwestergren approve this as well when he's back.

PR #1743, opened 2016-05-26
Removed half implemented multi-context support

davidpiuva:


davidpiuva:
It seems like we have tests that still use multiple OpenGL contexts but I don't know exactly which one Emil wanted gone.

PR #1744, opened 2016-05-26
Keyword

davidpiuva:
Changed two keyword identifiers. One of them caused text colors to malfunction in Atom.

Cells are also using the get keyword as an identifier but that is hard to change without version conflicts.

### DIFF
--- a/source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
+++ b/source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
@@ -23,10 +23,10 @@ Gles3ShaderProgram: class extends GLShaderProgram {
 		version(debugGL) { validateEnd("ShaderProgram dispose") }
 		super()
 	}
-	use: override func {
-		version(debugGL) { validateStart("ShaderProgram use") }
+	useProgram: override func {
+		version(debugGL) { validateStart("ShaderProgram useProgram") }
 		glUseProgram(this _backend)
-		version(debugGL) { validateEnd("ShaderProgram use") }
+		version(debugGL) { validateEnd("ShaderProgram useProgram") }
 	}
 	setUniform: override func ~Int (name: String, x: Int) {
 		version(debugGL) { validateStart("ShaderProgram setUniform~Int") }
--- a/source/draw/Map.ooc
+++ b/source/draw/Map.ooc
@@ -59,7 +59,7 @@ Map: abstract class {
 		this _bindings free()
 		super()
 	}
-	use: abstract func (forbiddenInput: Pointer, positionTransform, textureTransform: FloatTransform3D)
+	useProgram: abstract func (forbiddenInput: Pointer, positionTransform, textureTransform: FloatTransform3D)
 	add: func <T> (key: String, value: T) {
 		if (T inheritsFrom(Object))
 			this _bindings add(key, value as Object, false)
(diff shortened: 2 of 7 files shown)


davidpiuva:

marcusnaslund:
I might agree that `useProgram` is more descriptive than `use`. But bany more classes have a `get` implementation (`LinkedList`, `HashMap`, `CharBuffer`, `Env`) Isn't a better solution to ignore, or fix the syntax highlighter (at least for VS Code we can do that)? I'm a bit skeptical here. @fredrikbryntesson @thomasfanell ?

marcusnaslund:
I will leave this frozen over the weekend anyway, so there will be time to discuss it.

thomasfanell:
@marcusnaslund VS Code handle `get` and `set`, but fails on `use`. I've been meaning to improve the highlighter quite a bit, but that is on low priority right now.

davidpiuva:
The main problem is that using keywords as identifiers is against good parsing convention and might randomly fail if used in an ambiguous syntax. This is often the case with other languages that are unable to detect keywords overlapping identifiers leading to mysterious crashes.

thomasfanell:
@davidpiuva I agree, I also think it's a bad idea.

marcusnaslund:
> might randomly fail

marcusnaslund:
Still not convinced about renaming `get`. In C#, for example it is allowed to call a method `get` or `set` because they're different from other keywords:

"In some places in the grammar, specific identifiers have special meaning, but are not keywords. For example, within a property declaration, the "get" and "set" identifiers have special meaning (Section 10.6.2). An identifier other than get or set is never permitted in these locations, so this use does not conflict with a use of these words as identifiers."
( https://msdn.microsoft.com/en-us/library/aa664671%28v=vs.71%29.aspx )

marcusnaslund:
I agree about renaming `use`, though.

marcusnaslund:
After talking to @fredrikbryntesson and @sebastianbaginski we've decided to keep `get` and `set` as acceptable method names, partly for the same reasons as in C#.

So good thing you've made two separate commits, just rebase (to remove conflicts) and then remove the latest commit and `Map` will remain consistent with similar classes implementing get and set.

Put the final review flag back once it's done.

davidpiuva:
Rebased

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
```diff
@@ -23,7 +23,7 @@ Gles3ShaderProgram: class extends GLShaderProgram {
 		version(debugGL) { validateEnd("ShaderProgram dispose") }
 		super()
 	}
-	use: override func {
+	useProgram: override func {
 		version(debugGL) { validateStart("ShaderProgram use") }
 		glUseProgram(this _backend)
 		version(debugGL) { validateEnd("ShaderProgram use") }
```
marcusnaslund: These strings should be `useProgram` as well I suppose?

davidpiuva: Yes

source/draw/gpu/opengl/backend/gles3/Gles3ShaderProgram.ooc
```diff
@@ -23,10 +23,10 @@ Gles3ShaderProgram: class extends GLShaderProgram {
 		version(debugGL) { validateEnd("ShaderProgram dispose") }
 		super()
 	}
-	use: override func {
-		version(debugGL) { validateStart("ShaderProgram use") }
+	useProgram: override func {
+		version(debugGL) { validateStart("ShaderProgram useProgram") }
 		glUseProgram(this _backend)
-		version(debugGL) { validateEnd("ShaderProgram use") }
+		version(debugGL) { validateEnd("ShaderProgram useProgram") }
```
davidpiuva: Updated

PR #1745, opened 2016-05-26
Bugfix in DateTime toText()

marcusnaslund:
Peer review @thomasfanell ?


thomasfanell:
Took a while, but it sure looks good to me

PR #1749, opened 2016-05-30
Merged high-level close() and free() methods

### LINKED ISSUE #1740: Merge close and free methods
Freeing a, for example `FileWriter`, should also close it. Both calls should never be neccessary.

marcusnaslund:
Fixes #1740 
Breaks API but makes it a lot easier to use many I/O and Net classes.

`File*` and `Net` classes now only require a single `free` call (`close` mostly no longer exists, at least not for external use).
`close(Char)` still exists for `Pipe`, but `free` automatically calls it, so there is no need to be explicit about it except in certain circumstances.


marcusnaslund:

marcusnaslund:
Works perfectly locally, but PRInt hangs after the first gpu-off test, although all the kean tests run nicely. Since the Pipe classes have changed it's maybe something because of that. Any ideas @jonathanudd @fredrikbryntesson ?

marcusnaslund:
Ahh nevermind, I can reproduce some errors locally as well. Taking it off final review for now.

marcusnaslund:
Yay, it works. Let me do some more tests before I move this on to review.

marcusnaslund:

PR #1750, opened 2016-05-30
Moved preallocation and preregistering to render contexts

emilwestergren:
@davidpiuva peer review

### DIFF
--- a/source/draw/gpu/opengl/AndroidContext.ooc
+++ b/source/draw/gpu/opengl/AndroidContext.ooc
@@ -141,5 +141,6 @@ AndroidContext: class extends OpenGLContext {
 		map add("startY", startY)
 		DrawState new(target) setMap(map) draw()
 	}
+	preallocate: override func (resolution: IntVector2D) { GraphicBufferYuv420Semiplanar free~all() }
 }
 }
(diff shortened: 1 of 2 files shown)


davidpiuva:
ok

source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -141,5 +141,10 @@ AndroidContext: class extends OpenGLContext {
 		map add("startY", startY)
 		DrawState new(target) setMap(map) draw()
 	}
+	preallocate: override func (resolution: IntVector2D) { GraphicBufferYuv420Semiplanar free~all() }
+	preregister: override func (image: Image) {
+		if (image instanceOf(GraphicBufferYuv420Semiplanar))
+			(image as GraphicBufferYuv420Semiplanar) toRgba(this) referenceCount decrease()
```
davidpiuva: You might have to re-test this later since I am not sure if YUV to RGBA works right at this moment.

davidpiuva:
I don't see any obvious mistakes in the code so I am done reviewing.

source/draw/gpu/opengl/AndroidContext.ooc
emilwestergren: This is actually not a conversion, it's more a type cast. We simply interpret the YUV buffer as an RGBA buffer instead. It is then converted to YUV in the unpack/pack shaders.

davidpiuva: But if I flip YVU to YUV or BGR to RGB then the typecast may break from my changes if it depends on order.

PR #1752, opened 2016-05-30
Fixed egl parameter that caused issue with nvidia cards.

niklasborwell:
This was needed for compatibility with our new computers, but I'm not entirely sure what I just did. Maybe @emilwestergren can peer review and say if this is ok or not. Print seems fine at least.

### DIFF
--- a/source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
+++ b/source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
@@ -127,7 +127,7 @@ Gles3Context: class extends GLContext {
 			EGL_ALPHA_SIZE, 8,
 			EGL_SAMPLES, 0,
 			EGL_DEPTH_SIZE, 0,
-			EGL_BIND_TO_TEXTURE_RGBA, EGL_TRUE,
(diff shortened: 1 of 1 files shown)


emilwestergren:
Shouldn't affect anything but of course it should be tested on a device.

marcusnaslund:
> of course it should be tested on a device.

It doesn't crash and stabilization seems to work as before. That's all I can say.

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -127,7 +127,7 @@ Gles3Context: class extends GLContext {
 			EGL_ALPHA_SIZE, 8,
 			EGL_SAMPLES, 0,
 			EGL_DEPTH_SIZE, 0,
-			EGL_BIND_TO_TEXTURE_RGBA, EGL_TRUE,
+			EGL_BIND_TO_TEXTURE_RGBA, EGL_DONT_CARE,
```
marcusnaslund: *I don't care, I love it.* :notes:

PR #1753, opened 2016-05-30
Added a simple StringBuilder

marcusnaslund:
I think this is the only functionality we need at the moment.
Useful for replacing `TextBuilder` with the final aim of removing `Text`.

Peer review @thomasfanell ? Anything you feel is missing in this class?


marcusnaslund:
Oh yeah, he's not here. @sebastianbaginski ?

Issue #1724, opened 2016-05-18
Remove Text


marcusnaslund:
Yes. 

Since string literals don't leak anymore, and since people are mostly just confused by how `Text` ownership works, there is no need for `Text`. Right now we just have a lot of duplicate code with `toString` and `toText` in tons of classes.

We've also caused a lot of confusion for ourselves, where for example the `filename` arguments to `FileWriter` and `FileReader` are different, one of them expects a `String`, the other a `Text`...

We can do the major breaking changes during the summer when most people are on holiday.

Thoughts, objections? @sebastianbaginski @thomasfanell @fredrikbryntesson @emilwestergren @davidpiuva ?

EDIT: It's decided. Here is what to do.

- [x] `source/uri`
- [x] `DateTime` and `TimeSpan`
- [x] `Cell` (and its extensions)
- [x] `source/unit`
- [x] `source/math`
- [x] `source/geometry` (breaks API)
- [x] `source/draw` (breaks API)
- [x] `CsvReader` and `CsvWriter` (breaks API)
- [x] `ArgumentParser` (breaks API)
- [x] `TextBuilder`
- [x] Any other minor uses
- [ ] Remove `Text`, `TextBuffer`

emilwestergren:
I've never even used `Text` anyway so it's fine by me.

davidpiuva:
Great! Caller should be the owner unless explicitly stated otherwise.

thomasfanell:
Good! I always hated that thing.

sebastianbaginski:
I guess there is no other way.

marcusnaslund:
So it's decided. Good.

davidhesselbom:
> Since string literals don't leak anymore

Cool, how'd you fix it?

sebastianbaginski:
@davidhesselbom : https://github.com/magic-lang/ooc-kean/blob/develop/source/system/string/string_literal.c :/

davidhesselbom:
Oh, I see. Glad you finally found a way, that thing bothered me to bits...

marcusnaslund:
Everything but `Text` itself is gone now. It will disappear with #1835 , so closing this.

PR #1756, opened 2016-05-31
Removed use of Text in uri namespace

marcusnaslund:
Since it's not used anywhere ( @jonathanudd ? ) this is easy to do without breaking the API for anyone.

Also fixed memory leak in `StringBuilder` and `String split(String)`.

Part of the work on #1724

### DIFF
--- a/source/base/StringBuilder.ooc
+++ b/source/base/StringBuilder.ooc
@@ -7,8 +7,10 @@
  */
 
 StringBuilder: class {
-	_items := VectorList<String> new()
-	init: func
+	_items: VectorList<String>
+	init: func (capacity := 32) {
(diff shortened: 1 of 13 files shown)


jonathanudd:
It's not used anywhere as far as I know.

marcusnaslund:
Since you wrote the original, might I trouble you for a peer review @jonathanudd ?

Not everything in the tests is top notch, but I've confirmed they all pass and no memory leaks. For a namespace that isn't used at all, that's probably good enough.

jonathanudd:
I don't see any problems with this :)

marcusnaslund:
:tada:

PR #1758, opened 2016-05-31
Giving BGR to X11 window

davidpiuva:
I assume that the opposite has to be done for Windows but I can't test it on my computer.

### DIFF
--- a/source/ui/win32/Win32Window.ooc
+++ b/source/ui/win32/Win32Window.ooc
@@ -42,7 +42,6 @@ Win32Window: class extends NativeWindow {
 		super()
 	}
 	draw: func (image: RasterRgba) {
-		image swapRedBlue()
 		paintStruct: PaintStruct
 		bitmap: Bitmap
 		InvalidateRect(this backend as HWND, null, false)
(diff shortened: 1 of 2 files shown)


davidpiuva:
@emilwestergren Peer review?

source/ui/win32/Win32Window.ooc
```diff
@@ -42,7 +42,6 @@ Win32Window: class extends NativeWindow {
 		super()
 	}
 	draw: func (image: RasterRgba) {
-		image swapRedBlue()
```
davidpiuva: If BGR was given to X11 then I have to assume that the same goes for Win32 that wants BGR format.

source/ui/x11/X11Window.ooc
```diff
@@ -64,6 +64,7 @@ X11Window: class extends NativeWindow {
 	}
 	resize: func (size: IntVector2D) { XResizeWindow(this display, this backend, size x, size y) }
 	draw: func (image: RasterRgba) {
+		image swapRedBlue()
```
davidpiuva: Colors look normal with this change.

PR #1760, opened 2016-05-31
toText replaced with toString in DateTime and TimeSpan

marcusnaslund:
Not used in other projects, so doesn't break the API. :)
Also fixes `replaceAll` in `String`, it used to leak badly.

### DIFF
--- a/source/base/DateTime.ooc
+++ b/source/base/DateTime.ooc
@@ -79,24 +79,43 @@ DateTime: cover {
 	//	%mm - minute
 	//	%ss - second
 	//	%zzz - millisecond
-	toText: func (format := This defaultFormat) -> Text {
-		result := format copy()
+	toString: func (format := This defaultFormat) -> String {
 		data := This _ticksToDateTimeHelper(this ticks)
-		result = result replaceAll(t"%yyyy", t"%04d" format(data year))
-		result = result replaceAll(t"%yy", t"%02d" format(data year % 100))
-		result = result replaceAll(t"%y", t"%d" format(data year))
-		result = result replaceAll(t"%MM", t"%02d" format(data month))
-		result = result replaceAll(t"%dd", t"%02d" format(data day))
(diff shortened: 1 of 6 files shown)


source/base/DateTime.ooc
```diff
@@ -79,24 +79,43 @@ DateTime: cover {
...
+		second free()
+		second = result replaceAll("%h", hourString)
+		result free()
+		result = second replaceAll("%m", minuteString)
+		second free()
+		second = result replaceAll("%s", secondString)
+		result free()
+		result = second replaceAll("%zzz", millisecond3String)
+		second free()
+		second = result replaceAll("%z", millisecondString)
```
marcusnaslund: Yes this is ugly. We might consider implementing some `replaceAllAndFreeOriginal` methods (with a better name of course) and similarly for some other `String` methods... what do you think @sebastianbaginski @thomasfanell ?

thomasfanell: Is there a strong reason why we can't have `replaceAll` destroy the original `String`?

marcusnaslund: Consistency, I'd say, no other method destroys the original (I think?).

thomasfanell: Would a default `Bool` be prettier (defaults to false)? If true, then free the original. That would cater to the existing style and just add the option of an automatic free.

marcusnaslund: That works for me. I'd like that for `print` and `println` as well.

marcusnaslund: Let's be happy with this for now, I will create an issue to discuss this further.

marcusnaslund: #1763

marcusnaslund:

source/base/DateTime.ooc
sebastianbaginski: > what do you think @sebastianbaginski

I think we need to move away from ooc asap.

PR #1765, opened 2016-06-01
Replaced Text with String in source/math

marcusnaslund:
Also fixes leaks in `toString` of some `*List` classes.

0 leaks from these tests now.

### DIFF
--- a/test/math/BoolVectorListTest.ooc
+++ b/test/math/BoolVectorListTest.ooc
@@ -30,7 +30,7 @@ BoolVectorListTest: class extends Fixture {
 			(list, reversed) free()
 		})
 		this add("dilation", func {
-			list := this _createFromText(t"1100 0011 1000 1110")
+			list := this _createFromString("1100 0011 1000 1110")
 			dilated := list dilation(3)
 			expect(dilated[2])
 			expect(!dilated[3])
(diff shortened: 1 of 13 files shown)


test/math/BoolVectorListTest.ooc
```diff
@@ -84,30 +84,30 @@ BoolVectorListTest: class extends Fixture {
...
+			text := list toString()
+			expect(text, is equal to("true\nfalse\ntrue\n"))
 			(text, list) free()
 		})
 	}
-	_createFromText: func (content: Text) -> BoolVectorList {
+	_createFromString: func (content: String) -> BoolVectorList {
 		result := BoolVectorList new()
-		for (i in 0 .. content count) {
+		for (i in 0 .. content size) {
```
marcusnaslund: Funny story: `String` has a method called `count` which does something completely different from `Text count`. Before I changed this for-loop, `rock` would actually compile `for (i in 0 .. content count)` without any warnings or anything (i.e. the end of the loop was a pointer to a method?) but would of course crash horribly once the program ran.

marcusnaslund:

test/math/BoolVectorListTest.ooc
sebastianbaginski: Nice...

> the end of the loop was a pointer to a method

Implicitly converted to integer I assume :)

PR #1766, opened 2016-06-01
Removed Text from source/geometry

marcusnaslund:
- Removed `toText` and changed `parse` to accept `String` instead.
- Also removed not needed `toText` methods from `Numbers.ooc` and `types.ooc`.
- Moved implicit conversions from `Text` to `String`.

Confirmed 0 memory leaks.


marcusnaslund:
Peer review @thomasfanell ?

marcusnaslund:
Oh, this breaks the API more than I thought. Better wait a while with this one.

marcusnaslund:
Alright, ready now. Peer review @thomasfanell ?

PR #1771, opened 2016-06-02
Added parse() for IntVectorList and FloatVectorList

### LINKED ISSUE #1770: Add parse() method for the *List classes
And since `toString` takes a separator parameter, so should the parse method.

This applies for the `parse` methods that already exist in the geometry namespace.

marcusnaslund:
I think this is all we need to close #1770 

Peer review @bettinaselig or @patrikmalm ?


patrikmalm:
Looks good to me

PR #1774, opened 2016-06-05
Added unit test for String

### LINKED ISSUE #1772: Add unit test for String

marcusnaslund:
We've never had one.

(Also fixes the correct order of the string type error message in `Fixture`, and a leak in `capitalize`)

Fixes #1772

### DIFF
--- a/source/system/String.ooc
+++ b/source/system/String.ooc
@@ -104,11 +104,10 @@ String: class {
 		result
 	}
 	capitalize: func -> This {
-		match (this size) {
-			case 0 => this
-			case 1 => this toUpper()
-			case => this[0] toUpper() + this substring(1)
-		}
+		result := this clone()
+		if (result size > 0)
+			result _buffer[0] = result[0] toUpper()
+		result
 	}
 	cformat: final func ~str (...) -> This {
 		list: VaList
--- a/test/system/StringTest.ooc
+++ b/test/system/StringTest.ooc
@@ -1,0 +1,155 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use unit
+
+StringTest: class extends Fixture {
+	init: func {
+		super("String")
+		this add("length, substring", func {
+			string := "12345678"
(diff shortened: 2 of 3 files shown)


marcusnaslund:
There are some leaks in `prepend` and `append`, but I will fix those in a later PR (some overloads seem fixed already, but need to take a closer look).

source/system/String.ooc
```diff
@@ -107,7 +107,11 @@ String: class {
 		match (this size) {
 			case 0 => this
 			case 1 => this toUpper()
-			case => this[0] toUpper() + this substring(1)
+			case =>
+				end := this substring(1)
+				result := this[0] toUpper() + end
+				end free()
+				result
```
sebastianbaginski: I think this function is more complicated with this `case` than it should be. Maybe:
```ooc
capitalize: func -> This {
    result := this clone()
    if (result size > 0)
        result[0] = result[0] toUpper()
    result
}
```
?

marcusnaslund: `result _buffer[0] = result[0] toUpper()` works, but it also violates the immutability of `String`s. I have to agree it looks a lot better, though.

marcusnaslund: Do you think I should to it that way anyway?

sebastianbaginski: Immutability should affect only clients of the class, there is nothing wrong with bending the rules inside the implementation details. As long as it does not leak outside.

marcusnaslund: *A world without rules and controls, without borders or boundaries. A world where anything is possible.*

marcusnaslund: Updated

test/system/StringTest.ooc
```diff
@@ -0,0 +1,155 @@
...
+			string := "123"
+			times := string times(3)
+			expect(times, is equal to("123123123"))
+			times free()
+
+			prepend := string prepend("0")
+			append := prepend append("45")
+			expect(append, is equal to("012345"))
+			(prepend, append) free()
+		})
```
davidpiuva: I get a memory leak from this test.

sebastianbaginski: I think @marcusnaslund is going to fix them in another PR.

marcusnaslund: @davidpiuva If you mean `prepend` and `append` leak, then yes, I already mentioned that in a comment above.

Issue #1764, opened 2016-06-01
Freeing String literals


marcusnaslund:
Right now, calling `free` on a string literal will cause a segfault.

There should be a check in `free` to check whether the current object is a literal, and if so, not do anything. The `String` cleanup should be the only one which can free the string literals, perhaps via setting a "secret" flag in the `String` instance.

This will allow us to do stuff like `"literal" & variable` without problems. (Right now `"literal" << variable` must be used to not leak).

sebastianbaginski:
>  calling free on a string literal will cause a segfault.

How do I reproduce ? This does not crash
```ooc
use base

x := "test"
x free()

"a" free()

GlobalCleanup run()
```

davidpiuva:
Literals can be freed but they are only created once. Running a function with `"a" & "b"` should crash on second execution.

sebastianbaginski:
Ok, thanks.

PR #1776, opened 2016-06-07
Safe free

davidpiuva:
A `safeFree` method is used when it is not desired to free literals.

Fixes https://github.com/magic-lang/ooc-kean/issues/1764

### DIFF
--- a/source/system/String.ooc
+++ b/source/system/String.ooc
@@ -16,7 +16,7 @@ String: class {
 	// Avoid direct buffer access, as it breaks immutability.
 	_buffer: CharBuffer
 	size ::= this _buffer size
-
+	_isLiteral := false
 	init: func ~withBuffer (=_buffer)
 	init: func ~withCStr (s: CString) { init(s, s length()) }
 	init: func ~withCStrAndLength (s: CString, length: Int) { this _buffer = CharBuffer new(s, length) }
@@ -26,6 +26,7 @@ String: class {
 		this _buffer free()
 		super()
 	}
+	safeFree: func { if (!this _isLiteral) this free() } // Will not free literals automatically
 	equals: final func (other: This) -> Bool {
 		result := false
 		if (this == null)
@@ -45,7 +46,7 @@ String: class {
 	}
 	append: func ~char (other: Char) -> This {
 		newBuffer := (this _buffer clone(this size + 1)) . append(other)
-		this free()
+		this safeFree()
 		newBuffer toString()
 	}
 	append: func ~cStr (other: CString) -> This { (this _buffer clone(this size + other length())) append(other, other length()) . toString() }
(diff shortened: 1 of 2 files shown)


davidpiuva:

source/system/String.ooc
```diff
@@ -26,6 +26,7 @@ String: class {
 		this _buffer free()
 		super()
 	}
+	safeFree: func { if (!this _isLiteral) this free() } // Will not free literals automatically
```
sebastianbaginski: What happens to `CharBuffer` member for string literals ?

davidpiuva: They will be freed globally from the C code.

source/system/String.ooc
```diff
@@ -16,7 +16,7 @@ String: class {
 	// Avoid direct buffer access, as it breaks immutability.
 	_buffer: CharBuffer
 	size ::= this _buffer size
-
+	_isLiteral := false
```
sebastianbaginski: I think this is the same information as `this _buffer mallocAddr == 0`

davidpiuva: I could make a getter then.

sebastianbaginski: It is null if third argument to `CharBuffer` constructor is `true`, which seems to be the case only in `makeStringLiteral`.

source/system/String.ooc
sebastianbaginski: Does `safeFree` need to be public ?

davidpiuva: If someone gave me a string to keep, I would like to free only if it is not reused.

sebastianbaginski: I'd rather not have to worry about if it's reused or not. I think this should be hidden inside the implementation details. Interface for string objects should be the same if object is a literal or not (something like we have for recyclable buffers now).

davidpiuva: Another way is to only allow global termination to free literals.

sebastianbaginski: Yeah, I think this could be good.

davidpiuva: @marcusnaslund Should we make another attempt at this?

sebastianbaginski: Would this require much more than a private `_forceFree` kind of method, which gets called only by a static String cleanup for string literals ? So the code like `"a" free()` becomes a NOP for literals.

davidpiuva: We could probably block the regular free using a hidden lock flag but explicitly freeing literals is bad style if it does nothing and there were leaks last time I tried using a lock.

davidpiuva: I made it work in https://github.com/magic-lang/ooc-kean/pull/1777.

davidpiuva:
Replaced by https://github.com/magic-lang/ooc-kean/pull/1777/files

PR #1777, opened 2016-06-08
Locking literals from being freed until global termination

### LINKED ISSUE #1764: Freeing String literals

davidpiuva:
Alternative solution to https://github.com/magic-lang/ooc-kean/pull/1776

Fixes #1764

### DIFF
--- a/source/system/String.ooc
+++ b/source/system/String.ooc
@@ -17,7 +17,6 @@ String: class {
 	_buffer: CharBuffer
 	size ::= this _buffer size
 	_locked := false // Only global termination may unlock and free literals
-	_unlock: func { this _locked = false }
 	init: func ~withBuffer (=_buffer)
 	init: func ~withCStr (s: CString) { init(s, s length()) }
 	init: func ~withCStrAndLength (s: CString, length: Int) { this _buffer = CharBuffer new(s, length) }
@@ -29,6 +28,7 @@ String: class {
 			super()
 		}
 	}
+	_unlock: func { this _locked = false }
 	equals: final func (other: This) -> Bool {
 		result := false
 		if (this == null)
--- a/source/system/String.ooc
+++ b/source/system/String.ooc
@@ -16,15 +16,18 @@ String: class {
 	// Avoid direct buffer access, as it breaks immutability.
 	_buffer: CharBuffer
 	size ::= this _buffer size
-
+	_locked := false // Only global termination may unlock and free literals
+	_unlock: func { this _locked = false }
 	init: func ~withBuffer (=_buffer)
 	init: func ~withCStr (s: CString) { init(s, s length()) }
 	init: func ~withCStrAndLength (s: CString, length: Int) { this _buffer = CharBuffer new(s, length) }
 	free: override func {
-		if (this _buffer mallocAddr == 0)
-			string_literal_free(this)
-		this _buffer free()
-		super()
+		if (!this _locked) {
+			if (this _buffer mallocAddr == 0)
+				string_literal_free(this)
+			this _buffer free()
+			super()
+		}
 	}
 	equals: final func (other: This) -> Bool {
 		result := false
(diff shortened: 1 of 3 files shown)


marcusnaslund:
This is what I had in mind, actually.

Another alternative is to leave `free` as-is and have the concatenation operators `<<`, `>>`, `&` only call free on the argument(s) when they're not literals. Not sure that's any better though. @sebastianbaginski ?

davidpiuva:
I am still getting leaks when running this test:
```
result: String
for (i in 1 .. 100) {
	result = "ab" & "cd"
	expect(result, is equal to("abcd"))
}
```
Maybe something wrong with my global freeing.

sebastianbaginski:
I'd like to avoid having different interfaces for literals and regular strings. Actually I like this solution.

source/system/string/string_literal.c
```diff
@@ -59,8 +59,10 @@ void string_literal_free(void* ptr) {
 void string_literal_free_all() {
 	if (_literals) {
 		for (size_t i=0; i<_literalsCount; ++i)
-			if (_literals[i])
+			if (_literals[i]) {
+				String__String__unlock(_literals[i]);
```
davidpiuva: I don't know how to access the variable directly with mangled names so I made a setter method.

davidpiuva:
When I added debug printing, I got crashed instead of leaks when it suddenly started the test by freeing all literals globally.

marcusnaslund:
@davidpiuva Your code example should leak, since you're not freeing `result`.

davidpiuva:
That makes sense

davidpiuva:
I probably got a crash from debug printing because the message itself to print had already been freed when reaching the point to debug.

davidpiuva:

source/system/string/string_literal.c
sebastianbaginski: Don't bother, this is ok.

sebastianbaginski:
Nicely done.

test/system/StringTest.ooc
```diff
@@ -11,6 +11,15 @@ use unit
 StringTest: class extends Fixture {
 	init: func {
 		super("String")
+		this add("literal", func {
+			// The & operand tries to free the literals but if the lock works then they should resist the attempt
+			result: String
+			for (i in 1 .. 100) {
+				result = "ab" & "cd"
```
marcusnaslund: For even better effect you could declare `ab = "ab"` and `cd = "cd"` outside the loop and use them.

source/system/String.ooc
```diff
@@ -16,15 +16,18 @@ String: class {
 	// Avoid direct buffer access, as it breaks immutability.
 	_buffer: CharBuffer
 	size ::= this _buffer size
-
+	_locked := false // Only global termination may unlock and free literals
+	_unlock: func { this _locked = false }
```
marcusnaslund: No methods above `init`, put it somewhere else. Right below `free` is OK.

test/system/StringTest.ooc
marcusnaslund: Or both.

marcusnaslund:
That's all I've got. Well done.

test/system/StringTest.ooc
```diff
@@ -11,6 +11,22 @@ use unit
...
+				expect(result, is equal to("abcd"))
+				result free()
+			}
+			ab := "ab"
```
davidpiuva: Updated

PR #1778, opened 2016-06-08
Now using correct target in FBO.

emilwestergren:
Can be other than `GL_TEXTURE_2D`

@davidpiuva peer review


davidpiuva:
Looks good

PR #1781, opened 2016-06-08
Now locking with correct usage when reading/writing to GraphicBufferYuv420Semiplanar

emilwestergren:
Fixes flickering bug when using `toRaster` with a GraphicBufferYuv420Semiplanar as target with unoptimized OpenGLContext.
@erikhagglund peer review


erikhagglund:
very nice!

PR #1782, opened 2016-06-09
Remove multithread context code

emilwestergren:
Since we don't support use of `OpenGLContext` from different threads we might as well remove the code trying to achieve this.

@davidpiuva peer review


davidpiuva:
Looks safe.

PR #1783, opened 2016-06-09
Added optional freeing argument for String methods

### LINKED ISSUE #1763: Automatic free of strings
#1760 introduces very ugly code in `DateTime toString` and `TimeSpan toString` because of repeated `replaceAll` calls, where the intermediate string must be freed after each call.

We should implement some kind of automatic freeing for Strings. Two alternatives exist:

1) Besides `replaceAll`, add a `replaceAllAndFreeOriginal` (but with better name) which calls the former and then frees the original before returning.
2) Add a `Bool` argument to relevant functions, defaulted to false, which if true frees the current object.

This would also be nice for `print` and `println`, for example.

marcusnaslund:
Fixes #1763 

- Added a `freeOriginal` argument to relevant `String` methods
- Adapted `DateTime` and `TimeSpan` code to match
- Removed leaks in `String append` and `String prepend`


marcusnaslund:

marcusnaslund:
`./test.sh system/String -x && ./tools/memcheck.sh` yields 0 leaks.

source/base/TimeSpan.ooc
```diff
@@ -42,28 +42,19 @@ TimeSpan: cover {
...
-		second free()
-		second = result replaceAll("%m", minuteModString)
-		result free()
-		result = second replaceAll("%s", secondModString)
-		second free()
-		second = result replaceAll("%z", milliModString)
-		(result, weekString, dayString, hourString, minuteString, secondString, milliString, dayModString, hourModString, minuteModString, secondModString, milliModString) free()
-		second
+		result := format replaceAll("%w", weekString)
+		result = result replaceAll("%D", dayString, true, true)
```
sebastianbaginski: I'd prefer to see something like
```ooc
result = result replaceAll("%D", dayString, String CaseSensitive, String FreeArgument)
```
or sth. but we can start with `true, true` ;)

marcusnaslund: I agree `true, true` is bad. Let's consider it for the future.

sebastianbaginski:
true true

PR #1784, opened 2016-06-09
Implemented a basic SortedVectorList

marcusnaslund:
Calls `sort()` on *every* `add` and `append` which is very inefficient, but a smarter sorting function will have to wait until `primaryFeature1.4` has been merged.

### DIFF
--- a/source/collections/SortedVectorList.ooc
+++ b/source/collections/SortedVectorList.ooc
@@ -1,0 +1,28 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use collections
+
+SortedVectorList: class <T> extends VectorList<T> {
+	_defaultComparator: Func (T*, T*) -> Bool
+
+	init: func ~default (=_defaultComparator) { super() }
+	init: func ~heap (capacity: Int, =_defaultComparator, freeContent := true) { super(capacity, freeContent) }
+
+	add: override func (item: T) {
+		super(item)
+		this sort(this _defaultComparator)
+	}
+	append: override func (other: List<T>) {
+		super(other)
+		this sort(this _defaultComparator)
+	}
+
+	operator [] <T> (index: Int) -> T { this as VectorList<T> _vector[index] }
+	operator []= (index: Int, item: T) { this _vector[index] = item }
+}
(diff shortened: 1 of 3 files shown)


marcusnaslund:
@sebastianbaginski OK with reviewing this now (it's still efficient enough for the small amounts of data we thought to use it for) and then we can optimize it later when `primaryFeature1.4` is merged?

source/collections/SortedVectorList.ooc
```diff
@@ -0,0 +1,29 @@
...
+		super(item)
+		this sort(this _defaultComparator)
+	}
+	append: final override func (other: List<T>) {
+		super(other)
+		this sort(this _defaultComparator)
+	}
+
+	operator [] <T> (index: Int) -> T { this as VectorList<T> _vector[index] }
+	operator []= (index: Int, item: Int) { this _vector[index] = item }
```
sebastianbaginski: `item: Int` ?

source/collections/SortedVectorList.ooc
```diff
@@ -0,0 +1,29 @@
...
+use collections
+
+SortedVectorList: class extends VectorList<Int> {
+	_defaultComparator: Func (T*, T*) -> Bool
+
+	init: func ~default (=_defaultComparator) { super() }
+	init: func ~heap (capacity: Int, =_defaultComparator, freeContent := true) { super(capacity, freeContent) }
+	copy: func -> This { super() as This }
+
+	add: final override func (item: T) {
```
sebastianbaginski: Why are these methods `final` ?

sebastianbaginski:
OK let's have this for now. A test with `vector<class>` could be nice too.

source/collections/SortedVectorList.ooc
marcusnaslund: Oops.

source/collections/SortedVectorList.ooc
marcusnaslund: Because rock.

marcusnaslund:
> A test with `vector<class>`

comin' up

source/collections/SortedVectorList.ooc
sebastianbaginski: :confounded:

source/collections/SortedVectorList.ooc
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use collections
+
+SortedVectorList: class extends VectorList<Int> {
```
marcusnaslund: Gaaaaaaaaaaaaaaaaaaaaaaaaaaaah I had all sorts of weird problems until I discovered `extends VectorList<Int>`

marcusnaslund:
Updated with a test using a class too.

PR #1785, opened 2016-06-09
Updated VectorList sort() to match improved sort

marcusnaslund:
`lowerBound`, `upperBound` and `search` all take an `isLess` comparator, whereas the `sort` method takes a `greaterThan` comparator. The former, I believe, is a lot more common and so we should do it here as well.

### DIFF
--- a/source/system/structs/VectorList.ooc
+++ b/source/system/structs/VectorList.ooc
@@ -123,7 +123,7 @@ VectorList: class <T> extends List<T>{
 		while (!inOrder) {
 			inOrder = true
 			for (i in 0 .. this count - 1)
-				if (!isLess(this[i], this[i + 1])) {
+				if (isLess(this[i + 1], this[i])) {
 					inOrder = false
 					tmp := this[i]
 					this[i] = this[i + 1]
--- a/source/system/structs/VectorList.ooc
+++ b/source/system/structs/VectorList.ooc
@@ -118,12 +118,12 @@ VectorList: class <T> extends List<T>{
 			}
 		result
 	}
-	sort: override func (greaterThan: Func (T, T) -> Bool) {
+	sort: override func (isLess: Func (T, T) -> Bool) {
 		inOrder := false
 		while (!inOrder) {
 			inOrder = true
 			for (i in 0 .. this count - 1)
-				if (greaterThan(this[i], this[i + 1])) {
+				if (!isLess(this[i], this[i + 1])) {
 					inOrder = false
 					tmp := this[i]
 					this[i] = this[i + 1]
--- a/test/system/VectorListTest.ooc
+++ b/test/system/VectorListTest.ooc
@@ -200,7 +200,7 @@ VectorListTest: class extends Fixture {
 		})
 		this add("VectorList sort", func {
 			list := VectorList<Int> new()
-			list add(8) . add(16) . add(32)
+			list add(32) . add(16) . add(8)
 
 			sortedList := list copy()
 			sortedList sort(|v1, v2| v1 < v2)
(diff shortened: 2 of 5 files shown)


marcusnaslund:

source/system/structs/VectorList.ooc
```diff
@@ -118,12 +118,12 @@ VectorList: class <T> extends List<T>{
...
 		result
 	}
-	sort: override func (greaterThan: Func (T, T) -> Bool) {
+	sort: override func (isLess: Func (T, T) -> Bool) {
 		inOrder := false
 		while (!inOrder) {
 			inOrder = true
 			for (i in 0 .. this count - 1)
-				if (greaterThan(this[i], this[i + 1])) {
+				if (!isLess(this[i], this[i + 1])) {
```
sebastianbaginski: I'm not sure it is better in general, as it will shuffle equal elements around.

marcusnaslund: `isLess(this[i + 1], this[i])`?

sebastianbaginski:
```
vector<int> = {1, 1, 1, 1, 1};

! isLess(vector[i], vector[j]) == true {
    swap(i, j)
}
```

test/system/VectorListTest.ooc
```diff
@@ -200,7 +200,7 @@ VectorListTest: class extends Fixture {
 		})
 		this add("VectorList sort", func {
 			list := VectorList<Int> new()
-			list add(8) . add(16) . add(32)
+			list add(32) . add(16) . add(8)
 
 			sortedList := list copy()
 			sortedList sort(|v1, v2| v1 < v2)
```
marcusnaslund: It's interesting of course that the test already sent an `isLess` comparator @sebastianbaginski

sebastianbaginski: Nice :)

source/system/structs/VectorList.ooc
marcusnaslund: what?

sebastianbaginski: 1. `vector = {1, 1, 1, 1, 1}`
2. Call `vector sort()`
3. enter `for` loop, this[0] == 1, this[1] == 1
4. check condition  `!(isLess(1, 1))` == `!(1 < 1)` == `!(false)` == `true`
5. condition evaluates to true, swap elements `this[0] <-> this[1]`

marcusnaslund: Right.
But switching the location of the two arguments would fix it, no?

sebastianbaginski: And removing the negation :)

sebastianbaginski: Oh so there is no exclamation mark in your first response. Looks like I need to get some more sleep.

marcusnaslund: :sleeping:

source/system/structs/VectorList.ooc
```diff
@@ -118,12 +118,12 @@ VectorList: class <T> extends List<T>{
...
 		result
 	}
-	sort: override func (greaterThan: Func (T, T) -> Bool) {
+	sort: override func (isLess: Func (T, T) -> Bool) {
 		inOrder := false
 		while (!inOrder) {
 			inOrder = true
 			for (i in 0 .. this count - 1)
-				if (greaterThan(this[i], this[i + 1])) {
+				if (isLess(this[i + 1], this[i])) {
```
marcusnaslund: Updated.

PR #1788, opened 2016-06-09
Replaced Text with String in ArgumentParser and Csv* classes

marcusnaslund:
(Major) part of #1724 - removing use of `Text`.


marcusnaslund:

source/cli/ArgumentParser.ooc
```diff
@@ -48,68 +49,68 @@ ArgumentParser: class {
...
-				current = current slice(1)
-				for (j in 0 .. current count)
-					arguments add(current slice(j, 1))
-			}
+			current := input[i]
+			if (current startsWith("--"))
+				arguments add(current substring(2))
+			else if (current startsWith("-"))
+				for (j in 1 .. current size)
+					arguments add(current[j] as String)
```
sebastianbaginski: Why does it work ? Bracket operator in String class returns a Char : `operator [] (string: String, index: Int) -> Char { string _buffer [index] }` . Confusing :/

sebastianbaginski: Ah ok, Char has `as String` operator. Nevermind.

marcusnaslund: Because you can overload the `as` operator. So `current[j] as String` actually creates a new `String` of length 1 with `current[j]` as the only character.

source/io/CsvReader.ooc
```diff
@@ -59,38 +62,35 @@ CsvReader: class extends Iterator<VectorList<Text>> {
...
-						textBuilder append(readCharacter)
+						stringBuilder add(readCharacter)
 				}
 			}
-			result add(textBuilder toText())
-			textBuilder free()
+			result add(stringBuilder toString())
+			for (i in 0 .. stringBuilder count)
+				stringBuilder[i] free()
+			stringBuilder free()
```
sebastianbaginski: `stringBuilder freeAll()` ?

source/io/CsvReader.ooc
```diff
@@ -59,38 +62,35 @@ CsvReader: class extends Iterator<VectorList<Text>> {
...
+			stringBuilder add(rowData[position])
 			position += 1
 		}
 		position += 1
-		result = textBuilder toText()
-		textBuilder free()
+		result = stringBuilder toString()
+		for (i in 0 .. stringBuilder count)
+			stringBuilder[i] free()
+		stringBuilder free()
```
sebastianbaginski: As above.

source/io/CsvReader.ooc
marcusnaslund: Or, the `StringBuilder` constructor should take an argument of whether it should free the content or not. That will be another PR though.

sebastianbaginski:
Ok then.

marcusnaslund:
Thanks. Noted that this is green, taking it off until corresponding PR is ready.

source/io/CsvReader.ooc
marcusnaslund: It was way too easy to fix ( #1825 ) so I will be incorporating it here anyway.

marcusnaslund:
`./test.sh && ./tools/memcheck.sh` shows 0 new leaks, yay.

fredrikbryntesson:
OK

PR #1790, opened 2016-06-10
Added ctor for nanoseconds

thomasfanell:
@sebastianbaginski review?

### DIFF
--- a/source/base/DateTime.ooc
+++ b/source/base/DateTime.ooc
@@ -41,6 +41,7 @@ DateTime: cover {
 	_ticks: ULong = 0
 	ticks ::= this _ticks
 	init: func@ (=_ticks)
+	init: func@ ~fromNanoseconds(nanoseconds: ULong) { this _ticks = nanoseconds / This nanosecondsPerTick }
 	init: func@ ~fromYearMonthDay (year, month, day: Int) {
 		if (This dateIsValid(year, month, day))
 			this _ticks = This dateToTicks(year, month, day)
--- a/source/base/DateTime.ooc
+++ b/source/base/DateTime.ooc
@@ -41,7 +41,6 @@ DateTime: cover {
 	_ticks: ULong = 0
 	ticks ::= this _ticks
 	init: func@ (=_ticks)
-	init: func@ ~fromNanoseconds (nanoseconds: ULong) { this _ticks = nanoseconds / This nanosecondsPerTick }
 	init: func@ ~fromYearMonthDay (year, month, day: Int) {
 		if (This dateIsValid(year, month, day))
 			this _ticks = This dateToTicks(year, month, day)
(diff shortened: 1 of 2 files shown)


sebastianbaginski:
Looks good.

marcusnaslund:
PRInt fails.

marcusnaslund:
So first you complain about all the abbreviations in `rust`, and then you go ahead an write `ctor` instead of `constructor`... :)

thomasfanell:
@marcusnaslund stop lying...

source/base/DateTime.ooc
```diff
@@ -41,6 +41,7 @@ DateTime: cover {
 	_ticks: ULong = 0
 	ticks ::= this _ticks
 	init: func@ (=_ticks)
+	init: func@ ~fromNanoseconds (nanoseconds: ULong) { this _ticks = nanoseconds / This nanosecondsPerTick }
```
emilwestergren: This might be a bit unclear when you use it:
```ooc
time := DateTime new(1000)
```
Isn't it better to have a static function for it?
```ooc
time := DateTime fromNanoseconds(1000)
```

marcusnaslund: Aren't we removing `ticks` all together?

emilwestergren: Not really, the goal is to abstract what we use internally. It should be easy to change the definition of the backend counter without breaking the code.

marcusnaslund: But why stick to ticks internally? Do the extra two zeros from reading nanoseconds instead of ticks confuse anyone?

thomasfanell: @emilwestergren yeah, I thought about that... it might be better after all to have a factory method instead

@marcusnaslund not right now

emilwestergren: I mean `ticks` will be whatever we want it to be, but we can rename it to something better so people don't mistake it for .NET ticks.

marcusnaslund: Alright then.

source/base/TimeSpan.ooc
```diff
@@ -111,6 +109,7 @@ TimeSpan: cover {
 	hours: static func (count: Double) -> This { This new(DateTime ticksPerHour * count) }
 	days: static func (count: Double) -> This { This new(DateTime ticksPerDay * count) }
 	weeks: static func (count: Double) -> This { This new(DateTime ticksPerWeek * count) }
+	fromNanoseconds: static func (nanoseconds: Long) -> This { This new(nanoseconds / DateTime nanosecondsPerTick) }
```
emilwestergren: This is inconsistent with all the other create functions `milliseconds/seconds/minutes` etc

thomasfanell: Didn't even notice those, since their names suggests they are getters. Oh well, I'll fix and rename them.

marcusnaslund: Because it's used differently and takes a `Long` as argument.

thomasfanell: Anyway, I've renamed them to more appropriate names

thomasfanell: or perhaps I should just rename the nanoseconds one, since it would be a breaking change... *sigh*

emilwestergren: I think it's better to go with `nanoseconds` and change all of them later in a separate PR if we want to.

PR #1791, opened 2016-06-10
Simplified code for thread synchronization

emilwestergren:
@marcusnaslund what do you think of this?

### DIFF
--- a/source/concurrent/ThreadPool.ooc
+++ b/source/concurrent/ThreadPool.ooc
@@ -106,18 +106,7 @@ _TaskPromise: class extends Promise {
 		super()
 	}
 	wait: override func -> Bool { this _task wait() }
-	wait: override func ~timeout (time: TimeSpan) -> Bool {
-		timer := CpuTimer new() . start()
-		seconds := time toSeconds()
-		status := false
-		while (timer stop() / 1000.0 < seconds && !status) {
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Much better. Except the rock segfaulting.

emilwestergren:
Just add it to my PR debt

emilwestergren:
Let's begin with this.

source/concurrent/ThreadPool.ooc
```diff
@@ -37,6 +37,18 @@ _Task: abstract class {
...
 		this _state == _PromiseState Finished
 	}
+	wait: func ~timeout (time: TimeSpan) -> Bool {
+		timer := CpuTimer new() . start()
+		seconds := time toSeconds()
+		status := false
+		while (timer stop() / 1000.0 < seconds && !status) {
+			status = (this _state != _PromiseState Unfinished)
+			if (!status)
+				Time sleepMilli(seconds / 10 as Int)
```
marcusnaslund: `sleepMilli` with seconds argument?
Oh, it was like that before, but, why like this?

emilwestergren: No idea, I'm changing this in a later commit.

marcusnaslund:
Looks good.

PR #1793, opened 2016-06-13
Fixed naming inconsistency

thomasfanell:
Small enough to go directly to @marcusnaslund ?


thomasfanell:
@marcusnaslund hang on a sec, gotta change in  `DateTime` too, otherwise we end up with `DateTime fromNanoseconds` and `TimeSpan nanoseconds`.

thomasfanell:
There we go

PR #1794, opened 2016-06-13
Added removeRange function for VectorList.

emilwestergren:


source/system/structs/VectorList.ooc
```diff
@@ -51,9 +51,11 @@ VectorList: class <T> extends List<T>{
...
 		result
 	}
-	removeAt: override func (index: Int) {
-		this _vector copy(index + 1, index)
-		this _count -= 1
+	removeAt: override func (index: Int) { this removeRange(index, index + 1) }
+	removeRange: func (start, stop: Int) {
+		count := stop - start
+		this _vector copy(stop, start)
+		this _count -= count
```
sebastianbaginski: Maybe we should have a sanity check (`safe` block) ?

source/system/structs/VectorList.ooc
```diff
@@ -51,9 +51,11 @@ VectorList: class <T> extends List<T>{
 		this _count -= 1
 		result
 	}
-	removeAt: override func (index: Int) {
-		this _vector copy(index + 1, index)
-		this _count -= 1
+	removeAt: override func (index: Int) { this removeRange(index, index + 1) }
+	removeRange: func (start, stop: Int) {
```
sebastianbaginski: Most of the functions in this class takes `start, end: Int` but that does not matter for me (but maybe it does for @marcusnaslund :) )

sebastianbaginski:
Magnificient.

source/system/structs/VectorList.ooc
emilwestergren: My bad

source/system/structs/VectorList.ooc
emilwestergren: Sure

source/system/structs/VectorList.ooc
marcusnaslund: Good catch.

marcusnaslund: I think there should be a `removeRange: func ~range (range: Range)` overload as well.

source/system/structs/VectorList.ooc
```diff
@@ -51,9 +51,11 @@ VectorList: class <T> extends List<T>{
 		this _count -= 1
 		result
 	}
-	removeAt: override func (index: Int) {
-		this _vector copy(index + 1, index)
-		this _count -= 1
+	removeAt: override func (index: Int) { this removeRange(index, index + 1) }
```
marcusnaslund: No surrounding function is a one-liner, so this should according to the style guide not be a one-liner either.

source/system/structs/VectorList.ooc
emilwestergren: Yeah I was thinking of that, does that work? `list removeRange(0 .. 3)` ?

marcusnaslund: That would work then, yes.

emilwestergren: Also, do you prefer `removeAt ~range (range: Range)` ?

marcusnaslund: Now that you mention it, I do.

source/system/structs/VectorList.ooc
emilwestergren: What do you mean? I don't understand the rule.

emilwestergren:
Updated

emilwestergren:
Maybe these could be moved to the base class `List` as well, but I started out with only `VectorList`.

source/system/structs/VectorList.ooc
marcusnaslund: The point is that one-liners are OK if there's more than one of them together. Switching between the two styles too often impacts readability.

This is ok
```ooc
a: func { lala }
b: func { lala }
c: func {
    lolo
    lolo
}
d: func {
    lolo
    lolo
}
```
This is not
```ooc
a: func { lala }
b: func {
    lolo
    lolo
}
c: func { lala }
d: func {
    lolo
    lolo
}
```
We do break this rule a lot though, especially if, for example, `a` and `b` have the same name and `a` only calls `b` with some default arguments.

marcusnaslund: And now, one-liners are OK for both reasons. So ignore this.

source/system/structs/VectorList.ooc
```diff
@@ -51,9 +51,16 @@ VectorList: class <T> extends List<T>{
...
-	removeAt: override func (index: Int) {
-		this _vector copy(index + 1, index)
-		this _count -= 1
+	removeAt: override func (index: Int) { this removeAt(index, index + 1) }
+	removeAt: func ~range (range: Range) { this removeAt(range min, range max) }
+	removeAt: func ~indices (start, end: Int) {
+		count := end - start
+		version(safe) {
+			raise(count < 0 || start > this count || end > this count,
+				"Invalid range in VectorList::removeAt: start=%d end=%d (list count=%d)" format(start, end, this count))
```
marcusnaslund: This is something in ooc that makes no sense. You can declare a string somewhere, but not assign it to anything nor do anything with it. It just lives on forever.

(It should be `printfln`, not `format`)

source/system/structs/VectorList.ooc
```diff
@@ -51,9 +51,16 @@ VectorList: class <T> extends List<T>{
...
 		result
 	}
-	removeAt: override func (index: Int) {
-		this _vector copy(index + 1, index)
-		this _count -= 1
+	removeAt: override func (index: Int) { this removeAt(index, index + 1) }
+	removeAt: func ~range (range: Range) { this removeAt(range min, range max) }
+	removeAt: func ~indices (start, end: Int) {
+		count := end - start
+		version(safe) {
```
marcusnaslund: These braces aren't needed.

marcusnaslund:
> these could be moved to the base class `List` as well

Agreed. Not a pressing issue at the moment though, but created an issue for later: #1795

source/system/structs/VectorList.ooc
sebastianbaginski: It is not a complete nonsense, as string constructor could have side effects :smiling_imp:  (#evil #badstyle #sloppycode #microsoft)

emilwestergren: I'm not following you here, how do you mean I should raise?

marcusnaslund: The raise is fine, but `"Invalid range in VectorList::removeAt: start=%d end=%d (list count=%d)" format(start, end, this count))` is only creating a string, not printing it.
Replace `format` with `printfln`.

emilwestergren: It's a parameter to `raise`

marcusnaslund: Aaaaaaaaah nevermind.

marcusnaslund: This will leak memory, though, because in `-Dsafe` the string will be created regardless of whether the argument is true or not.

Better perhaps to use `if (count < 0 || ...) raise("Invalid range..." format(...))` which will only create the string if we're raising, at which point leaks don't matter.

emilwestergren: So that makes the `raise(condition, string)` function pretty useless?

marcusnaslund: No, `raise(condition, "I am a string literal")` still works fine, like
```ooc
version (safe)
			raise(x >= this width || x < 0 || y >= this height || y < 0, "Accessing Vector2D index out of range in get operator")
```

emilwestergren: Ok so it's the `format` that causes the leak?

emilwestergren:
Updated

source/system/structs/VectorList.ooc
marcusnaslund: Yes, because it creates a new String (which is not a literal).

PR #1798, opened 2016-06-16
Not for merge. Just testing.

niklasborwell:


niklasborwell:
Sorted vector list seems to fail for me. Is there an issue with this class?

marcusnaslund:
Not that I know of.

marcusnaslund:
It has never failed on `develop`...

niklasborwell:
@emilwestergren You've done some changes in VectorList in 1.4. See anything suspicious?

test/collections/SortedVectorListTest.ooc
```diff
@@ -0,0 +1,74 @@
...
+use unit
+
+SortedVectorListClassTester: class {
+	_value: Float
+	value ::= this _value
+	init: func (=_value)
+}
+
+sortedVectorListIntComparator := func (a, b: Int*) -> Bool { a@ > b@ }
+sortedVectorListClassComparator := func (a, b: SortedVectorListClassTester*) -> Bool { a@ value > b@ value }
```
marcusnaslund: @niklasborwell The problem is here. Change both `>` to `<`, and the test goes through.
The reason is that, in `primaryFeature1.4`, `VectorList sort()` has changed its argument from expecting a `greaterThan` comparator to an `isLess` comparator, in order to be consistent with other sorting-related methods.

marcusnaslund:
All kean tests passed now.

PR #1800, opened 2016-06-16
Standardized Point and Vector conversions

### LINKED ISSUE #1799: Standardize Vector and Point conversions
For example, an `IntPoint` cannot be converted to `FloatVector` without going through `IntVector` or `FloatPoint` first.

marcusnaslund:
Fixes #1799 

I would have preferred to use overloads of the `as` operator more, but this is how we're doing it and changing it everywhere isn't feasible.


marcusnaslund:

PR #1801, opened 2016-06-17
Added missing case in CPU line drawing algorithm

davidpiuva:
Added the case for when Y changes more than X for each drawn pixel.

### DIFF
--- a/source/draw/RasterImage.ooc
+++ b/source/draw/RasterImage.ooc
@@ -64,8 +64,7 @@ RasterImage: abstract class extends Image {
 			for (y in startY .. endY + 1)
 				this _drawPoint(start x, y, pen)
 		} else {
-			originalPen := pen
-			originalAlpha := originalPen alphaAsFloat
+			alpha := pen alphaAsFloat
 			if (Int abs(end x - start x) < Int abs(end y - start y)) {
 				slope := (end x - start x) as Float / (end y - start y) as Float
 				startY := start y minimum(end y)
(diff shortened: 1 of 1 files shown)


davidpiuva:
@emilwestergren Peer review?

source/draw/RasterImage.ooc
```diff
@@ -66,17 +66,32 @@ RasterImage: abstract class extends Image {
...
+				endX := start x maximum(end x)
+				for (x in startX .. endX + 1) {
+					idealY := slope * (x - start x) + start y
+					floor := idealY floor()
+					weight := (idealY - floor) abs()
+					pen setAlpha(originalAlpha * (1.0f - weight))
+					this _drawPoint(x, floor, pen)
+					pen setAlpha(originalAlpha * weight)
+					this _drawPoint(x, floor + 1, pen)
+				}
```
marcusnaslund: There's a lot of reused code here, both cases are pretty identical but switch x and y, as far as I can see. The same goes for the first two cases. Maybe we can clean that up.

davidpiuva: Anything I can think of would lead to performance loss.

marcusnaslund:
Frozen awaiting `primaryFeature1.4` merge.

source/draw/RasterImage.ooc
```diff
@@ -66,17 +66,32 @@ RasterImage: abstract class extends Image {
 		} else {
 			originalPen := pen
```
marcusnaslund: Is this line some residual of how `Pen` used to work? Looks like we can remove it.

davidpiuva:
Updated

PR #1806, opened 2016-06-22
Added function for checking if value is inside a certain range.

emilwestergren:
@marcusnaslund peer review

### DIFF
--- a/source/base/TimeSpan.ooc
+++ b/source/base/TimeSpan.ooc
@@ -21,6 +21,7 @@ TimeSpan: cover {
 	toHours: func -> Long { this ticks / DateTime ticksPerHour }
 	toDays: func -> Long { this ticks / DateTime ticksPerDay }
 	toWeeks: func -> Long { this ticks / DateTime ticksPerWeek }
+	inRange: func (start, end: This) -> Bool { this >= start && this <= end }
 
 	defaultFormat: static String = "%w weeks, %d days, %h hours, %m minutes, %s seconds, %z milliseconds"
 	// Supported formatting expressions:
--- a/test/base/TimeSpanTest.ooc
+++ b/test/base/TimeSpanTest.ooc
@@ -128,6 +128,13 @@ TimeSpanTest: class extends Fixture {
 			expect(Bstring == "0 0 text 0 60")
 			(defaultString, dayString, Astring, Bstring) free()
 		})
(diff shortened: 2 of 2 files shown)


marcusnaslund:
It's really not clear that this method should expect milliseconds. Another idea might be two other `TimeSpan` instances.

And if it should be numerical values, maybe of type `Long`, since that's what the `to...` methods return.

emilwestergren:
Updated

source/base/TimeSpan.ooc
```diff
@@ -21,6 +21,7 @@ TimeSpan: cover {
 	toHours: func -> Long { this ticks / DateTime ticksPerHour }
 	toDays: func -> Long { this ticks / DateTime ticksPerDay }
 	toWeeks: func -> Long { this ticks / DateTime ticksPerWeek }
+	inRange: func (start, end: This) -> Bool { this >= start && this <= end }
```
marcusnaslund: Yes, this makes more sense.

test/base/TimeSpanTest.ooc
```diff
@@ -128,6 +128,13 @@ TimeSpanTest: class extends Fixture {
 			expect(Bstring == "0 0 text 0 60")
 			(defaultString, dayString, Astring, Bstring) free()
 		})
+		this add("inRange", func {
+			span := TimeSpan milliseconds(30)
+			expect(!span inRange(TimeSpan milliseconds(1), TimeSpan milliseconds(25)))
+			expect(!span inRange(TimeSpan milliseconds(31), TimeSpan milliseconds(50)))
```
marcusnaslund: `expect(span inRange(TimeSpan milliseconds(31), TimeSpan milliseconds(50)), is false)` and similarly on the line above

emilwestergren:
updated

source/base/TimeSpan.ooc
marcusnaslund: We do have the similar `in` defined on numeric types (`6 in(2 .. 8) == true`) but using a keyword as method name isn't good practice, methinks. So if you're happy with `inRange` I won't bother you about it.

emilwestergren: This is fine. I will use it for sanity checking metadata, and almost all of them are `TimeSpan`

PR #1810, opened 2016-06-26
removeAt ~range is now part of List base class

### LINKED ISSUE #1795: Move removeAt(range) methods to the List base class
Introduced in #1794 and right now they are just additions on `VectorList`.

marcusnaslund:
This means it's now also implemented in the `SynchronizedList` class and by inheritance in the `SortedVectorList`.

Fixes #1795 


source/system/structs/List.ooc
```diff
@@ -16,6 +16,8 @@ List: abstract class <T> {
 	remove: abstract func ~last -> T
 	remove: abstract func ~atIndex (index: Int) -> T
 	removeAt: abstract func (index: Int)
+	removeAt: abstract func ~range (range: Range)
```
emilwestergren: This could have been
```ooc
removeAt: func ~range (range: Range) { this removeAt(range min, range max)) }
removeAt: abstract func ~indices (start, end: Int)
```
that way you don't have to implement it everywhere.

marcusnaslund: Ahh true.

PR #1811, opened 2016-06-27
remove access to private data in ReferenceCounter

sebastianbaginski:
No new functionality, I just wanted to update it so the private data stays private (easier to replace the implementation later).
@marcusnaslund

### DIFF
--- a/source/base/ReferenceCounter.ooc
+++ b/source/base/ReferenceCounter.ooc
@@ -22,6 +22,7 @@ ReferenceCounter: class {
 			}
 		}
 	}
+	count ::= this _count
 	init: func (target: Object, mutexType := MutexType Safe) {
 		this _target = target
 		this _lock = Mutex new(mutexType)
@@ -50,5 +51,6 @@ ReferenceCounter: class {
 	}
 	increase: func { this update(1) }
 	decrease: func { this update(-1) }
+	reset: func { this _count = 0 }
 	toString: func -> String { "Object ID: " << this _target as Pointer toString() >> " Count: " & this _count toString() }
 }
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -99,7 +99,7 @@ _RecyclableByteBuffer: class extends ByteBuffer {
 				version(debugByteBuffer) { Debug print("ByteBuffer bin full; freeing one ByteBuffer") }
 				bin remove(0) _forceFree()
 			}
-			this referenceCount _count = 0
+			this referenceCount reset()
 			bin add(this)
 			This _lock unlock()
 		}
@@ -118,7 +118,7 @@ _RecyclableByteBuffer: class extends ByteBuffer {
 		for (i in 0 .. bin count)
 			if ((bin[i] size) == size) {
 				buffer = bin remove(i)
-				buffer referenceCount _count = 0
+				buffer referenceCount reset()
 				break
 			}
 		This _lock unlock()
--- a/test/base/ByteBufferTest.ooc
+++ b/test/base/ByteBufferTest.ooc
@@ -53,9 +53,9 @@ ByteBufferTest: class extends Fixture {
 			yuv := ByteBuffer new(30000)
 			y := yuv slice(0, 20000)
 			uv := yuv slice(20000, 10000)
-			expect(yuv referenceCount _count, is equal to(2))
+			expect(yuv referenceCount count, is equal to(2))
 			y referenceCount decrease()
-			expect(yuv referenceCount _count, is equal to(1))
+			expect(yuv referenceCount count, is equal to(1))
 			uv referenceCount decrease()
 		})
 		this add("memset", func {


source/base/ReferenceCounter.ooc
```diff
@@ -50,5 +51,6 @@ ReferenceCounter: class {
 	}
 	increase: func { this update(1) }
 	decrease: func { this update(-1) }
+	reset: func { this _count = 0 }
```
marcusnaslund: Do you want to use this on new line 46 as well?

source/base/ReferenceCounter.ooc
```diff
@@ -22,6 +22,7 @@ ReferenceCounter: class {
 			}
 		}
 	}
+	count ::= this _count
```
marcusnaslund: Might be more readable if this is above `isSafe` rather than below.

marcusnaslund:
Accessing private variables is bad practice anyway, so good fix.

source/base/ReferenceCounter.ooc
sebastianbaginski: I'm not sure if this would improve readability, maybe I'll leave it.

source/base/ReferenceCounter.ooc
sebastianbaginski: Isn't `isSafe` also declaring a new data member ?

marcusnaslund: `isSafe` is a property, not a member. A specific `isSafe` value isn't stored anywhere.

source/base/ReferenceCounter.ooc
marcusnaslund: OK.

source/base/ReferenceCounter.ooc
sebastianbaginski: Oh that's true, it is so only for properties with empty get/set like `isSafe: Bool {get set}`

marcusnaslund: Exactly.

sebastianbaginski: But 
```C
struct _ReferenceCounter__ReferenceCounter {
    struct _types__Object __super__;
    types__Object* _target;
    Numbers__Int _count;
    types__Bool _kill;
    Mutex__Mutex* _lock;
    types__Bool isSafe;
};
```
it does declare a data member even now. Quite confusing actually.

marcusnaslund: Thanks rock. :/
Ah well, we can leave it, it doesn't really matter.

sebastianbaginski: I don't understand 
```
void ReferenceCounter__ReferenceCounter___setisSafe___impl(ReferenceCounter__ReferenceCounter* this, types__Bool value) {
    #line 19 "/home/seb/work/ooc-kean/source/base/ReferenceCounter.ooc"
    if (this->isSafe != value)
...
```
but I can't seem to find where `isSafe` is written to...

sebastianbaginski:
```ooc
Test: class {
	value: Int {
		get { return 2 }
		set (x) { 
			if (this value == 2)
				"boo" println()
			else
				"THE VALUE IS: %i" printfln(this value)
		 }
	}
	init: func
}

x := Test new()
x value = 10
```

prints
> THE VALUE IS: 0

sebastianbaginski: And again
```C
Numbers__Int props__Test___getvalue___impl(props__Test* this) {
    #line 3 "/home/seb/work/ooc-kean/props.ooc"
    return 2;
}

void props__Test___setvalue___impl(props__Test* this, Numbers__Int x) {
    #line 5 "/home/seb/work/ooc-kean/props.ooc"
    if (this->value == 2) {
...
```
but the `this->value` data member is never assigned any value.
Looks like it is not safe to use getters within setters ? Can you reproduce ?

marcusnaslund: Same here. But:
```ooc
Test: class {
    value: Int {
        get { return 2 }
        set (x) {
            "value=%i" printfln(this value)
			this value = x
         }
    }
    init: func
}

x := Test new()
x value = 2
x value = 3
x value = 4
```
prints
```
value=0
value=2
value=3
```

sebastianbaginski: Ok, I guess because the value is written to explicitly.
For example, this fails the ReferenceCounter test:
```occ
		counter isSafe = true
		expect(counter isSafe, is equal to(true))
		counter isSafe = false
		expect(counter isSafe, is equal to(false))
		counter isSafe = true
		expect(counter isSafe, is equal to(true))
```
to me it looks like a bug in rock.

sebastianbaginski:
Nice job @marcusnaslund , I think we've just found a serious flaw in properties :+1:

source/base/ReferenceCounter.ooc
marcusnaslund: > this fails the ReferenceCounter test

That's bad. Setters that call their corresponding getter should be banned at once.

marcusnaslund:
> Nice job

I think the credit is all yours, most people would have just said "OK" and moved the `count` property as I suggested and then moved on with their lives. :)

sebastianbaginski:
But, most people doing a PR would have never suggested moving the `count` property above `isSafe`, so teamwork +1.

marcusnaslund:
Alright then. :)

source/base/ReferenceCounter.ooc
sebastianbaginski: Yeah. Also, one could expect
```ooc
value: Int {
        set (x) {
            this value = x
         }
    }

...

t value = 0
```
to loop forever.

marcusnaslund:
We are done with this PR, though? I can merge this?

source/base/ReferenceCounter.ooc
marcusnaslund: https://github.com/ooc-lang/rock/issues/1002

sebastianbaginski:
I guess so. Let's not forget to fill a bug report, or just redirect the ooc staff to this thread.

marcusnaslund:
Someone needs to go through the setters and make sure we fix all setter/getter pairs like this, if any.

source/base/ReferenceCounter.ooc
sebastianbaginski: Great, thanks.

PR #1812, opened 2016-06-27
Added decimals argument for floating point toString()

### LINKED ISSUE #1804: Specify number of decimals in toString() for floating point types
`Float toString()` should take an `Int` argument. The same goes for `Double` and `LDouble`, and for methods using this `FloatVectorList toString()` for example.

marcusnaslund:
Fixes #1804 

Let's you call `toString()` with optional argument of the number of decimals. Useful when printing a `FloatVectorList` with arbitrary precision of each element, which sparked the original idea. Also added this feature for most classes in `math` and `geometry`.

### DIFF
--- a/source/geometry/FloatBox2D.ooc
+++ b/source/geometry/FloatBox2D.ooc
@@ -122,7 +122,7 @@ FloatBox2D: cover {
 		newSize := FloatVector2D mix(this size, other size, weight)
 		This createAround(newCenter, newSize)
 	}
-	toString: func -> String { (this leftTop toString() >> ", ") & this size toString() }
+	toString: func (decimals := 2) -> String { (this leftTop toString(decimals) >> ", ") & this size toString(decimals) }
 	toIntBox2D: func -> IntBox2D { IntBox2D new(this left, this top, this width, this height) }
 
 	operator + (other: This) -> This {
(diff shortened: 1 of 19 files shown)


marcusnaslund:

sebastianbaginski:
Looks ok. A test would be nice.

marcusnaslund:
>  A test would be nice.

It's actually already covered by tests since most `toString` tests expect 2 decimals, setting the default to something else breaks those tests. Should I add something more explicit?

Also, do you think I should add this to the geometry classes as well?

marcusnaslund:
Furthermore, looks like I forgot to push the last edits before moving to review. Take a look now.

sebastianbaginski:
I was thinking about testing something like `1.23456 toString(5) == "1.23456"`.

> Also, do you think I should add this to the geometry classes as well?

Why not.

marcusnaslund:
Very well, taking the flag off for now.

marcusnaslund:
Let's move to peer review again. Note that not all default `decimals` values are equal.

sebastianbaginski:
I agree, it makes sense to have more precision for transforms etc.
I have nothing more to add here.

marcusnaslund:
Ping @patrikmalm

PR #1814, opened 2016-06-28
wrapper covers for atomic bool & int

sebastianbaginski:


source/base/Atomic.ooc
```diff
@@ -0,0 +1,56 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+use system
```
marcusnaslund: Just stopping by because I was curious about your implementation and because I like commenting on stuff.
You don't need to explicitly import `system`.
It's also OK to write the methods below as one-liners.
See ya.

source/system/external/stdatomic.ooc
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+include ./atomic
+
+atomic_int: extern cover
```
shamanas: Huh, first time I've seem `extern cover` use, I knew it was valid syntax but wasn't too sure it would behave as expected.  

Any particular reason this is not `atomic_int: cover from atomic_int`?  
By the way, you could define the extern C functions directly inside your types if you did something like:  

```ooc
// atomic_bool never defined in ooc source before
AtomicBool: cover from atomic_bool {
    init: extern(atomic_init_bool) func@ (Bool)
    set: extern(atomic_store_bool) func@ (Bool)
    // Etc.
    // Operators could also be nice
    operator@ = (b: Bool) -> This {
        set(b)
        this
    }
}
```

Of course, this API is also useful to have defined, so I guess it boils down to stylistic choice.

source/base/Atomic.ooc
sebastianbaginski: Alright, thanks

source/system/external/stdatomic.ooc
sebastianbaginski: Thanks for the comments @shamanas :+1: 

The lack of operators is "by design", because I felt like this interface would be incomplete without the multiplication and division operators (as these operations would not be atomic), so I decided to not have any of them. It's not like people gonna use atomic integers in arithmetic expressions (if they are gonna use them at all), so I think this raw interface is ok for the specific situations where atomics are needed.

> Any particular reason this is not atomic_int: cover from atomic_int?

Good question :) There is no particular reason, we already have such syntax in `FileUnix` and `stdio` wrapper. It feels nice.

> you could define the extern C functions directly inside your types

thanks, this seems like a reasonable thing to do.

shamanas: This is probably the most idiomatic way to write bridges to C, since most libraries are written in a pseudo-OO way.  

Usually you do a cover from the pointer type and just define the extern functions as methods but in this case it makes more sense to have the cover be by value.  

If anyone is interested in some more advanced idiomatic ways to write bridges, I would recommend [oocui](https://github.com/shamanas/oocui), the IUP bindings, which include simple covers from pointer types as well passing Closures and C varargs.

marcusnaslund:
PRInt fails completely, I guess that's because of an old gcc version (we need 4.9+)?
We need to update everyone before going ahead with this. Is there a plan for that? @fredrikbryntesson

sebastianbaginski:
> old gcc version (we need 4.9+)?

Yes, exactly. But there is no rush with this, I guess.

fredrikbryntesson:
No we do not have a plan for that but it can easily be created :) Are there any problems with updating to 4.9? Will it break anything?

sebastianbaginski:
I am running gcc 5.3.1 and everything seems to work just fine. I think @marcusnaslund and @thomasfanell have already tested gcc 5 with our code base some time ago.

marcusnaslund:
I've tested gcc 5.x with Ubuntu 16.04 and it compiled and ran tests fine, but could not get vidhance to launch. Probably an issue with 16.04 rather than with gcc, though.

marcusnaslund:
Moving to gcc 4.9 should be perfectly fine, though. Nothing here applies to us: https://gcc.gnu.org/gcc-4.9/porting_to.html

emilwestergren:
This is great, can you tell me the results of the benchmark?

sebastianbaginski:
Atomic version is ~3x faster (if you run the benchmark from the 4th commit), but I will make some more tests just to be sure.

sebastianbaginski:
Tested again, average improvement seems to be between 2x - 3x when dominating operation is `increase / decrease`. Speed up ratio for `new / free` seems to be ~1.3x.

sebastianbaginski:
@marcusnaslund why no regular tests on PRInt for this PR ?

marcusnaslund:
PRInt will be back to normal in a few minutes. :)

source/draw/gpu/opengl/OpenGLPacked.ooc
```diff
@@ -158,6 +158,9 @@ OpenGLPacked: abstract class extends GpuImage {
 			this _backend upload(raster buffer pointer, raster stride)
 		}
 	}
-	onRecycle: func { this _renderTarget invalidate() }
+	onRecycle: func {
+		this referenceCount reset()
```
sebastianbaginski: This is needed now as the updated reference counter implementation does not allow to call `free` multiple times on the target. Previously `referenceCount decrease()` called `free()` each time the counter went below 0 (so the opengl images were recycled correctly). And it simply crashed for non-recycled classes (see the updated test). Now it's guarded by atomic flag, so in order to `free()` the same image multiple times we need to reset the counter explicitly when the image is recycled.

test/base/ReferenceCounterTest.ooc
```diff
@@ -37,6 +37,17 @@ ReferenceCounterTest: class extends Fixture {
...
 		})
 		this add("multiple threads", This _testMultipleThreads)
+		this add("multiple free", func {
+			isAlive: Bool
+			object := TestObject new(isAlive&)
+			counter := ReferenceCounter new(object)
+			counter increase()
+			counter decrease()
+			counter decrease()
+			counter decrease()
```
sebastianbaginski: This test crashes the old implementation of ref counter.

sebastianbaginski:
@thomasfanell peer review ?
@emilwestergren any comments about the `openglpacked` change ?

emilwestergren:
Looks good

thomasfanell:
Looks good!

marcusnaslund:
Looks good to me too.

I would just add some thoughts, but they're not for this PR:
- You *could* add some operator overloads (including `=` for `set()`) to these covers to better mimic "regular" variables. On the other hand it can also be helpful if they do behave somewhat different from their non-atomic counterparts and I don't know about any potential extra performance overhead.
- It is not possible to set the value of an `AtomicBool` with another `AtomicBool` without explicitly getting the value of the latter first, and likewise for `AtomicInt`.

PR #1818, opened 2016-06-29
fixes some small thread-related issues

sebastianbaginski:
- fixed destroying a locked mutex in `ThreadTest`
- removed race condition on static data generated by rock in `Pointer` class

@marcusnaslund

### DIFF
--- a/source/concurrent/SynchronizedQueue.ooc
+++ b/source/concurrent/SynchronizedQueue.ooc
@@ -63,8 +63,8 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 	}
 	cancel: func { this _waitLock with(|| this _canceled = true) . wakeAll() }
 	wait: func -> T {
-		result: T = null
 		this _waitLock lockWhen(func -> Bool { !this empty || this _canceled })
+		result: T = null
 		if (!this _canceled)
 			result = this _backend dequeue(result)
 		this _waitLock unlock()
--- a/test/concurrent/ThreadTest.ooc
+++ b/test/concurrent/ThreadTest.ooc
@@ -62,6 +62,7 @@ ThreadTest: class extends Fixture {
 		expect(value get(), is less than(expectedValue))
 		thread free()
 		startedCondition free()
+		mutex unlock()
 		mutex free()
 		value free()
 		(job as Closure) free()


source/concurrent/SynchronizedQueue.ooc
```diff
@@ -63,8 +63,8 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 	}
 	cancel: func { this _waitLock with(|| this _canceled = true) . wakeAll() }
 	wait: func -> T {
-		result: T = null
 		this _waitLock lockWhen(func -> Bool { !this empty || this _canceled })
+		result: T = null
```
sebastianbaginski: This innocent looking like can actually modify static data in Pointer class:
```
types__PointerClass *types__Pointer_class(){
    static _Bool __done__ = false;
...
__done__ = true;
...
}
```

marcusnaslund: What does this have to do with the placement of `result`?

marcusnaslund:
Did you try all tests or just this one?

marcusnaslund:
Oops, sorry.

source/concurrent/SynchronizedQueue.ooc
sebastianbaginski: Related to C code it generates.
This line was previously outside the critical section:
```C
Memory__memcpy(result, &(__SynchronizedQueue_unreferencable4), ((types__Class*)types__Pointer_class())->size);
```
and `types__Pointer_class()` modifies static data.

marcusnaslund: *sad*

sebastianbaginski:
> Did you try all tests or just this one?

There are many more issues, but we need to verify which of them are false positives. For now I wanted to fix only the problems I'm absolutely sure of.

marcusnaslund:
Alright.
We have problems with `PromiseTest` sometimes (especially when running memcheck.sh), there's definitely something wrong there.

sebastianbaginski:
> (...) there's definitely something wrong there

That's quite probable:

`ERROR SUMMARY: 117 errors from 12 contexts (suppressed: 312 from 44)`

Issue #1728, opened 2016-05-19
Adjust PromiseTest so that it passes the code coverage test


marcusnaslund:
Currently it often (always on nightly tests) fails the third case.

marcusnaslund:
Seems to be fixed by #1819 , will reopen if the problem pops up again.

PR #1819, opened 2016-06-30
added state mutex to Promise

sebastianbaginski:
Fixes race conditions on `_state` reported by helgrind.
@marcusnaslund

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -17,7 +17,12 @@ _PromiseState: enum {
 
 _Synchronizer: abstract class {
 	_state: _PromiseState
+	_stateMutex := Mutex new()
 	init: func
+	free: override func {
+		this _stateMutex free()
+		super()
+	}
 	wait: abstract func (time: TimeSpan) -> Bool
 	wait: func ~forever -> Bool { this wait(TimeSpan maximumValue) }
 	cancel: virtual func -> Bool { false }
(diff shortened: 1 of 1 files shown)


marcusnaslund:
Let's have @emilwestergren weigh in here too.

sebastianbaginski:
@marcusnaslund looks like PRInt machine has some problems with local settings file

marcusnaslund:
@sebastianbaginski Oops.

marcusnaslund:
Now it works.

sebastianbaginski:
Seems to be stuck for ~1h now : `09:29:22 SomeTest ...`

marcusnaslund:
“He that can have patience can have what he will.”
― Benjamin Franklin

marcusnaslund:
What happened here?

sebastianbaginski:
I deleted the branch locally and pushed the changes upstream.

marcusnaslund:
Less work for me to do. :)

emilwestergren:
You can use `mutex with(|| ...)` if you want to get rid of some lines.

marcusnaslund:
Let's hope this fixes #1728 too.

source/concurrent/Promise.ooc
```diff
@@ -131,19 +156,27 @@ _ThreadFuture: class <T> extends Future<T> {
...
 			if (this _thread wait(time toSeconds()))
 				this _threadAlive = false
-		this _state == _PromiseState Finished
+		this _stateMutex lock()
+		result := this _state == _PromiseState Finished
+		this _stateMutex unlock()
+		result
 	}
 	getResult: override func (defaultValue: T) -> T {
 		result := defaultValue
```
marcusnaslund: Does this line outside the lock also suffer problems?

sebastianbaginski: No, looks like it's just a memcpy with a constant size:
```ooc
Test: class <T> {
    init: func
    fun: func (value: T) {
		x := value
    }
}


void prob__Test_fun(prob__Test* this, uint8_t* value) {
    uint8_t* x = Memory__alloca(this->T->size);
    Memory__memcpy(x, value, this->T->size);
}
```
no "class constructor" calls here.

marcusnaslund: Thanks rock.

PR #1821, opened 2016-06-30
fixed race condition on a static data in Closure

sebastianbaginski:
Similar to https://github.com/magic-lang/ooc-kean/pull/1818
@marcusnaslund

### DIFF
--- a/source/concurrent/SynchronizedQueue.ooc
+++ b/source/concurrent/SynchronizedQueue.ooc
@@ -49,12 +49,16 @@ SynchronizedQueue: class <T> extends Queue<T> {
 BlockedQueue: class <T> extends SynchronizedQueue<T> {
 	_waitLock: WaitLock
 	_canceled := false
+	_lockFunc: Func -> Bool
 	init: func {
 		super()
 		this _waitLock = WaitLock new(this _mutex)
+		this _lockFunc = func -> Bool { !this empty || this _canceled }
+		this _lockFunc = ((this _lockFunc as Closure) take()) as Func -> Bool
 	}
 	free: override func {
 		this _waitLock free()
+		(this _lockFunc as Closure) give() free()
 		super()
 	}
 	enqueue: override func (item: T) {
@@ -63,7 +67,7 @@ BlockedQueue: class <T> extends SynchronizedQueue<T> {
 	}
 	cancel: func { this _waitLock with(|| this _canceled = true) . wakeAll() }
 	wait: func -> T {
-		this _waitLock lockWhen(func -> Bool { !this empty || this _canceled })
+		this _waitLock lockWhen(this _lockFunc)
 		result: T = null
 		if (!this _canceled)
 			result = this _backend dequeue(result)


source/concurrent/SynchronizedQueue.ooc
```diff
@@ -49,12 +49,16 @@ SynchronizedQueue: class <T> extends Queue<T> {
 BlockedQueue: class <T> extends SynchronizedQueue<T> {
 	_waitLock: WaitLock
 	_canceled := false
+	_lockFunc: Func -> Bool
 	init: func {
 		super()
 		this _waitLock = WaitLock new(this _mutex)
+		this _lockFunc = func -> Bool { !this empty || this _canceled }
+		this _lockFunc = ((this _lockFunc as Closure) take()) as Func -> Bool
```
marcusnaslund: Oh my.

sebastianbaginski: Yeah I know. Any other ideas ?

marcusnaslund: No, it's how it must be.

marcusnaslund:
What is the `static` data you mention here?

sebastianbaginski:
`static bool __done__;` in C code generated to initialize the Closure class.

marcusnaslund:
Without knowing much about what rock spits out, where's the "problem"?

sebastianbaginski:
In this line `__SynchronizedQueue_SynchronizedQueue_closure6_ctx* __SynchronizedQueue_ctx7 = Memory__calloc(1, ((types__Class*)__SynchronizedQueue_SynchronizedQueue_closure6_ctx_class()->size)` , it calls `__SynchronizedQueue_SynchronizedQueue_closure6_ctx_class` function which modifies static data.
Before this PR it could be executed simultaneously by two or more threads running the `wait` method.

marcusnaslund:
Is there any way of knowing this from looking at ooc code?

sebastianbaginski:
No, you need to know that there is a static data in the generate C code :(
Every single time you assign a generic like `myValue: T = null`, even a local variable, there is a static data access under the hood. Here is why:
```ooc
Test: class <T> {
	get: func -> T {
		result: T = null
		result
	}
}
```
generated C code for the method `get`:
```C
void gen__Test_get(gen__Test* this, uint8_t* __gen_genericReturn1) {
    uint8_t* result = Memory__alloca(this->T->size);
    types__Pointer __gen_unreferencable2 = NULL;
    Memory__memcpy(result, &(__gen_unreferencable2), ((types__Class*)types__Pointer_class())->size);
    if (__gen_genericReturn1) {
        Memory__memcpy(__gen_genericReturn1, result, this->T->size);
    }
    return;
}
```
and `types__Pointer_class()` in the memcpy call reads and writes to static `__done__` flag.

shamanas:
@marcusnaslund  

rock generates class "constructors" that fill out data information lazily, which is why the static bool variables exist.  

One way around it would be to just generate the data information as static const globals instead of using functions to get them, that would eliminate the need to write to static data.  

I don't have my setup available right now but I'm thinking of something like:  

```c

// Current output:
Class Foo_class() {
    static bool __done__ = false;
    static Class class;
    if (!__done__) {
        class = { .size = 42; .name = some_string_literal_nonsense };
        __done__ = true;    
    }
    return class;
}

// Somewhere in codebase:
ptr = alloc(Foo_class().size);

// Proposed output:
static const Class Foo_class = { .size = 42; .name = some_string_literal_nonsense };

// Somewhere in codebase:
ptr = alloc(Foo_class.size);

```
I will freely admit this is one of the problems I would never have thought of since the generated code is non-obvious.  

In fact, a smarter compiler would know the data layout and size of each type and just generate literals but that would not work for extern structs and would be platform specific.

shamanas:
That would also probably "break" some really hackish stuff (that shouldn't be valid anyways) like:

```ooc
adjustSize: func <T> {
    // Modify the class itself <- EVIL
    T size = 8
}
```

marcusnaslund:
Thanks for the great explanation @shamanas 

Getting rid of incredibly un-obvious errors like these would be fantastic. Glad you're looking into it @sebastianbaginski . I will just ping @thomasfanell and go ahead with this.

shamanas:
@marcusnaslund  

I will be updating the rock backend later today to something of the effect I mentioned, it pains me to see ooc code so convoluted for something that should be simple like the original code.

marcusnaslund:
@shamanas 

Grrrrrrrrrrrrrrrrrreat!

PR #1826, opened 2016-07-05
Correct distance metric in RasterMonochrome

### LINKED ISSUE #1824: Incorrect distance function for RasterMonochrome
It reads

```ooc
result /= (this size x squared + this size y squared as Float sqrt())
```

but should be

```ooc
result /= (this size x squared + this size y squared) as Float sqrt()
```

some tests rely on this incorrect metric, however, and need to be corrected as well.

marcusnaslund:
Fixes #1824 

Because dividing with `sizeX^2 + sqrt(sizeY^2)` gives a smaller number than dividing with `sqrt(sizeX^2 + sizeY^2)`, the distance between the two images is naturally slightly larger.

Also simplified a similar entry in `RasterUv distance()`.

### DIFF
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -115,7 +115,7 @@ RasterMonochrome: class extends RasterPacked {
 						result += distance sqrt()
 					}
 				}
-			result /= (sizeX squared + sizeY squared as Float sqrt())
+			result /= this size length
 		}
 		result
 	}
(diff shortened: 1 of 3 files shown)


sebastianbaginski:
Yes.

marcusnaslund:
Unfortunately fails tests in a dependant repo, probably for the same reason. Will take a look.

marcusnaslund:
By sheer magic the tests will pass now.

PR #1828, opened 2016-07-05
Fixed race condition in RecycleBin.

emilwestergren:
Replaced `SynchronizedList` with `VectorList` + `Mutex` to maintain thread safety.
@erikhagglund peer review

### DIFF
--- a/source/concurrent/RecycleBin.ooc
+++ b/source/concurrent/RecycleBin.ooc
@@ -10,30 +10,40 @@ use concurrent
 
 RecycleBin: class <T> {
 	_free: Func (T)
-	_list: SynchronizedList<T>
+	_list: VectorList<T>
 	_size: Int
 	count ::= this _list count
 	isFull ::= this count == this _size
 	isEmpty ::= this _list empty
-	init: func (=_size, =_free) { this _list = SynchronizedList<T> new(_size) }
+	_mutex := Mutex new()
+	init: func (=_size, =_free) { this _list = VectorList<T> new(_size) }
 	free: override func {
 		this clear()
 		this _list free()
 		(this _free as Closure) free(Owner Receiver)
+		this _mutex free()
 		super()
 	}
 	add: func (object: T) {
+		this _mutex lock()
 		if (this isFull)
 			this _free(this _list remove(0))
 		this _list add(object)
+		this _mutex unlock()
 	}
 	search: func (matches: Func (T) -> Bool) -> T {
-		index := this _list search(matches)
 		result: T = null
-		(index > -1) ? this _list remove(index) : result
+		this _mutex lock()
+		index := this _list search(matches)
+		if (index > -1)
+			result = this _list remove(index)
+		this _mutex unlock()
+		result
 	}
 	clear: func {
+		this _mutex lock()
 		while (!this isEmpty)
 			this _free(this _list remove())
+		this _mutex unlock()
 	}
 }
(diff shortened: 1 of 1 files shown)


source/concurrent/RecycleBin.ooc
```diff
@@ -10,30 +10,38 @@ use concurrent
...
 RecycleBin: class <T> {
 	_free: Func (T)
-	_list: SynchronizedList<T>
+	_list: VectorList<T>
 	_size: Int
 	count ::= this _list count
 	isFull ::= this count == this _size
 	isEmpty ::= this _list empty
-	init: func (=_size, =_free) { this _list = SynchronizedList<T> new(_size) }
+	_mutex := Mutex new()
```
erikhagglund: the count, isFull and isEmpty is not thread safe, change to private?

source/concurrent/RecycleBin.ooc
```diff
@@ -10,30 +10,38 @@ use concurrent
...
+	init: func (=_size, =_free) { this _list = VectorList<T> new(_size) }
 	free: override func {
 		this clear()
 		this _list free()
 		(this _free as Closure) free(Owner Receiver)
+		this _mutex free()
 		super()
 	}
 	add: func (object: T) {
+		this _mutex lock()
```
erikhagglund: this _mutex with(...) here as well?

emilwestergren: rock didn't accept that unfortunately

source/concurrent/RecycleBin.ooc
emilwestergren: sure

emilwestergren: On second thought, this class should be able to use without a mutex, then those properties should be public anyway.

source/concurrent/RecycleBin.ooc
```diff
@@ -10,30 +10,40 @@ use concurrent
...
 		if (this isFull)
 			this _free(this _list remove(0))
 		this _list add(object)
+		this _mutex unlock()
 	}
 	search: func (matches: Func (T) -> Bool) -> T {
-		index := this _list search(matches)
 		result: T = null
-		(index > -1) ? this _list remove(index) : result
+		this _mutex lock()
```
marcusnaslund: @sebastianbaginski Somewhere you changed to put `result: T = null` inside the lock, is that good as a general rule or was it a special case? I don't remember the specifics.

marcusnaslund:
This removes the only use of `SynchronizedList`... do we keep it?

source/concurrent/RecycleBin.ooc
sebastianbaginski: It was here https://github.com/magic-lang/ooc-kean/pull/1821#issuecomment-229643939 
Yes, it does matter for generics, because this line `result: T = null` can trigger the "constructor" of  the `Pointer_Class` due to lazy initialization. This constructor can then modify a static bool flag.

marcusnaslund: Thanks. Looks like it's being fixed, but won't make it to us any time soon I'm afraid. https://github.com/ooc-lang/rock/pull/1004
In the mean time, @emilwestergren - move line 35 below line 36.

emilwestergren:
I vote for keeping it.

emilwestergren:
Updated

PR #1829, opened 2016-07-06
Moved EGLRgba recycling to AndroidContext where it belongs

emilwestergren:
@erikhagglund peer review

### DIFF
--- a/source/draw/gpu/opengl/AndroidContext.ooc
+++ b/source/draw/gpu/opengl/AndroidContext.ooc
@@ -165,6 +165,6 @@ AndroidContext: class extends OpenGLContext {
 		}
 		result
 	}
-	recycle: func (image: EGLRgba) { this _eglBin add(image) }
+	recycleEGLRgba: func (image: EGLRgba) { this _eglBin add(image) }
 }
 }
--- a/source/draw/gpu/opengl/AndroidContext.ooc
+++ b/source/draw/gpu/opengl/AndroidContext.ooc
@@ -20,6 +20,7 @@ AndroidContext: class extends OpenGLContext {
 	_unpackRgbaToUv: OpenGLMap
 	_unpackRgbaToUvPadded: OpenGLMap
 	_packers := RecycleBin<EGLRgba> new(32, func (image: EGLRgba) { image free() })
+	_eglBin := RecycleBin<EGLRgba> new(100, |image| image _recyclable = false; image referenceCount decrease())
 	init: func (other: This = null) {
 		super(other)
 		this _unpackRgbaToMonochrome = OpenGLMap new(slurp("shaders/unpack.vert"), slurp("shaders/unpackRgbaToMonochrome.frag"), this)
@@ -28,6 +29,7 @@ AndroidContext: class extends OpenGLContext {
 	}
 	free: override func {
 		this _backend makeCurrent()
+		this _eglBin clear() . free()
 		(this _unpackRgbaToMonochrome, this _unpackRgbaToUv, this _unpackRgbaToUvPadded, this _packers) free()
 		super()
 	}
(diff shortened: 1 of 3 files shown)


source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -146,10 +148,23 @@ AndroidContext: class extends OpenGLContext {
...
+	createEGLRgba: func (source: GraphicBufferYuv420Semiplanar) -> EGLRgba {
+		result := this _eglBin search(|image| source buffer _handle == image buffer _handle)
+		if (result == null) {
+			padding := source uvOffset - source stride * source size y
+			extraRows := padding align(source stride) / source stride
+			height := source size y + source size y / 2 + extraRows
+			width := source stride / 4
+			rgbaBuffer := source buffer shallowCopy(IntVector2D new(width, height), width, GraphicBufferFormat Rgba8888, GraphicBufferUsage Texture | GraphicBufferUsage RenderTarget)
+			result = EGLRgba new(rgbaBuffer, this)
+		}
```
erikhagglund: What happened to referenceCount increase()?

source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -146,10 +148,23 @@ AndroidContext: class extends OpenGLContext {
...
+		result
+	}
+	recycle: func (image: EGLRgba) { this _eglBin add(image) }
```
erikhagglund: this will override the recycle bin in OpenGLContext, why have both if only one is used?

emilwestergren: No, this is an overloaded `recycle`, only for `EGLRgba`

source/draw/gpu/opengl/AndroidContext.ooc
emilwestergren: Previously we had to make sure the reference counter was always >= 1 so that the `EGLRgba` was never freed. Now we always want it to be freed since we moved the recycling to the `free` function in `EGLRgba`.

source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -146,10 +148,23 @@ AndroidContext: class extends OpenGLContext {
 		map add("startY", startY)
 		DrawState new(target) setMap(map) draw()
 	}
-	preallocate: override func (resolution: IntVector2D) { GraphicBufferYuv420Semiplanar free~all() }
+	preallocate: override func (resolution: IntVector2D) { this _eglBin clear() }
 	preregister: override func (image: Image) {
 		if (image instanceOf(GraphicBufferYuv420Semiplanar))
-			(image as GraphicBufferYuv420Semiplanar) toRgba(this) referenceCount decrease()
+			this createEGLRgba(image as GraphicBufferYuv420Semiplanar) referenceCount decrease()
```
erikhagglund: free()

source/draw/gpu/opengl/AndroidContext.ooc
erikhagglund: Why not reuse the same bin in this case?

emilwestergren: We can't invalidate these buffers anyway, that means OpenGL is free to overwrite the contents. In this case we share the memory with the buffer sent to the video encoder.

emilwestergren: I think it's cleaner to keep these separate since they are restricted to `AndroidContext`.

erikhagglund: So we could override the onRecycle in EGLRgba to do nothing and then reuse the same bin?

emilwestergren: You are right though about the `recycle` function, this doesn't look right.

emilwestergren: That's not gonna work, in `OpenGLContext` we will search for images of type `OpenGLRgba`. `EGLRgba` inherits that class, which means we will recycle `EGLRgba` like a normal `OpenGLRgba`.

emilwestergren:
Updated

PR #1831, opened 2016-07-06
Added function for printing EGL extensions

emilwestergren:
@marcusnaslund peer review


marcusnaslund:
`error: initialization discards ‘const’`

emilwestergren:
I don't get the error

marcusnaslund:
I think the problem is that the EGL function returns a `const char*`, but in ooc you store it as a `CString` which is a cover of `char*`, and thus ignore the `const` declaration.

Unfortunately the `const` keyword is meaningless in ooc, but I know I solved the same problem a while back, right now I don't remember exactly how but I will take a look.

marcusnaslund:
Oh yeah, look up `CONST_CHAR_CAST` in `draw/gpu/opengl/backend/gles3/external`.

emilwestergren:
Updated

marcusnaslund:
Yep, seems to work now.

PR #1832, opened 2016-07-06
Removed more Text usage

marcusnaslund:
- Removed `Text` support in `Numbers` and `HashMap`.
- Adapted `PromiseTest` and `ThreadPoolTest` to return `String` rather than `Text`.
- Removed `test/unit/TextTest.ooc` (there is still a `TextTest` in `test/system`).
- Removed `Text` support from `Fixture` and `Cell`.
- Removed `raise` overloads for `Text`.


marcusnaslund:

marcusnaslund:
*fast*

sebastianbaginski:
Nah, 3 minutes for changes that are mostly deletes, seems reasonable. Besides, we all _trust_ your code.

marcusnaslund:
I included the time it takes to tag you until I get a response. :)

PR #1833, opened 2016-07-06
FileReader takes String, not Text, argument

marcusnaslund:
And adapts other classes to match.

`FileReader` also does not store the `fileName` anymore.


marcusnaslund:

sebastianbaginski:
Are we gonna remove the `t""` extension from rock too ?

marcusnaslund:
I don't think it's worth messing with rock any more than neccessary, but we will certainly ban the use of it.

fredrikbryntesson:
Correct

PR #1835, opened 2016-07-07
Good bye, Text!

### LINKED ISSUE #1724: Remove Text

marcusnaslund:
Removes `Text`, `TextBuffer` and the support for text literals.

Will hold with this for a while.

Fixes #1724


sebastianbaginski:
Is `give() / take()` the next target ? ;)

marcusnaslund:
Well.. :)

emilwestergren:
So this entire project was just a huge waste of time?

sebastianbaginski:
I think it gave birth to the `Owned Buffer` concept. Not really aware of any other benefits (and if we consider `owner` a benefit).

marcusnaslund:
`Text` existed before I started, so I don't really know why it was created. I know Simon really pushed for its use and development, but noone who is here now seems to like it or really even understand its memory model very well. And with `String` working just great, it really brings nothing to the table.

marcusnaslund:
Trying to write `text := t"i am a string literal"` in the code will not cause an insane segfault in rock, it will just say that the `Text` class is missing. Sometimes it seems perfectly sane.

thomasfanell:
@marcusnaslund it was created because `String` used to leak memory, which is no longer the case (thank you, Sebastian)

marcusnaslund:
Peer review and/or any last-minute objections @sebastianbaginski @fredrikbryntesson ?

sebastianbaginski:
Nothing from me :+1:

fredrikbryntesson:
Great :)

PR #1844, opened 2016-07-13
Now raising error when calling OpenGL function outside of context.

emilwestergren:
@erikhagglund peer rev


emilwestergren:
@fredrikbryntesson merge

PR #1845, opened 2016-07-14
support for custom allocator in ByteBuffer

sebastianbaginski:
Adds possibility to allocate memory for byte buffers with something other than default `malloc`.
@thomasfanell or @emilwestergren review

### DIFF
--- a/source/base/Allocator.ooc
+++ b/source/base/Allocator.ooc
@@ -23,3 +23,14 @@ MallocAllocator: class extends AbstractAllocator {
 	allocate: override func (size: SizeT) -> Pointer { malloc(size) }
 	deallocate: override func (pointer: Pointer) { memfree(pointer) }
 }
+
+Allocator: abstract class {
+	_defaultAllocator: static AbstractAllocator = null
+	mallocAllocator := static MallocAllocator new(Owner Sender)
+	defaultAllocator: static func -> AbstractAllocator {
+		This _defaultAllocator ?? mallocAllocator as AbstractAllocator
+	}
+	free: static func ~all {
+		This mallocAllocator free(Owner Sender)
+	}
+}
--- a/source/base/Allocator.ooc
+++ b/source/base/Allocator.ooc
@@ -1,0 +1,18 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+AbstractAllocator: abstract class {
+	allocate: abstract func (size: SizeT) -> Void*
+	deallocate: abstract func (pointer: Void*)
+}
+
+MallocAllocator: class extends AbstractAllocator {
+	init: func
+	allocate: override func (size: SizeT) -> Void* { malloc(size) }
+	deallocate: override func (pointer: Void*) { memfree(pointer) }
+}
--- a/source/base/Allocator.ooc
+++ b/source/base/Allocator.ooc
@@ -14,12 +14,12 @@ AbstractAllocator: abstract class {
 		if (owner == this _owner)
 			this free()
 	}
-	allocate: abstract func (size: SizeT) -> Void*
-	deallocate: abstract func (pointer: Void*)
+	allocate: abstract func (size: SizeT) -> Pointer
+	deallocate: abstract func (pointer: Pointer)
 }
 
 MallocAllocator: class extends AbstractAllocator {
 	init: func (owner := Owner Receiver) { this _owner = owner }
-	allocate: override func (size: SizeT) -> Void* { malloc(size) }
-	deallocate: override func (pointer: Void*) { memfree(pointer) }
+	allocate: override func (size: SizeT) -> Pointer { malloc(size) }
+	deallocate: override func (pointer: Pointer) { memfree(pointer) }
 }
--- a/source/base/Allocator.ooc
+++ b/source/base/Allocator.ooc
@@ -6,13 +6,20 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
+import Owner
+
 AbstractAllocator: abstract class {
+	_owner: Owner
+	free: func ~withCriteria (owner: Owner) {
+		if (owner == this _owner)
+			this free()
+	}
 	allocate: abstract func (size: SizeT) -> Void*
 	deallocate: abstract func (pointer: Void*)
 }
 
 MallocAllocator: class extends AbstractAllocator {
-	init: func
+	init: func (owner := Owner Receiver) { this _owner = owner }
 	allocate: override func (size: SizeT) -> Void* { malloc(size) }
 	deallocate: override func (pointer: Void*) { memfree(pointer) }
 }
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -16,15 +16,15 @@ ByteBuffer: class {
 	_size: Int
 	_referenceCount: ReferenceCounter
 	_ownsMemory: Bool
-	_allocator: AbstractAllocator = null
+	_allocator: AbstractAllocator
 	pointer ::= this _pointer
 	size ::= this _size
 	referenceCount ::= this _referenceCount
 
-	init: func (=_pointer, =_size, ownsMemory := false, allocator := null as AbstractAllocator) {
+	init: func (=_pointer, =_size, ownsMemory := false, allocator := Allocator defaultAllocator()) {
 		this _referenceCount = ReferenceCounter new(this)
 		this _ownsMemory = ownsMemory
-		this _allocator = allocator ?? MallocAllocator new() as AbstractAllocator
+		this _allocator = allocator
 	}
 	free: override func {
 		if (this _referenceCount != null)
@@ -54,11 +54,14 @@ ByteBuffer: class {
 	copyTo: func (other: This, start: Int, destination: Int, length: Int) {
 		memcpy(other pointer + destination, this pointer + start, length)
 	}
-	new: static func ~size (size: Int, allocator: AbstractAllocator = null) -> This { _RecyclableByteBuffer new(size, allocator) }
+	new: static func ~size (size: Int, allocator := Allocator defaultAllocator()) -> This { _RecyclableByteBuffer new(size, allocator) }
 	new: static func ~recover (pointer: Byte*, size: Int, recover: Func (This) -> Bool) -> This {
 		_RecoverableByteBuffer new(pointer, size, recover)
 	}
-	free: static func ~all { _RecyclableByteBuffer _free~all() }
+	free: static func ~all {
+		_RecyclableByteBuffer _free~all()
+		Allocator free~all()
+	}
 }
 
 GlobalCleanup register(|| ByteBuffer free~all(), true)
@@ -90,7 +93,7 @@ _RecoverableByteBuffer: class extends ByteBuffer {
 }
 
 _RecyclableByteBuffer: class extends ByteBuffer {
-	init: func (pointer: Byte*, size: Int, allocator: AbstractAllocator = null) { super(pointer, size, true, allocator) }
+	init: func (pointer: Byte*, size: Int, allocator := Allocator defaultAllocator()) { super(pointer, size, true, allocator) }
 	_forceFree: func {
 		this _size = 0
 		this free()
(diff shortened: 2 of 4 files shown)


emilwestergren:
Should this really be static? What if you want several different allocators?

sebastianbaginski:
How would the api look like if this was per-instance ? `ByteBuffer new(128, |size| myAlloc(size) )` ?
Then how do we know when to return a recycled instance, as they could be allocated with different allocators ?
We can think about it later of course, but for now I'd like to allocate buffers by delegating allocation to some external library that has a pool of preallocated memory aligned on 16 byte boundary.
This is more flexible than hardcoded malloc anyway :)

sebastianbaginski:
Wait, I want a custom `free` too.

emilwestergren:
I guess all allocations would have to go through an allocator instance and not by simply using constructor.

thomasfanell:
Nice one. Are you thinking `_mm_malloc`?

sebastianbaginski:
Alright, so maybe
```ooc
AbstractAllocator: abstract class {
    allocate: abstract func (size: SIzeT) -> Void*
    deallocate: abstract func (pointer: Void*)
}

...

MallocAllocator: class extends AbstractAllocator {
    allocate: override func (size: SizeT) -> Void* { malloc(size) }
    deallocate: override func (pointer: Void*) { memfree(pointer) }
}

....
ByteBuffer: class {
    ...
    allocator := MallocAllocator new()

   init: func ~withAllocator (size: SizeT, allocator: AbstractAllocator)
}
```
should work ?

sebastianbaginski:
@thomasfanell : I need that for some external library, but I have no idea what they are using under the hood.

thomasfanell:
Would it be worth adding something like this then?

```ooc
AlignedMemoryAllocator: class extends AbstractAllocator {
	alignment: SizeT
	init: func (=alignment)
	allocate: override func (size: SizeT) -> Void* { _mm_malloc(size, alignment) }
	deallocate: override func (pointer: Void*) { _mm_free(pointer) }
}
```

sebastianbaginski:
I guess so, there is also `aligned_alloc` in C11 standard, so maybe we can use that after our default gcc version is updated.

emilwestergren:
You don't have to implement the Allocator instance here though, I was just thinking there will probably be times where you have different allocation requirements, and then a static allocator will not be enough.

sebastianbaginski:
Yeah, I agree. I'm almost done with this anyway.

source/base/ByteBuffer.ooc
```diff
@@ -15,21 +16,24 @@ ByteBuffer: class {
 	_size: Int
 	_referenceCount: ReferenceCounter
 	_ownsMemory: Bool
+	_allocator: AbstractAllocator = null
 	pointer ::= this _pointer
 	size ::= this _size
 	referenceCount ::= this _referenceCount
 
-	init: func (=_pointer, =_size, ownsMemory := false) {
+	init: func (=_pointer, =_size, ownsMemory := false, allocator := null as AbstractAllocator) {
```
sebastianbaginski: For some reason `allocator: AbstractAllocator = null` does not compile here, it is somehow affected by the previous `false` default argument:
> base/ByteBuffer.ooc:24:47 error Couldn't unwrap decl `false: AbstractAllocator` , trail = 
\_, Module ByteBuffer
  \_, ClassDecl ByteBuffer 
    \_, ClassDecl ByteBufferClass 
      \_, static ByteBuffer new(=_pointer, =_size, ownsMemory: <unknown type> = <CommaSequence>(false: AbstractAllocator, allocator: AbstractAllocator = null)) -> ByteBuffer
        \_, ownsMemory: <unknown type> = <CommaSequence>(false: AbstractAllocator, allocator: AbstractAllocator = null)
          \_, <CommaSequence>(false: AbstractAllocator, allocator: AbstractAllocator = null)

source/base/ByteBuffer.ooc
```diff
@@ -15,21 +16,24 @@ ByteBuffer: class {
...
 		this _referenceCount = ReferenceCounter new(this)
 		this _ownsMemory = ownsMemory
+		this _allocator = allocator ?? MallocAllocator new() as AbstractAllocator
```
sebastianbaginski: Cast is needed to prevent warning `pointer type mismatch in conditional expression` :/

sebastianbaginski:
Updated. @thomasfanell can you review again ?

source/base/ByteBuffer.ooc
```diff
@@ -15,21 +16,24 @@ ByteBuffer: class {
...
 	}
 	free: override func {
 		if (this _referenceCount != null)
 			this _referenceCount free()
 		this _referenceCount = null
 		if (this _ownsMemory)
-			memfree(this _pointer)
+			this _allocator deallocate(this _pointer)
 		this _pointer = null
+		this _allocator free()
```
emilwestergren: Should the allocator really be owned by the buffer?

sebastianbaginski: I guess that's easiest for now. Allocator could also have a `Owner` flag.

source/base/ByteBuffer.ooc
thomasfanell: Yes, rock does not handle this correctly: https://github.com/ooc-lang/rock/issues/990

source/base/ByteBuffer.ooc
thomasfanell: I think it's quite natural for the buffer to take ownership of the allocator. @emilwestergren can you think of a case were this would be undesirable?

sebastianbaginski:
Added `Owner` flag. Review again ? :)

source/base/ByteBuffer.ooc
emilwestergren: It's not uncommon to have an allocator as a singleton for example. In most cases you will maybe have a few different allocators, and to allocate and free these for every ByteBuffer seems like a waste. We could for example have a few static standard allocators in the ByteBuffer class.

```ooc
buffer := ByteBuffer new(size, Allocator malloc)
```

sebastianbaginski: Yes and a possibility to override the default allocator, as my point here was to allow custom memory allocation for byte buffers in raster image classes. This is currently not possible, but for the sake of code quality I happily sacrifice my own personal short-sighted goals.

test/base/ByteBufferTest.ooc
```diff
@@ -9,6 +9,18 @@
...
 use unit
 
+CustomAllocator: class extends AbstractAllocator {
+	_allocCount := 0
+	init: func
+	allocate: override func (size: SizeT) -> Void* {
+		this _allocCount += 1
+		malloc(size)
+	}
+	deallocate: override func (pointer: Void*) {
```
emilwestergren: `Void*` is called `Pointer` in ooc

source/base/ByteBuffer.ooc
thomasfanell: Good point

source/base/Allocator.ooc
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+import Owner
+
+AbstractAllocator: abstract class {
```
emilwestergren: Another way is to simply have a concrete `Allocator` class where you can supply allocate/deallocate closures. That way you don't have to inherit the class for every single allocator, only if you need to add more behaviour. Also, then we can add static default allocators in this class.

```ooc
Allocator: class 
malloc := static This new( func { malloc ... }, func { free... })
```

emilwestergren:
Looks good to me, just had some ideas/opinions but this is fine anyhow.

source/base/Allocator.ooc
sebastianbaginski: Good idea. I agree that allocating allocators for every object seems unnecessary. 

I shall add default allocators now to prevent this.

sebastianbaginski:
@thomasfanell @emilwestergren can you take a look at the latest changes ?

emilwestergren:
Looks good to me

thomasfanell:
Looks good

source/base/Allocator.ooc
shamanas: You can also use template classes to create composable allocators (I'm thinking about Andrei Alexandrescu's design of the D std.allocator), although they are quite broken right now.  

By the way, another nice way to expose allocators to any ooc code would be to add a global `current_allocator` variable and use it to allocate and free (in `Object alloc` and `Object free`) although that means that you then have to keep track of what variables were allocated with what allocator etc.  

With such a design, I expect you would use a single well tuned allocator for most allocations and switch allocators in specific places (for example, `current_allocator = SomePoolAllocator; /* allocate thousands of small objects here, use them, free them */; current_allocator = SystemWideAllocator`).

sebastianbaginski: Thanks @shamanas
`use template classes `
:arrow_up: that was the first thing I though of, but I gave up after fighting some compiler errors :/ This here is not perfect, but suits my needs for now. You have some nice ideas here, I think we may revisit this someday.

shamanas: @sebastianbaginski  
Yeah template classes definitely do need a whole lot of patching, althoug hthe only error I can currently think of right now (which is pretty big itself) is that modules that define them need to import all modules of the types they are instanciated with.

source/base/ByteBuffer.ooc
```diff
@@ -50,11 +54,14 @@ ByteBuffer: class {
...
 	}
-	new: static func ~size (size: Int) -> This { _RecyclableByteBuffer new(size) }
+	new: static func ~size (size: Int, allocator := Allocator defaultAllocator()) -> This { _RecyclableByteBuffer new(size, allocator) }
 	new: static func ~recover (pointer: Byte*, size: Int, recover: Func (This) -> Bool) -> This {
 		_RecoverableByteBuffer new(pointer, size, recover)
 	}
-	free: static func ~all { _RecyclableByteBuffer _free~all() }
+	free: static func ~all {
+		_RecyclableByteBuffer _free~all()
+		Allocator free~all()
```
marcusnaslund: Since it would be possible to use `Allocator` for other purposes besides `ByteBuffer`, I'd suggest not freeing it here but rather adding its own explicit register call to `GlobalCleanup`.
...the only bothersome thing is that the order is important (it needs to happen after `ByteBuffer free~all()`). 
`GlobalCleanup register(|| Allocator free~all(), true)` puts the call at the bottom of the (current) stack. If all else fails just put it after line 67 of this file.

test/base/ByteBufferTest.ooc
```diff
@@ -9,6 +9,18 @@
...
 
+CustomAllocator: class extends AbstractAllocator {
+	_allocCount := 0
+	init: func
+	allocate: override func (size: SizeT) -> Pointer {
+		this _allocCount += 1
+		malloc(size)
+	}
+	deallocate: override func (pointer: Pointer) {
+		memfree(pointer)
```
marcusnaslund: Maybe have a `freeCounter` as well and make sure that we indeed call `deallocate`.

source/base/ByteBuffer.ooc
sebastianbaginski: I think I've tried that and the order was messed up. Maybe we could use a `priority` integer value instead of `bool` to control the cleanup order.
But ok, for now I can put it after the line 67.

marcusnaslund: Thanks.
#1848

test/base/ByteBufferTest.ooc
sebastianbaginski: Updated.

source/base/Allocator.ooc
```diff
@@ -0,0 +1,36 @@
...
+}
+
+MallocAllocator: class extends AbstractAllocator {
+	init: func (owner := Owner Receiver) { this _owner = owner }
+	allocate: override func (size: SizeT) -> Pointer { malloc(size) }
+	deallocate: override func (pointer: Pointer) { memfree(pointer) }
+}
+
+Allocator: abstract class {
+	_defaultAllocator: static AbstractAllocator = null
```
marcusnaslund: Wait, I'm just slightly confused here. The *default* is null, and the fallback is `MallocAllocator` (which is the default today)?

sebastianbaginski: Yes, we could have 
```ooc
This _defaultAllocator = This mallocAllocator
```
but I don't know where to put this as it depends on the order of initialization of static variables.
The default allocator is meant to be overriden as 
```ooc
Allocator defaultAllocator = myAllocator
```

test/base/ByteBufferTest.ooc
```diff
@@ -9,6 +9,20 @@
...
+CustomAllocator: class extends AbstractAllocator {
+	_allocCount := 0
+	_freeCount := static 0
+	init: func
+	allocate: override func (size: SizeT) -> Pointer {
+		this _allocCount += 1
+		malloc(size)
+	}
+	deallocate: override func (pointer: Pointer) {
+		this _freeCount += 1
```
marcusnaslund: `This`

sebastianbaginski: Updated

source/base/Allocator.ooc
marcusnaslund: Alright then.

PR #1849, opened 2016-08-01
Reduced code duplication in Lists

### LINKED ISSUE #1839: Two removeAt()s in List do not need to be abstract
```ooc
removeAt: func ~range (range: Range) { this removeAt(range min, range max)) }
removeAt: abstract func ~indices (start, end: Int)
```

marcusnaslund:
Fixes #1839 


sebastianbaginski:
Great.

PR #1850, opened 2016-08-01
Added priority value to GlobalCleanup calls

### LINKED ISSUE #1848: Use a priority value for GlobalCleanup calls
The only option now is to put a call at the bottom of the stack, which still depends on the order that files are loaded. Use a priority int value instead.

marcusnaslund:
Fixes #1848 

I could also create a `Func`+priority pair and use the `SortedVectorList`, but we don't really need the complexity or performance. There shouldn't be that many items in this list, and if one registers a call with priority 123456789, one is just stupid.

Maybe the priority should be a `Byte`, we don't need the entire `Int` range, and then we won't have any problems.

Thoughts, peer review @sebastianbaginski ?

### DIFF
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -55,16 +55,12 @@ ByteBuffer: class {
 		memcpy(other pointer + destination, this pointer + start, length)
 	}
 	new: static func ~size (size: Int, allocator := Allocator defaultAllocator()) -> This { _RecyclableByteBuffer new(size, allocator) }
-	new: static func ~recover (pointer: Byte*, size: Int, recover: Func (This) -> Bool) -> This {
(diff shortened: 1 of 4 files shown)


sebastianbaginski:
` ... but we don't really need the complexity or performance`
I agree, the interface is most important, implementation can be always changed. Nice.

marcusnaslund:
Updated to handle conflicts from custom allocators in ByteBuffer.

marcusnaslund:
Gah, something's wrong.

sebastianbaginski:
Does it have something to do with the `removeAt` ? I saw your new PR for this.

marcusnaslund:
Turns out it's just my computer, let's see if this passes on PRInt.
But yes, waiting on #1854 before I merge this.

PR #1854, opened 2016-08-02
Fixed removeAt bugs in List

marcusnaslund:
Yes, this should work.
No, it does not.
I get `the arg 'index' should be of type 'Int' instead of 'Int'` when trying to use `removeAt(Int)`.


marcusnaslund:
@sebastianbaginski

sebastianbaginski:
I can't say I know what's going on here. I'll just set the flag to green.

marcusnaslund:
That's the spirit.

Issue #1857, opened 2016-08-02
Using a custom allocator/deallocator will not work if there already are bins of that size


marcusnaslund:
It will have been allocated with an incorrect allocator, and will be freed with the corresponding "incorrect" deallocator.

marcusnaslund:
I guess you will need to fix this to use the functionality properly, so can I bump it to you @sebastianbaginski ?

PR #1858, opened 2016-08-02
test for the allocator when returning recycled buffer

sebastianbaginski:
To prevent returning buffer allocated with other allocator when explicitly passing custom one.
Fixes https://github.com/magic-lang/ooc-kean/issues/1857
@marcusnaslund

### DIFF
--- a/source/base/ByteBuffer.ooc
+++ b/source/base/ByteBuffer.ooc
@@ -123,7 +123,7 @@ _RecyclableByteBuffer: class extends ByteBuffer {
 		bin := This _getBin(size)
 		This _lock lock()
 		for (i in 0 .. bin count)
-			if ((bin[i] size) == size) {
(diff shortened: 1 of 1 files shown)


source/base/ByteBuffer.ooc
```diff
@@ -123,7 +123,7 @@ _RecyclableByteBuffer: class extends ByteBuffer {
 		bin := This _getBin(size)
 		This _lock lock()
 		for (i in 0 .. bin count)
-			if ((bin[i] size) == size) {
+			if ((bin[i] size) == size && (bin[i] _allocator == allocator)) {
```
marcusnaslund: Oh, that was easy. While you're here, there are two not-needed pairs of parentheses.

sebastianbaginski: updated

marcusnaslund:
Thank you.

PR #1860, opened 2016-08-02
removed redundant cropInto implementation

sebastianbaginski:
Code was almost a duplicate of `resizeInto`.
I also think `resizeInto` is not that good name anymore, maybe rename to `copyInto` or `mapInto` (as this function can be used to perform crop without resizing) ?
@marcusnaslund


marcusnaslund:
`mapInto` sounds good to me.

PR #1866, opened 2016-08-04
Removed Synchronized<T>

### LINKED ISSUE #1837: Synchronized, to be or not to be?
This class would make more sense as an interface. But since we don't have interfaces (not sure why, since upstream ooc does?) it seems pretty redundant. It's not used in this repo and only twice in another project.

A class that needs a mutex can simply declare it as a member and call its `lock` and `unlock` methods directly, with no disadvantage (in fact the net result is less code).

marcusnaslund:
Not used anymore.
Fixes #1837


marcusnaslund:
The time has come. Peer review @sebastianbaginski

marcusnaslund:
You're calling me a bug?

sebastianbaginski:
sorry it was a missclick :sob:

PR #1867, opened 2016-08-04
Made toString() work for negative TimeSpans

### LINKED ISSUE #1836: ToString does not work for negative TimeSpans
~~Is there some logic to this difference, is it legacy from converted old C# code, or is it just an oversight that should be fixed?~~

ToString() only works for positive `TimeSpan`s at the moment.

If the ticks in `DateTime` would change to also be signed, we need the same fix there as well.

marcusnaslund:
Fixes #1836 
Simplest possible solution:

```ooc
if (this _ticks < 0L)
	result = "-(" << this negate() toString(format) >> ")"
else
	// ...
```

An easier-to-read diff is here: https://github.com/magic-lang/ooc-kean/pull/1867/files?w=1


marcusnaslund:

PR #1869, opened 2016-08-10
Added '-Ddeterministic' flag

marcusnaslund:
Running with `-Ddeterministic` will set permanent seeds on the random generators.

Why 73? Because googling `best number` has https://en.wikipedia.org/wiki/73_(number) as first result.

### DIFF
--- a/source/math/FloatRandomGenerator.ooc
+++ b/source/math/FloatRandomGenerator.ooc
@@ -17,6 +17,9 @@ FloatRandomGenerator: abstract class {
 	}
 }
 
+version (deterministic)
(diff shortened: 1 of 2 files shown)


sebastianbaginski:
Funny, I thought it's 42.

marcusnaslund:
73 is the smallest factor of the first composite generalized Fermat number in base 10 (10^4 + 1 = 10,001 = 73 × 137)

sebastianbaginski:
And 42 is 24 if written backwards.

marcusnaslund:
73 written backwards twice is 73, so checkmate.

PR #1871, opened 2016-08-16
fixed android build with the other compiler

sebastianbaginski:
Doesn't seem to like `const`.
@marcusnaslund

### DIFF
--- a/source/base/atomic.c
+++ b/source/base/atomic.c
@@ -8,7 +8,7 @@ void atomic_store_int(atomic_int* object, int value) {
 	atomic_store(object, value);
 }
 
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Putting the green flag yourself? You're quite a rebel.

sebastianbaginski:
I think the obvious PRs should go straight to you, which means green flag from the start ?

marcusnaslund:
> which means green flag from the start ?

Sure :) But you've usually put the peer review flag there even though you tagged me.

sebastianbaginski:
PRInt seems to be missing some files.

marcusnaslund:
Oops.

PR #1877, opened 2016-08-19
Moved _Synchronizer state variable

christopherlagerhult:


test/concurrent/ThreadPoolTest.ooc
```diff
@@ -27,6 +27,11 @@ ThreadPoolTest: class extends Fixture {
 			expect(promise3 wait())
 			expect(promise4 wait(), is false)
 			expect(promise5 wait(), is true)
+			expect(promise1 _state as Int, is equal to(_PromiseState Finished as Int))
+			expect(promise2 _state as Int, is equal to(_PromiseState Finished as Int))
+			expect(promise3 _state as Int, is equal to(_PromiseState Finished as Int))
+			expect(promise4 _state as Int, is equal to(_PromiseState Cancelled as Int))
+			expect(promise5 _state as Int, is equal to(_PromiseState Finished as Int))
```
christopherlagerhult: Should these expect calls return true, i.e. should the Promise state be changed for `ThreadPool` generated promises such as `_TaskPromise/_TaskFuture`? Currently they only change the state of their `_task`. @emilwestergren

emilwestergren: If they have a state then I assume this state should be changed. @marcusnaslund

marcusnaslund: Well, they have a state, but it's not public, so you are really "not doing it right" by testing this. Looks like we should maybe restructure these classes a bit.

christopherlagerhult: Yeah of course. The question if these should pass or not remains though. Is it a bug that they all fail or is it intended that way?

marcusnaslund: There should be some public (overrideable) method that returns the "appropriate" state. For the promises/futures returned by threadpool, the appropriate state is not this state but rather the one in their respective task.

This leads me to believe it may be incorrect that `_Synchronizer` has a state, and we should move it to `_ThreadPromise` and `_ThreadFuture<T>` classes. In these classes, `getState()` will return this state, and in the threadpool versions, `getState()` will return the state in their tasks.

christopherlagerhult: Sounds a lot better as moving ```_state``` to ```_ThreadPromise``` and ```ThreadFuture<T>``` hides it from ```_Threadpool```. Although I'm not sure we need to have a ```getState()``` method since no code (currently) make use of the state outside the class itself (neglecting my newly invented test case).

marcusnaslund: We've been fine without it for now, I was just thinking it would make your tests easier to perform.

marcusnaslund: Right now this code does not even compile.

christopherlagerhult: Ok. Yeah, these tests are probably redundant. Added them to visualize a potential bug/problem/something. I'll remove them now that we agreed on the state variable =)

fredrikbryntesson:
Tmp close

marcusnaslund:
Conflicts.

emilwestergren:
Merge in `develop` branch instead

PR #1878, opened 2016-08-22
Debug deadlock

christopherlagerhult:
Using the version(debugDeadlock) flag will now raise an error when a promise/mutexunix takes "too long" to finish.
Notes: 

- ~~Mutexes are ignored since Android uses ``std::mutex`` instead of our ``Mutex.ooc``.~~
- A timer is used for the timeout check since  ``_Synchronizer _state`` is currently very unreliable:  ``_Synchronizer`` objects created by ``Threadpool`` have tasks whose state are changed instead. ``OutputFrame`` seems to ignore the state variable as it uses a ``WaitCondition`` for the actual wait.
- ~~The timer uses ticks rather than seconds for the timeout due to the fact that ``_Task wait()`` is implemented that way.~~

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -19,7 +19,19 @@ _Synchronizer: abstract class {
 	_state: _PromiseState
 	init: func
 	wait: abstract func (time: TimeSpan) -> Bool
-	wait: func ~forever -> Bool { this wait(TimeSpan maximumValue) }
+	wait: func ~forever -> Bool {
+		version(debugDeadlock) {
+			timeLimit_msec := 2_000
+			timer := CpuTimer new() . start()
+			result := this wait(TimeSpan milliseconds(timeLimit_msec))
+			if (timer stop() > timeLimit_msec)
+				Debug error("Error! [" + this class name + " stalled for more than " + timeLimit_msec + " ticks and timed out]")
+			timer free()
+		}
+		else
+			result := this wait(TimeSpan maximumValue)
+		result
+	}
 	cancel: virtual func -> Bool { false }
 }
 
--- a/source/system/native/MutexUnix.ooc
+++ b/source/system/native/MutexUnix.ooc
@@ -19,7 +19,17 @@ version(unix || apple) {
 			super()
 		}
 		lock: override func {
-			pthread_mutex_lock(this _backend&)
+			version(debugDeadlock) {
+				timeLimit_sec := 1
+				deadline: TimeSpec
+				currentTime: TimeVal
+				gettimeofday(currentTime&, null)
+				(deadline tv_sec, deadline tv_nsec) = (currentTime tv_sec + (timeLimit_sec as TimeT), (currentTime tv_usec * 1000) as Int)
+				if (pthread_mutex_timedlock(this _backend&, deadline&) != 0)
+					raise("Error! [" + this class name + " stalled for more than " + timeLimit_sec + " seconds and timed out]")
+			}
+			else
+				pthread_mutex_lock(this _backend&)
 		}
 		unlock: override func {
 			pthread_mutex_unlock(this _backend&)
--- a/source/system/external/pthread.ooc
+++ b/source/system/external/pthread.ooc
@@ -23,6 +23,7 @@ pthread_cond_broadcast: extern func (cond: PThreadCond*) -> Int
 pthread_cond_wait: extern func (cond: PThreadCond*, mutex: PThreadMutex*) -> Int
 pthread_cond_destroy: extern func (cond: PThreadCond*) -> Int
 
+pthread_mutex_timedlock: extern func (PThreadMutex*, TimeSpec*) -> Int
 pthread_mutex_lock: extern func (PThreadMutex*)
 pthread_mutex_unlock: extern func (PThreadMutex*)
 pthread_mutex_init: extern func (PThreadMutex*, PThreadMutexAttr*)


christopherlagerhult:
@emilwestergren

source/concurrent/Promise.ooc
```diff
@@ -19,7 +19,19 @@ _Synchronizer: abstract class {
 	_state: _PromiseState
 	init: func
 	wait: abstract func (time: TimeSpan) -> Bool
-	wait: func ~forever -> Bool { this wait(TimeSpan maximumValue) }
+	wait: func ~forever -> Bool {
+		version(debugDeadlock) {
+			timeLimitTicks := 2_000
+			timer := CpuTimer new() . start()
+			result := this wait(TimeSpan milliseconds(timeLimitTicks))
+			if (timer stop() > timeLimitTicks)
```
emilwestergren: Isn't `timer stop()` in milliseconds?

source/concurrent/Promise.ooc
```diff
@@ -19,7 +19,19 @@ _Synchronizer: abstract class {
 	_state: _PromiseState
 	init: func
 	wait: abstract func (time: TimeSpan) -> Bool
-	wait: func ~forever -> Bool { this wait(TimeSpan maximumValue) }
+	wait: func ~forever -> Bool {
+		version(debugDeadlock) {
+			timeLimitTicks := 2_000
```
emilwestergren: If this is supposed to be 2000 ms:
```ooc
timeLimit := TimeSpan milliseconds(2000)
...
if (TimeSpan milliseconds(timer stop()) > timeLimit)
```

emilwestergren: Although I think in the long run `Timer` should return a `TimeSpan`

source/concurrent/Promise.ooc
christopherlagerhult: > Isn't timer stop() in milliseconds?

Using ``WallTimer`` you have milli/micro seconds. ``CpuTimer`` expresses the time in ticks.

emilwestergren: `this _result = 1000.0 * (this _endTime - this _startTime) / CLOCKS_PER_SEC`
Looks like ms to me

source/concurrent/Promise.ooc
christopherlagerhult: Agree on ``Timer`` returning a `TimeSpan`. The problem though of using "real time" for the `timer` in this case is that `_Task wait(TimeSpan)` uses ticks. 

```ooc
timer := WallTimer new() . start()
result := this wait(TimeSpan milliseconds(1000)) 
endTime := timer stop() 
```
``endTime`` in that case will probably be < 1000 since a cpu tick is probably counted faster than a "real" millisecond.

source/concurrent/Promise.ooc
christopherlagerhult: Might have missed that line =)

source/concurrent/Promise.ooc
```diff
@@ -19,7 +19,19 @@ _Synchronizer: abstract class {
...
+			timer := CpuTimer new() . start()
```
emilwestergren: Does this work? I mean instead of:

```ooc
timer := CpuTimer new()
timer start()
```

christopherlagerhult: Yes =)

source/concurrent/Promise.ooc
emilwestergren: Check out my code above ^

christopherlagerhult: Probably found inspiration from there

source/concurrent/Promise.ooc
```diff
@@ -19,7 +19,19 @@ _Synchronizer: abstract class {
...
 	init: func
 	wait: abstract func (time: TimeSpan) -> Bool
-	wait: func ~forever -> Bool { this wait(TimeSpan maximumValue) }
+	wait: func ~forever -> Bool {
+		version(debugDeadlock) {
+			timeLimit_msec := 2_000
+			timer := CpuTimer new() . start()
+			result := this wait(TimeSpan milliseconds(timeLimit_msec))
+			if (timer stop() > timeLimit_msec)
+				Debug error("Error! [" + this class name + " stalled for more than " + timeLimit_msec + " ticks and timed out]")
```
marcusnaslund: `timeLimitMilliseconds` is a better variable name, and you need to change from "ticks" to "milliseconds" in the text.

source/system/native/MutexUnix.ooc
```diff
@@ -19,7 +19,17 @@ version(unix || apple) {
 			super()
 		}
 		lock: override func {
-			pthread_mutex_lock(this _backend&)
+			version(debugDeadlock) {
+				timeLimit_sec := 1
```
marcusnaslund: Why is this in seconds and the one in `Promise` in milliseconds?

source/system/native/MutexUnix.ooc
```diff
@@ -19,7 +19,17 @@ version(unix || apple) {
...
+				deadline: TimeSpec
+				currentTime: TimeVal
+				gettimeofday(currentTime&, null)
+				(deadline tv_sec, deadline tv_nsec) = (currentTime tv_sec + (timeLimit_sec as TimeT), (currentTime tv_usec * 1000) as Int)
+				if (pthread_mutex_timedlock(this _backend&, deadline&) != 0)
+					raise("Error! [" + this class name + " stalled for more than " + timeLimit_sec + " seconds and timed out]")
```
marcusnaslund: Use `Debug error` rather than `raise` so we can read output on android too.

source/system/native/MutexUnix.ooc
```diff
@@ -19,7 +19,17 @@ version(unix || apple) {
 			super()
 		}
 		lock: override func {
-			pthread_mutex_lock(this _backend&)
+			version(debugDeadlock) {
+				timeLimit_sec := 1
+				deadline: TimeSpec
+				currentTime: TimeVal
+				gettimeofday(currentTime&, null)
+				(deadline tv_sec, deadline tv_nsec) = (currentTime tv_sec + (timeLimit_sec as TimeT), (currentTime tv_usec * 1000) as Int)
```
marcusnaslund: Why is the `tv_nsec` argument cast to `Int`, but not `tv_sec`?

source/system/native/MutexUnix.ooc
christopherlagerhult: Mostly due too convenience and I forgot to change that.

marcusnaslund: I suggest using milliseconds here as well.

source/system/native/MutexUnix.ooc
christopherlagerhult: Can't reach Debug class from `MutexUnix`. 

- Import makes the compiler complain as it "leaves" the system folder
- use (some use file with debug) creates a circular dependency as debug is included somewhere else. 
- Also tried creating a small Debug print class inside `MutexUnix` w/o success (as it didn't print anything). 
Are there any other way of reaching `Debug.ooc` as I don't know about?

source/system/native/MutexUnix.ooc
christopherlagerhult:
```TimeVal: cover from struct timeval {
		tv_sec: extern TimeT
		tv_usec: extern Int
	}

	TimeSpec: cover from struct timespec {
		tv_sec: extern TimeT
		tv_nsec: extern Long
	}```

File: source/system/native/MutexUnix.ooc (hunk 6 above)
Comment (marcusnaslund) [inline source/system/native/MutexUnix.ooc]: Ahhhh right. There's nothing we can do about it atm.

File: source/system/native/MutexUnix.ooc (hunk 7 above)
Comment (marcusnaslund) [inline source/system/native/MutexUnix.ooc]: I don't understand. `tv_usec` is an `Int`, but you still cast it to `Int`.
You are storing it in `tv_nsec`, which is a `Long`. It would make more sense to cast to `Long`, but you don't have to since `Int->Long` is implicit.

Comment (christopherlagerhult) [inline source/system/native/MutexUnix.ooc]: Oh you're absolutely right. Tired eyes today.

Comment (marcusnaslund) [inline source/system/native/MutexUnix.ooc]: Fridays easily have that effect.

File: source/system/native/MutexUnix.ooc (hunk 6 above)
Comment (marcusnaslund) [inline source/system/native/MutexUnix.ooc]: We might be able to fix it after #1883 - but it will go through some heavy testing first.

**Comment by christopherlagerhult** [comment]
Waiting for #1885 before continuing

File: source/concurrent/Promise.ooc [hunk 8]
```diff
@@ -19,7 +19,19 @@ _Synchronizer: abstract class {
 	_state: _PromiseState
 	init: func
 	wait: abstract func (time: TimeSpan) -> Bool
-	wait: func ~forever -> Bool { this wait(TimeSpan maximumValue) }
+	wait: func ~forever -> Bool {
+		version(debugDeadlock) {
+			timeLimit_msec := 2_000
+			timer := CpuTimer new() . start()
+			result := this wait(TimeSpan milliseconds(timeLimit_msec))
```
emilwestergren: You shouldn't need the timer for this right? Isn't it possible to do it like this:

```ooc
if (!this wait(TimeSpan milliseconds(2000))
   Debug error()...
```

marcusnaslund: Depends on what you want to measure. Doing it as you suggest will measure a wall time of 2000 ms, which isn't what the `CpuTimer` does.

christopherlagerhult: > You shouldn't need the timer for this right?

Won't `wait(TimeSpan)` trigger a deadlock then if cancelled?

emilwestergren: I think wall time should be fine?

emilwestergren: What about:
```ooc
if (!this wait(TimeSpan milliseconds(2000)) && this _state != cancelled)
   Debug error()...
```

emilwestergren: Maybe all the `wait` functions should return enum instead of bool so you know the reason.

christopherlagerhult: But `this _state` is never when a `Threadpool` is used. It modifes `_Task state` instead

marcusnaslund: That virtual `getState` method would help you here. :)

emilwestergren: Alright, it's not very important but it would clean up the code. What about `CpuTimer`, why not use `WallTimer` ?

christopherlagerhult: Yeah, but is it worth the trouble for a single debug feature? :)

emilwestergren:
Could you merge this in `develop` branch instead?

source/concurrent/Promise.ooc
christopherlagerhult: > Alright, it's not very important but it would clean up the code. What about CpuTimer, why not use WallTimer ?

`WallTimer` could work, but since `_Task wait(TimeSpan)` uses a `CpuTimer`, a timeout might not trigger a deadlock then right?

emilwestergren: Oh I see, why does `_Task wait` use `CpuTimer` @marcusnaslund ?

marcusnaslund: Well, ehm, the person who wrote that code was actually *you*. :) Commit: https://github.com/magic-lang/ooc-kean/commit/74b56f3ae8d7348de7b0ed398e31e5d81e45a36f

emilwestergren: I think it was copied from somewhere else though. Anyway, I don't know why it uses `CpuTimer` so should we change it?

christopherlagerhult:
Moved to develop branch as task #1887

source/concurrent/Promise.ooc
marcusnaslund: Makes sense to me.

PR #1879, opened 2016-08-25
Added 'absolute' property to FloatPoint2D and IntPoint2D

marcusnaslund:


marcusnaslund:
Created issue #1880 to do some more work along these lines, but this is all I need right now.

marcusnaslund:
*merging before PRInt because I'm the coolest guy around*

Issue #1880, opened 2016-08-25
Vector and Point covers differ a lot in what properties are available


marcusnaslund:
Maybe also some methods. This needs to be standardized.

- [x] `absolute` and `sign` properties in all covers
- [x] `distance` method to all covers
- [x] constructor with *one* argument to all covers
- [x] `length` should change name to `norm` (breaks API?)
- [x] Either `FloatPoint3D` should not have `normalized`, or *all* points should have `normalized`.
- [x] Add `isZero` property for vectors (and points?)

PR #1881, opened 2016-08-25
Standardized properties of vectors and points

marcusnaslund:
Fixes first three points of #1880 

Useful to clean up code in secondary feature 1.4.


marcusnaslund:
Peer review @thomasfanell

PR #1882, opened 2016-08-25
Cleaned up amongst String operators

marcusnaslund:
Removed `operator * (string: String, count: Int)` (not used) and four of the type `operator + (left: String, right: LLong)` (would leak memory if used!)

Most others were moved into `String`, where they belong.


marcusnaslund:
Peer review @thomasfanell

PR #1883, opened 2016-08-26
Moved Mutex to source/base

### LINKED ISSUE #1884: The location of Mutex is illogical
This PR https://github.com/magic-lang/ooc-kean/pull/1878#discussion-diff-76243900 should use `Debug error` instead of `raise` in an error checking inside `Mutex`, but cannot because `Mutex` resides in `system` and `Debug` resides in `base`.

PR #1883 shows it is possible to move `Mutex` out of `System`. It should reside in `concurrent`, but that is at present not possible because `ByteBuffer` uses it, and resides in `base`. 

This is what we should try:

1) Move `Mutex` to `base`.
2) Move `Debug` to `system`?

marcusnaslund:
`Mutex` is now lifted out of `system` and into `base`.

Fixes #1884 and does not break the API.


marcusnaslund:
Moves the ByteBuffer (since it uses a Mutex) and the Mutex classes to `concurrent`. Seems to cause no problems other than in the API test, at the moment not sure why, but should be fixable.

Let's us use `Debug error` instead of `raise` for debug code in `Mutex`, as per https://github.com/magic-lang/ooc-kean/pull/1878#discussion-diff-76243900.

What's your opinion @thomasfanell @fredrikbryntesson @emilwestergren @sebastianbaginski ?

thomasfanell:
I'm not too sure I would think to look in `concurrent` if I searched for functionality that `ByteBuffer` have to offer. If we do go ahead and change the location, perhaps we should also change its name?

marcusnaslund:
Another option is to put `Mutex` in `base`.

sebastianbaginski:
Dependency management with rock .use files is a mess. Everything is exposed, there is no such thing as "private" implementation in ooc - imagine how much easier would it be with just forward-declared mutex in byte buffer.
I agree with @thomasfanell  , `concurrent` is the last place I'd search for a byte buffer.
`Mutex` in `base` sounds a lot better :+1:

fredrikbryntesson:
Tmp close

source/base/native/MutexUnix.ooc
```diff
@@ -28,7 +28,7 @@ version(unix || apple) {
 				if (deadline tv_nsec >= 1_000_000_000)
 					(deadline tv_sec, deadline tv_nsec) = (deadline tv_sec + 1, deadline tv_nsec - 1_000_000_000)
 				if (pthread_mutex_timedlock(this _backend&, deadline&) != 0)
-					raise("Error! [" + this class name + " stalled for more than " + timeLimitMilliseconds + " milliseconds and timed out]")
+					Debug error("Error! [" + this class name + " stalled for more than " + timeLimitMilliseconds + " milliseconds and timed out]")
```
marcusnaslund: Now we can do this.

marcusnaslund:
Peer review @emilwestergren

emilwestergren:
Can we move it to `concurrent` in the long run?

marcusnaslund:
That's what I want. The only thing stopping us is the `ByteBuffer`.

PR #1885, opened 2016-08-26
Bug in _Task wait

christopherlagerhult:
Fixed a bug where a cancelled `_Task` returned false when calling `wait()`, while true when calling `wait(TimeSpan)`


christopherlagerhult:
@marcusnaslund

marcusnaslund:
Since this changes the behaviour of `Promise` maybe @emilwestergren should take a look as well

PR #1888, opened 2016-08-29
Moved `_state` from `_Synchronizer ` to `_ThreadPromise/_ThreadFuture`

christopherlagerhult:
Moved to develop branch (old: #1877)

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -16,7 +16,6 @@ _PromiseState: enum {
 }
 
 _Synchronizer: abstract class {
-	_state: _PromiseState
 	_stateMutex := Mutex new()
 	init: func
 	free: override func {
@@ -40,6 +39,7 @@ _EmptyPromise: class extends Promise {
 }
 
 _ThreadPromise: class extends Promise {
+	_state: _PromiseState
 	_action: Func
 	_thread: Thread
 	_threadAlive := true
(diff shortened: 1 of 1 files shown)


emilwestergren:
Reason for this?

christopherlagerhult:
Same as #1877. `_state` is only used in `_ThreadPromise/_ThreadFuture` which might make `_state` ambiguous when dealing with Promises created elsewhere(ex. `_TaskPromise`). 

for example:
```ooc
_Synchronizer: abstract class {
	_state: _PromiseState
	...
	wait: abstract func (time: TimeSpan) -> Bool { 
            this wait(time)
            this _state == _PromiseState Finished 
}
```
A `_TaskPromise` will always return false here.
Might be stupid to change it but it is still weird to have a state that is always `Unfinished`.

source/concurrent/Promise.ooc
```diff
@@ -118,6 +118,7 @@ Future: abstract class <T> extends _Synchronizer {
 }
 
 _ThreadFuture: class <T> extends Future<T> {
+	_state: _PromiseState
```
emilwestergren: This should probably not be named `_PromiseState` if it's used for both `Future` and `Promise`. Not related to this PR though and not that important.

marcusnaslund: #1889

marcusnaslund:
I'm not completely sure about this, but it removes confusion as you say, and we can revisit it later if need be.

PR #1890, opened 2016-08-29
_Task uses WallTimer instad of CpuTimer

christopherlagerhult:


christopherlagerhult:
Since `WallTimer stop()` returns microseconds instead of milliseconds, I had to change that too. Would prefer though to make `WallTimer` return milliseconds or extend `TimeSpan` with a function `toMicroseconds()`. review @emilwestergren @marcusnaslund

PR #1893, opened 2016-08-31
Merged develop.

emilwestergren:
@marcusnaslund peer review

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -16,7 +16,6 @@ _PromiseState: enum {
 }
 
 _Synchronizer: abstract class {
-	_state: _PromiseState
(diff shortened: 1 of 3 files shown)


marcusnaslund:
Looks good.

emilwestergren:
I think Print fails because it doesn't use the correct version of the wrapper.

marcusnaslund:
Hm, but if it used the wrong wrapper, PRs to this branch would have failed before too... @fredrikbryntesson ?

emilwestergren:
Yeah, it's the correct wrapper but for some reason it can't find the shared library.

emilwestergren:
Still fails with same error.

marcusnaslund:
Works now.

PR #1895, opened 2016-08-31
Cleaned EGLRgba recycling.

emilwestergren:
@erikhagglund peer review


marcusnaslund:
Not one, not two, not three, but *four* PRInts say this PR is good, so it must be amazing.

PR #1896, opened 2016-08-31
Primary feature1.5

emilwestergren:
@marcusnaslund can you have a look at the changes since it's not too much?


marcusnaslund:
Perfection.

marcusnaslund:
Actually, the `String.ooc` diff makes no sense. Those changes *came from* develop. What have you done? :)

PR #1904, opened 2016-09-06
check if vector is zero before normalizing

sebastianbaginski:
@marcusnaslund review

### DIFF
--- a/source/geometry/FloatPoint3D.ooc
+++ b/source/geometry/FloatPoint3D.ooc
@@ -17,7 +17,7 @@ FloatPoint3D: cover {
 	x, y, z: Float
 
 	norm ::= (this x squared + this y squared + this z squared) sqrt()
-	normalized ::= this / this norm
+	normalized ::= (this x equals(0.0f) && this y equals(0.0f) && this z equals(0.0f)) ? (this as This) : (this / this norm)
 	azimuth ::= this y atan2(this x)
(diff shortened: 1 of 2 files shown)


source/geometry/FloatPoint3D.ooc
```diff
@@ -17,7 +17,7 @@ FloatPoint3D: cover {
 	x, y, z: Float
 
 	norm ::= (this x squared + this y squared + this z squared) sqrt()
-	normalized ::= this / this norm
+	normalized ::= (this x equals(0.0f) && this y equals(0.0f) && this z equals(0.0f)) ? (this as FloatPoint3D) : (this / this norm)
```
sebastianbaginski: `(this as FloatPoint3D)` otherwise compiler error :/

sebastianbaginski:
@niklasborwell @tomasmahlberg check it out and comment as well

niklasborwell:
No comments from me, seems right.

source/geometry/FloatPoint3D.ooc
emilwestergren: Should have `isZero` property or something for zero vector.

marcusnaslund: I added an item for that on #1880

sebastianbaginski: @emilwestergren there is one already for `FloatVector3D` but rock fails if I use it here instead of this hard coded check.

PR #1907, opened 2016-09-13
Merge Secondaryfeature1.4

marcusnaslund:


marcusnaslund:
We can do this ahead of time, can I just get an OK from @sebastianbaginski ?

sebastianbaginski:
Looks ok.

PR #1916, opened 2016-09-19
Renamed _PromiseState to TaskState

### LINKED ISSUE #1889: Rename _PromiseState

marcusnaslund:
The name was confusing since it was also used in `_ThreadFuture<T>` and `Task`.

Also removed the `_` since it's used in multiple classes - we cannot really consider it "private".

Fixes #1889

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -9,7 +9,7 @@
 use base
 use concurrent
 
-_PromiseState: enum {
+TaskState: enum {
 	Unfinished
 	Finished
 	Cancelled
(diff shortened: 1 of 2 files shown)


marcusnaslund:
OK with you @emilwestergren ?

emilwestergren:
I think it's still confusing that we have a class `Task`, a state called `TaskState`, and this state is also used in `Promise` which neither is a task or holds a task.

marcusnaslund:
Maybe we should move `Task` out of `ThreadPool` into its own class, and use `Task` internally for `_ThreadPromise` and `_ThreadFuture`?

emilwestergren:
I don't know, it's clear these classes need some refactoring.

marcusnaslund:
Yeah.. let me get back to this when I have some more time.

PR #1917, opened 2016-09-19
Cleanup in Timer classes

### LINKED ISSUE #1894: Walltimer stop() and CpuTimer stop() return Double instead of Long
The value is in milliseconds, so this really makes no sense. Should they return seconds?

Also, it might make more sense to rename `stop` to `elapsed` or something like that.

Also, an `isRunning()` method to see if `start()` ever has been called would be nice.

marcusnaslund:
Fixes #1894 

`Timer::stop` returns a `TimeSpan`. Added an `isRunning` method. Also some minor cleanup in the `Timer` classes.

### DIFF
--- a/source/base/Profiler.ooc
+++ b/source/base/Profiler.ooc
@@ -53,15 +53,15 @@ Profiler: class {
 	benchmark: static func (task: Func, timeoutMilliseconds := 1000.0) -> Double {
 		numberOfIterations := 0
 		timer := WallTimer new()
-		totalTime := 0.0
-		while (totalTime < timeoutMilliseconds * 1000.0) {
+		totalTime := TimeSpan new(0)
+		while (totalTime toMilliseconds() < timeoutMilliseconds) {
 			numberOfIterations += 1
 			timer start()
 			task()
 			totalTime += timer stop()
 		}
 		timer free()
-		totalTime / (numberOfIterations * 1000)
+		totalTime toMilliseconds() as Double / (numberOfIterations * 1000)
 	}
 	_logResults: static func (logMethod: Func (String)) {
 		for (i in 0 .. This _profilers count) {
--- a/source/base/Profiler.ooc
+++ b/source/base/Profiler.ooc
@@ -61,7 +61,7 @@ Profiler: class {
 			totalTime += timer stop()
 		}
 		timer free()
-		totalTime toMilliseconds() as Double / (numberOfIterations * 1000)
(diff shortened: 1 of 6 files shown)


marcusnaslund:

source/base/Profiler.ooc
```diff
@@ -53,15 +53,15 @@ Profiler: class {
...
+		totalTime := TimeSpan new(0)
+		while (totalTime toMilliseconds() < timeoutMilliseconds) {
 			numberOfIterations += 1
 			timer start()
 			task()
 			totalTime += timer stop()
 		}
 		timer free()
-		totalTime / (numberOfIterations * 1000)
+		totalTime toMilliseconds() as Double / (numberOfIterations * 1000)
```
sebastianbaginski: Can you `totalTime / (numberOfIterations * 1000.0)` instead of cast ? Or not possible because `ooc` ?

source/base/Timer.ooc
```diff
@@ -40,34 +40,28 @@ Timer: abstract class {
...
 	}
 	reset: virtual func {
-		this _startTime = 0
-		this _endTime = 0
-		this _result = 0
-		this _total = 0
-		this _count = 0
-		this _min = 0
-		this _average = 0
+		(this _min, this _max, this _startTime, this _endTime, this _result, this _total, this _count, this _average) = (Double positiveInfinity, 0, 0, 0, 0, 0, 0, 0)
```
sebastianbaginski: I'm sure @bettinaselig is gonna love this line :)

source/unit/Fixture.ooc
```diff
@@ -53,7 +53,7 @@ Fixture: abstract class {
 			This failureNames add(this name clone())
 		}
 		This _print(result ? " done" : " failed")
-		testTime := timer stop() / 1_000_000.0
+		testTime := timer stop() toSeconds()
```
sebastianbaginski: So much better

source/base/Profiler.ooc
marcusnaslund: I think it should work the way you suggest. There used to be problems with ooc implicit casts, but it should work now. I will try it.

source/base/Timer.ooc
marcusnaslund: Schhhhhhh!

marcusnaslund: I put the nonzero assignemtn on its own line. Maybe she won't complain too much now. :)

bettinaselig: Haha! :-P

PR #1918, opened 2016-09-19
Color distances are now Euclidean

### LINKED ISSUE #1827: Color distances are not euclidean...should they be?
And some look plain wrong.

Most seem to suffer from a pair of missing parentheses, as is the case for `ColorUv`, `ColorYuv`, `ColorRgb`:
`((this u - other u) as Float pow(2) + (this v - other v) as Float pow(2)) / 2.0f sqrt()`
should be
`(((this u - other u) as Float pow(2) + (this v - other v) as Float pow(2)) / 2.0f) sqrt()`

The error in `ColorRgba` is even more severe. Currently it is:
`(this toRgb() distance(other toRgb()) * 3.0f + (this a - other a) as Float pow(2)) / 4.0f sqrt()`
but should be
`((this toRgb() distance(other toRgb()) pow(2) * 3.0f + (this a - other a) as Float pow(2)) / 4.0f) sqrt()`

Finally, `ColorYuva` has no distance method at all.

marcusnaslund:
I don't know who wrote these or what metric is considered "good" for color spaces, but these functions made no sense to me. Especially bad was `ColorRgba`. Also added a missing `distance` method for `ColorYuva` in the same way as for `ColorRgba`.

Am I wrong to think that euclidean distance is easier to handle or is this actually better? Peer review @davidpiuva 

Fixes #1827

### DIFF
--- a/source/draw/Color.ooc
+++ b/source/draw/Color.ooc
@@ -110,7 +110,7 @@ ColorYuva: cover {
 	mix: static func (a, b: This, ratio: Float) -> This {
 		This new(Float mix(a y, b y, ratio) as Byte, Float mix(a u, b u, ratio) as Byte, Float mix(a v, b v, ratio) as Byte, Float mix(a a, b a, ratio) as Byte)
 	}
-	distance: func (other: This) -> Float { ((this toYuv() distance(other toYuv()) squared * 3.0f + (this a - other a) squared) / 4.0f) sqrt() }
+	distance: func (other: This) -> Float { (((this y - other y) squared + (this u - other u) squared + (this v - other v) squared + (this a - other a) squared) / 4.0f) sqrt() }
 	operator == (other: This) -> Bool { this equals(other) }
 	operator != (other: This) -> Bool { !this equals(other) }
 }
@@ -164,7 +164,7 @@ ColorRgba: cover {
 	mix: static func (a, b: This, ratio: Float) -> This {
 		This new(Float mix(a r, b r, ratio) as Byte, Float mix(a g, b g, ratio) as Byte, Float mix(a b, b b, ratio) as Byte, Float mix(a a, b a, ratio) as Byte)
 	}
-	distance: func (other: This) -> Float { ((this toRgb() distance(other toRgb()) squared * 3.0f + (this a - other a) squared) / 4.0f) sqrt() }
+	distance: func (other: This) -> Float { (((this r - other r) squared + (this g - other g) squared + (this b - other b) squared + (this a - other a) squared) / 4.0f) sqrt() }
 	operator == (other: This) -> Bool { this equals(other) }
 	operator != (other: This) -> Bool { !this equals(other) }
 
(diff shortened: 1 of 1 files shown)


davidpiuva:
Euclidean distance makes sense but we want the same distance for all color formats for consistency with test cases. Black to white is not further away just because we use RGB to describe it.

source/draw/Color.ooc
```diff
@@ -163,7 +164,7 @@ ColorRgba: cover {
 	mix: static func (a, b: This, ratio: Float) -> This {
 		This new(Float mix(a r, b r, ratio) as Byte, Float mix(a g, b g, ratio) as Byte, Float mix(a b, b b, ratio) as Byte, Float mix(a a, b a, ratio) as Byte)
 	}
-	distance: func (other: This) -> Float { (this toRgb() distance(other toRgb()) * 3.0f + (this a - other a) as Float pow(2)) / 4.0f sqrt() }
+	distance: func (other: This) -> Float { ((this toRgb() distance(other toRgb()) as Float pow(2) * 3.0f + (this a - other a) as Float pow(2)) / 4.0f) sqrt() }
```
davidpiuva: Why not just copy the distance function from RGB and extend it from the inside?
`sqrt(r^2 + g^2 + b^2 + a^2)`

source/draw/Color.ooc
```diff
@@ -135,7 +136,7 @@ ColorRgb: cover {
 		This new(Float mix(a r, b r, ratio) as Byte, Float mix(a g, b g, ratio) as Byte, Float mix(a b, b b, ratio) as Byte)
 	}
 	distance: func (other: This) -> Float {
-		((this r - other r) as Float pow(2) + (this g - other g) as Float pow(2) + (this b - other b) as Float pow(2)) / 3.0f sqrt()
+		((this r - other r) as Float pow(2) + (this g - other g) as Float pow(2) + (this b - other b) as Float pow(2)) sqrt()
```
davidpiuva: `distance(a, b)` should equal `distance(a toRgba(), b toRgba())` because it is still the same perceived colors. The problem comes when one adds a uniform channel and wants to have the same distance.

davidpiuva:
@niklasborwell What do you think about this?

source/draw/Color.ooc
davidpiuva: Maybe we can divide RGB by 3 but keep alpha in full scale.

marcusnaslund: Sure, I can do that. But why should alpha be in full scale?

davidpiuva: If you write a logo using alpha instead of monochrome then alpha would be like intensity but it all depends on the use case of the channels.

marcusnaslund:
I think I'm done fixing/cleaning and all tests pass. @davidpiuva

davidpiuva:
There may be objections to whether this is the way to compare images but at least it is consistent with the name now.

marcusnaslund:
I'll give @tomasmahlberg @niklasborwell @emilwestergren some time to object before I merge this.

emilwestergren:
lewks gewd

PR #1919, opened 2016-09-19
Fixed inconsistencies in source/geometry

### LINKED ISSUE #1880: Vector and Point covers differ a lot in what properties are available

marcusnaslund:
Fixes #1880

### DIFF
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -115,7 +115,7 @@ RasterMonochrome: class extends RasterPacked {
 						result += distance sqrt()
 					}
 				}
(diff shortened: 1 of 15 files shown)


marcusnaslund:

PR #1922, opened 2016-09-20
Added simple chaining for Promise and Future classes

### LINKED ISSUE #1906: (Re)implement ThreadPool::waitAll()

marcusnaslund:
Fixes #1906 

Allows us to simplify code like 

```ooc
promises: Promise[count]
for (i in 0 .. count)
   promises[i] = pool getPromise(f)
...
for (i in 0 .. count)
   promises[i] wait() . free()
promises free()
```

to

```ooc
promises := VectorList<Promise>
for (i in 0 .. count)
    promises add(threadPool getPromise(f))
...
Promise wait(promises)
promises free()
```

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -36,6 +36,12 @@ _Synchronizer: abstract class {
 			result := this wait(TimeSpan maximumValue)
 		result
 	}
+	wait: static func ~timed (items: VectorList<This>, time := TimeSpan maximumValue) -> Bool {
+		result := true
+		for (i in 0 .. items count)
+			result = result && (time == TimeSpan maximumValue ? items[i] wait() : items[i] wait(time))
+		result
+	}
 	cancel: virtual func -> Bool { false }
 }
 
--- a/test/concurrent/PromiseTest.ooc
+++ b/test/concurrent/PromiseTest.ooc
@@ -88,6 +88,12 @@ PromiseTest: class extends Fixture {
 			promise2 = Promise start(this counter)
 			promise2 wait() . free()
 		})
+		this add("many promises", func {
+			list := VectorList<Promise> new()
+			list add(Promise start(this counter)) . add(Promise start(this counter)) . add(Promise start(this counter))
+			expect(Promise wait(list))
+			list free()
+		})
 	}
 }
 


marcusnaslund:
Is this what you want? @emilwestergren

sebastianbaginski:
Why baking a linked list functionality into thread-related type ? Isn't it better to have a separate classes for `promise` and `collection of promises` instead of combining them two (you know, single class, single responsibility :) ) ?
Code would not get longer than your example:
```ooc
promises := PromiseCollection new()
for (i in 0 .. count)
    promises add(threadPool getPromise(f))
promises wait() . free()
```
With this "embedded list" approach you will have to implement from scratch every single function from a `vector / list` api (if ever needed) - counting promises in the chain, querying promise at given index etc. What is your motivation for this ? I understand that it comes from the old "operator +="- based approach ?

marcusnaslund:
From what I understood we only needed a way to wait and free on several promises. This solves that with a rather simple syntax, unfortunately the diff eventually got large because I renamed `wait` to `_wait`. Maybe there's a better way, but anyhow.

Sure - I can do a `SynchronizeVectorList extends VectorList<_Synchronize>` with an additional `waitAll` method or something like that instead. The thing is we *had* one of those, and it was never really used. And is there any reasonable probability that we will ever need more functionality on a chain of `Promise`s? Not saying your suggestion is bad, we won't have to change any existing code and it's certainly more future-proof, I just don't know how far into the future we need to speculate.

sebastianbaginski:
I was just curious about this choice. If both of you are happy with this, then I'm fine too.

marcusnaslund:
I'm fine either way, it wouldn't take many minutes to change it. @emilwestergren will be the one using this, so the decision is his.

emilwestergren:
They all work, but if we want the least amount of code then what about the simplest of all suggestions:
```ooc
promises := VectorList<Promise> new() //or array
for (i in 0 .. count)
    promises add(pool getPromise(f))
Promise wait(promises)
```

If the `PromiseCollection` class only contains a `wait` function then I vote for the above solution, however if we want to extend the functionality then it makes sense to have a separate class.

marcusnaslund:
I'm fine with that too. :+1:

marcusnaslund:
How about this? @emilwestergren

source/concurrent/Promise.ooc
```diff
@@ -36,6 +36,18 @@ _Synchronizer: abstract class {
 			result := this wait(TimeSpan maximumValue)
 		result
 	}
+	wait: static func ~many (items: VectorList<This>) -> Bool {
+		result := true
+		for (i in 0 .. items count)
+			result = result && items[i] wait()
+		result
+	}
+	wait: static func ~timedMany (items: VectorList<This>, time: TimeSpan) -> Bool {
```
emilwestergren: Can't you do 
```ooc
wait: static func ~timedMany (items: VectorList<This> time := TimeSpan maximumValue) -> Bool {
```
and remove the other function

marcusnaslund: Oh there was a `TimeSpan maximumValue`. I was going to suggest we implement i. Very good.

source/concurrent/Promise.ooc
```diff
@@ -36,6 +36,12 @@ _Synchronizer: abstract class {
 			result := this wait(TimeSpan maximumValue)
 		result
 	}
+	wait: static func ~timed (items: VectorList<This>, time := TimeSpan maximumValue) -> Bool {
+		result := true
+		for (i in 0 .. items count)
+			result = result && (time == TimeSpan maximumValue ? items[i] wait() : items[i] wait(time))
```
marcusnaslund: Add this check so we don't bypass the `debugDeadlock` version block in `wait~forever`.

PR #1923, opened 2016-09-21
Fixed toRaster for padded output buffers.

emilwestergren:
It is now possible to use glReadPixels to copy from texture to padded output buffer.

@davidpiuva peer review


davidpiuva:
Looks okay. Just need to fix what magic complained about.

PR #1924, opened 2016-09-21
Added support for RenderMode Cpu

christopherlagerhult:
Includes changes for ~~[link removed]~~ to work.
Review @emilwestergren (and maybe someone else too)

### DIFF
--- a/source/draw/DrawContext.ooc
+++ b/source/draw/DrawContext.ooc
@@ -67,4 +67,6 @@ DrawContext: abstract class {
 	getStringPixelBounds: static func (message: String, fontAtlas: Image) -> IntVector2D {
 		This getStringIndexBounds(message) * This getFontSize(fontAtlas)
 	}
+	preregister: virtual func (image: Image)
+	preallocate: virtual func (resolution: IntVector2D)
 }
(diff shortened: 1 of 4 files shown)


marcusnaslund:
Edited your post. Please don't link to (or mention) private repos from here - instead link from the other PR to this one. It's safer. :)

source/draw/gpu/opengl/backend/gles3/external/gles3.ooc
```diff
@@ -6,7 +6,6 @@
  * of the MIT license.  See the LICENSE file for details.
  */
 
-version(!gpuOff) {
```
emilwestergren: Do you really need to remove this? This kind of code we want inside `!gpuOff`

christopherlagerhult: Not if the "dependent" parts are wrapped inside version flags. I will reinstate those flags

christopherlagerhult: Updated

emilwestergren:
Much better :)

PR #1925, opened 2016-09-22
Add operator * for multiplying FloatTransform2D and 3D with each other.

bettinaselig:
@marcusnaslund Peer review?

### DIFF
--- a/source/geometry/FloatTransform3D.ooc
+++ b/source/geometry/FloatTransform3D.ooc
@@ -205,6 +205,10 @@ FloatTransform3D: cover {
 			this c * other x + this g * other y + this k * other z + this o
 		)
 	}
+	operator * (other: FloatTransform2D) -> This {
+		otherTransform3D := This new(other)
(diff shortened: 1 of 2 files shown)


marcusnaslund:
I think you can just write `new` without the `~` suffix?

source/geometry/FloatTransform3D.ooc
```diff
@@ -205,6 +205,10 @@ FloatTransform3D: cover {
 			this c * other x + this g * other y + this k * other z + this o
 		)
 	}
+	operator * (other: FloatTransform2D) -> This {
+		otherTransform3D := this new~fromFloatTransform2D(other)
```
marcusnaslund: Also, it should be `This` not `this`.

bettinaselig:
Updated.

marcusnaslund:
Looks good. You should probably make the pull request towards `secondaryFeature1.5`, though.

PR #1928, opened 2016-09-22
Override list functions

thomasfanell:
`rock 1.0.22` will prohibit the use of `super()` in shadowed functions (functions not marked as override).
@marcusnaslund


marcusnaslund:
Is there a way to turn that prohibition off? If we want to run older versions of the code base, I mean.

thomasfanell:
@marcusnaslund I actually thought of that, a command line switch will be included in the next rock release (1.0.22)

thomasfanell:
I might as well add it right away in this first batch of changes...

thomasfanell:
`rock --allow-super-when-shadowing` now disables it

marcusnaslund:
Perfect

PR #1930, opened 2016-09-23
Merged develop.

emilwestergren:
@marcusnaslund


marcusnaslund:
Some test failed. Can't see why from home

marcusnaslund:
And now it passed.

emilwestergren:
I think the test started before I pushed the corresponding branch in the other repo

marcusnaslund:
Ah that makes sense.

PR #1932, opened 2016-10-03
Fixed race condition in Gles3Context free

erikhagglund:
If free was called simultaneously for two contexts eglTerminate
might not have been called

review @emilwestergren

### DIFF
--- a/source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
+++ b/source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
@@ -30,12 +30,14 @@ Gles3Context: class extends GLContext {
 		status = eglDestroySurface(this _eglDisplay, this _eglSurface)
 		if (status != EGL_TRUE)
 			Debug print("eglDestroySurface failed with error code %d" format(status))
-		if (This _contextCount == 1) {
-			status = eglTerminate(this _eglDisplay)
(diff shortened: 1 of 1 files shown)


source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -30,12 +30,14 @@ Gles3Context: class extends GLContext {
...
 		if (status != EGL_TRUE)
 			Debug print("eglDestroySurface failed with error code %d" format(status))
-		if (This _contextCount == 1) {
-			status = eglTerminate(this _eglDisplay)
-			if (status != EGL_TRUE)
-				Debug print("eglTerminate failed with error code %d" format(status))
-		}
-		This _mutex with(|| This _contextCount -= 1)
+		This _mutex with(||
+			if (This _contextCount == 1) {
```
emilwestergren: This is an ugly hack to fix a driver bug where it crashed when `eglTerminate` was called several times, even though it should be allowed according to the specification: https://www.khronos.org/registry/egl/sdk/docs/man/html/eglTerminate.xhtml

> Terminating an already terminated EGL display connection has no effect. A terminated display may be re-initialized by calling eglInitialize again.

emilwestergren:
Actually should have been merged to `develop`

PR #1933, opened 2016-10-04
Added ConditionPromise.

emilwestergren:
@erikhagglund peer review

### DIFF
--- a/source/concurrent/Promise.ooc
+++ b/source/concurrent/Promise.ooc
@@ -61,6 +61,30 @@ ClosurePromise: class extends Promise {
 	wait: override func (time: TimeSpan) -> Bool { this _wait(time) }
 }
 
+ConditionPromise: class extends Promise {
+	_completed := false
+	_condition := WaitCondition new()
+	_mutex := Mutex new()
+	init: func { super() }
+	free: override func {
+		(this _mutex, this _condition) free()
+		super()
+	}
+	signal: func {
+		this _mutex with(||
+			this _completed = true
+			this _condition broadcast()
+		)
+	}
+	wait: override func (time: TimeSpan) -> Bool {
+		this _mutex lock()
+		if (!this _completed)
+			this _condition wait(this _mutex)
+		this _mutex unlock()
+		this _completed
+	}
+}
+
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Add test plz

emilwestergren:
Added test

source/concurrent/Promise.ooc
```diff
@@ -61,6 +61,30 @@ ClosurePromise: class extends Promise {
...
+		(this _mutex, this _condition) free()
+		super()
+	}
+	signal: func {
+		this _mutex with(||
+			this _completed = true
+			this _condition broadcast()
+		)
+	}
+	wait: override func (time: TimeSpan) -> Bool {
```
marcusnaslund: I like how this function takes a time argument but never uses it.

marcusnaslund:
...but since I locked you in the office yesterday, I have no comments.

source/concurrent/Promise.ooc
emilwestergren: Yeah, I guess I could do it like this instead:
```ooc
wait: override func (time: TimeSpan) -> Bool { Debug error("Timeout wait not supported for ConditionPromise") }
wait: override func -> Bool { ... }
```

marcusnaslund: That would make more sense, but it would override the `debugDeadlock` code. You decide.

emilwestergren: How about this:
```ooc
if (time != TimeSpan maximumValue)
   Debug error("Timed wait not supported...")
```

emilwestergren: That won't work with `debugDeadlock` though...

marcusnaslund: the timed wait should give an error if the argument is not `TimeSpan maximumValue`, but keep all the code you have now.
Then you can always call `wait()` without argument, and everyone is happy.

emilwestergren:
Updated

source/concurrent/Promise.ooc
marcusnaslund: Ah, true

PR #1939, opened 2016-10-06
Primary feature1.6

emilwestergren:


source/concurrent/Promise.ooc
```diff
@@ -51,6 +51,41 @@ Promise: abstract class extends _Synchronizer {
 	empty: static This { get { _EmptyPromise new() } }
 }
 
+ClosurePromise: class extends Promise {
+	_wait: Func (TimeSpan) -> Bool
+	init: func (=_wait)
```
emilwestergren: Missing `super()` here

PR #1954, opened 2016-10-19
fixed leaks related to exception context

sebastianbaginski:
I think @niklasborwell does not have so much time to look into this, but I guess this is what he used in elSystemGyro branch.

- global `exceptionContext` was not released
- `ExceptionContext` class was missing a proper free function

@marcusnaslund review

### DIFF
--- a/source/system/Exception.ooc
+++ b/source/system/Exception.ooc
@@ -38,6 +38,15 @@ ExceptionContext: class {
 		set (value) { this _currentException = value }
 	}
 	init: func
+	free: override func {
+		while (!this _stackFramesToDelete isEmpty)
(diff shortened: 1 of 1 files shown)


source/system/Exception.ooc
```diff
@@ -38,6 +38,15 @@ ExceptionContext: class {
 		set (value) { this _currentException = value }
 	}
 	init: func
+	free: override func {
+		while (! this _stackFramesToDelete isEmpty)
+			this _stackFramesToDelete pop() free()
+		while (! this _localFrames isEmpty)
```
sebastianbaginski: Fixed.

PR #1960, opened 2016-10-27
Added missing math operands

davidpiuva:
Useful when doing math on array dimensions and borders.

### DIFF
--- a/source/geometry/FloatPoint2D.ooc
+++ b/source/geometry/FloatPoint2D.ooc
@@ -65,6 +65,8 @@ FloatPoint2D: cover {
 	operator - (other: FloatVector2D) -> This { This new(this x - other x, this y - other y) }
 	operator * (other: FloatVector2D) -> This { This new(this x * other x, this y * other y) }
 	operator / (other: FloatVector2D) -> This { This new(this x / other x, this y / other y) }
+	operator + (other: Float) -> This { This new(this x + other, this y + other) }
+	operator - (other: Float) -> This { This new(this x - other, this y - other) }
 	operator * (other: Float) -> This { This new(this x * other, this y * other) }
 	operator / (other: Float) -> This { This new(this x / other, this y / other) }
 
--- a/source/geometry/FloatPoint3D.ooc
+++ b/source/geometry/FloatPoint3D.ooc
@@ -70,6 +70,8 @@ FloatPoint3D: cover {
 	operator - (other: FloatVector3D) -> This { This new(this x - other x, this y - other y, this z - other z) }
 	operator * (other: FloatVector3D) -> This { This new(this x * other x, this y * other y, this z * other z) }
 	operator / (other: FloatVector3D) -> This { This new(this x / other x, this y / other y, this z / other z) }
+	operator + (other: Float) -> This { This new(this x + other, this y + other, this z + other) }
+	operator - (other: Float) -> This { This new(this x - other, this y - other, this z - other) }
 	operator * (other: Float) -> This { This new(this x * other, this y * other, this z * other) }
 	operator / (other: Float) -> This { This new(this x / other, this y / other, this z / other) }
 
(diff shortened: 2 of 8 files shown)


davidpiuva:
@christopherlagerhult Peer review?

davidpiuva:
We use vectors as size for images and matrices. Adding 1 to a size adds a border of one row and column.

marcusnaslund:
I have the same reservations as @christopherlagerhult ... I understand this may sometimes be *useful*, but vector + scalar doesn't make sense mathematically the same way vector * scalar does.

I'd like some input @sebastianbaginski @tomasmahlberg @bettinaselig

marcusnaslund:
(Nothing wrong with this PR, just taking the green flag off until we have a decision)

sebastianbaginski:
It looks useful for sure. I agree that maybe this is not mathematically correct, but neither is `0.1+0.1+0.1 != 0.3` and yet we can live with it ;)
I've looked into documentation of some existing "math" libraries / packages (OpenCV, GMTL, Eigen, Lapack, Qt) and I couldn't find one that allows adding a scalar to a vector directly. Which of course doesn't mean that we shouldn't do it.
Let's look how would it look like:
```ooc
doSomething: func (vec: IntVector2D) -> IntVector2D {
    result := vec + 7
    ...
    result
}
```
Hmm could be in fact confusing...

Maybe explicit functions for this ?
```ooc
doSomething: func (vec: IntVector2D) -> IntVector2D {
    result := vec addComponentWise(7) // addCWise ?
    ...
    result
}
```
Not a fan myself though.

Can we define these functions to take tuples instead ? I think this could make it quite clear:
```ooc
doSomething: func (vec: IntVector2D) -> IntVector2D {
    result := vec + (7,7)
    ...
    result
}
```

davidpiuva:
Both GLSL and HLSL have it built in since it is supported in hardware. `r0.xyzw = r1.xyzw + r2.xxxx`

sebastianbaginski:
Yes but r2.xxxx is a vec4 type, you can't do `vec4 y = r2.xxxx + 1.0f;`.

sebastianbaginski:
OK looks like actually you can, I've had invalid types :D

davidpiuva:
Scalars are automatically packed into lanes of vector registers when you write 1.0 in GLSL or HLSL.
HLSL:
float MyScalar = 1.0f;
float4 MyResult = MyVector + MyScalar;
GLSL:
float MyScalar = 1.0f;
vec4 MyResult = MyVector + MyScalar;
Instruction:
r0.xyzw = r1.xyzw + r2.xxxx

sebastianbaginski:
Yes it works, it didn't compile for me as I used float instead of double and got a syntax error from the compiler.

bettinaselig:
I feel the same as @christopherlagerhult and @marcusnaslund. Just because we use vectors as size, does not mean that we need to add weird things to the vector class. 
If you want to increase the size of a box I think it is acceptable to write `size + IntVector2D new(1,1)`. We aim for readable code. If you find somewhere `size + 1`, it is not completely intuitive what it means. It produces overhead because you need to look into the vector class to see what it it actually does.

bettinaselig:
If you want to have this functionality I prefer an additional function like `addComponentWise()` as @sebastianbaginski suggested.

davidpiuva:
To GPU programmers, it is just a well defined standard addition.

davidpiuva:
I just leave the totally broken API then.

sebastianbaginski:
Why not `vector + (1,1)` then ? I like the functionality proposed in this PR.

davidpiuva:
The initial problem is when you write a formula and suddenly, the operation you requested does not exist and you always have to look up in the documentation to find out which addition is supported since it is very random. Like having a large hole in the road that distracts from the initial problem.

PR #1962, opened 2016-10-28
Added target transform in DrawState

davidpiuva:
The math is quite messy but it looks correct because a projection matrix does not set the w axis like a regular translation point. Translation in 3D is usually done by multiplying the constant one with x, y, z in the w axis so that the new translation is multiplied by one and simply added to the previous translation. The projection matrix does not have that constant one, so translation is instead applied to the tilt of the Z axis expressed in normalized space.

### DIFF
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -150,11 +150,19 @@ DrawState: cover {
 		this
 	}
 	getTransformNormalized: func -> FloatTransform3D { this _transformNormalized }
-	setFlatTransform: func (flatTransform: FloatTransform2D) -> This {
+	setFlatTransformNormalized: func (flatTransform: FloatTransform2D) -> This {
 		this flatTransform = flatTransform
 		this
 	}
-	getFlatTransform: func -> FloatTransform2D { this flatTransform }
+	setFlatTransformReference: func ~ExplicitIntSize (flatTransform: FloatTransform2D, imageSize: IntVector2D) -> This {
+		this flatTransform = flatTransform referenceToNormalized(imageSize)
+		this
+	}
+	setFlatTransformReference: func ~ExplicitFloatSize (flatTransform: FloatTransform2D, imageSize: FloatVector2D) -> This {
+		this flatTransform = flatTransform referenceToNormalized(imageSize)
+		this
+	}
+	getFlatTransformNormalized: func -> FloatTransform2D { this flatTransform }
 	setOrigin: func (origin: FloatPoint2D) -> This {
 		this _originReference = origin
 		this
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -36,6 +36,7 @@ DrawState: cover {
 	_destinationNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_focalLengthNormalized := 0.0f // Relative to image width
+	flatTransform := FloatTransform2D identity // 2D scaling done after projection matrix but before w division
 	init: func@ ~default
 	init: func@ ~target (=target)
 	setTarget: func (target: Image) -> This {
@@ -149,6 +150,11 @@ DrawState: cover {
 		this
 	}
 	getTransformNormalized: func -> FloatTransform3D { this _transformNormalized }
+	setFlatTransform: func (flatTransform: FloatTransform2D) -> This {
+		this flatTransform = flatTransform
+		this
+	}
+	getFlatTransform: func -> FloatTransform2D { this flatTransform }
 	setOrigin: func (origin: FloatPoint2D) -> This {
 		this _originReference = origin
 		this
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -36,7 +36,7 @@ DrawState: cover {
 	_destinationNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_focalLengthNormalized := 0.0f // Relative to image width
-	flatTransform := FloatTransform2D identity // 2D scaling done after projection matrix but before w division
+	targetTransform := FloatTransform2D identity // 2D scaling done after projection matrix but before w division
 	init: func@ ~default
 	init: func@ ~target (=target)
 	setTarget: func (target: Image) -> This {
@@ -150,19 +150,19 @@ DrawState: cover {
 		this
 	}
 	getTransformNormalized: func -> FloatTransform3D { this _transformNormalized }
-	setFlatTransformNormalized: func (flatTransform: FloatTransform2D) -> This {
-		this flatTransform = flatTransform
+	setTargetTransformNormalized: func (targetTransform: FloatTransform2D) -> This {
+		this targetTransform = targetTransform
 		this
 	}
-	setFlatTransformReference: func ~ExplicitIntSize (flatTransform: FloatTransform2D, imageSize: IntVector2D) -> This {
-		this flatTransform = flatTransform referenceToNormalized(imageSize)
+	setTargetTransformReference: func ~ExplicitIntSize (targetTransform: FloatTransform2D, imageSize: IntVector2D) -> This {
+		this targetTransform = targetTransform referenceToNormalized(imageSize)
 		this
 	}
-	setFlatTransformReference: func ~ExplicitFloatSize (flatTransform: FloatTransform2D, imageSize: FloatVector2D) -> This {
-		this flatTransform = flatTransform referenceToNormalized(imageSize)
+	setTargetTransformReference: func ~ExplicitFloatSize (targetTransform: FloatTransform2D, imageSize: FloatVector2D) -> This {
+		this targetTransform = targetTransform referenceToNormalized(imageSize)
 		this
 	}
-	getFlatTransformNormalized: func -> FloatTransform2D { this flatTransform }
+	getTargetTransformNormalized: func -> FloatTransform2D { this targetTransform }
 	setOrigin: func (origin: FloatPoint2D) -> This {
 		this _originReference = origin
 		this
--- a/source/draw/README.md
+++ b/source/draw/README.md
@@ -17,8 +17,8 @@ The destination is similar to the viewport by defining where to draw. The differ
 ## Transform
 Transform is the inverse camera matrix applied after the destination scaling.
 
-## FlatTransform
-FlatTransform can be used to translate and scale in image space. The difference from translating in camera space is that no parallax is applied because vertices at all depths moves equally in two dimensions. The center point of FlatTransform defines the center of perspective where no geometry is stretched from the planar projection.
+## TargetTransform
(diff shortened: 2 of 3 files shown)


davidpiuva:
@emilwestergren Peer review?

source/draw/gpu/opengl/OpenGLPacked.ooc
```diff
@@ -84,7 +85,7 @@ OpenGLPacked: abstract class extends GpuImage {
 					gpuMap add("texture0", image)
 			}
 		}
-		gpuMap useProgram(this, projection * view * model, textureTransform)
+		gpuMap useProgram(this, flat * projection * view * model, textureTransform)
```
davidpiuva: Applying a translation to a camera space point would translate in 3D with parallax.
Applying a translation to a projection matrix will not translate it, just tilt the axis system.
Others have tried this before: http://stackoverflow.com/questions/28018458/how-to-translate-the-projected-object-in-screen-in-opengl

emilwestergren: But this will not work because it's before the perspective division?

davidpiuva: These matrices confirm that the assumption works for basic cropping and translation in normalized coordinates.
```
projection =
1.62414968, 0.00000000, 0.00000000, 0.00000000
0.00000000, -2.88737726, 0.00000000, 0.00000000
0.00000000, 0.00000000, 1.00000155, 0.02000002
0.00000000, 0.00000000, -1.00000000, 0.00000000

flat =
2.500000, 0.000000, 0.000000
0.000000, 2.500000, 0.000000
-0.781250, -0.925926, 1.000000	

flat * projection =
4.06037426, 0.00000000, 0.78125000, 0.00000000
0.00000000, -7.21844292, 0.92592591, 0.00000000
0.00000000, 0.00000000, 1.00000155, 0.02000002
0.00000000, 0.00000000, -1.00000000, 0.00000000
```

davidpiuva: That is why we use normalized coordinates from -1 to +1 for axis tilt coefficients.

davidpiuva: We can test some prototypes before making a simpler interface for translation.

emilwestergren: And how do you calculate these normalized coordinates?

davidpiuva: Since reference coordinates goes from -half size to +half size and they need to go from -1 to +1, we simply divide by (inputSize * 0.5). If the offset is in read coordinates then we take the inverse of the read transfrom as the write transform.

davidpiuva:
Tested with hard rotations and the result is the same as when using a viewport to translate in 2D.

source/draw/DrawState.ooc
```diff
@@ -149,6 +150,20 @@ DrawState: cover {
...
+		this
+	}
+	setFlatTransformReference: func ~ExplicitIntSize (flatTransform: FloatTransform2D, imageSize: IntVector2D) -> This {
+		this flatTransform = flatTransform referenceToNormalized(imageSize)
+		this
+	}
+	setFlatTransformReference: func ~ExplicitFloatSize (flatTransform: FloatTransform2D, imageSize: FloatVector2D) -> This {
+		this flatTransform = flatTransform referenceToNormalized(imageSize)
+		this
+	}
```
davidpiuva: Updated

davidpiuva:
All checks passed.

source/draw/DrawState.ooc
```diff
@@ -36,6 +36,7 @@ DrawState: cover {
 	_destinationNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_focalLengthNormalized := 0.0f // Relative to image width
+	flatTransform := FloatTransform2D identity // 2D scaling done after projection matrix but before w division
```
emilwestergren: Not sure if this is a good name, I would have no idea of what a "flat transform" is. `projectedTransform/imageTransform` ?

source/draw/DrawState.ooc
```diff
@@ -149,6 +150,20 @@ DrawState: cover {
...
+	}
+	setFlatTransformReference: func ~ExplicitIntSize (flatTransform: FloatTransform2D, imageSize: IntVector2D) -> This {
+		this flatTransform = flatTransform referenceToNormalized(imageSize)
+		this
+	}
+	setFlatTransformReference: func ~ExplicitFloatSize (flatTransform: FloatTransform2D, imageSize: FloatVector2D) -> This {
+		this flatTransform = flatTransform referenceToNormalized(imageSize)
+		this
+	}
+	//meshTransform flatTransform referenceToNormalized(inputImage size)
```
emilwestergren: What's this?

source/draw/README.md
```diff
@@ -17,6 +17,11 @@ The destination is similar to the viewport by defining where to draw. The differ
 ## Transform
 Transform is the inverse camera matrix applied after the destination scaling.
 
+## FlatTransform
+FlatTransform is applied after the projection in 2D.
```
emilwestergren: I don't really like this description, it only complicates things. I would prefer something like this:
> FlatTransform is a 2D transform applied after the projection matrix, i.e. in the projected coordinate system. This can be useful when 2D-operations like translation and cropping should be applied as a final step on the rendered image.

davidpiuva: We want to make things abstract so it should not be mixed with the projection matrix. It is a 2D operation in its own 2D space.

source/draw/DrawState.ooc
davidpiuva: I cannot think of a better name for now.
Using 3D terms like projection to describe a 2D operation does not sound abstract.
ImageTransform was used before to describe the source coordinates.

source/draw/README.md
```diff
@@ -17,6 +17,11 @@ The destination is similar to the viewport by defining where to draw. The differ
...
+It works without parallax by tilting the Z axis in the projection matrix.
+This is done by multiplying the projection matrix with a normalized translation which will skew instead of translate the projection matrix.
```
davidpiuva: I had to add this only to make sure that no one takes back the viewport hack believing that the parallax problem still exists.

source/draw/README.md
```diff
@@ -17,6 +17,11 @@ The destination is similar to the viewport by defining where to draw. The differ
 ## Transform
 Transform is the inverse camera matrix applied after the destination scaling.
 
+## FlatTransform
+FlatTransform can be used to translate and scale in image space. The difference from translating in camera space is that no parallax is applied because vertices at all depths moves equally in two dimensions. The center point of FlatTransform defines the center of perspective where no geometry is stretched from the planar projection.
+
+* This works as if the projection was applied after depth division by multiplying the projection matrix with a normalized translation. If we want to add a constant to something that will later be divided by depth then we simply add the constant multiplied by depth. For a perspective projection matrix, the translation will be multiplied by -Z and thereby tilt the Z axis. For an orthogonal projection matrix, the translation will be multiplied by 1 as usual.
```
davidpiuva: Now I have divided between the abstract and the in depth explanation.

source/draw/DrawState.ooc
emilwestergren: `postTransform`?

davidpiuva: That name is already taken.

source/draw/README.md
emilwestergren: orthographic projection matrix*

source/draw/DrawState.ooc
emilwestergren: Or this can be implicit by having overloaded `setTransform(transform: FloatTransform2D)` and `setTransform(transform: FloatTransform3D` ?

davidpiuva: Setting different variables with the same name leads to confusion since we might use both at the same time and then someone erases the first assignment believing that it is dead code.

source/draw/DrawState.ooc
```diff
@@ -36,6 +36,7 @@ DrawState: cover {
 	_destinationNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_focalLengthNormalized := 0.0f // Relative to image width
+	targetTransform := FloatTransform2D identity // 2D scaling done after projection matrix but before w division
```
davidpiuva: Updated

davidpiuva:

source/draw/README.md
```diff
@@ -17,6 +17,11 @@ The destination is similar to the viewport by defining where to draw. The differ
 ## Transform
 Transform is the inverse camera matrix applied after the destination scaling.
 
+## TargetTransform
+TargetTransform can be used to translate and scale in image space. The difference from translating in camera space is that no parallax is applied because vertices at all depths moves equally in two dimensions. The center point of TargetTransform defines the center of perspective where no geometry is stretched from the planar projection.
+
+* This works as if the projection was applied after depth division by multiplying the projection matrix with a normalized translation. If we want to add a constant to something that will later be divided by depth then we simply add the constant multiplied by depth. For a perspective projection matrix, the translation will be multiplied by -Z and thereby tilt the Z axis. For an orthographic projection matrix, the translation will be multiplied by 1 as usual.
```
marcusnaslund: I think it looks better if you remove the `* ` and just let this be a second paragraph.

source/draw/README.md
```diff
@@ -17,6 +17,11 @@ The destination is similar to the viewport by defining where to draw. The differ
 ## Transform
 Transform is the inverse camera matrix applied after the destination scaling.
 
+## TargetTransform
+TargetTransform can be used to translate and scale in image space. The difference from translating in camera space is that no parallax is applied because vertices at all depths moves equally in two dimensions. The center point of TargetTransform defines the center of perspective where no geometry is stretched from the planar projection.
```
marcusnaslund: "vertices at all depths *move*"

source/draw/gpu/opengl/OpenGLPacked.ooc
marcusnaslund: Since the PR is green I assume this discussion has been resolved?

davidpiuva: Yes

source/draw/README.md
```diff
@@ -17,6 +17,11 @@ The destination is similar to the viewport by defining where to draw. The differ
 ## Transform
 Transform is the inverse camera matrix applied after the destination scaling.
 
+## TargetTransform
+TargetTransform can be used to translate and scale in image space. The difference from translating in camera space is that no parallax is applied because vertices at all depths move equally in two dimensions. The center point of TargetTransform defines the center of perspective where no geometry is stretched from the planar projection.
+
+This works as if the projection was applied after depth division by multiplying the projection matrix with a normalized translation. If we want to add a constant to something that will later be divided by depth then we simply add the constant multiplied by depth. For a perspective projection matrix, the translation will be multiplied by -Z and thereby tilt the Z axis. For an orthographic projection matrix, the translation will be multiplied by 1 as usual.
```
davidpiuva: A second paragraph now.

PR #1966, opened 2016-11-01
Fixed memory leak of triangle index buffer

davidpiuva:


davidpiuva:
@emilwestergren Peer review?

emilwestergren:
Who can be so sloppy and miss something like this? :angry:

PR #1971, opened 2016-11-03
Runtime GPU check

davidpiuva:
Instead of not being able to mention a GPU type without getting a compiler error in -DgpuOff, it is now possible to mention the generic GPU types in -DgpuOff as long as no GpuContext is being created and no internal OpenGL types are mentioned.

When no GpuContext can be created without a GPU, no GpuImage can be created without a GPU because a context is needed for creation.


davidpiuva:
@emilwestergren Peer review?

PR #1974, opened 2016-11-07
Fixed EGLYuv recycling

erikhagglund:
Recycling was previously done on the native buffer window instead of the buffer handle. This caused the recyle bin to get full immediately

review and verify I am not doing something stupid now, i guess there was some reason for using the nativeBuffer to begin with but I cannot remember the details though @emilwestergren

### DIFF
--- a/source/draw/gpu/opengl/EGLYuv.ooc
+++ b/source/draw/gpu/opengl/EGLYuv.ooc
@@ -15,9 +15,10 @@ import backend/[GLTexture, EGLImage]
 
 version(!gpuOff) {
 EGLYuv: class extends OpenGLPacked {
-	_nativeBuffer: Pointer
-	nativeBuffer ::= this _nativeBuffer
+	_handle: Pointer
+	handle ::= this _handle
 	init: func (_buffer: GraphicBuffer, context: OpenGLContext) {
+		_handle = _buffer handle
 		super(EGLImage create(TextureType External, _buffer size, _buffer nativeBuffer, context backend), 3, context)
 	}
 	toRasterDefault: override func -> RasterImage { Debug error("toRasterDefault unimplemented for EGLYuv"); null }
--- a/source/draw/gpu/opengl/EGLYuv.ooc
+++ b/source/draw/gpu/opengl/EGLYuv.ooc
@@ -17,11 +17,11 @@ version(!gpuOff) {
 EGLYuv: class extends OpenGLPacked {
 	_handle: Pointer
 	handle ::= this _handle
-	init: func (_buffer: GraphicBuffer, context: OpenGLContext) {
-		_handle = _buffer handle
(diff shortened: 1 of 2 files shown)


emilwestergren:
I can't remember why we would use `nativeBuffer`, as long as you have tested this and it fixed a problem it seems fine.

source/draw/gpu/opengl/EGLYuv.ooc
```diff
@@ -15,9 +15,10 @@ import backend/[GLTexture, EGLImage]
 
 version(!gpuOff) {
 EGLYuv: class extends OpenGLPacked {
-	_nativeBuffer: Pointer
-	nativeBuffer ::= this _nativeBuffer
+	_handle: Pointer
+	handle ::= this _handle
 	init: func (_buffer: GraphicBuffer, context: OpenGLContext) {
+		_handle = _buffer handle
```
marcusnaslund: Missing `this`

marcusnaslund: Also, you don't need the `_` for `buffer` anymore

erikhagglund:
updated

source/draw/gpu/opengl/EGLYuv.ooc
```diff
@@ -15,12 +15,13 @@ import backend/[GLTexture, EGLImage]
...
-		super(EGLImage create(TextureType External, _buffer size, _buffer nativeBuffer, context backend), 3, context)
+	_handle: Pointer
+	handle ::= this _handle
+	init: func (buffer: GraphicBuffer, context: OpenGLContext) {
+		this _handle = buffer handle
+		super(EGLImage create(TextureType External, buffer size, buffer nativeBuffer, context backend), 3, context)
 	}
 	toRasterDefault: override func -> RasterImage { Debug error("toRasterDefault unimplemented for EGLYuv"); null }
-	toRasterDefault: override func ~target (target: RasterImage) { Debug error("toRasterDefault~target unimplemented for EGLYuv"); null }
+	toRasterDefault: override func ~target (target: RasterImage) { Debug error("toRasterDefault~target unimplemented for EGLYuv")}
```
marcusnaslund: Magic complains about missing space between `)` and `}`

erikhagglund: haha, my review script complained so I fixed it but forgot to commit

erikhagglund:
fixed Magic error

source/draw/gpu/opengl/EGLYuv.ooc
marcusnaslund: > I fixed it but forgot to commit

Yeah yeah, I've heard that one before, "my dog ate my homework" etc. etc. :)

PR #1977, opened 2016-11-09
Set pointer to array data to null after free

thomasfanell:
@sebastianbaginski

### DIFF
--- a/source/system/Array.h
+++ b/source/system/Array.h
@@ -20,7 +20,7 @@
     (((type*) array.data)[index] = value)
 #endif
 
-#define _lang_array__Array_free(array) { array_free(array.data); }
+#define _lang_array__Array_free(array) { array_free(array.data); array.data = NULL; array.length = 0; }
 
 typedef struct {
     size_t length;


sebastianbaginski:
Do you want to have this always or just in the "safe" version ? Do we clear the memory of other objects after freeing them ? What's the motivation here ?

marcusnaslund:
> Do we clear the memory of other objects after freeing them

We do, especially in the other project (`if (object) { object free(); object = null }`).

sebastianbaginski:
Sure, but I mean automatically like this. Don't get me wrong, I think it's ok.

thomasfanell:
Why would we ever want dangling pointers to array data?

sebastianbaginski:
I mean from the user's point of view there is no difference between
```
object := Object new()
object free() // garbage now
object free() // crash

array := Int[10] new()
array free() // set to null
array free() // no crash
```

But I get it now after your "private" argument :)

marcusnaslund:
I'm sorry to say this broke the android build with `-DfastCv` flag. :(
I don't know if @sebastianbaginski has an easy fix, otherwise I will revert this before today's code freeze and we'll make a new shot it next week.

thomasfanell:
How is that even possible?

thomasfanell:
Let's revert and look at it... better to be safe than sorry

PR #1981, opened 2016-11-10
Revert "Set pointer to array data to null after free"

marcusnaslund:
Reverts magic-lang/ooc-kean#1977

Better safe than sorry.

@thomasfanell @sebastianbaginski

### DIFF
--- a/source/system/Array.h
+++ b/source/system/Array.h
@@ -20,7 +20,7 @@
     (((type*) array.data)[index] = value)
 #endif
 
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
Can we hold with this until 12.00 ? Maybe we can fix it in the other repo ?

marcusnaslund:
Even if we do there may be other build configs which fail from this, so I choose to revert this anyway. We'll have more time to test next week.
There's nothing critical about this until next rock version, so we can wait.

PR #1982, opened 2016-11-10
VectorList resizes by *2

### LINKED ISSUE #1980: VectorList resizes with a constant increment each time
From `VectorList.ooc` (two places):

```ooc
if (this _vector capacity <= this _count)
	this _vector resize(this _vector capacity + 8)
```

We should investigate what the reason is for this and if we should change `+8` to something more common, like `*1.5` or `*2` (https://en.wikipedia.org/wiki/Dynamic_array#Growth_factor) if it doesn't have any performance impacts on other projects.

marcusnaslund:
Fixes #1980 

We have to enforce a positive size in the constructor, the other project is creating a 0 size list sometimes. That's probably a bug and should be fixed, but this seems like a smart fix anyway.

Otherwise seems to work well, we make a lot fewer resizes and `test` runtimes are not affected.

### DIFF
--- a/source/system/structs/Vector.ooc
+++ b/source/system/structs/Vector.ooc
@@ -12,9 +12,7 @@ Vector: abstract class <T> {
 	_freeContent: Bool
 	capacity ::= this _capacity
 	init: func ~preallocated (=_backend, =_capacity, freeContent := true)
-	init: func (=_capacity, freeContent := true) {
-		this _freeContent = freeContent
-	}
+	init: func (=_capacity, freeContent := true) { this _freeContent = freeContent }
 	free: override func {
 		memfree(this _backend)
 		super()
--- a/source/system/structs/Vector.ooc
+++ b/source/system/structs/Vector.ooc
@@ -33,6 +33,9 @@ Vector: abstract class <T> {
 		else if (capacity > this capacity)
 			this _capacity = capacity
 	}
+	grow: func {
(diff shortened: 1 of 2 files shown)


marcusnaslund:
@sebastianbaginski ?

source/system/structs/VectorList.ooc
```diff
@@ -23,7 +23,7 @@ VectorList: class <T> extends List<T> {
 	}
 	add: override func (item: T) {
 		if (this _vector capacity <= this _count)
-			this _vector resize(this _vector capacity + 8)
+			this _vector resize(this _vector capacity * 2)
```
sebastianbaginski: `_grow: func { this resize(this capacity * 2) }` ?

sebastianbaginski:
Aren't you afraid the exponential growth ? It always scares me.

marcusnaslund:
Large numbers don't scare me.

source/system/structs/Vector.ooc
```diff
@@ -35,6 +33,9 @@ Vector: abstract class <T> {
 		else if (capacity > this capacity)
 			this _capacity = capacity
 	}
+	grow: func {
+		this resize(this capacity < 1024 ? this capacity * 2 : this capacity + 1024)
```
marcusnaslund: Hopefully this scares you a little less @sebastianbaginski

PR #1993, opened 2016-11-15
Free static array in GlobalCleanup

thomasfanell:
This particular array cause no memory leak, but is visible in the `still reachable` section when running valgrind. @sebastianbaginski


sebastianbaginski:
:+1:

PR #1995, opened 2016-11-15
More sense among DebugLevels

marcusnaslund:
- Changed states to mimic LogCat types and order.
- Changed `print` to print out everything with *at least* the set level, not just the current level. Except `Silent`.
- `error()` prints message with `Fatal` level

A later PR will allow `test.sh` (and maybe `build.sh` in other projects?) to select which level is default - which will help clean up debug output from tests. Default could be `Error` and the nightly test can run with a lower level.

Peer review @sebastianbaginski @erikhagglund

### DIFF
--- a/source/base/Debug.ooc
+++ b/source/base/Debug.ooc
@@ -9,35 +9,39 @@
 import io/FileWriter
 
 DebugLevel: enum {
-	Everything
+	Verbose
 	Debug
-	Notification
+	Info
 	Warning
-	Recoverable
-	Message
-	Critical
+	Error
+	Fatal
+	Silent
 }
 
 Debug: class {
-	_level: static DebugLevel = DebugLevel Everything
+	_level: static DebugLevel = DebugLevel Verbose
+	level: static DebugLevel {
+		get { This _level }
+		set (value) { This _level = value }
+	}
 	_printFunction: static Func (String) = func (s: String) { println(s) }
 	initialize: static func (f: Func (String)) {
 		(This _printFunction as Closure) free()
 		This _printFunction = f
 	}
-	print: static func (string: String, level := DebugLevel Everything) {
-		if (This _level == level || This _level == DebugLevel Everything)
+	print: static func (string: String, level := DebugLevel Debug) {
+		if (level as Int >= This _level as Int && This _level != DebugLevel Silent)
 			This _printFunction(string)
 	}
-	print: static func ~free (string: String, level := DebugLevel Everything) {
+	print: static func ~free (string: String, level := DebugLevel Debug) {
 		This print(string, level)
 		string free()
 	}
 	error: static func (message: String, origin: Class = null) {
 		if (origin)
-			This print~free("%s: %s" format(origin name toCString(), message))
+			This print~free("%s: %s" format(origin name toCString(), message), DebugLevel Fatal)
 		else
-			This print(message)
+			This print(message, DebugLevel Fatal)
 		raise(message, origin)
 	}
 	error: static func ~assert (condition: Bool, message: String, origin: Class = null) {
--- a/test/base/DebugTest.ooc
+++ b/test/base/DebugTest.ooc
@@ -11,6 +11,7 @@ use unit
 
 DebugTest: class extends Fixture {
 	outputString: String = null
+	oldLevel: DebugLevel
 
 	init: func {
 		super("Debug")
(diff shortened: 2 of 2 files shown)


source/base/Debug.ooc
```diff
@@ -35,7 +33,7 @@ Debug: class {
 	}
 	error: static func (message: String, origin: Class = null) {
 		if (origin)
-			This print~free("%s: %s" format(origin name toCString(), message))
+			This print~free("%s: %s" format(origin name toCString(), message), DebugLevel Critical)
 		else
 			This print(message)
```
marcusnaslund: oh right

marcusnaslund:
Please come with input @emilwestergren @niklasborwell @tomasmahlberg @christopherlagerhult

source/base/Debug.ooc
```diff
@@ -10,11 +10,9 @@ import io/FileWriter
 
 DebugLevel: enum {
 	Everything
+	Message
 	Debug
-	Notification
 	Warning
```
christopherlagerhult: To many levels is just confusing in my opinion, hence, I put my vote on not having "Error". It should be possible to place such stuff in either warning or critical.

marcusnaslund: Or rename `Critical` to `Error`.

marcusnaslund:
As a reference, the LogCat levels are:

V — Verbose (lowest priority)
D — Debug
I — Info
W — Warning
E — Error
F — Fatal
S — Silent (highest priority, on which nothing is ever printed)

Maybe we should mimic that one.

sebastianbaginski:
I vote for having some level which turns off every single printout.

sebastianbaginski:
For reference, this is how it's in Qt:
```c++
QtDebugMsg	- A message generated by the qDebug() function.
QtInfoMsg	- A message generated by the qInfo() function.
QtWarningMsg	 - A message generated by the qWarning() function.
QtCriticalMsg	 - A message generated by the qCritical() function.
QtFatalMsg	- A message generated by the qFatal() function.
QtSystemMsg = QtCriticalMsg
```
levels are not explicitly exposed in the API, user needs to use one of the predefined functions listed above (`qDebug() << "message"; qFatal() << "message"; etc.)
However, users are allowed to overwrite default message handler by calling `qtInstallMessageHandler(mycallback)` with the error level type, context and message values passed into the callback.

christopherlagerhult:
> Maybe we should mimic that one.

Accually it makes sense to mimic the levels since most of us are used to them. +1 for a silent mode

niklasborwell:
> Accually it makes sense to mimic the levels since most of us are used to them. +1 for a silent mode
I agree!

marcusnaslund:
Updated to reflect LogCat.

source/base/Debug.ooc
```diff
@@ -9,35 +9,39 @@
...
+	}
 	_printFunction: static Func (String) = func (s: String) { println(s) }
 	initialize: static func (f: Func (String)) {
 		(This _printFunction as Closure) free()
 		This _printFunction = f
 	}
-	print: static func (string: String, level := DebugLevel Everything) {
-		if (This _level == level || This _level == DebugLevel Everything)
+	print: static func (string: String, level := DebugLevel Debug) {
+		if (level as Int >= This _level as Int && This _level != DebugLevel Silent)
```
christopherlagerhult: Is the last condition needed? (`This _level != DebugLevel Silent`)

sebastianbaginski: Can we have a `version(!noDebugOutput)` (or alike) around the print callback call ? I think it could be nice to disable every printout with a single build switch.

marcusnaslund: That would happen if you set `Debug level = DebugLevel Silent`

marcusnaslund: As I understood it, *nothing* is printed on `Silent`, not even messages tagged with `Silent`. Is that incorrect?

sebastianbaginski: Yes, but I mean to have a compile time solution, without changing anything in the code.

marcusnaslund: > compile time solution,

I was thinking of sending a debug level flag to the test/build scripts. Maybe I can do it with version flags.

christopherlagerhult: I guess. Although, it seems super awkward adding `Debug print(msg, DebugLevel Silent)` in the first place.

marcusnaslund:
Is this OK for now? There will be another PR to actually do something with this. @christopherlagerhult ?

source/base/Debug.ooc
marcusnaslund: It would, but should still check.

christopherlagerhult:
I'm super happy if everyone else is @emilwestergren @niklasborwell @tomasmahlberg @erikhagglund

emilwestergren:
Sure

sebastianbaginski:
So we're doing that compile-time disabling of printouts in a separate PR ?

marcusnaslund:
Yeah

PR #1996, opened 2016-11-16
Made clip plane settings for DrawState

davidpiuva:
Allows changing clip planes for special cases.

### DIFF
--- a/source/draw/DrawState.ooc
+++ b/source/draw/DrawState.ooc
@@ -38,6 +38,8 @@ DrawState: cover {
 	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_focalLengthNormalized := 0.0f // Relative to image width
 	targetTransform := FloatTransform2D identity // 2D scaling done after projection matrix but before w division
+	nearPlane: Float = 0.01f
+	farPlane: Float = 12500.0f
 	init: func@ ~default
 	init: func@ ~target (=target)
 	setTarget: func (target: Image) -> This {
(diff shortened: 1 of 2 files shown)


davidpiuva:
@christopherlagerhult Peer review?

source/draw/DrawState.ooc
```diff
@@ -38,6 +38,8 @@ DrawState: cover {
 	_sourceNormalized := FloatBox2D new(0.0f, 0.0f, 1.0f, 1.0f)
 	_focalLengthNormalized := 0.0f // Relative to image width
 	targetTransform := FloatTransform2D identity // 2D scaling done after projection matrix but before w division
+	nearPlane: Float = 0.01f
+	farPlane: Float = 12500.0f
```
christopherlagerhult: Just a though. Do you really need to initialize these variables if you're going to have them defaulted in `_createProjection` with the same value?

christopherlagerhult:
Other then the comment, I'm happy

source/draw/DrawState.ooc
davidpiuva: If not set, the uninitialized data will be given to `_createProjection` and override the default values with uninitialized values (possibly NaN causing a once in a million crash).

christopherlagerhult: Yeah but I was thinking more in term of that `_createProjection` could be called w/o those parameters then. Never mind though, just a thought.

marcusnaslund: But maybe you want to call `_createProjection` with custom argument values?

source/draw/gpu/opengl/OpenGLPacked.ooc
```diff
@@ -115,14 +115,12 @@ OpenGLPacked: abstract class extends GpuImage {
 	_createView: func (targetSize: FloatVector2D, normalizedView: FloatTransform3D) -> FloatTransform3D {
 		this _toLocal * normalizedView normalizedToReference(targetSize) * this _toLocal
 	}
-	_createProjection: func (targetSize: FloatVector2D, focalLengthPerWidth: Float) -> FloatTransform3D {
+	_createProjection: func (targetSize: FloatVector2D, focalLengthPerWidth, nearPlane, farPlane: Float) -> FloatTransform3D {
```
davidpiuva: Updated

PR #1997, opened 2016-11-17
Adapt to rock 1.0.22

thomasfanell:
Regarding **`GlobalCleanup unload()`**
Since we want to support unload/reload of modules and `Memory` must be unloaded last, rock will not emit calls to `Memory_unload()`, so we must call it manually. `GlobalCleanup unload()` should be the last instruction before we enter the de-initialized state.

@sebastianbaginski review

### DIFF
--- a/source/system/string/string_literal.c
+++ b/source/system/string/string_literal.c
@@ -7,9 +7,10 @@
 #if defined(__WIN32__) || defined(__WIN64__)
 
 typedef HANDLE mutex_t;
+static HANDLE mutex_init() { return CreateMutex(NULL, FALSE, NULL); }
 static void mutex_lock(mutex_t* m) {
 	if (*m == NULL) {
-		HANDLE mutex = CreateMutex(NULL, FALSE, NULL);
+		HANDLE mutex = mutex_init();
 		if (InterlockedCompareExchangePointer(m, mutex, NULL) != NULL)
 			CloseHandle(mutex);
 	}
--- a/source/system/string/string_literal.c
+++ b/source/system/string/string_literal.c
@@ -7,7 +7,6 @@
 #if defined(__WIN32__) || defined(__WIN64__)
 
 typedef HANDLE mutex_t;
-#define MUTEX_INIT NULL
 static void mutex_lock(mutex_t* m) {
 	if (*m == NULL) {
 		HANDLE mutex = CreateMutex(NULL, FALSE, NULL);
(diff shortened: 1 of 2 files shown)


source/system/string/string_literal.c
```diff
@@ -21,20 +21,25 @@ static void mutex_free(mutex_t* m) { CloseHandle(*m); }
 #else
 
 typedef pthread_mutex_t mutex_t;
-#define MUTEX_INIT PTHREAD_MUTEX_INITIALIZER
 static void mutex_lock(mutex_t* m) { pthread_mutex_lock(m); }
 static void mutex_unlock(mutex_t* m) { pthread_mutex_unlock(m); }
 static void mutex_free(mutex_t*  m) { pthread_mutex_destroy(m); }
+static void mutex_init(mutex_t* m) { pthread_mutex_init(m, NULL); }
```
sebastianbaginski: I know windows is not our major target platform, but can you add this function for the windows build too ? Just for the sake of completeness

thomasfanell: Will do.

thomasfanell:
Also, this depends on rock 1.0.22, I just wanted to get feedback before the rock release on Monday.

sebastianbaginski:
This looks ok. Can you add your comment about `Memory_unload()` to the source file ?

marcusnaslund:
Can't you add the stringliteral mutex init fix to a separate PR we can merge before-hand as well?

source/system/string/string_literal.c
```diff
@@ -1,15 +1,16 @@
...
 
 #if defined(__WIN32__) || defined(__WIN64__)
 
 typedef HANDLE mutex_t;
-#define MUTEX_INIT NULL
+static HANDLE mutex_init() { return CreateMutex(NULL, FALSE, NULL); }
 static void mutex_lock(mutex_t* m) {
 	if (*m == NULL) {
-		HANDLE mutex = CreateMutex(NULL, FALSE, NULL);
+		HANDLE mutex = mutex_init();
```
sebastianbaginski: I think it should be something like
```
static void mutex_init(mutex_t* m) {
    *m = CreateMutex(NULL, FALSE, NULL);
}
```

because of how it's used in line 41

thomasfanell: Yes, my mistake. I'll fix.

thomasfanell:
@marcusnaslund yeah, this was a poorly executed PR on my part, sorry about that.

PR #1998, opened 2016-11-17
Added print

christopherlagerhult:


emilwestergren:
I think this is included in this one https://github.com/magic-lang/ooc-kean/pull/2000

PR #2001, opened 2016-11-17
Compile-time debug output

marcusnaslund:
Added version blocks to set `Debug level` at compile time. Please suggest better names if you have any. The order is so that a lower level overrides a higher one, so that calling `./test.sh -DdebugOutputVerbose` will print everything, and the last one is silent so that `-DdebugOutputSilent` will override any flags and be silent.

Also set the default level in `test.sh`. Would prefer to not have `Fatal` as the default, perhaps, but having another level will mean calling the script with `fatal` flag doesn't change anything... not without much more bash wizardry anyway.

Comments @sebastianbaginski ?

### DIFF
--- a/source/system/Debug.ooc
+++ b/source/system/Debug.ooc
@@ -19,7 +19,7 @@ DebugLevel: enum {
 }
 
 Debug: class {
-	_level: static DebugLevel = DebugLevel Verbose
+	_level: static DebugLevel = DebugLevel Debug
 	level: static DebugLevel {
 		get { This _level }
 		set (value) { This _level = value }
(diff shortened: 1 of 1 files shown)


sebastianbaginski:
I think this is ok. I wanted to have a possibility to suppress every printout at compile time and this PR introduces this.

marcusnaslund:
Thanks, but I want to think this through some more

marcusnaslund:
Updated: Default is `Debug`, just as before. There's also a flag for `Verbose`, should you wish it.
(Other PRs will try to separate the output levels of different debug calls)
Review again @sebastianbaginski ?

PR #2004, opened 2016-11-21
Added init from array to ConvexHull2D.

niklasborwell:
@tomasmahlberg peer review?


source/geometry/FloatConvexHull2D.ooc
```diff
@@ -24,6 +24,13 @@ FloatConvexHull2D: class {
 		if (computeHull)
 			this computeHull()
 	}
+	init: func ~fromArray (array: FloatPoint2D[], computeHull := true) {
+		this _points = VectorList<FloatPoint2D> new(array length)
+		for (i in 0 .. array length)
+			this _points add(array[i])
+		if (computeHull)
+			this computeHull()
```
marcusnaslund: I prefer
```ooc
list := VectorList<FloatPoint2D> new(array length)
for (i in 0 .. array length)
	list add(array[i])
this init(list, computeHull)
```

niklasborwell: That sounds reasonable. But I really don't feel like wrestling with ooc right now, so let's just add it here where we need it right now

niklasborwell:
updated

marcusnaslund:
Looks good

niklasborwell:
Rebased.

PR #2014, opened 2016-11-30
Correctly create closure for eglExtensions

erikhagglund:
review @emilwestergren  @christopherlagerhult

what about "This eglDupNativeFenceFDANDROID" shouln't we create a closure around it aswell?


christopherlagerhult:
:+1:

PR #2018, opened 2016-11-30
Removed overly complicated code for projection in FloatTransform3D.

niklasborwell:
@marcusnaslund Don't think we need a peer review for this. It does exactly the same thing as before.


marcusnaslund:
Is there a test which covers this function? :)

niklasborwell:
Yeah of course

marcusnaslund:
I take your word for it

PR #2019, opened 2016-11-30
function eglDupNativeFenceFDANDROID is now loaded correctly

christopherlagerhult:
review @emilwestergren

### DIFF
--- a/source/draw/gpu/opengl/OpenGLContext.ooc
+++ b/source/draw/gpu/opengl/OpenGLContext.ooc
@@ -14,7 +14,7 @@ use collections
 use concurrent
 import OpenGLPacked, OpenGLMonochrome, OpenGLRgb, OpenGLRgba, OpenGLUv, OpenGLMesh, OpenGLPromise, GraphicBufferYuv420Semiplanar, GraphicBuffer
 import OpenGLMap
-import backend/[GLContext, GLRenderer]
+import backend/[GLContext, GLRenderer, GLExtensions]
 
 version(!gpuOff) {
 _FenceToRasterFuture: class extends ToRasterFuture {
@@ -106,7 +106,13 @@ OpenGLContext: class extends GpuContext {
 		this _recycleBin search(|image| image instanceOf(type) && image size == size)
 	}
 	createFence: override func -> Promise {
-		OpenGLPromise new(this)
+		result: OpenGLPromise
+		if (GLExtensions nativeFenceSupported)
+			result = OpenGLNativeFencePromise new(this)
+		else
+			result = OpenGLPromise new(this)
+		result sync()
+		result
 	}
 	createMonochrome: override func (size: IntVector2D) -> GpuImage {
 		result := this _searchImageBin(OpenGLMonochrome, size)
--- a/source/draw/gpu/opengl/OpenGLContext.ooc
+++ b/source/draw/gpu/opengl/OpenGLContext.ooc
@@ -105,6 +105,9 @@ OpenGLContext: class extends GpuContext {
 	_searchImageBin: func (type: Class, size: IntVector2D) -> GpuImage {
 		this _recycleBin search(|image| image instanceOf(type) && image size == size)
 	}
+	createFence: override func -> Promise {
+		OpenGLPromise new(this)
+	}
 	createMonochrome: override func (size: IntVector2D) -> GpuImage {
 		result := this _searchImageBin(OpenGLMonochrome, size)
 		result == null ? OpenGLMonochrome new(size, this) as GpuImage : result
--- a/source/draw/gpu/opengl/backend/GLExtensions.ooc
+++ b/source/draw/gpu/opengl/backend/GLExtensions.ooc
@@ -42,7 +42,7 @@ GLExtensions: class {
 	_load: static func (name: String) -> Closure {
 		result := eglGetProcAddress(name toCString())
 		if (result == null)
-			Debug print("Failed to load OpenGL extension function: " << name)
+			Debug print~free("Failed to load OpenGL extension function: " << name)
 		Closure fromPointer(result)
 	}
 }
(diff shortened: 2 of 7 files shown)


source/draw/gpu/opengl/backend/GLExtensions.ooc
```diff
@@ -29,8 +29,13 @@ GLExtensions: class {
 			This eglClientWaitSyncKHR = This _load("eglClientWaitSyncKHR") as Func (Pointer, Pointer, Int, ULong) -> Bool
 			version(!windows) {
 				//For some reason this function can't be loaded with eglGetProcAddress so we load it with dlsym instead
-				handle := null //This should be RTLD_DEFAULT, defined in dlfcn.h
-				This eglDupNativeFenceFDANDROID = Closure fromPointer(dlsym(handle, "eglDupNativeFenceFDANDROID")) as Func(Pointer, Pointer) -> Int
+				//dlsym can also be called with the special flag RTLD_DEFAULT, defined in dlfcn.h
```
christopherlagerhult: Could probably remove this comment but I saved it for the moment as we might want to use it instead.

source/draw/gpu/opengl/backend/GLExtensions.ooc
```diff
@@ -29,8 +29,13 @@ GLExtensions: class {
...
+				dynlib := Dynlib load("/system/lib/libEGL.so")
+				result := dynlib symbol("eglDupNativeFenceFDANDROID")
+				dynlib close()
+				if (result == null)
+					Debug print("Failed to load OpenGL extension function: eglDupNativeFenceFDANDROID")
```
marcusnaslund: `Debug print` or `Debug error`?

christopherlagerhult: The other functions initialized inside `_load`, which also uses `Debug print` is something fails. I believe the intensio is to have some sort of fallback in the future if a function fails.

christopherlagerhult:
As @erikhagglund mentioned. Maybe it's better to use the `RTLD_DEFAULT` approach as it is a more generic solution and won't fail when compiling for 64-bit.

marcusnaslund:
I don't follow here. Is there more to be done? This is green-flagged but the comments indicate there's more to do.

christopherlagerhult:
I believe we agreed on changing to a RTLD_DEFAULT search

christopherlagerhult:
Updated

marcusnaslund:
`error: ‘RTLD_DEFAULT’ undeclared (first use in this function) result := dlsym(RTLD_DEFAULT, "eglDupNativeFenceFDANDROID")`

christopherlagerhult:
> RTLD_DEFAULT: extern Pointer

What?! I'm compiling it right now locally

emilwestergren:
Both with `build.sh` and `/tools/android/make.sh` ?

christopherlagerhult:
forgot to cleanup rock, but why did it find it before ... :(

christopherlagerhult:
Added flags during lunchbreak @emilwestergren ? Please verify the latest additions as well

source/draw/gpu/opengl/OpenGLContext.ooc
```diff
@@ -105,6 +105,9 @@ OpenGLContext: class extends GpuContext {
 	_searchImageBin: func (type: Class, size: IntVector2D) -> GpuImage {
 		this _recycleBin search(|image| image instanceOf(type) && image size == size)
 	}
+	createFence: override func -> Promise {
+		OpenGLPromise new(this)
```
emilwestergren: Single line

emilwestergren: Should we sync the fence here as well, then we don't need to cast to OpenGLPromise

christopherlagerhult: Though it would make sense to let a `OpenGLContext` object create an `OpenGLPromise`, why else do we have the class `OpenGLPromise`

emilwestergren: `OpenGLPromise` is a promise that can synchronize renderings on the GPU, but the idea is that we don't need to know what kind of Promise it is when we wait for it.

```ooc
createFence: override func -> Promise {
    result := OpenGLPromise new()
    result sync()
    result
}

//Using it:
promise := context createFence()
promise wait() . free()

Comment (christopherlagerhult) [inline source/draw/gpu/opengl/OpenGLContext.ooc]: Oh believe I missunderstood you there :)

**Comment by marcusnaslund** [comment]
`AndroidContext.ooc:37:184: error: pointer type mismatch in conditional expression [-Werror] GLExtensions allFunctionsLoaded ? OpenGLNativeFencePromise new(this) : super()`

Does anyone actually try to compile around here? :)

**Comment by marcusnaslund** [comment]
`OpenGLContext.ooc:108:173: error: pointer type mismatch in conditional expression [-Werror] result := GLExtensions nativeFenceSupported ? OpenGLNativeFencePromise new(this) : OpenGLPromise new(this)`

I can do this all day.

File: source/draw/gpu/opengl/backend/GLExtensions.ooc [hunk 4]
```diff
@@ -27,18 +28,21 @@ GLExtensions: class {
...
+				This eglDupNativeFenceFDANDROID = Closure fromPointer(result) as Func(Pointer, Pointer) -> Int
 			}
 			This _initialized = true
 		}
 	}
 	_load: static func (name: String) -> Closure {
 		result := eglGetProcAddress(name toCString())
 		if (result == null)
-			Debug print("Failed to load OpenGL extension function: " + name)
+			Debug print("Failed to load OpenGL extension function: " << name)
```
marcusnaslund: Should be `print~free`

christopherlagerhult: I'm on it

PR #2020, opened 2016-12-01
Removed rebinding of context on free

emilwestergren:
It is illegal to bind the context to another thread when it is already bound. 
@davidpiuva


davidpiuva:
Taking a look.

davidpiuva:
Seems okay.

PR #2021, opened 2016-12-01
Normalized input to quaternion SLERP

tomasmahlberg:
This requires the input quaternions to be normalized. We want to be on the safe side here and always do this, as this has caused some unexpected problems when any of the inputs has been "almost normalized".


tomasmahlberg:
@niklasborwell review?

marcusnaslund:
Requires some updated tests

marcusnaslund:
`FloatRotation3D ....f.. failed -> sphericalLinearInterpolation : expected equal to 0.21 was 0.21 [tolerance: 0.00010000]`

tomasmahlberg:
Oops, missed that one. Fixed now. As a side note, we're planning to get rid of that meaningless wrapper class at some point.

marcusnaslund:
Sounds good to me

PR #2023, opened 2016-12-02
Moved EGL into a separate context

davidpiuva:
First step of separating EGL from OpenGL.

### DIFF
--- a/source/draw/gpu/opengl/backend/EglDisplayContext.ooc
+++ b/source/draw/gpu/opengl/backend/EglDisplayContext.ooc
@@ -1,0 +1,139 @@
+/* This file is part of magic-sdk, an sdk for the open source programming language magic.
+ *
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+version(!gpuOff) {
+use base
+use geometry
+import egl/egl
+
+EglDisplayContext: class {
+	_eglDisplay: Pointer
+	_eglContext: Pointer
+	_eglSurface: Pointer
+	_contextCount := static 0
+	_mutex := static Mutex new()
+	init: func
+	free: override func {
+		This validate(eglMakeCurrent(this _eglDisplay, null, null, null), EGL_TRUE, "eglMakeCurrent")
+		This validate(eglDestroyContext(this _eglDisplay, this _eglContext), EGL_TRUE, "eglDestroyContext")
+		This validate(eglDestroySurface(this _eglDisplay, this _eglSurface), EGL_TRUE, "eglDestroySurface")
+		This _mutex with(||
+			if (This _contextCount == 1)
+				This validate(eglTerminate(this _eglDisplay), EGL_TRUE, "eglTerminate")
+			This _contextCount -= 1
+		)
+		super()
+	}
+	getDisplay: func -> Pointer { this _eglDisplay }
+	printExtensions: func {
+		extensions := eglQueryString(this _eglDisplay, EGL_EXTENSIONS) as CString
+		extensionsString := String new(extensions, extensions length())
+		array := extensionsString split(' ')
+		extensionsString free()
+		Debug print("EGL Extensions: ")
+		for (i in 0 .. array count)
+			Debug print(array[i])
+		array free()
+	}
+	swapBuffers: func { eglSwapBuffers(this _eglDisplay, this _eglSurface) }
+	_chooseConfig: func (configAttribs: Int*) -> Pointer {
+		numConfigs: Int
+		eglChooseConfig(this _eglDisplay, configAttribs, null, 10, numConfigs&)
+		matchingConfigs: Pointer[numConfigs]
+		eglChooseConfig(this _eglDisplay, configAttribs, matchingConfigs[0]&, numConfigs, numConfigs&)
+		chosenConfig: Pointer = null
+
+		for (i in 0 .. numConfigs) {
+			success: UInt
+			red, green, blue, alpha: Int
+			success = eglGetConfigAttrib(this _eglDisplay, matchingConfigs[i], EGL_RED_SIZE, red&)
+			success &= eglGetConfigAttrib(this _eglDisplay, matchingConfigs[i], EGL_BLUE_SIZE, blue&)
+			success &= eglGetConfigAttrib(this _eglDisplay, matchingConfigs[i], EGL_GREEN_SIZE, green&)
+			success &= eglGetConfigAttrib(this _eglDisplay, matchingConfigs[i], EGL_ALPHA_SIZE, alpha&)
+
+			if (success && red == 8 && blue == 8 && green == 8 && alpha == 8) {
+				chosenConfig = matchingConfigs[i]
+				break
+			}
+		}
+		chosenConfig
+	}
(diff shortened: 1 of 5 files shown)


davidpiuva:
@emilwestergren Peer review?

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -9,126 +9,20 @@
...
 import external/gles3
 import Gles3Debug, Gles3Fence, Gles3FramebufferObject, Gles3Quad, Gles3Renderer, Gles3ShaderProgram, Gles3Texture, Gles3VolumeTexture, Gles3VertexArrayObject, Gles3IndexBufferObject
 
 Gles3Context: class extends GLContext {
-	_eglContext: Pointer
-	_eglSurface: Pointer
-	_contextCount := static 0
-	_mutex := static Mutex new()
+	displayContext := EglDisplayContext new()
+	getDisplayContextSafely: func -> EglDisplayContext { this == null ? null : this displayContext }
```
davidpiuva: A bit ugly to compare this to null but the alternatives are to duplicate the code or make the getter static.

source/draw/gpu/opengl/backend/EglDisplayContext.ooc
```diff
@@ -0,0 +1,139 @@
...
+		}
+		This validate(eglMakeCurrent(this _eglDisplay, this _eglSurface, this _eglSurface, this _eglContext), EGL_TRUE, "eglMakeCurrent")
+		true
+	}
+	_initializeDisplay: func (display: Pointer) {
+		this _eglDisplay = eglGetDisplay(display)
+		This validate(this _eglDisplay != EGL_NO_DISPLAY, "eglGetDisplay")
+		This validate(eglInitialize(this _eglDisplay, null, null), EGL_TRUE, "eglInitialize")
+	}
+	_generate: func (display: Pointer, nativeBackend: Long, sharedContext: This) -> Bool {
```
davidpiuva: I do not know if these should be private or public when only called from the OpenGL context creation.

emilwestergren: If you want we can replace the "return null on failure" philosophy with good error checking that throws exception instead (we don't handle the null return anyway) Then you don't need these ugly `_generate` functions, you can simply overload constructors.

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
emilwestergren: Ok, as a temporary solution I can approve it.

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -9,126 +9,20 @@
...
+import ../[GLContext, GLTexture, GLVertexArrayObject, GLIndexBufferObject, EglDisplayContext]
 import external/gles3
 import Gles3Debug, Gles3Fence, Gles3FramebufferObject, Gles3Quad, Gles3Renderer, Gles3ShaderProgram, Gles3Texture, Gles3VolumeTexture, Gles3VertexArrayObject, Gles3IndexBufferObject
 
 Gles3Context: class extends GLContext {
-	_eglContext: Pointer
-	_eglSurface: Pointer
-	_contextCount := static 0
-	_mutex := static Mutex new()
+	displayContext := EglDisplayContext new()
```
emilwestergren: Public member variable? Should be private or have a getter property.

source/draw/gpu/opengl/backend/EglDisplayContext.ooc
emilwestergren: Keep shared context feature, try to create a some nice constructor overloads. Either by defaulting shared context to null or overload constructors and demand that the user makes the choice of the constructor.

davidpiuva: I tried overloads for `Gles3Context create` but in order to use them, I had to duplicate `GLContext createContext` to have any use of the new overloads. Default values for internal things like EGL might not be needed since it should be hidden inside the OpenGL context.

davidpiuva: I might try just having constructors for Gles3Context then since _generate is a mess.

emilwestergren: `Gles3Context create` can be replaced with constructor as well.

davidpiuva: That would be cleaner.

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -191,25 +85,13 @@ Gles3Context: class extends GLContext {
...
 		version(debugGL) { Debug print("Creating OpenGL Context") }
-		result := This new()
-		result _generate(display, nativeBackend, sharedContext) ? result : null
+		result := This new(EglDisplayContext new(display, nativeBackend, sharedContext getDisplayContextSafely()))
 	}
 	create: static func ~pbufferShared (sharedContext: This = null) -> This {
 		version(debugGL) { Debug print("Creating OpenGL Context") }
-		result := This new()
-		result _generate(sharedContext) ? result : null
+		result := This new(EglDisplayContext new(sharedContext getDisplayContextSafely()))
```
davidpiuva: This looks cleaner but I do not know if this will work when GLES3 has to fall back to GLES2 since the constructor cannot have a boolean return value.

emilwestergren: So keep the `create` for Gles3Context

davidpiuva: Okay

source/draw/gpu/opengl/backend/gles3/Gles3Context.ooc
```diff
@@ -9,126 +9,20 @@
 version(!gpuOff) {
 use base
 use geometry
-import ../egl/egl
-import ../[GLContext, GLTexture, GLVertexArrayObject, GLIndexBufferObject]
+import ../[GLContext, GLTexture, GLVertexArrayObject, GLIndexBufferObject, EglDisplayContext]
 import external/gles3
 import Gles3Debug, Gles3Fence, Gles3FramebufferObject, Gles3Quad, Gles3Renderer, Gles3ShaderProgram, Gles3Texture, Gles3VolumeTexture, Gles3VertexArrayObject, Gles3IndexBufferObject
```
marcusnaslund: Huh, I guess that's not true always

source/draw/gpu/opengl/backend/EglDisplayContext.ooc
```diff
@@ -0,0 +1,137 @@
...
+ * Copyright (C) 2016 magic-lang
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+version(!gpuOff) {
+use base
+use geometry
+import egl/egl
```
davidpiuva: It was like that from where I copied it from so any change in behaviour should be a separate PR.

marcusnaslund: yeah and as my next comment shows, maybe "generally outside in other files" isn't even true

PR #2026, opened 2016-12-05
Added T[] constructor to VectorList

marcusnaslund:
This is a more general solution.

### DIFF
--- a/source/geometry/FloatConvexHull2D.ooc
+++ b/source/geometry/FloatConvexHull2D.ooc
@@ -25,9 +25,7 @@ FloatConvexHull2D: class {
 			this computeHull()
 	}
 	init: func ~fromArray (array: FloatPoint2D[], computeHull := true) {
-		list := VectorList<FloatPoint2D> new(array length)
-		for (i in 0 .. array length)
-			list add(array[i])
+		list := VectorList<FloatPoint2D> new(array)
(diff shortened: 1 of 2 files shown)


marcusnaslund:
Why is this so impossible? :( @thomasfanell

source/geometry/FloatConvexHull2D.ooc
```diff
@@ -25,9 +25,7 @@ FloatConvexHull2D: class {
 			this computeHull()
 	}
 	init: func ~fromArray (array: FloatPoint2D[], computeHull := true) {
-		list := VectorList<FloatPoint2D> new(array length)
-		for (i in 0 .. array length)
-			list add(array[i])
+		list := VectorList<FloatPoint2D> new(array)
```
sebastianbaginski: Are you sure the correct constructor is called, like ooc is not interpreting "array" as Pointer, converting to Int and using this
```ooc
init: func ~heap (capacity: Int, freeContent := true) {
 		this init(HeapVector<T> new(1 maximum(capacity), freeContent))
 	}
```
instead ?

thomasfanell:
Seems to that `T[]` requires `T` to be a class, not a cover. :heart:

marcusnaslund:
C++, you say?

PR #2027, opened 2016-12-05
Moved Debug to System namespace

marcusnaslund:
This allows us to use `Debug error` instead of `raise` in classes in `system/` as well, and won't force anyone to add `use base` whenever you want to use `Debug`.


marcusnaslund:

sebastianbaginski:
Great :+1:

PR #2028, opened 2016-12-06
Thread check

davidpiuva:
Asserting that the same thread is not used to create multiple OpenGL contexts.

### DIFF
--- a/source/draw/gpu/opengl/backend/EglDisplayContext.ooc
+++ b/source/draw/gpu/opengl/backend/EglDisplayContext.ooc
@@ -56,8 +56,11 @@ EglDisplayContext: class {
 		this _generateContext(sharedContext ? sharedContext _eglContext : null, chosenConfig)
 	}
 	free: override func {
-		if (eglGetCurrentContext() == EGL_NO_CONTEXT)
+		currentContext := eglGetCurrentContext()
+		if (currentContext == EGL_NO_CONTEXT)
 			Debug error("Trying to unbind OpenGL context from a thread without a context!")
+		else if (currentContext != this _eglContext)
+			Debug error("Trying to unbind OpenGL context from another thread!")
 		This validate(eglMakeCurrent(this _eglDisplay, null, null, null), EGL_TRUE, "eglMakeCurrent")
 		This validate(eglDestroyContext(this _eglDisplay, this _eglContext), EGL_TRUE, "eglDestroyContext")
 		This validate(eglDestroySurface(this _eglDisplay, this _eglSurface), EGL_TRUE, "eglDestroySurface")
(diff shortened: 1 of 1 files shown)


davidpiuva:
Restarting the test.

davidpiuva:
@emilwestergren Peer review?

source/draw/gpu/opengl/backend/EglDisplayContext.ooc
```diff
@@ -56,6 +56,8 @@ EglDisplayContext: class {
 		this _generateContext(sharedContext ? sharedContext _eglContext : null, chosenConfig)
 	}
 	free: override func {
+		if (eglGetCurrentContext() == EGL_NO_CONTEXT)
+			Debug error("Trying to unbind OpenGL context from a thread without a context!")
```
davidpiuva: Maybe there is a way to also check if it is the same context as the one being freed but if it fails then we get false positives.

source/draw/gpu/opengl/backend/EglDisplayContext.ooc
```diff
@@ -56,6 +56,12 @@ EglDisplayContext: class {
 		this _generateContext(sharedContext ? sharedContext _eglContext : null, chosenConfig)
 	}
 	free: override func {
+		if (eglGetCurrentContext() != this _eglContext) {
+			if (eglGetCurrentContext() == EGL_NO_CONTEXT)
+				Debug error("Trying to unbind OpenGL context from a thread without a context!")
+			else
+				Debug error("Trying to unbind OpenGL context from another thread!")
+		}
```
davidpiuva: Updated with a check against freeing with a thread that has a context but not the same context.

emilwestergren:
```ooc
currentContext := eglGetCurrentContext()
if (currentContext == EGL_NO_CONTEXT)
    Debug error(...)
else if (currentContext != this _eglContext)
    Debug error(...)
```

source/draw/gpu/opengl/backend/EglDisplayContext.ooc
```diff
@@ -56,6 +56,11 @@ EglDisplayContext: class {
 		this _generateContext(sharedContext ? sharedContext _eglContext : null, chosenConfig)
 	}
 	free: override func {
+		currentContext := eglGetCurrentContext()
+		if (currentContext == EGL_NO_CONTEXT)
+			Debug error("Trying to unbind OpenGL context from a thread without a context!")
+		else if (currentContext != this _eglContext)
+			Debug error("Trying to unbind OpenGL context from another thread!")
```
davidpiuva: Updated

davidpiuva:
The test keeps failing from not finding media.

PR #2031, opened 2016-12-07
Added missing this in GraphicBufferYuv420Semiplanar

davidpiuva:


davidpiuva:
@emilwestergren Peer review?

davidpiuva:
Adding missing `this` should not be able to cause a crash and the same error was seen on another PR.

PR #2033, opened 2016-12-08
Optimized FloatConvexHull2D

marcusnaslund:
Maybe doesn't help much. We'll see.


sebastianbaginski:
I'd also think about removing dynamic allocations from `_findHull` and maybe testing if it'd be faster when changed into a cover with a constant max hull size (== no calloc / free at all).

If you need it to be even faster then I'd suggest looking into `fcvFindConvexHull` https://developer.qualcomm.com/docs/fastcv/api/group___structural___analysis__and___drawing.html#gaa2c09403262c9dfb19e17f343156932c

marcusnaslund:
> fcvFindConvexHull

Thanks, I'll look it up if we need to :+1:

marcusnaslund:
Closing this, will reopen if it turns out we need it

PR #2034, opened 2016-12-08
Cleaned up FloatConvexHull2D, simple optimizations

marcusnaslund:
- Some cleanup with tuple syntax
- Inlined a function used in only one place
- Split a for-loop to avoid the constant branching
- Set a size for each new vectorlist to avoid resizing it again

### DIFF
--- a/source/geometry/FloatConvexHull2D.ooc
+++ b/source/geometry/FloatConvexHull2D.ooc
@@ -67,26 +67,21 @@ FloatConvexHull2D: class {
 			// Uses the Quickhull algorithm if more than two points, average complexity O(n log n)
 			// https://en.wikipedia.org/wiki/Quickhull
 			points := this _points
-			this _points = VectorList<FloatPoint2D> new()
-			leftMostIndex := 0
-			rightMostIndex := 0
+			this _points = VectorList<FloatPoint2D> new(points _count)
+			(leftMostIndex, rightMostIndex) := (0, 0)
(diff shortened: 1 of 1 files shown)


source/geometry/FloatConvexHull2D.ooc
```diff
@@ -102,24 +97,26 @@ FloatConvexHull2D: class {
...
-			outsideRight := VectorList<FloatPoint2D> new()
-			for (i in 0 .. currentSet count)
-				if (i != maximumDistanceIndex)
-					if (This _isOnLeft(leftPoint, maximumDistancePoint, currentSet[i]))
-						outsideLeft add(currentSet[i])
-					else if (This _isOnLeft(maximumDistancePoint, rightPoint, currentSet[i]))
-						outsideRight add(currentSet[i])
+			outsideLeft := VectorList<FloatPoint2D> new(currentSet _count)
+			outsideRight := VectorList<FloatPoint2D> new(currentSet _count)
+			for (i in 0 .. maximumDistanceIndex)
```
sebastianbaginski: Have you tested that the compiler does not actually do that with -O3 for you ? I think previous loop code was more readable

sebastianbaginski: You can try  generating asm code with -O3 for the function before and after your changes and check if there are any differences (or profile it ?). If you want you can just merge the changes, I don't look into this code very often, so it does not matter that much to me :)

PR #2036, opened 2016-12-12
Added onRecycle.

emilwestergren:
`onRecycle` must be called when recycling images to reset the reference counter
@erikhagglund peer review


source/draw/gpu/opengl/AndroidContext.ooc
```diff
@@ -165,7 +165,9 @@ AndroidContext: class extends OpenGLContext {
 	}
 	recycle: override func (image: OpenGLPacked) {
 		match (image) {
-			case i: EGLRgba => this _eglBin add(i)
+			case i: EGLRgba =>
```
marcusnaslund: Name it `eglRgba` and we have a deal.

emilwestergren:
Updated

PR #2038, opened 2016-12-13
corrected pad function name

sebastianbaginski:


source/geometry/IntBox2D.ooc
```diff
@@ -58,7 +58,7 @@ IntBox2D: cover {
 	pad: func (left, right, top, bottom: Int) -> This {
 		This new(IntPoint2D new(this left - left, this top - top), IntVector2D new(this width + left + right, this height + top + bottom))
 	}
-	pad: func ~fromFloat (pad: Int) -> This { this pad(pad, pad, pad, pad) }
```
marcusnaslund: Is it correct in `FloatBox`?

sebastianbaginski: yes, I think someone copy pasted this from there

marcusnaslund: `Blame David Hesselbom ( 28 months ago )`
wow, it's hard to find an older bug than that

PR #2044, opened 2016-12-16
Removed FloatRotation3D wrapper class

tomasmahlberg:


marcusnaslund:
Why was this class created in the first place? Did it use to do more?

tomasmahlberg:
I'm not sure exactly what the purpose of this class was and no, it never used to do anything more than what it does now. @niklasborwell Any input here?

niklasborwell:
It never had a purpose. Good riddance!

tomasmahlberg:
@niklasborwell

PR #2049, opened 2016-12-20
Added close and dup to unistd

christopherlagerhult:
... in order to support `close/dup`


source/system/native/FileUnix.ooc
```diff
@@ -62,7 +62,6 @@ lstat: extern func (CString, FileStat*) -> Int
 fstat: extern func (Int, FileStat*) -> Int
 chmod: extern func (CString, ModeT) -> Int
 fchmod: extern func (Int, ModeT) -> Int
-close: extern func (Int) -> Int
```
emilwestergren: Why did you move this?

PR #2050, opened 2016-12-22
fixed some race conditions

sebastianbaginski:


source/concurrent/ThreadPool.ooc
```diff
@@ -50,7 +50,10 @@ _Task: abstract class {
 			}
 			timer free()
 		}
-		this _state == _PromiseState Finished
+		this _mutex lock()
+		status = this _state == _PromiseState Finished
+		this _mutex unlock()
+		status
```
sebastianbaginski: This one was reported when running helgrind on `ThreadPoolTest`

source/concurrent/Promise.ooc
```diff
@@ -71,18 +71,19 @@ ConditionPromise: class extends Promise {
...
 	}
 	wait: override func (time: TimeSpan) -> Bool { Debug error("Timed wait is not supported for ConditionPromise"); false }
 	wait: override func ~forever -> Bool {
 		this _mutex lock()
 		if (!this _completed)
 			this _condition wait(this _mutex)
+		status := this _completed
 		this _mutex unlock()
-		this _completed
+		status
```
sebastianbaginski: This one was not reported, but if we're protecting `_completed` with the mutex, let's always do that.

source/concurrent/Promise.ooc
```diff
@@ -71,18 +71,19 @@ ConditionPromise: class extends Promise {
...
 	}
 	signal: func {
-		this _mutex with(||
-			this _completed = true
-			this _condition broadcast()
-		)
+		this _mutex lock()
+		this _completed = true
+		this _condition broadcast()
+		this _mutex unlock()
```
sebastianbaginski: This one puzzles me. When using `with()`, one of the tests in other repo reports more race conditions.
But the main reason I changed - I think it's not worth creating / deleting a temporary closure object with `calloc / free` just to save few characters (not even a line).

source/concurrent/ThreadPool.ooc
emilwestergren: Here is a good case to use `this _mutex with(|| status = this _state == _PromiseState Finished)` though, but doesn't matter to me.

sebastianbaginski: Hmm that's true, it's much cleaner. @marcusnaslund ?

marcusnaslund: Agree with @emilwestergren

marcusnaslund: unless there's a cost of doing it

emilwestergren: There is, a closure needs to allocated and freed.

sebastianbaginski: `calloc / free` on a closure

PR #2051, opened 2016-12-22
Timedwait for wait condition

sebastianbaginski:
We have `timedlock` for mutex.
Also cleans up some code in ThreadUnix.

depends on https://github.com/magic-lang/ooc-kean/pull/2050

### DIFF
--- a/source/concurrent/WaitCondition.ooc
+++ b/source/concurrent/WaitCondition.ooc
@@ -11,6 +11,7 @@ import native/[ConditionUnix, ConditionWin32]
 
 WaitCondition: abstract class {
 	wait: abstract func (mutex: Mutex) -> Bool
+	wait: abstract func ~timed (mutex: Mutex, seconds: Double) -> Bool
 	signal: abstract func -> Bool
 	broadcast: abstract func -> Bool
 	new: static func -> This {
--- a/source/concurrent/native/ConditionUnix.ooc
+++ b/source/concurrent/native/ConditionUnix.ooc
@@ -29,6 +29,12 @@ ConditionUnix: class extends WaitCondition {
 			Debug error(mutex class != MutexUnix, "ConditionUnix can work only with MutexUnix")
 		pthread_cond_wait(this _backend, mutex as MutexUnix _backend&) == 0
 	}
+	wait: override func ~timed (mutex: Mutex, seconds: Double) -> Bool {
+		version(safe)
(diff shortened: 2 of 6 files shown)


source/concurrent/native/ConditionUnix.ooc
```diff
@@ -29,6 +29,12 @@ ConditionUnix: class extends WaitCondition {
 			Debug error(mutex class != MutexUnix, "ConditionUnix can work only with MutexUnix")
 		pthread_cond_wait(this _backend, mutex as MutexUnix _backend&) == 0
 	}
+	wait: override func ~timed (mutex: Mutex, seconds: Double) -> Bool {
+		version(safe)
+			Debug error(mutex class != MutexUnix, "ConditionUnix can work only with MutexUnix")
```
christopherlagerhult: I believe `intstanceOf` is preferred

sebastianbaginski: Oh ok, I just copied it from up there :/

source/concurrent/WaitCondition.ooc
```diff
@@ -11,6 +11,7 @@ import native/[ConditionUnix, ConditionWin32]
 
 WaitCondition: abstract class {
 	wait: abstract func (mutex: Mutex) -> Bool
+	wait: abstract func ~timed (mutex: Mutex, seconds: Double) -> Bool
```
emilwestergren: `TimeSpan` ?

sebastianbaginski: but we use Double for Threads:
```ooc
ThreadUnix: ...
....
wait: override func ~timed (seconds: Double) -> Bool 
...
```
I just followed that convention

christopherlagerhult: Speaking of, why do we use seconds instead of milliseconds?

source/concurrent/native/ConditionUnix.ooc
sebastianbaginski: It's corrected now.

source/concurrent/WaitCondition.ooc
sebastianbaginski: No idea. Should we make a card for this ?

marcusnaslund: It's a leftover from the original SDK which uses seconds.

PR #2056, opened 2017-01-03
Now flushing after glFenceSync

erikhagglund:
4.1.2 Signaling
Footnote 3: The simple flushing behavior defined by SYNC_FLUSH_COMMANDS_BIT
will not help when waiting for a fence command issued in another context’s
command stream to complete. Applications which block on a fence sync object
must take additional steps to assure that the context from which the
corresponding fence command was issued has flushed that command to the
graphics pipeline.

### DIFF
--- a/source/draw/gpu/opengl/backend/EGLNativeFence.ooc
+++ b/source/draw/gpu/opengl/backend/EGLNativeFence.ooc
@@ -8,7 +8,7 @@
 
 use base
 import egl/egl
-import GLExtensions, GLFence, gles3/Gles3Debug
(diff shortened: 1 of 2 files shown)


erikhagglund:
Tried with removing "FLUSH_COMMAND_BIT" as well but I got lower performance, 27.5 fps instead of 29.7 so kept it.

I also noticed that enabling "fences"  vidhance_settings_set_synchronized(false) costs 5 fps in 4K on urchin, can only get 25 fps

emilwestergren:
That's a damn shame, son

PR #2057, opened 2017-01-04
Added fitAspectRatio function.

emilwestergren:
I am sick of doing these calculations so I think it would be a good idea to implement them somewhere here.
@marcusnaslund @erikhagglund peer review


emilwestergren:
I agree, but where else?

emilwestergren:
I guess we could create a help class `AspectRatio` with some functions.
```ooc
result := AspectRatio fit(input, output)
scaleVector := AspectRatio getScaler(input, output)
```

marcusnaslund:
That sounds much more reasonable.

PR #2062, opened 2017-01-17
Reimplemented gpuOff.

emilwestergren:
`gpuOff` is now (basically) only implemented in the lowest layer where the actual dependency to OpenGL lies.

### DIFF
--- a/source/draw/gpu/opengl/OpenGLPromise.ooc
+++ b/source/draw/gpu/opengl/OpenGLPromise.ooc
@@ -22,9 +22,8 @@ OpenGLPromise: class extends Promise {
 	}
 	sync: func { this _fence sync() }
 	wait: override func (time: TimeSpan) -> Bool { this _fence clientWait(time toNanoseconds()) }
+	duplicateFileDescriptor: func -> Int { this _fence duplicateFileDescriptor() }
 }
-
 OpenGLNativeFencePromise: class extends OpenGLPromise {
-	init: func (context: OpenGLContext) { super(EGLNativeFence new(context backend getDisplay())) }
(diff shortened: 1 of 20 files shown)


emilwestergren:
@marcusnaslund @sebastianbaginski at least with this we can keep it at the actual OpenGL level. Do you have any suggestions of how we can improve it even more?

marcusnaslund:
Yes, this is where it should always (only) have been.

sebastianbaginski:
Keeping it at lower level seems cleaner. How do you use it in the other project though ? Trying to create opengl context will return a null pointer with `gpuOff`, so we still need something like
```
createContext: func -> DrawContext {
    version(gpuOff)
       CpuContext
    else
      GLESContext...
}
``` ?

marcusnaslund:
Isn't that what `RenderMode Cpu` is for?

emilwestergren:
Yes, we force `RenderMode Cpu`

sebastianbaginski:
I don't think that every component checks the mode. There are many hard-coded gpu dependencies. But this is not a problem with this PR anyway.

emilwestergren:
Yeah we probably need to adapt the code, but the idea is to reimplement this from scratch without having gpuOff everywhere in the code base.

marcusnaslund:
Any news here?

emilwestergren:
Not sure if we want this merged until we need it?

marcusnaslund:
Sure, my guess was that you started this _because_ we needed it :)

marcusnaslund:
You should probably rebase on the latest code to keep things up to date, at least

emilwestergren:
Sure

marcusnaslund:
Closing. Reopen when needed.

PR #2064, opened 2017-01-18
Added blend modes to blitter

davidpiuva:
Added fill and add blend modes for blitting. These modes are bit exact when used on the CPU blitter.
Only the luma channel is affected by blitting to keep it fast.

The default image comparison used for handling the oddities of OpenGL had too much room for incorrect offsets so it is not used to test CPU blitting anymore.

### DIFF
--- a/source/draw/Image.ooc
+++ b/source/draw/Image.ooc
@@ -47,12 +47,12 @@ Image: abstract class {
 		this drawLines(positions, pen)
 		positions free()
 	}
-	blitWhiteSource: virtual func (source: This, offset: IntVector2D, sourceBox: IntBox2D) {
+	blitSource: virtual func (source: This, offset: IntVector2D, sourceBox: IntBox2D, blendMode: BlendMode) {
 		normalizedSourceBox := (sourceBox toFloatBox2D()) / (source size toFloatVector2D())
-		DrawState new(this) setBlendMode(BlendMode White) setInputImage(source) setSourceNormalized(normalizedSourceBox) setViewport(IntBox2D new(offset, sourceBox size)) draw()
+		DrawState new(this) setBlendMode(blendMode) setInputImage(source) setSourceNormalized(normalizedSourceBox) setViewport(IntBox2D new(offset, sourceBox size)) draw()
 	}
-	blitWhite: func (source: This, offset: IntVector2D) {
-		this blitWhiteSource(source as This, offset, IntBox2D new(source size))
+	blit: func (source: This, offset: IntVector2D, blendMode: BlendMode) {
+		this blitSource(source as This, offset, IntBox2D new(source size), blendMode)
 	}
 	fill: virtual func (color: ColorRgba) { Debug error("fill unimplemented for class %s!" format(this class name)) }
 	draw: virtual func ~DrawState (drawState: DrawState) { Debug error("draw~DrawState unimplemented for class %s!" format(this class name)) }
@@ -67,8 +67,8 @@ Image: abstract class {
 	distance: virtual abstract func (other: This) -> Float
 	equals: func (other: This) -> Bool { this size == other size && this distance(other) < 10 * Float epsilon }
 	isValidIn: func (x, y: Int) -> Bool { x >= 0 && x < this size x && y >= 0 && y < this size y }
-	// Writes white text on the existing image
-	write: func (message: String, fontAtlas: This, localOrigin: IntPoint2D) {
+	// Writes white text on the existing image (with black background if solid is true)
+	write: func (message: String, fontAtlas: This, localOrigin: IntPoint2D, blendMode: BlendMode = BlendMode White) {
 		skippedRows := 2
 		columns := 16
 		fontSize := DrawContext getFontSize(fontAtlas)
@@ -78,8 +78,8 @@ Image: abstract class {
 			sourceX := charCode % columns
 			sourceY := (charCode / columns) - skippedRows
 			source := IntBox2D new(sourceX * fontSize x, sourceY * fontSize y, fontSize x, fontSize y)
-			if ((charCode as Char) graph())
-				this blitWhiteSource(fontAtlas, (localOrigin + (targetOffset * fontSize)) toIntVector2D(), source)
+			if (blendMode == BlendMode Fill || (charCode as Char) graph())
+				this blitSource(fontAtlas, (localOrigin + (targetOffset * fontSize)) toIntVector2D(), source, blendMode)
 			targetOffset x += 1
 			if (charCode == '\n') {
 				targetOffset x = 0 // Carriage return
--- a/source/draw/Image.ooc
+++ b/source/draw/Image.ooc
@@ -67,7 +67,7 @@ Image: abstract class {
 	distance: virtual abstract func (other: This) -> Float
 	equals: func (other: This) -> Bool { this size == other size && this distance(other) < 10 * Float epsilon }
 	isValidIn: func (x, y: Int) -> Bool { x >= 0 && x < this size x && y >= 0 && y < this size y }
-	// Writes white text on the existing image (with black background if solid is true)
+	// Writes white text on the existing image (with black background if blendMode is Fill)
 	write: func (message: String, fontAtlas: This, localOrigin: IntPoint2D, blendMode: BlendMode = BlendMode White) {
 		skippedRows := 2
 		columns := 16
--- a/source/draw/RasterMonochrome.ooc
+++ b/source/draw/RasterMonochrome.ooc
@@ -43,25 +43,56 @@ RasterMonochrome: class extends RasterPacked {
 		if (monochrome != image)
 			monochrome referenceCount decrease()
 	}
-	// Precondition: Do not draw out of bound
-	_blitWhiteRaw: static func (target, source: Byte*, startSrcX, startSrcY, startDstX, startDstY, width, height, targetStride, sourceStride: Int) {
-		for (y in 0 .. height) {
-			dstY := startDstY + y
-			srcY := startSrcY + y
-			writer := target + startDstX + (dstY * targetStride)
-			reader := source + startSrcX + (srcY * sourceStride)
-			for (x in 0 .. width) {
-				sourceColor: Float = reader@
-				targetColor: Float = writer@
-				blendColor: Int = (targetColor * (1.0f - (sourceColor / 255.0f))) + sourceColor // Source intensity as alpha
-				if (blendColor > 255) blendColor = 255 // Limit intensity
-				writer@ = blendColor // Write to target
-				reader += 1
-				writer += 1
+	// Preconditions:
+	//   Do not draw out of bound.
+	//   Do not let the target region overlap the source region.
+	_blitRaw: static func (target, source: Byte*, startSrcX, startSrcY, startDstX, startDstY, width, height, targetStride, sourceStride: Int, blendMode: BlendMode) {
+		if (width > 0) {
+			upperLeftTarget := target + startDstX + (startDstY * targetStride)
+			upperLeftSource := source + startSrcX + (startSrcY * sourceStride)
+			writerLineStart := upperLeftTarget
+			readerLineStart := upperLeftSource
+			match (blendMode) {
+				case BlendMode White =>
+					for (y in 0 .. height) {
+						writer := writerLineStart
+						reader := readerLineStart
+						for (x in 0 .. width) {
+							sourceColor: Float = reader@
+							targetColor: Float = writer@
+							blendColor: Int = (targetColor * (1.0f - (sourceColor / 255.0f))) + sourceColor // Source intensity as alpha
+							if (blendColor > 255) blendColor = 255 // Limit intensity
+							writer@ = blendColor // Write to target
+							reader += 1
+							writer += 1
+						}
+						writerLineStart += targetStride
+						readerLineStart += sourceStride
+					}
+				case BlendMode Add =>
+					for (y in 0 .. height) {
+						writer := writerLineStart
+						reader := readerLineStart
+						for (x in 0 .. width) {
+							blendColor: Int = reader@ + writer@ // Simple addition
+							if (blendColor > 255) blendColor = 255 // Limit intensity
+							writer@ = blendColor // Write to target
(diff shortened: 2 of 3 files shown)


davidpiuva:
@emilwestergren Peer review?

test/draw/gpu/BlitTest.ooc
```diff
@@ -91,17 +91,73 @@ BlitTest: class extends Fixture {
...
+				<    XXXXXXXX  >
+				<              >
+				<              >"
+			)
+			this testBlend(this sourceImage, correctImage, IntVector2D new(0, 0), IntVector2D new(4, 2), BlendMode Fill, 0) // Bit exact
+			correctImage free()
+		})
 	}
 	testOffset: func (source, expected: RasterMonochrome, offset: IntVector2D) {
+		this testOffsetMode(source, expected, offset, BlendMode Fill, 0) // Bit exact
```
davidpiuva: There is no reason to have a tolerance for rounding errors when only using integer assignments.

test/draw/gpu/BlitTest.ooc
```diff
@@ -91,17 +91,73 @@ BlitTest: class extends Fixture {
...
+		this testOffsetMode(source, expected, offset, BlendMode Add, 0) // Bit exact
```
davidpiuva: This is just a clamped integer addition so no rounding errors possible there either.

test/draw/gpu/BlitTest.ooc
```diff
@@ -91,17 +91,73 @@ BlitTest: class extends Fixture {
...
+				<XXXXXXXX      >
+				<X*    *X      >
+				<X*  XXXXXXXX  >
+				<X*  X*    *X  >
+				<X***X*    *X  >
+				<XXXXX*   XXX  >
+				<    X****XXX  >
+				<    XXXXXXXX  >
+				<              >
+				<              >"
```
davidpiuva: The second sprite is covering the first sprite.

test/draw/gpu/BlitTest.ooc
```diff
@@ -91,17 +91,73 @@ BlitTest: class extends Fixture {
...
+				<X*    *X       >
+				<X*   XXX       >
+				<X****XXXXXXXX  >
+				<XXXXXXXX   *X  >
+				<     X*    *X  >
+				<     X*   XXX  >
+				<     X****XXX  >
+				<     XXXXXXXX  >
+				<               >
+				<               >"
```
davidpiuva: This is valid for both white and additive blending because they second sprite only overlap with black and white pixels from the first sprite.

source/draw/RasterMonochrome.ooc
```diff
@@ -79,7 +110,34 @@ RasterMonochrome: class extends RasterPacked {
...
+					thisColor: Int = thisPixel@
+					otherColor: Int = otherPixel@
+					result += thisColor > otherColor ? thisColor - otherColor : otherColor - thisColor
+					thisPixel += 1
+					otherPixel += 1
+				}
+				thisLineStart += thisStride
+				otherLineStart += otherStride
+			}
+			return result;
```
davidpiuva: This will not tolerate any movement so that CPU blitting can have more strict testing.

source/draw/RasterMonochrome.ooc
```diff
@@ -43,25 +43,56 @@ RasterMonochrome: class extends RasterPacked {
...
+							targetColor: Float = writer@
+							blendColor: Int = (targetColor * (1.0f - (sourceColor / 255.0f))) + sourceColor // Source intensity as alpha
+							if (blendColor > 255) blendColor = 255 // Limit intensity
+							writer@ = blendColor // Write to target
+							reader += 1
+							writer += 1
+						}
+						writerLineStart += targetStride
+						readerLineStart += sourceStride
+					}
```
davidpiuva: Slowest blend mode using floats to avoid the hard clamping seen in additive blending.

source/draw/RasterMonochrome.ooc
```diff
@@ -43,25 +43,56 @@ RasterMonochrome: class extends RasterPacked {
...
+						for (x in 0 .. width) {
+							blendColor: Int = reader@ + writer@ // Simple addition
+							if (blendColor > 255) blendColor = 255 // Limit intensity
+							writer@ = blendColor // Write to target
+							reader += 1
+							writer += 1
+						}
+						writerLineStart += targetStride
+						readerLineStart += sourceStride
+					}
```
davidpiuva: Bit-exact integer addition.

source/draw/RasterMonochrome.ooc
```diff
@@ -43,25 +43,56 @@ RasterMonochrome: class extends RasterPacked {
...
+				case => // Fill or unhandled
+					for (y in 0 .. height) {
+						memcpy(writerLineStart, readerLineStart, width) // Direct memory copy
+						writerLineStart += targetStride
+						readerLineStart += sourceStride
+					}
```
davidpiuva: Bit-exact assignment to completely ignore the background.

source/draw/RasterMonochrome.ooc
davidpiuva: Bit-exact integer addition.

source/draw/RasterMonochrome.ooc
```diff
@@ -43,25 +43,56 @@ RasterMonochrome: class extends RasterPacked {
...
+	_blitRaw: static func (target, source: Byte*, startSrcX, startSrcY, startDstX, startDstY, width, height, targetStride, sourceStride: Int, blendMode: BlendMode) {
+		if (width > 0) {
+			upperLeftTarget := target + startDstX + (startDstY * targetStride)
+			upperLeftSource := source + startSrcX + (startSrcY * sourceStride)
+			writerLineStart := upperLeftTarget
+			readerLineStart := upperLeftSource
+			match (blendMode) {
+				case BlendMode White =>
+					for (y in 0 .. height) {
+						writer := writerLineStart
```
emilwestergren: Looks like a lot of duplicate code here. Can you move the blending to a function or closure?

source/draw/RasterMonochrome.ooc
```diff
@@ -79,7 +110,34 @@ RasterMonochrome: class extends RasterPacked {
 		dimensions := endDst - startDst
 		endSrc := startSrc + dimensions
 		if (startSrc x >= 0 && startSrc y >= 0 && endSrc x <= source size x && endSrc y <= source size y && startDst x >= 0 && startDst y >= 0 && endDst x <= this size x && endDst y <= this size y) {
-			this _blitWhiteRaw(this buffer pointer, (source as This) buffer pointer, startSrc x, startSrc y, startDst x, startDst y, dimensions x, dimensions y, this stride, (source as This) stride)
+			this _blitRaw(this buffer pointer, (source as This) buffer pointer, startSrc x, startSrc y, startDst x, startDst y, dimensions x, dimensions y, this stride, (source as This) stride, blendMode)
+		}
+	}
+	exactDistance: func (other: This) -> Int {
+		if (this size != other size) {
+			return 2147483647;
```
emilwestergren: What is this value?

davidpiuva: Maximum integer. I could not find the name for it.

source/draw/RasterMonochrome.ooc
davidpiuva: No, that would probably make it 100 times as slow.

source/draw/RasterMonochrome.ooc
emilwestergren: `Int maximumValue` ?

source/draw/RasterMonochrome.ooc
```diff
@@ -43,25 +43,56 @@ RasterMonochrome: class extends RasterPacked {
...
+			match (blendMode) {
+				case BlendMode White =>
+					for (y in 0 .. height) {
+						writer := writerLineStart
+						reader := readerLineStart
+						for (x in 0 .. width) {
+							sourceColor: Float = reader@
+							targetColor: Float = writer@
+							blendColor: Int = (targetColor * (1.0f - (sourceColor / 255.0f))) + sourceColor // Source intensity as alpha
+							if (blendColor > 255) blendColor = 255 // Limit intensity
```
emilwestergren: Clamp?

source/draw/RasterMonochrome.ooc
emilwestergren: You have one(!) line of code that is different for the two cases (White/Add). I would at least start with a closure and then if it's really 100 times faster maybe considering optimizing it in the future.

davidpiuva: I have tried it before with profiling and it really is around 100 times as slow. Any function call inside the loop makes it non-realtime by flushing the whole execution window for every pixel.

emilwestergren: What about if-statement? Shouldn't branch prediction improve it?

source/draw/RasterMonochrome.ooc
davidpiuva: You told me to make it fast and the C compiler cannot inline the function call into a conditional move instruction if I do that.

source/draw/RasterMonochrome.ooc
```diff
@@ -79,7 +110,34 @@ RasterMonochrome: class extends RasterPacked {
 		dimensions := endDst - startDst
 		endSrc := startSrc + dimensions
 		if (startSrc x >= 0 && startSrc y >= 0 && endSrc x <= source size x && endSrc y <= source size y && startDst x >= 0 && startDst y >= 0 && endDst x <= this size x && endDst y <= this size y) {
-			this _blitWhiteRaw(this buffer pointer, (source as This) buffer pointer, startSrc x, startSrc y, startDst x, startDst y, dimensions x, dimensions y, this stride, (source as This) stride)
+			this _blitRaw(this buffer pointer, (source as This) buffer pointer, startSrc x, startSrc y, startDst x, startDst y, dimensions x, dimensions y, this stride, (source as This) stride, blendMode)
+		}
+	}
+	exactDistance: func (other: This) -> Int {
+		if (this size != other size) {
```
emilwestergren: No brackets needed

source/draw/RasterMonochrome.ooc
davidpiuva: We can do style optimizations after we have ported it to C++ and have access to macros.

emilwestergren: Did you try if-statement?

source/draw/RasterMonochrome.ooc
emilwestergren: Ok

source/draw/RasterMonochrome.ooc
davidpiuva: I tried using if statements in pixel loops before and the only type of if statement that was fast enough for realtime use was the conditional move instructions taking one cycle. All others force the instruction scheduler to stop and think about what to do with the excess ALUs since it is not within the same execution window.

davidpiuva: On Little clusters, there would less significance of branching but this is most likely placed on the Big cluster that has out of order execution.

emilwestergren: Maybe put the blending modes in different functions at least.

davidpiuva: I would have to read the assembly output and check if the C compiler inlines them since C compilers are very careful about inlining.

davidpiuva: Maybe put comments around the differing sections?

source/draw/RasterMonochrome.ooc
```diff
@@ -43,25 +43,60 @@ RasterMonochrome: class extends RasterPacked {
...
+			upperLeftSource := source + startSrcX + (startSrcY * sourceStride)
+			writerLineStart := upperLeftTarget
+			readerLineStart := upperLeftSource
+			match (blendMode) {
+				case BlendMode White =>
+					for (y in 0 .. height) {
+						writer := writerLineStart
+						reader := readerLineStart
+						for (x in 0 .. width) {
+							// White function
```
davidpiuva: Updated

source/draw/RasterMonochrome.ooc
emilwestergren: I don't think the comment adds anything, I was more trying to reduce the size of this function. Image we have 5 different modes, this function will be huge.

davidpiuva: The C++ port will probably be more compact.

source/draw/Image.ooc
```diff
@@ -67,8 +67,8 @@ Image: abstract class {
 	distance: virtual abstract func (other: This) -> Float
 	equals: func (other: This) -> Bool { this size == other size && this distance(other) < 10 * Float epsilon }
 	isValidIn: func (x, y: Int) -> Bool { x >= 0 && x < this size x && y >= 0 && y < this size y }
-	// Writes white text on the existing image
-	write: func (message: String, fontAtlas: This, localOrigin: IntPoint2D) {
+	// Writes white text on the existing image (with black background if solid is true)
```
marcusnaslund: `solid`?

source/draw/RasterMonochrome.ooc
```diff
@@ -79,7 +114,34 @@ RasterMonochrome: class extends RasterPacked {
...
+			thisUpperLeft := this buffer pointer
+			otherUpperLeft := other buffer pointer
+			thisStride := this stride
+			otherStride := other stride
+			thisLineStart := thisUpperLeft
+			otherLineStart := otherUpperLeft
+			for (y in 0 .. this size y) {
+				thisPixel := thisLineStart
+				otherPixel := otherLineStart
+				for (x in 0 .. this size x) {
```
marcusnaslund: Since performance is critical here you should definitely break out `size x` and `size y` into variables

davidpiuva: I can probably optimize it all the way here too but Emil might dislike that since this comparison function is only executed for regression tests. I just copied the render loop to save some time on regression tests.

marcusnaslund:
Alright. Looks reasonable, there is code duplication but I understand the unfortunate reasons (ooc) for it.

PR #2067, opened 2017-01-24
Blitting to YUV will blit to Y

davidpiuva:
Solves a bug where DrawState was used on the CPU instead of the faster blitter.


davidpiuva:
@emilwestergren Peer review?

PR #2068, opened 2017-01-31
Removed Shell2D classes

marcusnaslund:
May have served a purpose at some point?


marcusnaslund:
the license checker has collapsed for some reason, otherwise PRInt is happy. These classes aren't used anymore.

PR #2071, opened 2017-02-08
Now using custom vertex shader when drawing lines.

emilwestergren:
Drawing lines now works. @christopherlagerhult


christopherlagerhult:
I should probably add a version flag for my new libs.

emilwestergren:
It works on the other starfish at least, haven't tested on eagle

christopherlagerhult:
Tested it and it works for eagle

marcusnaslund:
There should be no review flags for the starfish branch right?

emilwestergren:
Should I close this @christopherlagerhult ?

christopherlagerhult:
I guess I can add this by rebasing so yes

PR #2072, opened 2017-02-08
Now using custom vertex shader when drawing lines.

emilwestergren:
Using the generic transform vertex shader caused some OpenGL drivers to fail when drawing lines.
@christopherlagerhult


marcusnaslund:
I'm happy too, but PRInt is not

emilwestergren:
Traceback (most recent call last): File "licensecheck.py", line 39, in <module> check_file(filepath) File "licensecheck.py", line 10, in check_file lines = f.readlines() File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode return codecs.ascii_decode(input, self.errors)[0] UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 3937: ordinal not in range(128)

What's wrong?

marcusnaslund:
Oh it's just that? Then we're OK.
@fredrikbryntesson @rikardwiklander can you please either remove the "License check" step or fix whatever is preventing it from working properly?

PR #2073, opened 2017-02-09
Added flush after glFenceSync

christopherlagerhult:
It solves the endless-wait-problem when `glClientWaitSync` is beeing called on a different thread than `glFenceSync`


emilwestergren:
Is this the fix that is already in develop? Do you think this branch can sync with develop any time soon?

christopherlagerhult:
Oh..didn't know that. I will try to rebase it towards develop instead

PR #2075, opened 2017-02-14
Now supports drawing from EGLYuv to OpenGLMonochrome and OpenGLUv.

emilwestergren:
@erikhagglund @christopherlagerhult

### DIFF
--- a/source/draw/gpu/opengl/YuvOptimizedContext.ooc
+++ b/source/draw/gpu/opengl/YuvOptimizedContext.ooc
@@ -61,9 +61,9 @@ YuvOptimizedContext: class extends OpenGLContext {
 		result: OpenGLMap
 		if (input instanceOf(EGLYuv) && output instanceOf(EGLYuv))
 			result = this _yuvShader
-		else if (output instanceOf(EGLYuv) && (input instanceOf(OpenGLMonochrome) || input instanceOf(RasterMonochrome)))
+		else if ((input instanceOf(OpenGLMonochrome) || input instanceOf(RasterMonochrome)) && output instanceOf(EGLYuv))
 			result = this _monochromeToYuv
-		else if (output instanceOf(EGLYuv) && input instanceOf(GpuYuv420Semiplanar))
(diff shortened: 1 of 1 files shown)


source/draw/gpu/opengl/YuvOptimizedContext.ooc
```diff
@@ -64,6 +65,10 @@ YuvOptimizedContext: class extends OpenGLContext {
 			result = this _monochromeToYuv
 		else if (output instanceOf(EGLYuv) && input instanceOf(GpuYuv420Semiplanar))
 			result = this _compositeYuvToNativeYuv
+		else if (input instanceOf(EGLYuv) && output instanceOf(OpenGLMonochrome))
```
erikhagglund: If you feel like it you could place all ```input instanceOf(EGLYuv)``` after one another.

Is there any need for support of render to RasterMonochrome? (I guess we simply add when needed)

emilwestergren: > If you feel like it you could place all input instanceOf(EGLYuv) after one another.

I can do that.

> Is there any need for support of render to RasterMonochrome? (I guess we simply add when needed)

Not that I know of, but I guess it would be simple to implement if necessary, just render to a temp GPU image and then download the result.

emilwestergren:
Updated

source/draw/gpu/opengl/YuvOptimizedContext.ooc
erikhagglund: >>If you feel like it you could place all ```input instanceOf(EGLYuv)``` after one another.

>I can do that.

Not exactly what i intended, but ok ;)

marcusnaslund:
*(gentle reminder)* It is up to the owner of the PR to set the ready flag once there is approval from PRInt and peers.

emilwestergren:
Is that really necessary? You can see if the PR is approved and Print passed in the PR list.

marcusnaslund:
The label serves not just to highlight "This is accepted" but also "I am finished and won't update anymore, please go ahead." which is not always implied by the first.

PR #2076, opened 2017-02-27
Cast to int

christopherlagerhult:
Solves **Invalid Write, Uninitialized value, Conditional jump** errors reported by Valgrind 

I'm not sure why this problem occurs, and neither why this PR solves it. It's almost the same C code except this one is broken down to smaller pieces.

### DIFF
--- a/source/system/Numbers.ooc
+++ b/source/system/Numbers.ooc
@@ -100,7 +100,9 @@ LDouble: cover from long double {
 		result
 	}
 	in: func (range: Range) -> Bool {
-		this as Int >= range min && this as Int < range max
+		(min, max) := (range min as This, range max as This)
+		(aboveMin, belowMax) := (this >= min, this < max)
+		aboveMin && belowMax
 	}
 }
 
--- a/source/system/Numbers.ooc
+++ b/source/system/Numbers.ooc
@@ -100,7 +100,7 @@ LDouble: cover from long double {
 		result
 	}
 	in: func (range: Range) -> Bool {
-		this >= range min && this < range max
+		this as Int >= range min && this as Int < range max
 	}
 }
 


source/system/Numbers.ooc
```diff
@@ -100,7 +100,7 @@ LDouble: cover from long double {
 		result
 	}
 	in: func (range: Range) -> Bool {
-		this >= range min && this < range max
+		this as Int >= range min && this as Int < range max
```
erikhagglund: Since `range min` and `range max` is Int this should be same as previously.

However what does "as Int" mean, should it be truncated first? `this min(Int maximumValue) as Int >= range min && this as this max(Int minimumValue) < range max`

christopherlagerhult: > Since range min and range max is Int this should be same as previously.
I assume that previously, `range min/range max` was converted to `This`, while it now truncates `this` before comparison. 

> However what does "as Int" mean
cast to int

I felt it was inreasonable to expect `2.1 in (1 .. 2)` to be invalid. Coming to think of it, I would expect this to succeed so I wil update it :D

christopherlagerhult: The question here I guess is if the comparison should be `... this as Int <= range max` rather than `... this as Int < range max`. I leave it to @marcusnaslund to decide

erikhagglund:
```
value := Int minimumValue - 1
value in(Int minimumValue .. Int maximumValue)
```
will return true if we round `this`, so we cannot convert this to Int, but must solver the `(long) range min` problem somehow

marcusnaslund: It should be <

christopherlagerhult:
Yes, for once I'm not accually sure OOC is to be blamed.

christopherlagerhult:
Not sure why Print PC / magic never finishes

PR #2077, opened 2017-03-06
Added a function to check on which side of a vector a point lies

alexisboucharin:


alexisboucharin:
Fixed!

marcusnaslund:
Looks great now but PRInt Android failed for unknown reasons

PR #2085, opened 2017-04-07
Removed debug init() call in QuaternionTest

marcusnaslund:
Git tells me to blame @thomasfanell 20 months ago, so you get to do the review. :)


marcusnaslund:
Haha I think I know why now. This was hiding a completely different bug.

marcusnaslund:
Yep, now PRInt passes. See my latest commit

PR #2088, opened 2017-04-27
Changed crop box from float to int

tomasmahlberg:
This should of course be of an int type, not float. Sorry about that.


marcusnaslund:
`ready`?

PR #2095, opened 2017-06-14
register callbacks for GraphicBuffer

sebastianbaginski:
This will make our lives much easier in the other-other project. Otherwise we'd have to go through lenghty workarounds, as these function pointers are not properties (no associated setter function).
I can explain in more detail in private conversation.


sebastianbaginski:
I'm not really sure who should review this, feel free to comment @christopherlagerhult @erikhagglund @ziquanyu @emilwestergren

sebastianbaginski:
@marcusnaslund what's wrong here ? tests on the other PR went through, could you restart it ?

PR #2100, opened 2017-08-08
Moved save to Image class.

emilwestergren:
Don't see why not all images could potentially support this.


emilwestergren:
I will merge this since it's such a small change and I have double approval

PR #2102, opened 2017-08-16
Added changes from El Toro

emilwestergren:


emilwestergren:
Replaced with https://github.com/magic-lang/ooc-kean/pull/2103

PR #2109, opened 2017-09-19
Added create func in Image.

firog:
Convenience function to be able to easier create an image of the same type as the current and the same size.


marcusnaslund:
Tests pass.

marcusnaslund:
Thank you for your contribution, @firog !

PR #2112, opened 2017-10-11
Introduced squared norm to avoid floating roundoff

marcusnaslund:
Also using `squared` avoids casting to/from `double` (in C).

### DIFF
--- a/source/geometry/Quaternion.ooc
+++ b/source/geometry/Quaternion.ooc
@@ -24,10 +24,11 @@ Quaternion: cover {
 	isValid ::= this real isNumber && this imaginary isValid
 	isIdentity ::= this real equals(1.f) && this imaginary isZero
 	isZero ::= this real equals(0.f) && this imaginary isZero
-	norm ::= (this real squared + this imaginary norm squared) sqrt()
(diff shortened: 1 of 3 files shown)


source/geometry/Quaternion.ooc
```diff
@@ -24,10 +24,11 @@ Quaternion: cover {
 	isValid ::= this real isNumber && this imaginary isValid
 	isIdentity ::= this real equals(1.f) && this imaginary isZero
 	isZero ::= this real equals(0.f) && this imaginary isZero
-	norm ::= (this real squared + this imaginary norm squared) sqrt()
+	norm ::= (this real squared + this imaginary squaredNorm) sqrt()
+	squaredNorm ::= this real squared + this imaginary squaredNorm
 	normalized ::= this / this norm
 	conjugate ::= This new(this real, -this imaginary)
-	inverse ::= this conjugate / (this real squared + this imaginary norm squared)
+	inverse ::= this conjugate / squaredNorm
```
marcusnaslund: now dividing by `(w^2+x^2+y^2+z^2)` instead of `(w^2+sqrt(x^2+y^2+z^2)^2)`

PR #2118, opened 2017-12-20
Now possible to set both min/mag filter in a single call.

emilwestergren:
We always do them in pair anyway, will remove unnecessary rebinding of the texture.


marcusnaslund:
needs porting to c++?

emilwestergren:
Yes, should I do it?

marcusnaslund:
Please :)

PR #2130, opened 2018-03-07
Added identifier

christopherlagerhult:
I have ignored yuv and opengloptimized contexts as they are to be removed...right?

### DIFF
--- a/source/draw/gpu/opengl/EGLYuv.ooc
+++ b/source/draw/gpu/opengl/EGLYuv.ooc
@@ -16,9 +16,12 @@ import backend/egl/EGLImage
 
 EGLYuv: class extends OpenGLPacked {
 	_handle: Pointer
+	_identifier: Pointer
 	handle ::= this _handle
+	identifier ::= this _identifier ?? this _handle
 	init: func (buffer: GraphicBuffer, context: OpenGLContext) {
 		this _handle = buffer handle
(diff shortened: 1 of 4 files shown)


source/draw/gpu/opengl/GraphicBuffer.ooc
```diff
@@ -46,8 +47,10 @@ GraphicBuffer: class {
 	length ::= this stride * this size y
 	nativeBuffer ::= this _nativeBuffer
 	handle ::= this _handle
+	identifier ::= this _identifier ?? this _handle
 
 	init: func (=_backend, =_nativeBuffer, =_handle, =_size, =_pixelStride, =_format, =_userData)
+	init: func ~identifier (=_backend, =_nativeBuffer, =_handle, =_size, =_pixelStride, =_format, =_userData, =_identifier)
```
emilwestergren: You should call this constructor from the other one with `_handle` as identifier

source/draw/gpu/opengl/EGLYuv.ooc
```diff
@@ -16,9 +16,12 @@ import backend/egl/EGLImage
 
 EGLYuv: class extends OpenGLPacked {
 	_handle: Pointer
```
emilwestergren: This is not needed anymore

PR #2141, opened 2018-09-20
Rewrote LDouble::in in the same style as LLong and ULLong

marcusnaslund:
Hello everyone, what's new? :wave:


ghost (dismissed):
range can be const

